PR #6, opened 2017-06-27
Implement `@reaction_network`

alanderos91:


src/DiffEqBiological.jl
```diff
@@ -7,11 +7,15 @@ using DiffEqJump
 using Compat
 @compat abstract type AbstractReaction end
 
+import DataStructures: OrderedDict
```
ChrisRackauckas: This needs a dep in REQUIRE

test/reaction_network.jl
```diff
@@ -0,0 +1,16 @@
+using DifferentialEquations
```
ChrisRackauckas: This can just be `using DiffEqBiological`

ChrisRackauckas:
thanks!

PR #7, opened 2017-10-10
Higher order reactions

weilandtd:
To my opinion is the current calculation of the reaction rate is not correct for the case of higher with order > 1. 
For example for a dimerization the current implementation would consider for the reaction: 
Reaction(k,[1,1], [(1,-2), (2,1)])
The rate 
v = k*u[1]*u[1] 
but it should be 
 v = k*u[1]*(u[1]-1)
since a single monomer cannot dimerize with itself. 

The propsed changes in the code should correct this.
I am not sure if it is a nice solution but it works for me.

### DIFF
--- a/src/reactions.jl
+++ b/src/reactions.jl
@@ -59,8 +59,16 @@ ReactionSet(args...) = ReactionSet(tuple(args...))
 function build_jumps_from_reaction(r::Reaction;save_positions=(false,true))
   rate = function (t,u)
     val = r.rate_constant
+    # for higher order interactions i.e. reactants [1,1] vaule = k*u[1]*(u[1]-1)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Makes sense. Can you add tests?

weilandtd:
Tests added, and also a small fix for the use of mixed reactants, like [1,1,2]

weilandtd:
Just realized you have on open issue with this but I think this would fix the problem since it is only applying this rule if a jump problem is created, which is a discrete problem and this case it should be consideried to avoid the creation of mass from nothing.

ChrisRackauckas:
This is great, thanks!

PR #15, opened 2018-02-08
Reaction Network Remake

Gaussia:
Uploading an alternative DSL for reading reaction networks. Currently it does not replace anything of the old DSL, rather it is added in parallel to the package. The DiffEqBiological module will now also load file "ReactionNetwork.jl" (which contains all(ish) the new code). It will also export the macro @reaction_network_new.

We had some problem properly creating dispatch for Ode/SDE/Jump Problems. I don't think it ishard to fix (if one knows how) but for now there is just replacement functions called newODE/SDE/JumpProblem. These are located in the problem file.

Initially I had a base macro for generating expressions and functions as I described. About the time I contacted you about having my own version I got @korsbo interested (he sits next to be at the department) and he have added some extra functionality to the final structure (like Jacobians and integrated support in his Latexify package). He also suggested a much better way to include parameters which was added. There were also a few more things I wanted to include that I figured would be independent of your feedback, so I figured we might as well wait with sending it over to you until we had something we were satisfied with. 

The basic functionality of the DSL is pretty much like this:
```julia
rn = @reaction_network_new rType begin #Creates a reaction network of type rType.
        2.0, X + Y --> XY            #This will have reaction rate corresponding to 2.0*[X][Y]
        2.0, XY ← X + Y              #Identical to 2.0, X + Y --> XY
        (2.0,1.0), X + Y ↔ XY     #Identical to reactions (2.0, X + Y --> XY) and (1.0, XY --> X + Y).
        2.0, X + Y ⟾ XY            #Ignores mass kinetics. This will have reaction rate corresponding to 2.0*[X].
        hill(XY,2,2,2), X + Y --> XY    #Reaction inis activated by XY according to a hill function. hill(x,v,K,N).
        mm(XY,2,2), X + Y --> XY     #Reaction inis activated by XY according to a m.m. function. 
        2.0, (X,Y) --> 0                   #This corresponds to both X and Y degrading at rate 2.0.
        (2.0, 1.0), (X,Y) --> 0          #This corresponds to X and Y degrading at rates 2.0 and 1.0, respectively.
        2.0, (X1,Y1) --> (X2,Y2)      #X1 and Y1 becomes X2 and Y2, respectively, at rate 2.0.
    end
```
Can also use parmaters:
```julia
    kB = 2.0; kD = 1.0
    p = [kB, kD]
    p = []
    rn = @reaction_network_new type begin
        (kB, kD), X + Y --> XY       #Lets you define parameters outside on network. Parameters can be changed without recalling the network.  
    end kB, kD
```
(This is just fetched from a small documentation inside the main file)

You can decide if you think you can include this in you big package in some way. Either case I would always appreciate feedback :)
cheers
Torkel


src/ReactionNetwork.jl
```diff
@@ -0,0 +1,341 @@
...
+    coordinate(name, ex, p)
+end
+
+#Declare various arrow types symbols used for the empty set (also 0).
+empty_set = Set{Symbol}([:∅])
+fwd_arrows = Set{Symbol}([:>, :→, :↣, :↦, :⇾, :⟶, :⟼, :⥟, :⥟, :⇀, :⇁, :⇒, :⟾])
+bwd_arrows = Set{Symbol}([:<, :←, :↢, :↤, :⇽, :⟵, :⟻, :⥚, :⥞, :↼, :↽, :⇐, :⟽])
+double_arrows = Set{Symbol}([:↔, :⟷, :⥎, :⥐, :⇄, :⇆, :⇋, :⇌, :⇔, :⟺])
+no_mass_arrows = Set{Symbol}([:⇐, :⟽, :⇒, :⟾, :⇔, :⟺])      #Using this arrows will disable the program from multiplying reaction rates with the substrate concentrations. Gives user full control of reaction rates.
+disallowed_reactants = Set{Symbol}([:du, :u, :p, :t])           #These are not allowed since they are used in "return :((du,u,p,t) -> $system)", if a variable these gets replaced with e.g. u[1], which is bad. To Do: Make change so that these can be used.
```
ChrisRackauckas: ParameterizedFunctions.jl avoids this issue by picking names that won't hit user names. It appends, `internal_var___`, i.e. `internal_var___du`, with then the only issues being that the user should know to use `t` for time. You probably want to make use of something similar here to reduce the number of "sacred names" that will commonly be hit.

Gaussia: @korsbo pointed this out to me as well (both one and two times). I agree it is better that way (I will update it sometime today probably).

src/ReactionNetwork.jl
```diff
@@ -0,0 +1,341 @@
...
+    reactant::Symbol
+    stoichiometry::Int64
+end
+
+#Structure containing information about one Reaction. Contain all its substrates and products as well as its rate. Contains an specialized constructor.
+struct ReactionStruct
+    substrates::Vector{ReactantStruct}
+    products::Vector{ReactantStruct}
+    rate::Any
+    function ReactionStruct(sub_line::Any, prod_line::Any, rate::Any, use_mass_kin::Bool)
```
ChrisRackauckas: there's no need to `::Any`. That's the same thing as leaving it off. Just a style thing.

src/DiffEqBiological.jl
```diff
@@ -2,20 +2,28 @@ __precompile__()
 
 module DiffEqBiological
 
-using DiffEqJump
+using Reexport
+using DiffEqBase
+using SymEngine
```
ChrisRackauckas: This needs to be added to the Require file.

src/problem.jl
```diff
@@ -1,7 +1,20 @@
...
                  JumpProblem(prob,aggregator,build_jumps_from_reaction(rs...);kwargs...)
-                 
+
 GillespieProblem(prob,aggregator::AbstractAggregatorAlgorithm,
                  r::ReactionSet;kwargs...) =
                  JumpProblem(prob,aggregator,build_jumps_from_reaction(r);kwargs...)
+
+### Added by Me ###
+### ODEProblem ###
+newODEProblem(ar::AbstractReaction, args...; kwargs...) =
```
ChrisRackauckas: To overload, do `DiffEqBase.ODEProblem`, etc.

src/ReactionNetwork.jl
```diff
@@ -0,0 +1,341 @@
...
+    return expr
+end
+
+#Makes the Jacobian.
+function calculate_jac(f_expr::Vector{Expr}, syms)
+    symjac = Matrix{SymEngine.Basic}( length(syms), length(syms))
+    symfuncs = [SymEngine.Basic(f) for f in f_expr]
+    for i in eachindex(f_expr)
+        for j in eachindex(syms)
+          symjac[i,j] = diff(symfuncs[i],syms[j])
```
ChrisRackauckas: just for reference: this looping is done in row major form so you usually want to flip the ordering in Julia. Performance won't matter here, but it's something to keep in mind.

ChrisRackauckas:
Wow, that's a really clean read. Very well done! A few things to note:

1. Instead of using recursion like ParameterizedFunctions.jl, you might want to make use of MacroTools.jl's `postwalk`:

https://github.com/MikeInnes/MacroTools.jl#expression-walking

that can be cleaner, though that's just a style thing. There's nothing wrong with the recursion.

2. I didn't see how higher order terms were handled, but it would be nice to address https://github.com/JuliaDiffEq/DiffEqBiological.jl/issues/10 . I saw where they were counted but not some form of handling on this.

3. You might as well create the inverse W and inverse W_t matrices if possible, though that can always be added later.

4. This needs unit tests.

5. Once again, great job. This is good work.

ChrisRackauckas:
@alanderos91 if you have time, would you mind taking a look at this?

Gaussia:
Cheers :)
Will be looking at the feedback and make the appropriate changes.

When I made my initial version I didn't actually think of this issue, it was not until I looked closer at he DiffEqBiological implementation that I saw the issue. While I understand the motivation I was a bit uncertain of the consensus and decided to not do anything until I could get more input.

When defining the reaction rates for the ConstantRateJumps it would not be hard to replace X^n with bin(X,n) (X is a discrete amount) and the product rate * R1^n1 * R2^n2 * ... becomes rate * bin(R1,n1) * bin(R2,n2) * ...

ChrisRackauckas:
When it's a discrete value, it should divide by `k!`. When it's a continuous value (a concentration), it should not. The DSL should just by default assume that for a jump problem they are all discrete, and in ODEs/SDEs they are all continuous. That solution will take us quite far, and anything else will likely need to build the rates directly.

Gaussia:
I see. So for Guillespie one would have: X^n -> X(X-1)...(X-n)/n!
For SDE and ODE would be mean that one have X^n -> X(X-1)...(X-n). Anyone trying to run an ODE for low values would get a nasty surprise. Ofcourse one would then have to make sure to run the simulation with values corresponding to actual copy numbers. Just checking so I understand, the actual implementation is straightforward.

ChrisRackauckas:
>Anyone trying to run an ODE for low values would get a nasty surprise.

The convergence result isn't true for small values of the ODE though. Gillespie's convergence result is true as the copy number goes to infinity: concentrations close to zero violate this. There's not really much you can do there except extrapolate the mass action laws to this domain. The crucial difference is that the ODEs/SDEs are written in terms of concentration instead of copy number, and thus dividing by `k!` doesn't make sense with the units anyways. But also, since it's in the limit where copy number goes to infinity, the `k!` constant based on small integer copies erodes away.

ChrisRackauckas:
Actually hmm, looking at the convergence result, the constants might have the binomial term from the repeated parts of the summation. Can someone double check this?

Gaussia:
I've modified the program accordingly (mostly minor changes really).
For now I have used bionomials in when creating the reaction rates for the ConstantRateJumps and made no changes to the rates for ODEs and SDEs.

src/ReactionNetwork.jl
```diff
@@ -0,0 +1,343 @@
...
+Most arrows accepted (both right, left and bi drectional arrows).
+Note that while --> is a correct arrow, neither <-- nor <--> works.
+Using non-filled arrows (⇐, ⟽, ⇒, ⟾, ⇔, ⟺) will disable mass kinetics and lets you cutomize reaction rates yourself.
+Use 0 or ∅ for degradation/creation to/from nothing.
+Example systems:
+    rn = @reaction_network_new rType begin #Creates a reaction network of type rType.
+        2.0, X + Y --> XY                  #This will have reaction rate corresponding to 2.0*[X][Y]
+        2.0, XY ← X + Y                    #Identical to 2.0, X + Y --> XY
+        (2.0,1.0), X + Y ↔ XY              #Identical to reactions (2.0, X + Y --> XY) and (1.0, XY --> X + Y).
+        2.0, X + Y ⟾ XY                   #Ignores mass kinetics. This will have reaction rate corresponding to 2.0*[X].
```
alanderos91: There is a minor ambiguity in the meaning of the `⟾` arrow. The example suggests the reaction's total rate depends only on the first substrate. I think this is a reasonable convention, but should be made explicit to a user.

Otherwise, (as a contrived example) we may have to think about what `2.0, X + Y + Z ⟾ XYZ` ought to mean.

Gaussia: Ok, this is actually a problem in my documentation (I think I changed the expression at some point to make them as simple as possible, but forgot to change some of the text.

The reaction rate in ´2.0, X + Y ⟾ XY´ will become 2.0. Basically it will only take whatever you write as reaction rate and not make any modifications. You can still write the reaction as ´2.0*X, X + Y ⟾ XY´ if you want reaction rate ´2.0*X´ (that is actually another non obvious thing that I probably should add...). You can put pretty much anything in the reaction rate as long as it do not contain variables not passed as parameters or functions which you have defined yourself. E.g.
´´´julia
rn = @reaction_network_new type begin
        log(X,2.0), X + Y --> XY      
end 
´´´
should work fine (although I am unsure when you would want that). However
´´´julia
hill2(x,v,K) = v*x*x/(K*K+x*x)
rn = @reaction_network_new type begin
        hill2(X,3.0,1.5), X + Y --> XY      
end 
´´´
will not work. the hill function and m.m. function works because I have defined those internally in the package.

I will update it so that it actually says what ´⟾´ does, as well as what you can put as rates.

ChrisRackauckas:
The convergence result yields the SDE in equation 23.

http://aip.scitation.org/doi/pdf/10.1063/1.481811

korsbo:
I just added a way for users to provide their own functions. The syntax is best shown by example: 

```julia
using DiffEqBiological

@add_reaction_func my_hill_repression(x, v, k, n) = v*k^n/(k^n+x^n)

r = @reaction_network_new MyReactionType begin
    my_hill_repression(x, v_x, k_x, n_x), 0 --> x
end v_x k_x n_x

```

korsbo:
And now it errors nicely when you try to use an undefined function.

korsbo:
Should we rename `@add_reaction_func`? I don't want people to think that it is just adding a function, like 
```julia
myFunc(x) = x
@add_reaction_func myFunc
```
which will not work. 

is just plain `@reaction_func` better?

ChrisRackauckas:
>is just plain @reaction_func better?

I like this.

ChrisRackauckas:
Great. The user-defined functions answer the question of how to handle multiple regulation terms together.

src/ReactionNetwork.jl
ChrisRackauckas: I just noticed that mistake is from my code 📦

korsbo:
Cool, what else needs to be fixed before we can merge? 
The things that I can see are: 

- tests
- decide what to do with the old stuff/clean up
- docs in DiffEqDocs

Things that I would like to see fixed at some point would be: 
- expand symbolic calculations and actually use them (now or later?).
- allow for u0 = Matrix() or other multi-cell simulations (MultiScaleArrays?).
- allow for diffusion between cells, etc.

ChrisRackauckas:
I think those last 3 points should get an issue and be handled in future PRs. I prefer shortening PRs so that way things finish: going for the perfect PR tends to make them stall!

I think that this is enough of an extension of the previous DSL that we can comfortably just upgrade. But make sure you address @alanderos91 's points about the arrows. (Could you explain the `KineticLaw` part a little more though? I didn't get it)

src/ReactionNetwork.jl
korsbo: yup, we stole that and the type definition from ParameterizedFunctions.jl

ChrisRackauckas:
For spatial models, I have a `ReactionDiffusionProblem` in mind that I want to create so that way it has special tie-ins to the IIF integrators. Let's wait on those before extending to space.

korsbo:
Both me and @Gaussia are unsure of how to act on the arrow comments. Neither of us are experts in theoretical chemistry and we do not fully understand what the suggestions mean.

ChrisRackauckas:
One thing to find out: can there be a type-based way to plug into this DSL? Usually something type-based is easier to programmatically construct than trying to construct an expression for a macro. However, I'm not sure of how a type-based method could handle parameters, so I'm not sure there is an answer there.

korsbo:
I have thought about this for a few minutes and now my brain hurts. I do not have a good solution at the moment.

alanderos91:
> Both me and @Gaussia are unsure of how to act on the arrow comments.

I'm sorry for the confusion. I only meant to point out that, for example, `⇌` is used to make a statement at equilibrium. This is not a practical concern; only experts may be miffed about it.

 > (Could you explain the KineticLaw part a little more though? I didn't get it)

> One thing to find out: can there be a type-based way to plug into this DSL?

That's essentially what I was getting at. I thought using a type-based system might be helpful in letting the macro figure out whether `X * (X - 1) / 2` or `X^2` should be used (or, alternatively, for the user to specify what they want). This is a different issue altogether though: I think total rate like `X * (X - 1) / 2` should only ever be used in a `JumpProblem` or `GillespieProblem`. Maybe I'm wrong about this?

All that aside, I think `@add_reaction_func` is a better approach. Kudos!

ChrisRackauckas:
I agree with @korsbo that my head hurts: I am not sure it's possible to write some of this without a macro, unless the API is something horrid where the user states which indices of the parameters to use.

REQUIRE
```diff
@@ -1,5 +1,8 @@
 julia 0.6
 DiffEqJump 1.0.0
-DiffEqBase 3.0.0
+DiffEqBase 4.0.0
```
ChrisRackauckas: This doesn't exist.

ChrisRackauckas:
For the old DSL, since it cannot be encompassed by the new one, it should probably be kept but throw a depwarn on usage (to then be deleted a few months later).

REQUIRE
korsbo: Fixed, I set it to 3.3.0.

korsbo:
@Gaussia  has already cleaned out the old code, should do some git magic to get it back, or should we just create an overload to simulate the old DSL with the new code?

The overload 
```julia
macro reaction_network(expr)
    coordinate(:Reaction, expr, Tuple{Float64}())
end
```
would imitate the old DSL macro. However, I don't think that the backwards-compatibility would be perfect (we would for example not have any VariableRateReaction types).

ChrisRackauckas:
Yeah, backwards compatibility would be too hard, but it would be good to throw deprecation warnings so people know how to update.

ChrisRackauckas:
I added an error for the old Reaction type to make reference to the docs. I think the remaining stuff is:

1. Delete the equilibrium arrows.
2. Make sure the constants for higher order terms is correct.
3. Update the tests.

ChrisRackauckas:
What's this noise scaling thing?

Gaussia:
If one declares a system as normal it will not appear (if I have implemented it correctly), but if one declare a reaction like:
```julia
rn2 = @reaction_network rn η begin
    1, 0 --> X
end
```
All noise entries in g will be multiplied with the scaling factor η, which is be default set to be the last parameter in your parameter array p. Basically the idea is that if one are playing around with a system but no clue on the actual amounts involved this will offer a simple way to adjust the amount of noise.

ChrisRackauckas:
>Basically the idea is that if one are playing around with a system but no clue on the actual amounts involved this will offer a simple way to adjust the amount of noise.

IMO the nicer handling of this would be to allow for affine (additive + multiplicative) noise to be added. If the noise is just phenomenological and not following the convergence theorem, then people usually use additive or multiplicative noise, and we have a bunch of special integrators to do these fast.

Gaussia:
Sounds good to me. Will check that out.

ChrisRackauckas:
BTW, thanks for taking this on. This is quite a big and scary PR to be your first!

Gaussia:
It is a good but very chaotic way to learn new things...

test/runtests.jl
```diff
@@ -2,8 +2,13 @@ using DiffEqBiological
 using Base.Test
 
 tic()
-@time @testset "Gillespie Tests" begin include("gillespie.jl") end
-@time @testset "Variable Rate Reaction Tests" begin include("variable_rate_reactions.jl") end
-@time @testset "@reaction_network Tests" begin include("reaction_network.jl") end
-@time @testset "Higher order reaction Tests" begin include("higher_order_reactions.jl") end
+#@time @testset "Gillespie Tests" begin include("gillespie.jl") end
+#@time @testset "Variable Rate Reaction Tests" begin include("variable_rate_reactions.jl") end
```
ChrisRackauckas: You should make these tests, when applicable, use the new DSL to make sure that the old models can still be solved.

src/DiffEqBiological.jl
```diff
@@ -22,12 +22,4 @@ export AbstractReaction
 export @reaction_network, @reaction_func
 export ODEProblem, SDEProblem, JumpProblem
 
-Reaction(args...) = error("""
-The old Reaction DSL is deprecated for a new 
-macro-based DSL which supports parameters, regulation,
```
ChrisRackauckas: This got deleted in the merge.

test/test_gillespie.jl
```diff
@@ -48,28 +51,3 @@ nums = Int[]
   push!(nums,sol[end][3])
 end
 println("Direct Jumps: $(mean(nums))")
-
-
-using Gillespie
```
ChrisRackauckas: This is for testing against Gillespie.jl. Please leave it in.

test/variable_rate_reactions.jl
```diff
@@ -1,48 +1,35 @@
 using DiffEqBiological, DiffEqJump, DiffEqBase, OrdinaryDiffEq, StochasticDiffEq
 using Base.Test
+using tmpMod
```
ChrisRackauckas: What's this?

test/variable_rate_reactions.jl
```diff
@@ -1,48 +1,35 @@
...
 
-r1 = VariableRateReaction(1e-4,(1,2),((1,-1),(2,1)))
-r2 = VariableRateReaction(0.01,[2],[(2,-1),(3,1)])
+sir_model = @reaction_network rn begin
+    0.1/1000, s + i --> 2i
+    0.01, i --> r
```
ChrisRackauckas: Is this making use of variable rate reactions?

test/test_gillespie.jl
Gaussia: Ok, figured it had something to do with the old GuillespieProblems.

test/runtests.jl
```diff
@@ -3,7 +3,12 @@ using Base.Test
 
 tic()
 @time @testset "Gillespie Tests" begin include("gillespie.jl") end
+@time @testset "Run Another Gillespie Tests" begin include("test_gillespie.jl") end
 @time @testset "Variable Rate Reaction Tests" begin include("variable_rate_reactions.jl") end
-@time @testset "@reaction_network Tests" begin include("reaction_network.jl") end
-@time @testset "Higher order reaction Tests" begin include("higher_order_reactions.jl") end
+#@time @testset "@reaction_network Tests" begin include("reaction_network.jl") end
```
ChrisRackauckas: This one can be deleted.

test/runtests.jl
```diff
@@ -3,7 +3,12 @@ using Base.Test
...
+#@time @testset "Higher order reaction Tests" begin include("higher_order_reactions.jl") end
```
ChrisRackauckas: This test we will need to have updated to make sure we get the constants correct.

test/variable_rate_reactions.jl
Gaussia: It has to do with me not really being very good with git and understanding the loading of packages in different orders. Basically I could not use the new reaction_network by loading DiffEqBiological, so I renamed it as a module tmpMod so I can load it and try it. 

This is only a temporary I had there while running the code checking that it worked, forgot to remove it.

test/variable_rate_reactions.jl
Gaussia: No

ChrisRackauckas: We'll need a way to set that if we want to do this kind of mixing then, or not support it.

test/runtests.jl
Gaussia: Yes. Right now only the Guillespie/Jump simulations are updated for Higher Order Terms. If we want to have similar modifications in SDE it would make sense (?) to test that here as well.

Gaussia:
In an effort to finally fixing the higher order reaction business I have had a think trying to figure out how it should be. Starting with the discrete simulations (using absolute copy number), if we have the reaction:
```julia
k, 2X --> ...
```
then the rate should be k multiplied with X*(X-1)/2! which is simply the number of combinations of X molecules (generalized to bin(X1,n1)*bin(X2,n2)*... if someone decides to put a lot of reactants in a single reaction, rather than several).

For SDE and ODE it is anyway assumed that X is large, and hence X(X-1)...(X-N+1) ~ X^N, which is convenient since people will not always use real numbers of X corresponding to the actual copy number when simulating. However there should be no real reason to discard n! right? When rereading the 2000 Gillespie paper it seems to agree (second page, left side, the paragraph starting with "The function h_j(x)"). It only include a very short example and do include the X(X-1)... which I want to drop. However it also uses the correction of 2!.

My current suggestion is to have X*(X-1)/2! for discrete simulations and X*X/2! for ODE/SDE. Of course there are situations were one would want to customize this factors yourself (like when replicating the models of people who have not used the factor...). The use of the ⟾ and similar would allow for this by giving you full control over the reaction rate. However in the future we might want to discuss if there are several approaches "correct" enough for them to have individual arrows.

comments?

Gaussia:
I've added a documentation file as well. Honestly have no real experience in these so simply wrote down all the details about the DSL. Will definitely need some reworking and formatting though, including making sure that it is actually rendered prettily.

Gaussia:
I get the impression that we might be wrong whatever we pick? Would our approach to try and pick whichever is the simplest one, or try to decide which one is least wrong?

My fear with using binomials is that then one have to ensure that the values corresponds to the actual numbers, which might be a healthy things but sometimes it is convenient to skip. This is definitely one thing that would crash a lot of peoples programs since there is not good way to find the error and throw a proper warning (?).

ChrisRackauckas:
Sorry, was gone all day.

>For SDE and ODE it is anyway assumed that X is large, and hence X(X-1)...(X-N+1) ~ X^N, which is convenient since people will not always use real numbers of X corresponding to the actual copy number when simulating.

Yes. A better way to think about it is that the units of `X` when an SDE or ODE is concentration, so `-1` doesn't make sense. In fact, `X-1` can be negative just by the choice of units for concentrations, even if `X` has a really large copy number (since then the amount requires knowledge of volume). So using the `X^N` approximation is the only consistent way to handle it.

So then

>My current suggestion is to have X*(X-1)/2! for discrete simulations and X*X/2! for ODE/SDE. Of course there are situations were one would want to customize this factors yourself (like when replicating the models of people who have not used the factor...). The use of the ⟾ and similar would allow for this by giving you full control over the reaction rate. However in the future we might want to discuss if there are several approaches "correct" enough for them to have individual arrows.

Yes, I agree.

>I've added a documentation file as well. Honestly have no real experience in these so simply wrote down all the details about the DSL. Will definitely need some reworking and formatting though, including making sure that it is actually rendered prettily.

This should just be the README, after the note to the DiffEqDocs (and we should do something with the docs there too).

korsbo:
>  X*X/2! for ODE/SDE.

what's up with the 2 factorial? I do not see why this would belong in a DE. With DEs I would usually have x represent a concentration.  Is this correction term correct?

korsbo:
Ok, I think I see the 2!. But I may well have missed this when translating backa and forth between this DSL and maths. It looks like we are in danger of misleading the user. Or is this just me?

ChrisRackauckas:
>Ok, I think I see the 2!. But I may well have missed this when translating backa and forth between this DSL and maths. It looks like we are in danger of misleading the user. Or is this just me?

Doesn't it need to have the `2!` in order to get the means/steady states correct? It should be clearly documented, but it needs to be consistent and requiring that the means are all the same is quite necessary.

test/higher_order_reactions.jl
```diff
@@ -1,40 +1,38 @@
...
-
-
-# Solve the dimerization problem
-prob = DiscreteProblem([1,0],(0.0,250.0))
-jump_prob = GillespieProblem(prob,Direct(),r2)
-sol = solve(jump_prob,Discrete())
-@test find( x-> x!=0 ,[u[2] for u in sol.u]) == []
+using DifferentialEquations
+using Base.Test
+using tmpMod
```
ChrisRackauckas: Remove tmpMod. BTW, you might want to check out Revise.jl for doing interactive development, or using Juno.

test/higher_order_reactions.jl
```diff
@@ -1,40 +1,38 @@
...
+function tmp_sol2vec(sol,j)
```
ChrisRackauckas: This is just `vec(convert(Array,sol))`?

test/higher_order_reactions.jl
```diff
@@ -1,40 +1,38 @@
...
+    vect = Vector{Float64}(length(sol.u))
+    for i = 1:length(sol.u)
+        vect[i] = sol.u[i][j]
+    end
+    return vect
+end
+
+ho_model1 = @reaction_network rn begin
+    (1.0,1.0), X + 2X ↔ Y + X
```
ChrisRackauckas: This. This is a great test!

test/runtests.jl
```diff
@@ -3,7 +3,11 @@ using Base.Test
 
 tic()
 @time @testset "Gillespie Tests" begin include("gillespie.jl") end
+@time @testset "Run Another Gillespie Tests" begin include("test_gillespie.jl") end
```
ChrisRackauckas: No, keep the file but don't run it. Maybe it should move to benchmarking repos later, but for not it's here.

ChrisRackauckas:
Getting very close. Remaining points:

- [ ] The reaction deprecation is gone again. Please bring that back to make sure the transition is smooth!
- [ ] Move the new documentation to the README
- [ ] Remove `test_gillespie.jl` from the `runtests.jl` file.
- [ ] Clean up `tmpMod` usage.
- [ ] Find out what to do about variable rate reactions.

The docs are beautiful BTW.

test/higher_order_reactions.jl
```diff
@@ -1,40 +1,38 @@
...
-  jump12.rate([1,1],nothing,0.0) == 0
-@test  jump12.rate([2,1],nothing,0.0) == k*2*1*1
-
-
-# Solve the dimerization problem
-prob = DiscreteProblem([1,0],(0.0,250.0))
-jump_prob = GillespieProblem(prob,Direct(),r2)
-sol = solve(jump_prob,Discrete())
-@test find( x-> x!=0 ,[u[2] for u in sol.u]) == []
+using DifferentialEquations
```
ChrisRackauckas: This doesn't need the full DiffEq. I think you just need `using DiffEqBiological, DiffEqJump, OrdinaryDiffEq` and you'll be fine. This is currently causing the test failure.

test/higher_order_reactions.jl
```diff
@@ -1,40 +1,38 @@
...
+prob_det2 = ODEProblem(ho_model2, u0, tspan)
+sol_det1 = solve(prob_det1)
+sol_det2 = solve(prob_det2)
+@test tmp_sol2vec(sol_det1,1) == tmp_sol2vec(sol_det2,1)
+@test tmp_sol2vec(sol_det1,2) == tmp_sol2vec(sol_det2,2)
+
+disc_prob =  DiscreteProblem([1000,1000],(0.,1000.))
+jump_prob1 = JumpProblem(disc_prob,Direct(),ho_model1)
+jump_prob2 = JumpProblem(disc_prob,Direct(),ho_model2)
+sol_jump1 = solve(jump_prob1,Discrete())
```
ChrisRackauckas: Uses of `Discrete` should change to `FunctionMap`. That will be throwing a deprecation warning soon, but [the docs](http://docs.juliadiffeq.org/latest/solvers/discrete_solve.html) show that it'll do exactly what `Discrete` used to do here (overlaps with Distributions.jl were the main reason for this change).

ChrisRackauckas:
Okay, constant vs variable reaction rates. The standard methods on Gillespie reactions assumes that the reaction rates are constant between each jump. I.e., between every reaction, the reaction rates like `2*X*Y` are constant. That is true for jump-only cases of course. When you stick an ODE on there, if `X` is a continuous quantity, that's not necessarily true anymore. In that case, it needs to be treated very differently. Long story short, if you write out the time to reaction as an ODE instead of as a result of an exponential random variable, it can handle continuous rates and thus the implementation needs to know when it should do this (to extend the user's ODE and do ...). 

But, it looks from this setup that it won't be mixing ODEs and jumps at all (at least for now). That means we'd be safe always using constant reaction rate routes, so the variable reaction rate stuff can just be deleted (and users who want it can still directly define jumps).

test/higher_order_reactions.jl
Gaussia: @korsbo mentioned Revise this very morning actually, good stuff.

test/higher_order_reactions.jl
Gaussia: Your probably right here.

Gaussia:
I understand(ish) but happy not to have to think closer about it right now.

ChrisRackauckas:
I added a commit that fixes up most of the tests. The only non-cosmetic change was that I made the higher order reaction test shorter in tspan. The reason is because there's a reaction just about every `1e-8`, making it take awhile to get to t=1000. Since it's tested against the open arrow it seems like it's correct, so that seems to just be the nature of the chosen problem meaning we just need to make the test a little easier.

The one last thing is that the `make_model_test.jl` still fails. The definition of `network2` throws an error, and `@test network4.jumps[2].rate(u,p,t) == network5.jumps[2].rate(u,p,t)` fails. Once those are fixed we're probably good to go.

Please make sure when you merge that you accept these values and do not overwrite them. I would recommend using a tool like GitKraken to help you out.

korsbo:
I'm creating a method for AbstractReactionNetwork in Latexify.jl. Will the type definition remain here, or will you move it to DiffEqBase?

ChrisRackauckas:
>I'm creating a method for AbstractReactionNetwork in Latexify.jl. Will the type definition remain here, or will you move it to DiffEqBase?

Let's move it to DiffEqBase.

korsbo:
Nice!
`latexify()` and `latexalign()` should now be able to handle AbstractReactionNetworks. Currently you need to checkout the master, but I'll tag a new release when this PR is merged.

ChrisRackauckas:
Alright. I just tagged DiffEqBase and upped the minimum version here. Let's figure out those last failing tests and we're good.

Gaussia:
>  The definition of `network2` throws an error
Simple, forgot a comma, this:
`((1.0,2,3.)(P1,X2,X1)), X2 ↔ (X2+X4,X1,X2)`
should become this:
`((1.0,2,3.), (P1,X2,X1)), X2 ↔ (X2+X4,X1,X2)`

> `@test network4.jumps[2].rate(u,p,t) == network5.jumps[2].rate(u,p,t)` fails
I have been unable to reproduce this issue on my machine so far.

ChrisRackauckas:
```julia
@test network4.jumps[2].rate(u,p,t) == network5.jumps[2].rate(u,p,t)
```

throws

```julia
Error During Test
  Test threw an exception of type UndefVarError
  Expression: network4.jumps[2].rate(u, p, t) == network5.jumps[2].rate(u, p, t)
  UndefVarError: p1 not defined
  Stacktrace:
   [1] (::##227#243)(::Array{Float64,1}, ::Array{Float64,1}, ::Float64) at C:\Users\Chris\.julia\v0.
6\DiffEqBiological\src\ReactionNetwork.jl:312
```

ChrisRackauckas:
```julia
julia> network4.jump_rate_expr
(:(X * Y), :(p1 * XY), :(2Z))
```

ChrisRackauckas:
Looks like it's because the `input_rate` isn't replaced like the `rate` is.

Gaussia:
Found it, I think. In the coordinate function 
`jumps = get_jumps(jump_rate_expr, jump_affect_expr,reactants,parameters)`
should be replaced with
`jumps = get_jumps(jump_rate_expr, jump_affect_expr,reactants,reactants)`

ChrisRackauckas:
Thanks. I'll finish this up.

Gaussia:
OK, I tried merging and putting it up. I hope I succeeded.

ChrisRackauckas:
No you did not. You rejected all of the changes. You need to make sure that when you merge you go through and pick the latest code bits. It's okay, I'll finish it up.

Gaussia:
I am sorry. I pulled it and got some changes which I thought I merged into the files, apparantly not. Sorry for the hassle.

ChrisRackauckas:
It's okay, it's far from the first time I've had to handle this. The changes are too extensive for me to want to lose them though, and this PR is pretty big so I'm taking the safe route of just finishing it up. Please don't get discouraged though: this is great work and using Git correctly takes time (and patience)!

ChrisRackauckas:
Tests pass locally. Merging when CI passes.

Gaussia:
Cheers.
(the cross country skiing just started in the Olympics so I am slightly distracted right now anyway)

korsbo:
I ran a Pkg.test before our commits @ChrisRackauckas. It ate all my memory and crashed my computer.

ChrisRackauckas:
It's that higher order reaction test

```julia
julia> length(sol_jump1)
27137579
```

That crashed Travis. I'm making it a big smaller once more. We really need tau leaping for that.

ChrisRackauckas:
Travis passed. Cheers!

Gaussia:
Cheers, thanks for all the help :)

PR #17, opened 2018-02-11
Added support for reaction order in the DSL

saolof:
Needs some testing.  The new parser code should hopefully compute the orders of the reaction properly.

More importantly, I added an extra field to the Reaction type which keeps track of the order of the reaction, and afaik this is necessary to properly describe the behaviour of the reaction network. As a bonus, this makes a backend to a symbolic ODE system much easier.


ChrisRackauckas:
Hey, sorry we are remaking the DSL in #15. One of the remaining questions though is about handling higher order reactions, so if you'd like to help there it would be much appreciated. Sorry for not getting to your issue earlier!

PR #34, opened 2018-03-22
Proper usage of the time variable, t

Gaussia:
- t now reserved for time.
- t will now be properly detected as time.


Gaussia:
update with new test is coming as well.

PR #35, opened 2018-03-22
added mass_action_reactions field to reaction_network

isaacsas:
added `mass_action_reaction` field to generated `reaction_network`. This field simply stores the `Vector{ReactionStruct}` generated in parsing the network, which contains additional stochiometric info.

### DIFF
--- a/src/maketype.jl
+++ b/src/maketype.jl
@@ -12,6 +12,7 @@ function maketype(name,
                   params = Symbol[],
                   pfuncs=Vector{Expr}(0),
                   symjac=Matrix{SymEngine.Basic}(0,0),
+                  reactions=Vector{ReactionStruct}()
                   )
 
     typeex = :(mutable struct $name <: AbstractReactionNetwork
@@ -27,6 +28,7 @@ function maketype(name,
         syms::Vector{Symbol}
         params::Vector{Symbol}
         symjac::Matrix{SymEngine.Basic}
+        reactions::Vector{ReactionStruct}
     end)
     # Make the default constructor
     constructorex = :($(name)(;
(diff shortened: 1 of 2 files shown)


korsbo:
Congratulations on your first PR, and thanks for your help! 

This looks good to me, but maybe we should rename the `mass_action_reaction field`. My reasoning is that the name is long and that all reactions are not mass action. Could we for example just use `reactions`?

isaacsas:
Ok, I renamed it as suggested. Thanks!

ChrisRackauckas:
Looks like this needs to be rebased to master.

isaacsas:
How do I go about doing that?

ChrisRackauckas:
I use GitKraken (there's a free version) to make Git easier, and then just drag and drop the master branch onto my current branch and it runs the merge tool (or you can ask for the rebase tool, but I guess start with merges since they are a little easier). You'll need to make sure you keep the "no Jacobian calculation" changes that were merged before, and then you'll be good.

From CLI git? Uhh, good luck and may the force be with you 👍 .

isaacsas:
This is fixed in the latest push, but I'm not sure how to get rid of the check failing... That line will be changed since I've added the `reactions` field.

ChrisRackauckas:
This is because the merge still needs to happen.

isaacsas:
Just wanted to check up on this. Is there more that needs to be addressed or is it ok now?

ChrisRackauckas:
Thanks for the bump. I didn't realize you rebased it. Good job!

isaacsas:
No worries, figured that was that case. Thanks for the help with the process!

PR #37, opened 2018-03-23
Added usage of VariableRateJumps in addition to ConstantRateJumps

Gaussia:
I think I know what I am doing, I have never really used the VariableRateJumps before. But after reading some I think I understood that those should be used when e.g. the rate depends on time.

Previously only ContantRateJumps were used (for SSA). Now it will be checked whenever a reaction has a reaction rate which depends on t. If that is the case a VariableRateJump will be used instead of a ConstantRateJump for that reaction.

If such a reaction is present (and hence a VariableRateJump) something like
```julia
prob = ODEProblem(model,u0,tspan)
jump_prob = JumpProblem(prob,Direct(),model)
sol = solve(jump_prob,Tsit5())
```
should be used, rather than:
```julia

prob = DiscreteProblem(u0,tspan)
jump_prob = JumpProblem(prob,Direct(),model)
sol = solve(jump_prob,FunctionMap())
```
(I am right here? The docs do not directly say that DiscreteProblems cannot be used, but I got strange behaviour when I tried it)

Since this seems like a likely error to do (if it actually is an error) I added a message for anyone trying to declare a jump problem using a DiscreteProblem and VariableRateJumps.

### DIFF
--- a/src/maketype.jl
+++ b/src/maketype.jl
@@ -20,7 +20,7 @@ function maketype(name,
         f_symfuncs::Matrix{SymEngine.Basic}
         g::Function
         g_func::Vector{Any}
-        jumps::Tuple{ConstantRateJump,Vararg{ConstantRateJump}}
+        jumps::Tuple{AbstractJump,Vararg{AbstractJump}}
         jump_rate_expr::Tuple{Any,Vararg{Any}}
         jump_affect_expr::Tuple{Vector{Expr},Vararg{Vector{Expr}}}
         p_matrix::Array{Float64,2}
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
>(I am right here? The docs do not directly say that DiscreteProblems cannot be used, but I got strange behaviour when I tried it)

Yes, it is correct. A VariableRateJump actually has to solve an ODE. We should add an error in solve.jl of DiffEqJump.jl to make this safer.

ChrisRackauckas:
>Since this seems like a likely error to do (if it actually is an error) I added a message for anyone trying to declare a jump problem using a DiscreteProblem and VariableRateJumps.

That works too.

Gaussia:
Good. When the error/warning is added at a higher level we will remove it from here.

PR #38, opened 2018-03-26
Add inbounds

TorkelE:
Added `@inbounds` to the `f`, `g` function as well as to the `rate` and `affect` functions used by jump problems.

### DIFF
--- a/src/reaction_network.jl
+++ b/src/reaction_network.jl
@@ -82,17 +82,16 @@ function coordinate(name, ex::Expr, p, scale_noise)
     syms = collect(keys(reactants))
     params = collect(keys(parameters))
     (in(:t,union(syms,params))) && error("t is reserved for the time variable and may neither be used as a reactant nor a parameter")
-
+    println("HERE1")
     f_expr = get_f(reactions, reactants)
     f = make_func(f_expr, reactants, parameters)
-
-    g_expr = get_g(reactions, reactants, scale_noise)
+    println("HERE2")    g_expr = get_g(reactions, reactants, scale_noise)
     g = make_func(g_expr, reactants, parameters)
     p_matrix = zeros(length(reactants), length(reactions))
-
+    println("HERE3")
     (jump_rate_expr, jump_affect_expr) = get_jump_expr(reactions, reactants)
     jumps = get_jumps(jump_rate_expr, jump_affect_expr,reactants,parameters)
-
+    println("HERE4")
     f_rhs = [element.args[2] for element in f_expr]
     symjac = Expr(:quote, calculate_jac(f_rhs, syms))
     f_symfuncs = hcat([SymEngine.Basic(f) for f in f_rhs])
--- a/src/reaction_network.jl
+++ b/src/reaction_network.jl
@@ -82,16 +82,17 @@ function coordinate(name, ex::Expr, p, scale_noise)
     syms = collect(keys(reactants))
     params = collect(keys(parameters))
     (in(:t,union(syms,params))) && error("t is reserved for the time variable and may neither be used as a reactant nor a parameter")
-    println("HERE1")
+
     f_expr = get_f(reactions, reactants)
     f = make_func(f_expr, reactants, parameters)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
Looks good, I just didn't know the lines to do it at. Thanks. 

This needs to be rebased onto the current master.

Gaussia:
Yeah, seems I forgot to base it of the latest version. It is nothing complicated to fix though. Did something, hope that was it.

ChrisRackauckas:
```julia
A = [1,2,3]
@inbounds A[1]
```

returns nothing. That's really ugh... The workaround is:

```julia
A = [1,2,3]

function f(A)
  @inbounds x = A[1]
   x
end
f(A)
```

src/reaction_network.jl
```diff
@@ -281,7 +281,7 @@ function make_func(func_expr::Vector{Expr},reactants::OrderedDict{Symbol,Int},pa
     system = Expr(:block)
     for func_line in deepcopy(func_expr)
         tmp_line = recursive_replace!(func_line, (reactants,:internal_var___u), (parameters, :internal_var___p))
-        push!(system.args, :(@inbounds $tmp_line))
+        push!(system.args, :(inbounds_func($tmp_line)))
```
ChrisRackauckas: this doesn't seem to make sense. If you already are doing the `tmp` line, then you already indexed and thus already needed to `@inbounds`?

Gaussia: Correct me if I am wrong (well somewhere I am obviously wrong of you ask Travis). Here `tmp_line` is an expression which have not actually been executed yet. `tmp_line` might be something like
```julia
internal_var___du[1] = internal_var___p[1] - internal_var___p[2]*internal_var___u[1]
```
Next the following expression
```julia
inbounds_func(internal_var___du[1] = internal_var___p[1] - internal_var___p[2]*internal_var___u[1])
```
is pushed on an expression array. But that actual line is not execute until the last line of the entire `@reaction_network` macro.

I have not been able to determine what the error is here on git, but I can run stuff on my machine and get a `UndefVarError: inbounds_func not defined` error when using the reaction network as input to a solver. This suggests that the program is trying to execute the ` inbounds_fun` after the model has been made.

Could be solved by exporting the ` inbounds_fun` function. However, it I have understood anything then the `@inbounds` stuff is supposed to be a direction to the compiler? Then should it actually have to be called in the solver?

ChrisRackauckas: This would execute the argument expression `tmp_line` and then execute the function, but when `tmp_line` is executed that's when it will need `@inbounds`. It won't put that expression into the function. But I found out last night that the `@inbounds` not returning bug was fixed on v0.7, so we might want to just pull back a commit and then postpone for v0.7.

Gaussia: Understood.

I agree, if the bug is fixed in a later version then we wait for that. Is there a simple way to remove the last commit from the chain. Or should I simply redo and add another commit?

ChrisRackauckas: You can give a revert commit or branch back and force push. A revert commit is probably easier (and safer)

Gaussia:
With the usual caveat of me not knowing what I am doing I tried to revert the earlier commits, then we wait and see for the eventual appearance of 0.7, which might take a while.

If things are messed up with merges and etc. after that it would probably take me less than 5min to scrap this pull request, add the `@inbounds` again make another pull request. But if we are lucky maybe we will not have to deal with that.

(also, if this is attached to the `ModelingToolkit` package there is a chance that that might do the `@inbounds` part?)

ChrisRackauckas:
>(also, if this is attached to the ModelingToolkit package there is a chance that that might do the @inbounds part?)

Yes it would.

isaacsas:
@TorkelE Any chance you might be willing to resurrect this now? It probably makes sense to just make a new PR and close this one (rather than try to rebase this one).

TorkelE:
Yes, no sense keeping this one around for ever, only vaguely remember what it was all about anyway. I will close this one and open a corresponding issue in the meantime.

PR #40, opened 2018-03-29
More diverse input options

TorkelE:
Till allows you to
- Only input a noise scaling parameter.
- Input Noise scaling and a custom type in any order.

Also:
- The noise scaling parameter is now required to be among the parameters (where it earlier simply was optional.

E.g. this becomes a possible input:
```julia
equi_model = @reaction_network η typeR begin
    1, 0 --> X
    k, X + Y --> Z
    Z, Z --> X
end η k
```
```julia
equi_model = @reaction_network η begin
    1, 0 --> X
    2, X + Y --> Z
    Z, Z --> X
end η
```

### DIFF
--- a/src/reaction_network.jl
+++ b/src/reaction_network.jl
@@ -55,8 +55,10 @@ end
 
 #Macro to create a reaction network model. Multiple dispatch is used to allow for SDE noise scalling.
 macro reaction_network(sym1, sym2, ex::Expr, p...)
+    println(typeof(sym1)," ",typeof(sym2), " ",typeof(p))
+    println(sym1," ",sym2, " ",p)
     in(sym1,p) $ in(sym2,p) || error("Two initial options have been given (designating type and noise scaling). However, exactly one of these must also be given as a parameter")
-    in(sym1,p) && coordinate(sym2,ex,p,sym1) : coordinate(sym1,ex,p,sym2)
+    in(sym1,p) ? coordinate(sym2,ex,p,sym1) : coordinate(sym1,ex,p,sym2)
 end
 
 #Used to give a warning if someone uses the old macro.
--- a/src/reaction_network.jl
+++ b/src/reaction_network.jl
@@ -54,7 +54,7 @@ macro reaction_network(sym, ex::Expr, p...)
 end
 
 #Macro to create a reaction network model. Multiple dispatch is used to allow for SDE noise scalling.
-macro reaction_network(name, scale_noise, ex::Expr, p...)
+macro reaction_network(sym1, sym2, ex::Expr, p...)
     in(sym1,p) $ in(sym2,p) || error("Two initial options have been given (designating type and noise scaling). However, exactly one of these must also be given as a parameter")
     in(sym1,p) && coordinate(sym2,ex,p,sym1) : coordinate(sym1,ex,p,sym2)
 end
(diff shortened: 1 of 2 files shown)


korsbo:
> The noise scaling parameter is now required to be among the parameters (where it earlier simply was optional.

This feels iffy to me. It seems that you now need to specify the same thing twice. Also, you are trying to ensure that you do not need to specify a type name if you do not need one, but you are forcing the user to specify a noise parameter, even if they do not need it (or am I wrong?).

ChrisRackauckas:
>Input Noise scaling and a custom type in any order.

I'm not sure this is a great idea. Sometimes some rigidity makes code easier to read. Keyword arguments are fine because the name is written there. Here it just feels random.

>The noise scaling parameter is now required to be among the parameters (where it earlier simply was optional.

This is a breaking change. Is it really a good one? I'm skeptical that we're gaining much for such a loss.

Gaussia:
Not necessarily. I am mainly looking for a way to input noise scaling without having to designate a customize type. e.g. if I write
´´´julia
equi_model = @reaction_network η begin
    1, 0 --> X
    2, X + Y --> Z
    Z, Z --> X
end
´´´
the DSL cannot know whenever η is a type or a parameter fro noise scaling. Another thing which is funny with the current system is that the actual symbol used for noise scaling doesn't really matter (unless you want to print the equations in Latexify.

Anyone have any alternative suggestions. I do think that there should be some way of inputting the noise calling without also designating a type.

(@korsbo No, you will not need to specify a noise parameter)

isaacsas:
@TorkelE Don't know if you are still interested in this functionality, but it would be easy to modify `addsdes!` to take the `noise_scaling` as an optional keyword parameter for a `min_reaction_network`.

TorkelE:
Allowing for noise scaling to be designated in addsdes! should be a useful thing (otherwise one is unable to actually scale the noise when using minimal reaction networks?).

isaacsas:
The current functionality of the macro should work with `min_reaction_network`, i.e. if you include a noise scaling when calling the macro, then the scaling should propagate through and get used internally when one calls `addsdes!`. What I was suggesting is that we could also allow an option so that one doesn't need to specify the scaling when calling `min_reaction_network`, and instead just specifies it when calling `addsdes!`. I've always found it a bit out of place that the scaling goes at the start of the block with the type parameter. Why should that one parameter be distinguished and included in the macro, but not other parameters?

TorkelE:
The reason is that it kinda acts as a flag, telling the macro that noise scaling is desired. The alternative was to simply allow the user to feed a parameter vector which is one step longer into the SDEProblem. However it was decided that this might be problematic if someone not knowing about noise scaling accidentally had that vector be one number to long. Similar with simply adding it as a parameter at the end, which is then unused in the actual expression.

Since these two options (designate a custom type and noise scaling) currently only are the only ones available to the macro it is simple to just input them like this, rather than designating specific function for them. If we ever start coming up with loads of options we might need to rethink, but currently that feels a bit away.

TorkelE:
This one has been dormant for a while and is partially implemented already. What is left of it (adding noise scaling without specifying a type) is not very important and I agree that this is not the way to do it. Still might have to think of good way to input more options to the `@reaction_network` macro, but this is not the place for that.

PR #47, opened 2018-04-18
Function overload

### LINKED ISSUE #18: Variable names to Plots.jl
With ParameterizedFunctions.jl, the variable names are fed to the legends generated by Plots.jl, it would be nice if DiffEqBiological could do the same.

korsbo:
Overload the reaction network type with the generated deterministic function  

```julia
(f::myReactionTypeName)(du, u, p, t) = f.f(du,u,p,t)
```

This mirrors what has been done in ParameterizedFunctions.jl and allows the full reaction network object to be passed along in the problems and solutions. 

This automatically allows the plot recipe to use the variable names. Closes #18.

I also updated the DiffEqJump requirement since this was causing an issue for me when testing.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.6
-DiffEqJump 1.0.0
+DiffEqJump 4.2.0
 DiffEqBase 3.4.0
 Compat 0.17.0
 DataStructures 0.4.6
--- a/src/DiffEqBiological.jl
+++ b/src/DiffEqBiological.jl
@@ -17,7 +17,7 @@ include("maketype.jl")
 include("problem.jl")
 
 export @reaction_network, @reaction_func
-export ODEProblem, SDEProblem, JumpProblem, SteadyStateProblem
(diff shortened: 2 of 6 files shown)


korsbo:
I see that you sorted out the problem in the SteadyStateDiffEq REQUIRE @ChrisRackauckas, cheers!

korsbo:
One sec and I'll add a line for testing too.

ChrisRackauckas:
It's been a two days of just fixing test problems so that way other contributors can see green again 👍 . Bunch of upstream changes but everything should be good again.

korsbo:
We might also consider whether to let AbstractReactionNetwork inherit from Function. 
This is what AbstractParameterizedFunction does, and I see no point in having them do different things.

ChrisRackauckas:
>We might also consider whether to let AbstractReactionNetwork inherit from Function. This is what AbstractParameterizedFunction does, and I see no point in having them do different things.

Yeah it's a good idea.

PR #52, opened 2018-07-13
Add jacobian

Gaussia:
Re adds the code which calculates a symbolic Jacobian. Also ensured that it can deal with reactant names such as `I` (the previous problem).


Gaussia:
Sounds good (or maybe "under control" is a better term...). I trust you know how to deal with the technicalities of the 0.7 and later 1.0 updates, just say if there is anything I can help with.

PR #57, opened 2018-07-31
Depricate default

Gaussia:
Currently declaring a reaction network like
```julia
model = @reaction_network begin
    1.0, X --> Y
end
```
gives a deprecation warning for the old DSL. Eventually this should be changed to create a reaction network with some default type. This PR does just that and can be merged when we think the time is right.

### DIFF
--- a/src/reaction_network.jl
+++ b/src/reaction_network.jl
@@ -59,9 +59,9 @@ macro reaction_network(name, scale_noise, ex::Expr, p...)
     coordinate(name, MacroTools.striplines(ex), p, scale_noise)
 end
 
-#Used to give a warning if someone uses the old macro.
-macro reaction_network(ex::Expr)
-    error("The Reaction Network DSL have been deprecated in favor for a new one. With only slight modification old code can be made to work with the new DSL. In addition the new one provides lots of additional functionality. Please view the documentation for more information.")
+#If no type name is given, creates a network with a default name.
+macro reaction_network(ex::Expr, p...)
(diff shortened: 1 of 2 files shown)


src/reaction_network.jl
```diff
@@ -59,9 +59,9 @@ macro reaction_network(name, scale_noise, ex::Expr, p...)
     coordinate(name, MacroTools.striplines(ex), p, scale_noise)
 end
 
-#Used to give a warning if someone uses the old macro.
-macro reaction_network(ex::Expr)
-    error("The Reaction Network DSL have been deprecated in favor for a new one. With only slight modification old code can be made to work with the new DSL. In addition the new one provides lots of additional functionality. Please view the documentation for more information.")
+#If no type name is given, creates a network with a default name.
+macro reaction_network(ex::Expr, p...)
+    coordinate(:reaction_network, MacroTools.striplines(ex), p, :no___noise___scaling)
```
ChrisRackauckas: The default name should be a `gensym` since types cannot be recreated in the same session

ChrisRackauckas:
The time is now. We need to get everything updated for v0.7 though. For some reason it

ChrisRackauckas:
It looks like this cases a test failure.

Gaussia:
Cheers. As so often I am loosing track of what is happening, is there anything I can do?

ChrisRackauckas:
Looks like it's fixed by adding an import. Cheers!

Gaussia:
Splendid

PR #63, opened 2018-10-29
pass SSAs species to dependent reaction maps if needed

isaacsas:
This should only be merged after my DiffEqJump RSSA PR, as it relies on a function to query if SSAs need dependency maps from jumps to species (and vice-versa). It won't pass integration checks until it can use that PR.

Some SSAs I'm working on require different dependency info (namely species to dependent reaction maps). This sets up DiffEqBiological to pass the required info.

### DIFF
--- a/src/massaction_jump_utils.jl
+++ b/src/massaction_jump_utils.jl
@@ -1,10 +1,10 @@
-# collection of utility functions for getting mass action jumps out of 
+# collection of utility functions for getting mass action jumps out of
 # DiffEqBiological reaction network structures
 
 # return a map from species symbol to species index
 function species_to_indices(network)
     specs = network.syms
-    Dict( specs[i] => i for i in eachindex(specs) )    
+    Dict( specs[i] => i for i in eachindex(specs) )
 end
 
 # return a map from reaction param symbol to rate index
@@ -13,7 +13,7 @@ function rate_to_indices(network)
     Dict( rates[i] => i for i in eachindex(rates) )
 end
 
-# get substrate stoichiometry for a reaction 
+# get substrate stoichiometry for a reaction
 function get_substrate_stoich(rs, specmap)
     reactant_stoich = Vector{Pair{Int,Int}}()
     for substrate in rs.substrates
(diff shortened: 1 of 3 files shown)


src/massaction_jump_utils.jl
```diff
@@ -134,26 +148,17 @@ function spec_to_dep_jumps_map(rn, specmap, rxidxs_to_jidxs)
...
-    numrxs          = length(rn.reactions)
-    numspec         = length(specmap)
-    num_majumps     = get_num_majumps(jset.massaction_jump)
-
-    # map reaction indices to jump indices
-    rxidxs_to_jidxs = rxidxs_to_jidxs_map(rn, num_majumps)
-
-    # map from species to jumps that depend on it
-    spec_to_dep_jumps = spec_to_dep_jumps_map(rn, specmap, rxidxs_to_jidxs)
+function depgraph_from_network(rn, specmap, jset, rxidxs_to_jidxs, spec_to_dep_jumps)
```
ChrisRackauckas: why add the new arguments?

src/problem.jl
```diff
@@ -17,14 +17,30 @@ function DiffEqJump.JumpProblem(prob,aggregator,rn::DiffEqBase.AbstractReactionN
     # get a JumpSet of the possible jumps
     jset = network_to_jumpset(rn, spec_to_idx, param_to_idx, prob.p)
 
-    # # construct dependency graph if needed by aggregator
+    # construct map from species index to indices of reactions that depend on it
+    if needs_vartojumps_map(aggregator) || needs_depgraph(aggregator)
```
ChrisRackauckas: would it be a good idea to precompute some of this? This is re-computed each time solve is called. I don't know if that's a big deal though.

isaacsas: Just to clarify; this is called each time a `JumpProblem` is made, not when `solve` is called right? We definitely don't want to redo this every time `solve` is called; it could be very expensive on larger networks. For `JumpProblem`s it would probably make sense to have optional kwargs that pass in the needed maps / dependency graphs, and only generate them if they aren't passed in, but I imagine only more advanced users with make use of that since one has to understand what is required for each specific SSA and how to get it from these routines. I'll can update to allow that possibility.

isaacsas: I'm going to leave this for a future PR. We might want at some point to make a struct to encapsulate various dependency info, rather than passing around lots of different things. In that case the user could sent that into the `JumpProblem` if they already have it.

PR #65, opened 2018-11-08
- Added a 1* in jacobian entries to ensure expression type.

TorkelE:
Should fix bug.


isaacsas:
Great, thanks! Looks like it is now passing tests so I'll merge.

PR #67, opened 2018-11-09
- Add default functions for hill and mm repressors.

TorkelE:
Added hillr and mmr functions so hill repressors and mm repressors can be written in easier
(e.g. v*(K^n)/(K^n+X^n))

### DIFF
--- a/test/func_test.jl
+++ b/test/func_test.jl
@@ -6,21 +6,25 @@ using DiffEqBiological, StochasticDiffEq, Statistics
 
 network1 = @reaction_network rn begin
     hill(X,4,3,2), X + Y --> Z1
-    3X^2+1, X + Y --> Z2
-    exp(Y), X + Y --> Z3
-    hillR(X,2.0,3.0,2), X + Y --> Z4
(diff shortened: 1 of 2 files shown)


test/func_test.jl
```diff
@@ -8,11 +8,13 @@ network1 = @reaction_network rn begin
     hill(X,4,3,2), X + Y --> Z1
     3X^2+1, X + Y --> Z2
     exp(Y), X + Y --> Z3
+    hillR(X,2.0,3.0,2), X + Y --> Z4
 end
 network2 = @reaction_network rn begin
     new_hill(X,4,3,2), X + Y --> Z1
     new_poly(X), X + Y --> Z2
     new_exp(Y), X + Y --> Z3
+    2.0*(3.0^2)/(3.0^2+X^2), X + Y --> Z4
```
ChrisRackauckas: no test for MMR?

TorkelE: Good point. On the same not we seem to lack any test for the normal michaelis menten function as well. Away a few days but will add them both in when I get home.

PR #69, opened 2018-12-31
Allow mass action rate consts involving parameter expressions

### LINKED ISSUE #64: Unable to create jump problem with reaction rate containg parameter multiplication
This code:
```julia
using DifferentialEquations
rn = @reaction_network begin
    p1*p2, X --> Y
end p1 p2
p = [1.,2.]
dProb = DiscreteProblem([10,10],(0,10.),p)
jProb = JumpProblem(dProb,Direct(),rn)
```
generates the error 
```
UndefVarError: p1 not defined
in top-level scope at base/none
in JumpProblem at N4fWp/src/problem.jl:7
in #JumpProblem#43 at N4fWp/src/problem.jl:18
in network_to_jumpset at N4fWp/src/massaction_jump_utils.jl:84
in make_majump at N4fWp/src/massaction_jump_utils.jl:65
in eval at N4fWp/src/DiffEqBiological.jl:3 
in eval at base/boot.jl:319 
in top-level scope at base/none
```
It seems to have to do with multiplying p1 and p2 in the reaction rate and creating the jump problem.
e.g. these pieces of code works fine:
```julia
using DifferentialEquations
rn = @reaction_network begin
    p1*p2, X --> Y
end p1 p2
p = [1.,2.]
ODEProblem(rn,[10.,10.],(0.,10.),p)
```
and 
```julia
rn = @reaction_network begin
    p1, X --> Y
end p1 p2
p = [1.,2.]
dProb = DiscreteProblem([10,10],(0,10.),p)
jProb = JumpProblem(dProb,Direct(),rn)
```

spcornelius:
Fixes #64 

Previously, within the rxn network DSL, legitimate mass action reactions whose rate constants were non-trivial expressions of the system parameters like

`2.0 * r, 0 --> X`
`p1*p2, X --> Y`

...would fail to parse upon instantiation of a `JumpProblem`/`SDEProblem`/etc. Indeed, it seemed that parameters could only be used within the rate constants if either i) they appeared standalone

`r, X --> 2X`

or ii) were adorned by a species--e.g.

`X^2/(X^2 + S^2), X ---> 2*X`

...but in that case, the reaction simply ceases to be regarded as mass action.

### DIFF
--- a/src/massaction_jump_utils.jl
+++ b/src/massaction_jump_utils.jl
@@ -52,24 +52,18 @@ function get_net_stoich(rs, specmap)
 end
 
 # given a ReactionStruct and a species map construct a MassActionJump
-function make_majump(rs, specmap, ratemap, params)
+function make_majump(rs, specmap, ratemap, params, param_context)
     reactant_stoich = get_substrate_stoich(rs, specmap)
     net_stoich      = get_net_stoich(rs, specmap)
     if isempty(net_stoich)
         error("Empty net stoichiometry vectors for mass action reactions are not allowed.")
     end
 
-    # populate dummy context with params as local variables
-    Dummy = Module()
-    for (param, index) in ratemap
-        Base.eval(Dummy, :($param = $(params[index])))
-    end
-
     if typeof(rs.rate_org) == Symbol
         rateconst = params[ratemap[rs.rate_org]]
     elseif typeof(rs.rate_org) == Expr
-        # eval in Dummy, in case Expr depends on params
-        rateconst = Base.eval(Dummy, rs.rate_org)
+        # eval in param_context, in case Expr depends on params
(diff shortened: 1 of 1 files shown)


src/massaction_jump_utils.jl
```diff
@@ -59,10 +59,17 @@ function make_majump(rs, specmap, ratemap, params)
         error("Empty net stoichiometry vectors for mass action reactions are not allowed.")
     end
 
+    # populate dummy context with params as local variables
```
isaacsas: Thanks for the help! Since populating the dummy module only needs to be done once, would it not make sense to move lines 62-66 to a higher level in the call tree? (Even just to `network_to_jumpset`). You could then pass the module into `make_majump`. Also, in this case maybe use a more descriptive name for the parameter context module input variable.

PR #87, opened 2019-02-05
abs the sqrt in the CLE

ChrisRackauckas:
This approximation is from Higham

https://strathprints.strath.ac.uk/13549/1/siamssa.pdf

If a value ever goes negative it's unphysical, but this can make it so you can finish the step. Then rootfinding can occur to approximate the zero. This would fix the domain errors that @TorkelE had and allow domain callbacks to work.

### DIFF
--- a/src/reaction_network.jl
+++ b/src/reaction_network.jl
@@ -341,7 +341,7 @@ function get_g(reactions::Vector{ReactionStruct}, reactants::OrderedDict{Symbol,
     g = Vector{Expr}(undef,length(reactions)*length(reactants))
     idx = 0
     for reactant in keys(reactants), i = 1:length(reactions)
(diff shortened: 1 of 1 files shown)


isaacsas:
Non-negativity of the CLE (and tau-leaping) is very much an open area of research, and an issue with almost all methods. Anderson's recent paper on the constrained CLE has a nice summary of the literature on this issue for the CLE, see [http://www.math.ucsd.edu/~williams/biochem/constrainedLE.pdf](http://www.math.ucsd.edu/~williams/biochem/constrainedLE.pdf).

Would it make sense to also have the `SDEProblem` generated from the `reaction_network` by default include the appropriate domain callback?

ChrisRackauckas:
>Would it make sense to also have the `SDEProblem` generated from the `reaction_network` by default include the appropriate domain callback?

Hmm, maybe? I wonder how performance and robustness changes if we always include `PositiveDomain()` by default. I'm going to punt on that for now and open an issue to investigate it more.

PR #90, opened 2019-02-08
add DiscreteProblem overload for reaction_networks

isaacsas:
Added an overload so that we can do:
```
dprob = DiscreteProblem(rn, u0, tspan, p)
```
This allows plots of jump simulations to actually get the species syms in the legend. For example, the repressilator as a jump process:
syms

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 1.0
 DiffEqJump 5.6.0
-DiffEqBase 3.11.0
+DiffEqBase 5.1.0
 Compat 0.17.0
 DataStructures 0.8.1
 Reexport 0.1.0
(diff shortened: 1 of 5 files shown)


isaacsas:
This relies on DiffEqBase [187](https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/187), so probably won’t pass tests till a new release of it is tagged.

isaacsas:
Looks like the registry still hasn't picked up the DiffEqBase 5.1... I'll bump this once it updates so tests re-run...

isaacsas:
It needs the `DiscreteFunction` that takes `syms`, doesn't that require 5.1 or greater? I thought that was why Travis was failing (at least why it was failing prior to the whole registry issue today).

ChrisRackauckas:
Oh yes it does need the bound. My bad.

PR #92, opened 2019-02-13
Add jac

TorkelE:
Previously only the symbolic Jacobian was calculated. Added so that whenever the symbolic Jacobian is calculated a Jacobian function is also calculated (of the type `jac = (J,u,p,t) -> (J[1,1] = t; J[2,2] = t; J)`).

I looked at `ParameterizedFunctions` to get some inspirations and it also had a `paramjac`, which would be doable to calculate (but is it actually used anywhere? For now I have not added it). There is also `jac_prototype` field. However, that one always seem to contain `nothing`? For now I do create a `jac_prototype` field, but it will always remain `nothing`.

I am not sure whenever this is useful to improve some of the solvers, and if some extra step is required for the Jacobian to be used there. However, it is needed for the stability analysis of fixed points so it felt good to include it for at least that reason.


isaacsas:
If you add the Jacobian to the ODEFunction that `addodes!` makes, see [here](http://docs.juliadiffeq.org/latest/features/performance_overloads.html#Declaring-Explicit-Jacobians-for-ODEs-1), it will then get passed along to ODE solvers. The only issue is that I think that means solvers that use Jacobians will then always use this Jacobian (as opposed to AD or finite differences)? I'm not sure if there would be any reason not to do this. @ChrisRackauckas is there any reason not to pass along the Jacobian when it gets constructed?

ChrisRackauckas:
>I am not sure whenever this is useful to improve some of the solvers, and if some extra step is required for the Jacobian to be used there. However, it is needed for the stability analysis of fixed points so it felt good to include it for at least that reason.

You should always do it if you have it. It will be more optimized.

>but is it actually used anywhere?

Sensitivity equations

>There is also `jac_prototype` field. However, that one always seem to contain `nothing`? For now I do create a `jac_prototype` field, but it will always remain `nothing`.

Sparsity pattern. `nothing` means dense.

isaacsas:
@TorkelE If adding the param Jacobian is simple, feel free to do so. Otherwise let's open an issue so we keep in mind it would be good to calculate too.

TorkelE:
It was quick work to take the code for generating the normal jacobian and make a parameter version of it, so now we have that one. The parameter Jacobian is calculated with size nxm (n = number of reactants, m=number of parameters). and is of the form (pJ, u, p, t) -> J[1,1] = ...

I also added the jacobian (and `jac_prototype=nothing`) to the ode/sde functions which is are passed into the problem.

TorkelE:
I am a bit unsure why appveyor is unhappy, unless anyone have any direct idea I will investigate over the weekend.

ChrisRackauckas:
Because of merge conflicts

src/maketype.jl
```diff
@@ -125,25 +137,27 @@ end
...
 
-    (f_expr, f, f_rhs, symjac, f_symfuncs) = genode_exprs(reactions, syms_to_ints, params_to_ints, syms; kwargs...)
+    (f_expr, f, f_rhs, symjac, jac, paramjac, f_symfuncs) = genode_exprs(reactions, syms_to_ints, params_to_ints, syms; kwargs...)
     rn.f          = eval(f)
     rn.f_func     = f_rhs
+    rn.jac        = eval(jac)
+    rn.paramjac   = eval(paramjac)
     rn.symjac     = eval(symjac)
     rn.f_symfuncs = f_symfuncs
     rn.odefun     = ODEFunction(rn.f; syms=rn.syms)
```
isaacsas: Can you add the jacobians here too so we have parallel behavior of the min network with `addodes!`?

Similarly, add them where the SDEFunction is made in `addsdes!`?

isaacsas:
Your branch is behind master, so you need to merge it in or rebase.

TorkelE:
Think I managed to fix the merge. Appveyor still seem unhappy, but only on the nightly julia_version, version 1 seems ok.

My computer (Ubuntu 16) also have troubles running the discreteproblem_test.jl file. I get errors on two lines, 
```julia
execute_test(prob, u0, tf, rates, rs, Nsims, expected_avg, 1, "Nonlinear rx test (syms)")
```
(line 65)
And
```julia
execute_test(prob, u0, tf, rates, rs, Nsims, expected_avg, 1, "Nonlinear rx test (syms)")
```
(line 50)

I get an error that there are to many parameters in the line (16) from the problem.jl file
```julia
df = DiscreteFunction{true,typeof(f),Nothing,typeof(rn.syms)}(f,nothing,rn.syms)
```
However, Travis does not seem to have an issue with this.

isaacsas:
Have you updated DiffEqBase lately? That definition was only added there ~10 days ago. 

Are you finished now? If so I'll check if it runs through ok locally for me with the latest DiffEqBase, and then merge if it goes ok.

TorkelE:
updating DiffEqBase seems like a good idea.  Yes, unless there are bugs to be fixed this should be done.

isaacsas:
Goes through fine for me, so I'll merge.

PR #96, opened 2019-02-22
DiscreteFunction creation fix

isaacsas:
DiffEqBiological was using the exact parametric constructor form, which has now changed due to the addition of DEStats. This switches to using the more general form with kwargs.

### DIFF
--- a/src/problem.jl
+++ b/src/problem.jl
@@ -12,7 +12,7 @@ end
 
 ### DiscreteProblem, passing through syms
 function DiffEqBase.DiscreteProblem(rn::DiffEqBase.AbstractReactionNetwork, u0, tspan::Tuple, p=nothing; kwargs...)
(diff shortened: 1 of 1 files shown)


isaacsas:
Except now syms have disappeared from legends again...

ChrisRackauckas:
Interesting, wouldn't this give true? https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/diffeqfunction.jl#L494

isaacsas:
```
sir_model = @reaction_network SIR begin
           c1, s + i --> 2i
           c2, i --> r
       end c1 c2
df = DiscreteFunction(DiffEqBase.DISCRETE_INPLACE_DEFAULT,syms=sir_model.syms)
```

It's the `DISCRETE_INPLACE_DEFAULT` that messes it up. Using `sir_model.f` instead causes everything to work ok. Probably why I used the explicit parametric constructor before...

PR #97, opened 2019-02-23
Equilibrate

TorkelE:
This contains various changes for finding steady states and plotting bifurcation diagrams. I will try to list everything that is going on:

**General Changes**
- I have created a file "test_networks.jl" which creates 3 sets of 10 different reaction networks to be used in tests (currently only used in the steady states tests, but could potentially be used for other stuff). The first set is just standard reaction networks (of these there exists a minimal reaction network version), the second set contains reaction networks with hill functions (so were some parameters is required to be integer). The final set contains reaction networks were some reactants are neither created nor destroyed (like `X1 + X2 ↔ X3`). This will have an infinite amount of potential steady states, which can become finite by introducing additional equations or initial conditions.
- The macro, instead of returning the reaction network as last argument returns the reaction network as input to the `manage_reaction_network!()` function (which then outputs the reaction network. This have no real changes, but allows to make some post-creation manipulations of the reaction network without dealing with meta programming. This possibility is currently only used by the equilibration functionality. The function is currently not called by minimum reaction network (because they currently does not need any such processing). However, since the `addodes!`, `addsdes!`... functions uses eval these can not use the `manage_reaction_network!()` function, which have to be used separately. As a result of this, the way to get equilibrium features to you minimal reaction network is:
```julia
rn = @min_reaction_network begin
....
end p1 p2 p3
p = [1.,2.5,4.5]
addequi!(rn)
manage_reaction_network!(rn)
steady_states(rn,p)
```
- I have also added a separate file, `equilibrate_utils.jl`, containing all the steady state stuff.

**New required packages**
`DiffEqBiological` now requires `HomotopyContinuation` (does the actual solving), `Plots` (just by the functions which plots bifurcation diagrams), and `DynamicPolynomials` (since HC deals in polynomials). In addition `DynamicPolynomials` is now `@reexport`'ed by `DiffEqBiological` (I tried to solve stuff without this, but failed. Might still be possible).

**Homotopy Continuation**
Currently the only solver for finding steady states and doing bifurcations I have implemented is homotopy continuations. Basically it works very well for solving multivariate polynomial systems of n variables and _n_ and _n_ equations. So as long as the reaction rates only depends on other reactants as michaelis-menten functions and hill functions it should be fine (or some other rational polynomial expression). I believe this covers a very large portion of biochemical reaction network models.

While homotopy continuation is not really designed for the purpose it can be used to track the steady states solutions from one system to another, essentially creating bifurcation diagrams. Some times it can have problems tracking solutions through bifurcations (I had some problem with some saddle-node bifurcations, but sometimes these works as well). Right now I have a simple heuristics for this (checking if any solutions are lost while tracking from p1 to p2. If that is the case I track from p2 to p1 as well to get the missing paths). Right now this has worked for the system I have tried with. If we get problem with new systems it should be easy to introduce further heuristics (It is quite easy to realise when something funny is going on, probably a bifurcation. Just move the tracker forward a little bit, past the bifurcation, and continue).

I also looked at other methods for tracking bifurcations in Julia, but the docs were kinda confusing and I knew how to get it done with homotopy continuation so went with that. In the future we might want to switch this though.

Awesome work by the guys at https://github.com/JuliaHomotopyContinuation/HomotopyContinuation.jl for actually implementing this algorithm. We#ll have to add their paper to the reference list.

**How the update work**

- To find steady states using HC we need a polynomial. Ideally we would want to create this polynomial when the reaction network is first created. However, parameters are represented as variables in the polynomial until they are set. This means that if one have a reaction rate with a reactant concentration to the power of a parameter (typically a hill equation) a polynomial cannot be created initially since it would have a variable to a power of a variable. Instead the macro creates a function, `make_polynomial` which takes parameters as input and creates the polynomial (this function is primarily accessed by another function). However, if no such parameters are present in the network a polynomial will be created and stored in the `equilibratium_polynomial` field.
- If some parameters are exponents which needs to be set afterwards this is done with the `fix_parameter` function, e.g.
```julia
rn = @reaction_network begin
(Y^n1,X^n1), X ↔ Y
end n1 n2
fix_parameter(rn,n1=2,n2=3)
steady_states(rn,p)
```
It can also be done by inserting a whole parameter vector (but then the whole vector needs to be inserted), e.g.
```julia
rn = @reaction_network begin
(k1*Y^n1,k2*X^n1), X ↔ Y
end k1 k2 n1 n2
p= [2.,3.,2,2]
fix_parameter(rn,p)
steady_states(rn)
```
- Certain systems, like `(k1,k2), X ↔ Y`, have an infinite amount of steady states (since neither reactant is created nor degraded). Here the equation system to solve is just the same equation two times over. To solve this additional information is needed. I have added the `@fixed_concentration` macro for this. The code to solve the system becomes:
```julia
rn = @reaction_network begin
(k1,k2), X ↔ Y
end k1 k2 
p = [1.0, 2.0]
@fixed_concentration rn X+Y=5.
steady_states(rn,p)
```
The fixed concentrations can also include parameters (which should be declared as a part of the network):
```julia
rn = @reaction_network begin
(k1,k2), X ↔ Y
end k1 k2 xyTot
p = [1.0, 2.0, 5.]
@fixed_concentration rn X+Y=xyTot
steady_states(rn,p)
```
The fixed concentrations are stored in a dictionary in the field `fixed_concentrations`. When the fixed concentrations are set the corresponding equations are replaced in the equilibrium polynomial. If the equilibrium polynomial is created after the fixed concentrations the fixed concentrations equations will be fetched from the `fixed_concentrations` fields and inserted automatically.

It should be possible to compute these relations from the reactions itself, and then the steady states should be computable by just giving some initial conditions of the system, instead of giving the algorithms this extra information. I would be surprised if someone already have looked into how this is done (or if anyone of you knows about it). In the future we might want to add that in, but this works ok for now.
- HC works much faster if it has a solution to a similar system to work with. So if you solve it for some parameter values it then goes very fast to solve it for another set. For some reason this works best if you just select random (complex) parameters instead of just taking the first set of parameters solved by the system. This HC template is stored in the `homotopy_continuation_template` field and is initially nothing. The first time you solve a system a template is created and stored here (a tuple of a random parameter set and the corresponding solutions). Alternatively the template can be initialised using a macro or function (`@make_hc_template rn`, `make_hc_template(rn)`).
- Once everything is prepared the function `steady_states(reaction_network::DiffEqBase.AbstractReactionNetwork, params=Vector{Float64}()::Vector{Float64}, solver=HcSteadyStateSolver::Function)`. For the future one can define other solvers, but right now there is only the homotopy continuation one.
- Finally there is also a function `stability(solutions,params,reaction_network)` which returns true if the inserted fix point is stable).

**Bifurcations**

- If everything is ready to calculate steady states everything is also there for bifurcations. The function is simply: `bifurcations(reaction_network,parameters,param,range)`. `param` is the bifurcation parameter and `range` is a tuple with the range were you want to do the diagram.
(there is also an option for selecting an algorithm, with only one available). The output is a `bifurcation_diagram` structure (containing the parameter in question, the range, and a set of `bifurcation_path`s). The `bifurcation_path`s are a set of parameter value and corresponding fixed points, as well as the value of the corresponding eigenvalues.
- I have also added the option of making grid like bifurcation diagrams (`bifurcations_grid` and `bifurcations_grid_s2`). These does not actually track paths but simply solves the system at various grid points. 

**Plotting**

- I initially wanted to simply overload the `plot` function, but was a bit unsure how to do it properly (my initial attempts failed). Now I have simple added two functions, `bif_plot` and `bif_plot!` which take bifurcation diagrams as input and plots them. 
- Wanted to push this out now, but figure proper functions can be set of soon-ish.
- Rights now the color in the bifurcation diagrams depends on the eigenvalues of the corresponding fixed point (red if unstable with real eigenvalues, orange if unstable with imaginary ones, cyan if stable with imaginary ones, blue if stable with real ones).
- Some work might be needed to improve the plot recipes (e.g. better labelling), but that is for alter.

**Finally**
Still think the tests return some errors (which I think is due to a minor bug, will fix when I get time). The test got a bit long, would rewrite it a little bit more but got to go now, and figured it might be nice to have this up finally. Will update with more details.


isaacsas:
This looks really great, thanks. Since it is a big PR, and I’m a bit busy the next week, my comments may come in slowly as I process everything. (So just a heads up I may comment throughout the week.)

At a high level, one thing I’m been thinking is that the generated `reaction_network` structure is getting really big. I wonder if it may make sense to create an `EquilibriumProperties`structure that stores all the needed info, and then save just it as one additional field within the `reaction_network`? (I’d advocate we should maybe do some refactoring soon to do this more broadly for network features that make sense to group together, but that’s another discussion.)

isaacsas:
One other quick comment, maybe use `addequilibria!` instead (or `addsteadystates!`)?

src/equilibrate_utils.jl
```diff
@@ -0,0 +1,394 @@
...
+function fix_parameters(reaction_network::DiffEqBase.AbstractReactionNetwork, params=Vector{Number}()::Vector{Number}; kwargs...)
+    check_is_polynomial(reaction_network)
+    reaction_network.equilibratium_polynomial = reaction_network.make_polynomial(reaction_network;internal___var___polyvector=params,kwargs...)
+    !(typeof(reaction_network.equilibratium_polynomial[1])<:Polynomial) && (reaction_network.equilibratium_polynomial = map(pol->pol.num,reaction_network.equilibratium_polynomial))
+    foreach(sym -> reaction_network.equilibratium_polynomial[findfirst(reaction_network.syms.==sym)] = reaction_network.fixed_concentrations[sym], keys(reaction_network.fixed_concentrations))
+end
+
+#In some networks some linear combinations of concentrations remain fixed. This supplies this information directly to the network.
+macro fixed_concentration(reaction_network::Symbol, fixed_conc::Expr...)
```
isaacsas: Another suggestion, instead of:

```
@fixed_concentration rn X+Y=xyTot
```

what about

```
@add_constraints rn begin
X + Y = xyTot
Z + W = zwTot
end
```
The idea being that in the future we could build out this macro to support more general network constraints, like quasi-equilibrium constraints (leading to DAE systems). Using `begin` to `end` parallels the `reaction_network` macro too, which maybe makes it a bit easy to remember for users.

This might also make sense to live in `reaction_network.jl` with the idea we could use it more generally in the future.

I agree that in the future we should add auto-detection methods for conservation laws. There does appear to be some literature on this. For example, with just a couple minutes of searching I found [this](https://arxiv.org/abs/1109.1505).

isaacsas: Also, does the order of the equation matter here? i.e. could we instead say:
```
xyTot = X + Y
```
which seems more like a definition of `xyTot` to me, so I find it a bit clearer.

TorkelE: This type of definition would work.

the begin/end definition would also work. I think that in many places only one fixed state would be given, making the neater one more preferably. But I agree with your argument, especially if there are several such conditions. Could implement both?

TorkelE:
Yes, this is quite a big one, so lets give it some time.

src/equilibrate_utils.jl
isaacsas: How about `add_constraint` for adding one (without begin/end) and `add_constraints` with begin/end for adding multiple. 

Then the only question is whether we should also support calling `add_constraint` multiple times (which we probably should).

isaacsas:
I don't think you should need a manifest or project toml file. We haven't updated to using them yet.

TorkelE:
> I don't think you should need a manifest or project toml file. We haven't updated to using them yet.

Yeah, I removed it again on my computer (adding them gave some weird crash). Got these warnings that I was using packages that weren't in the dependencies and tried a lot of stuff to get it to go away. Didn't really think of adding new files with this one, just entered a few commands in the Pkg manager. They will be removed next time I update the PR.

TorkelE:
Hopefully the tests should work now. Systems with time as a variable now also works, if used however it should be fixed as a parameter with `fix_parameter(rn,t=0.,...)`. Implemented 
```julia
@fixed_concentration rn X+Y=xyTot
```
and
```julia
@add_constraints rn begin
X + Y = xyTot
Z + W = zwTot
end
```

isaacsas:
Maybe remove the DifferentialEquations package dependence and just use DiffEqBase with OrdinaryDiffEq. That would be lighter on the testing. I think you also need to update REQUIRES to include dependencies that only appear within tests (hence the CI failures).

TorkelE:
Think one would require the SDESolvers as well. I will try to draw out the test dependencies to all be declared in the runtests file, and also try scaling away what I can of differentialequations. Will see if this works first though.

isaacsas:
@TorkelE So I've been trying to test this out on an example where I know the answer but not having much luck with the bifurcation plot. The steady-state value seems correct when I test that. Can you see anything I'm doing wrong?

The code is at:
https://gist.github.com/isaacsas/3af31e1b10f3966467f48d1ede2b022c

and I've put a picture of the correct bifurcation plot up as a comment at 
https://gist.github.com/isaacsas/b354c84095b26949797d4a2ce5c9a0fc#file-ek-cell-cycle-png

isaacsas:
Ok, by eliminating the constraint and fixing all the parameters but the bifurcation parameter things seem to work. Not sure why this worked when the other case didn't though.

https://gist.github.com/isaacsas/fd313a2f1921b62128a0cc21a6d67772

isaacsas:
Last comment; the working example still sometimes misses the middle, unstable, branch. Using a parameter interval of `(.1,.6)` it isn't caught, but using `(.01,.65)` it is detected.

TorkelE:
I used a mostly similar network for some of my tests, which worked fine (but forced me to introduce some heuristics to get the unstable branch). Cannot look at the example right now, but hopefully be back in a few days, if required to I will implement the next step in the heuristics as well. That should solve it.

isaacsas:
Sounds good! When you look at the gist, you'll see that it appears that
simply reordering the parameters when calling `fix_parameters` can lead to
the unstable branch disappearing. (But it could be I'm doing something
wrong, as I found it strange I had to fix all parameters but the
bifurcation parameter for it to work at all...)

TorkelE:
Will have a look, but sometimes these kinda of bifurcations behave weirdly, so I am not overly surprised. I am quite sure that with an extra afternoon of coding we should be able to fix these bifurcations without failing though.

TorkelE:
Fixed the issue that you were having. It was due to a curious combination of a quirk of `Polynomials` and a (now fixed) bug. The essence can be found in this code:
```julia
using DynamicPolynomials
@polyvar a b c
p1 = 2a^2+1
p2 = 2a^2/b+1
p3 = (2a^2/b)/c+1
```
the last one causes an error. Although it is a rational polynomial, and should be ok, dynamic polynomial seems to get problem when you start stacking fractions. We represent parameters as polynomial variables until they are fixed. The problem is in the reaction rate `mr(P,k4*m/J4,J4)` (and similar one for `J3`). Somehow `J4` becomes tangled in such a triple fraction somewhere, causing an error. To solve this the variables need to be fixed one start everything (just like coefficient parameter need to be, in e.g. hill functions). This will be important to keep in mind when writing the docs at some point.

(why this then generated some strange behaviours is due to the now fixed bug)

isaacsas:
Happy to hear there was a bug that was the issue with the strange behaviors.

I think I'm still a bit confused about the recommended workflow for bifurcations. Should one first fix all non-bifurcation parameters, and then call `bifurcations` over the one non-fixed parameter? If this is case, is there a reason to not just fix the other parameters within the `bifurcations` function?

isaacsas:
I'll try to test the latest branch version some more in the next few days.

TorkelE:
Basically you only have to fix parameters that are an exponent, so in
```julia
rn = @reaction_network begin
    (k1*Y^n1,k2*X^n1),     X ↔ Y
end k1 k2 n1 n2
```
you have to fix `n1` and `n2`, but not `k1` and `k2`. The reason is that when the polynomial is generated parameters are kinda variables, and you cannot have a variable to the power of a variable in a polynomial, hence it has to become set to a number before the polynomial is created.

TorkelE:
I've added a 2nd HC based bifrucation solver. This one should solve systems like cell cycle network more reliably, and the expense of some speed and small gaps around the bifurcation points. At some point I hope to implement a third solver, fixing these issues as well.

Two different solvers are currently exported, `HcBifurcationSolver1` (the old one, also default), and `HcBifurcationSolver2`, the new one. The second can be selected with keyword argument:
```julia
bifurcations(reaction_network,parameters,:p,(1.0,10.0),solver=HcBifurcationSolver2)
```

You should now also be able to calculate steady states without first fixing the required parameters (but if you are doing a lot of calculating, fixing the parameters are very recommended for speed). This is currently not possible for bifurcation diagrams.

isaacsas:
travis seems to be timing out in steady_state_tests.jl. Maybe split those tests across several files to see if that gets rid of the time out error? Otherwise we'll have to look at bumping the travis timeout. Do tests finish ok locally for you?

I think we're ultimately going to have to introduce a parallel build structure when we are ready to merge this -- builds are otherwise going to get too long it seems. If you look at the travis file and runtests.jl in OrdinaryDiffEq.jl you can see how they define groups to split up the tests across several processors.

TorkelE:
Yeah, it works on my machine. Sounds like a good plan, will move it to different files in the meantime.

TorkelE:
There seems to be a problem only occurring on the windows version were HomotopyContinuation runs into an honest error. Slight delay while I try to set stuff up on a windows machine, figure it out and possibly sending a report to the people at HomotopyContinuation. Travis have some weird new error, but hopefully no biggie.

Going travelling for two weeks, so probably not fixed until after that.

isaacsas:
No worries; I've been meaning to play around with testing this on more examples anyways, but have been pretty sick the past few weeks... Have a good trip!

TorkelE:
Some things came in between, have been mostly off for a while.
Back in action, trying to finish this off now.

TorkelE:
Okay, we are pretty much there now. Things should be a little bit smoother now and you don't have to call so many function before, these are called automatically instead. Just `steady_states(rn,p)` and `bifurcations(rn,p,:param,(p1,p2))` is enough. There are some specific cases when a performance boost can be gained by using pre-functions, but I will write down all the details later.

All the stuff required for steady state solving have now been bundle up into an `equilibrate_content` field containing an `EquilibrateContent` structure.

We found some cases for which the latest HomotopyContinuation are having troubles, but that should be resolved soon. Otherwise the biggest trouble is that the `addequi!()` function used to add equilibrate functionality to minimum reaction networks does not work. I've basically worked it down to the `eval()` call being placed in a function (and being evaluated in global scope?). I'm not really keen on starting to mess with those kinds of errors right now, but will get there eventually.

isaacsas:
@TorkelE Sounds good. Is the issue that you are calling `eval` in that function, and then trying within the same function (or a sub-function) to use what was just `eval`'ed? That's the usual way I get tripped up.

src/maketype.jl
```diff
@@ -424,3 +428,21 @@ function addjumps!(rn::DiffEqBase.AbstractReactionNetwork;
 
     nothing
 end
+
+function addequi!(rn::DiffEqBase.AbstractReactionNetwork)
+    @unpack params, reactions, syms_to_ints, params_to_ints, scale_noise = rn
+
+    # first construct an ODE reaction network (required for getting f_func).
+    if rn.f == nothing
+        addodes!(rn)
```
isaacsas: Could this be what is causing your eval issues? Maybe require `addodes!` to be called before calling `addequi!` by the user? Then you could just throw an error here saying to call `addodes!` first if `rn.f` is not defined.

TorkelE:
> @TorkelE Sounds good. Is the issue that you are calling `eval` in that function, and then trying within the same function (or a sub-function) to use what was just `eval`'ed? That's the usual way I get tripped up.

Yeah, your right. This is the issue.

src/maketype.jl
TorkelE: This part is fine, I tried removing this part and calling before and the issue persisted.

TorkelE:
Ok, it works now. It was as you suggested with doing eval and using it in the same function. I've split `addequi!` into two functions, `addequi!1` and `addequi!2`. Maybe there is a better way which do not require the splitting, but at least it works now.

TorkelE:
Quick demo of how it works:
```juliua
using DiffEqBiological, Plots #Plots required to plot bifurcation diagrams

brusselator_network = @reaction_network begin
    A, ∅ → X
    1, 2X + Y → 3X
    B, X → Y
    1, X → ∅
end A B;
brusselator_p = [1.,4.]

ss_brusselator = steady_states(brusselator_network,brusselator_p)
bif_brusselator = bifurcations(brusselator_network,brusselator_p,:B,(1.,4.))
plot(bif_brusselator,2) #The 2 means that we plot the value of the second varriable (Y).

#This network contains a parameter in an exponent.
#This parameter needs to be fixed before steady states can be found.
σ_network = @reaction_network begin
    v0 + hill(σ,v,K,n), ∅ → (σ+A)
    deg, (σ,A,Aσ) → ∅
    (kB,kD), A + σ ↔ Aσ
    S*kC, Aσ → σ
end v0 v K n kD kB kC deg S;
fix_parameters(σ_network,n=4)       #We fix n to 4.
σ_p = [0.005, 0.1, 2.8, 4, 10, 100, 0.1, 0.01, 0.]

ss_σ = steady_states(σ_network,σ_p)
bif_σ = bifurcations(σ_network,σ_p,:S,(0.,2.))
plot(bif_σ)

#The steady states of this network depends on initial concentration.
#Additional information is required in the form of some contraint.
cc = @reaction_network begin
  k1, 0 --> Y
  k2p, Y --> 0
  k2pp*P, Y --> 0
  (k3p+k3pp*A)/(J3+Po), Po-->P
  (k4*m)/(J4+P), Y + P --> Y + Po
end k1 k2p k2pp k3p k3pp A J3 k4 m J4
p = [0.04,0.04,1.,1.,10.0,0.,0.04,35.,.1,.04]
@add_constraint cc P+Po=1     #The sum P+Po is always constant.

ss_cc = steady_states(cc,p)
bif_cc = bifurcations(cc,p,:m,(.01,.65))
plot(bif_cc)
```

isaacsas:
@TorkelE @korsbo What's the status of this? Are you both happy with it now, or are you planning more changes?

I'm traveling today, if you feel it is done I'll try to check it out myself in the next few days once I get back before merging.

korsbo:
I think that it is very close now. It needs a few docstrings and a minor change so that relative, rather than absolute, concentrations are used for some comparisons. @TorkelE started his vacation today but he indicated that he should have time to address this in not too long.

isaacsas:
Sounds good. One other small change would be to try to come up with more descriptive names in the min reaction network case than `addequi1!` and `addequi2!`.

korsbo:
Hehe, I must have missed that part of the API when I did a review/minor restructuring yesterday. I'll have a look and see if I can create a nicely descriptive function like `if_you_want_to_equlibrate_your_minimal_reaction_network_then_please_call_this_function_first!` :smile:

isaacsas:
Sounds like the perfect name!

korsbo:
I'm not sure why there currently are two functions for this. It seems that `addequi2` simply removes information added by `addequi1`. I'm inclined to remove this as well as the added constructor to `EquilibrateContent` but this is @TorkelE's domain and I will defer this to him.

isaacsas:
My understanding was there were `eval` related (world age?) issues that required a return to global scope before running what is in `addequil2`, but @TorkelE can comment better.

src/maketype.jl
```diff
@@ -414,11 +407,38 @@ function addjumps!(rn::DiffEqBase.AbstractReactionNetwork;
...
     nothing
 end
+
+#Adds stuff required to do homotopy continuation based equilibration calculations.
+function addequi1!(rn::DiffEqBase.AbstractReactionNetwork)
+    @unpack params, reactions, syms_to_ints, params_to_ints, scale_noise = rn
+
+    # first construct an ODE reaction network (required for getting f_func).
+    if rn.f == nothing
+        addodes!(rn)
```
isaacsas: I wonder if simply requiring the user to call `addodes!` before calling `addequil` would allow merging `addequil1` and `addequil2`? It seems reasonable to require this seeing as one is looking properties of the SS ODE solution.

isaacsas: Nevermind, I missed that @TorkelE tried this and it didn't work...

TorkelE:
> Hehe, I must have missed that part of the API when I did a review/minor restructuring yesterday. I'll have a look and see if I can create a nicely descriptive function like `if_you_want_to_equlibrate_your_minimal_reaction_network_then_please_call_this_function_first!` 😄

I love how you managed to squeeze in twin uses of the exclamation mark here. There might be some prettier way to solve the world age issue, but that probably requires us to call in a top-notch expert. Will have a chat about it if I encounters one, but even then there's a fundamental problem here that might be  to some extent un-solvable.


As @korsbo says this is more or less done, with the exceptions he mentioned (stuff that came up in his revision). Currently bifurcations are tracked with HomotopyContinuation.jl, in the long run we might want to look into using some of the other methods, however, right now I was well familiar with that package, it was well documented, and the authors very helpful. The results are also very good. I'll try a small write up on how it is actually put together later in the weak.

TorkelE:
Also, it should be noted that the line `fix_parameters(σ_network,n=4)` is not strictly necessary in the second example, and that this works:
```julia
#This network contains a parameter in an exponent.
#This parameter needs to be fixed before steady states can be found.
σ_network = @reaction_network begin
    v0 + hill(σ,v,K,n), ∅ → (σ+A)
    deg, (σ,A,Aσ) → ∅
    (kB,kD), A + σ ↔ Aσ
    S*kC, Aσ → σ
end v0 v K n kD kB kC deg S;
σ_p = [0.005, 0.1, 2.8, 4, 10, 100, 0.1, 0.01, 0.]
```
However, if you are going to solve several instances of the network (of course all with n=4), then this will provide a speed-up.

Project.toml
```diff
@@ -23,6 +27,7 @@ OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 SteadyStateDiffEq = "9672c7b4-1e72-59bd-8a11-6ac3964bc41f"
 StochasticDiffEq = "789caeaf-c7a9-5a7d-9973-96adeb23e2a0"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 
 [targets]
-test = ["Test", "OrdinaryDiffEq", "StochasticDiffEq", "SteadyStateDiffEq"]
+test = ["Test", "OrdinaryDiffEq", "StochasticDiffEq", "SteadyStateDiffEq", "Plots"]
```
ChrisRackauckas: why?

src/DiffEqBiological.jl
```diff
@@ -5,6 +5,8 @@ module DiffEqBiological
 using Compat, DataStructures, MacroTools, Parameters, Reexport, SparseArrays, SymEngine
 using DiffEqBase, DiffEqJump
 @reexport using DiffEqBase, DiffEqJump
+using HomotopyContinuation, DynamicPolynomials, LinearAlgebra, RecipesBase
+@reexport using DynamicPolynomials
```
ChrisRackauckas: why reexport?

src/reaction_network.jl
```diff
@@ -479,7 +483,7 @@ function massaction_ratelaw_deriv(spec::Symbol, scoef::Int, rate::ExprValues, su
     denom = one(scoef)
     for sub in subs
         name = sub.reactant
-        stoich = sub.stoichiometry
```
ChrisRackauckas: Please do "clean up" PRs separate from big changes...

test/equilibrate_test_4.jl
```diff
@@ -0,0 +1,68 @@
...
+    range_tupple = (1.,rand(5.:10.))
+    range1 = 1.: rand(5.:10.)
+    range2 = 1.: rand(5.:10.)
+    for p in p_vals
+        bif1 = bifurcations(rn,p,param1,range_tupple)
+        bif2 = bifurcations(SimpleHCBifurcationSolver(), rn,p,param1,range_tupple)
+        bif_grid = bifurcations_grid(rn,p,param1,range1)
+        bif_grid_2d = bifurcations_grid_2d(rn,p,param1,range1,param2,range2)
+        bif_grid_dia = bifurcations_diagram_grid(rn,p,param1,range1,param2,range_tupple)
+        plot(bif1); plot(bif2); plot(bif_grid); plot(bif_grid_2d); plot(bif_grid_dia);
```
ChrisRackauckas: do we need to plot in the test?

test/equilibrate_test_4.jl
```diff
@@ -0,0 +1,68 @@
...
+    param1 = rand(rn.params)
+    param2 = rand(rn.params)
+    range_tupple = (1.,rand(5.:10.))
+    range1 = 1.: rand(5.:10.)
+    range2 = 1.: rand(5.:10.)
+    for p in p_vals
+        bif1 = bifurcations(rn,p,param1,range_tupple)
+        bif2 = bifurcations(SimpleHCBifurcationSolver(), rn,p,param1,range_tupple)
+        bif_grid = bifurcations_grid(rn,p,param1,range1)
+        bif_grid_2d = bifurcations_grid_2d(rn,p,param1,range1,param2,range2)
```
ChrisRackauckas: why so many API functions? Can this not just work by dispatch?

src/equilibrate_utils.jl
```diff
@@ -0,0 +1,726 @@
...
+
+#Generates a number between 0 and 3 corresponing to some stability type (0=unstable, 1=stable, 2=unstable with imaginary eigenvalues, 3=stable with imaginary eigenvalues).
+function stability_type(eigenvalues::Vector{Any})
+    stab_type = Int8(0)
+    (maximum(real(eigenvalues))<1e-6)&&(stab_type+=1)
+    any(imag(eigenvalues).>1e-6)&&(stab_type+=2)
+    return stab_type
+end
+#For a specific stability types, returns a color (to be used to distinguish the types in plots).
+function stab_color(stab_type::Int8)
```
ChrisRackauckas: this should be with the plot recipes

src/equilibrate_utils.jl
```diff
@@ -0,0 +1,726 @@
...
+### Functions used during the construction of the reaction network. ###
+
+#Equilibrate content structure, with all the stuff required for homotopy continuation base equilibration calculations.
+mutable struct EquilibrateContent
+    make_polynomial::Function
+    constraints::Vector{Polynomial{true,Float64}}
+    homotopy_continuation_templates::Vector{NamedTuple{(:p, :sol),Tuple{Vector{Complex{Float64}},Vector{Vector{Complex{Float64}}}}}}
+    equilibrium_polynomial::Union{Vector{Polynomial{true,Float64}},Nothing}
+    is_polynomial_system::Bool
+    polyvars::NamedTuple{(:x, :t, :p),Tuple{Vector{PolyVar{true}},PolyVar{true},Vector{PolyVar{true}}}}
```
ChrisRackauckas: x?

src/equilibrate_utils.jl
```diff
@@ -0,0 +1,726 @@
...
+
+#Generates the expression which is then converted into a function which generates polynomials for given parameters. Parameters not given can be given at a later stage, but parameters in the exponential must be given here. Also checks whenever the system is polynomial.
+function get_equilibration(params::Vector{Symbol}, reactants::OrderedDict{Symbol,Int}, f_expr::Vector{ExprValues})
+    func_body = Expr(:block)
+    push!(func_body.args,Expr(:if,:(length(internal___var___paramvals) != 0),Expr(:block)))
+    for i = 1:length(params)
+        push!(func_body.args[1].args[2].args, :((internal___var___paramvals___exemption!=$i) && ($(params[i])=(isinteger(internal___var___paramvals[$i]) ? Int64(internal___var___paramvals[$i]) : internal___var___paramvals[$i]))))
+    end
+    push!(func_body.args,:([]))
+    foreach(poly -> push!(func_body.args[2].args, recursive_replace!(poly,(reactants,:internal___polyvar___x),(OrderedDict(:t=>1),:internal___polyvar___t))), deepcopy(f_expr))
```
ChrisRackauckas: Is x this the parameter being changed?

src/equilibrate_utils.jl
```diff
@@ -0,0 +1,726 @@
...
+-  constraints: Constraints, typically given as X + Y = C. Both parameters, varriables, and numbers can be given. However, parameters much be declared previously in the reaction network.
+"""
+macro add_constraint(rn::Symbol, constraints::Expr...)
+    func_expr = Expr(:escape,:(internal___add___constraint!($rn)))
+    foreach(constraint -> push!(func_expr.args[1].args,recursive_replace_vars!(balance_poly(constraint), rn)),constraints)
+    return func_expr
+end
+#In some networks some linear combinations of concentrations remain fixed. This supplies this information directly to the network.
+"""
+    @add_contraints(reaction_network, constraints)
```
ChrisRackauckas: If we add constraints we should get a DAE, but I don't see a mass matrix being added to the `ODEProblem`?

src/equilibrate_utils.jl
```diff
@@ -0,0 +1,726 @@
...
+        (get_equi_poly(rn)[i]==-get_equi_poly(rn)[j])&&deleteat!(get_equi_poly(rn),i)
+    end
+end
+
+#--- Makes homotopy continuation templates to be used at a later stage when solving for fixed points ---#
+#Can be called separatly, but will otherwise be called first time a steady state is to be found. Solves the system once using random parameters. Saves the solution as a template to be used for further solving.
+"""
+    make_hc_template(reaction_network)
+
+    Makes a template for solving the riven reaction networks steady states using homotopy continuation. Generally not needed as this is genrated automatically when netowkr is first solved *if one does not already exists). However, it can be sued to replace the current template, for whatever reason.
```
ChrisRackauckas:
```suggestion
    Makes a template for solving for the given reaction networks steady states using homotopy continuation. Generally not needed as this is generated automatically when network is first solved *if one does not already exists*. However, it can be sued to replace the current template, for whatever reason.
```

src/equilibrate_utils.jl
```diff
@@ -0,0 +1,726 @@
...
+ - parameter_symbol: the parameter which is to be varried.
+ - parameter_range: a range with all the values of the parameter for which steady states are to be solved.
+
+## kwargs
+ - potential arguments for steady state solvers (however, only current solver does not have any such arguments).
+"""
+function bifurcations_grid(rn::DiffEqBase.AbstractReactionNetwork, args...)
+    return bifurcations_grid(HCSteadyStateSolver(), rn, args...)
+end
+function bifurcations_grid(solver::AbstractSteadyStateSolver, rn::DiffEqBase.AbstractReactionNetwork,p::Vector{Float64},param::Symbol,range::AbstractRange)
```
ChrisRackauckas: network then solver to match conventions?

src/equilibrate_utils.jl
```diff
@@ -0,0 +1,726 @@
...
+ - parameter_values: a vector which specifies the parameter values for which the steady states are to be found. Not required in case the reaction network lacks steady states.
+ - parameter_symbo1l: the first parameter which is to be varried.
+ - parameter_range1: a range with all the values of the first parameter for which steady states are to be solved.
+ - parameter_symbol2: the second parameter which is to be varried.
+ - parameter_range2: a range with all the values of the second parameter for which steady states are to be solved.
+
+## kwargs
+ - potential arguments for steady state solvers (however, only current solver does not have any such arguments).
+"""
+function bifurcations_grid_2d(rn::DiffEqBase.AbstractReactionNetwork, args...)
```
ChrisRackauckas: don't need a second function. This is just a different dispatch of `bifurcations_grid`

src/equilibrate_utils.jl
```diff
@@ -0,0 +1,726 @@
...
+            deleteat!(paths,i)
+            return
+        end
+    end
+end
+
+#--- Function for the main HC bifurcation solver ---#
+#Main bifurcation solver. Track paths from the first to the last parameter values. Detects whenever a solution is lost due to a complicated bifurcation. In this case it resumes the path tracking just after that bifurcation.
+#Δp = how long step in parameter value to take after a bifrucation is encountered (should be smaller than the minimum expected distance between two bifurcations).
+#Δx = distance between two solutions for them to be considered identical. Should larger than the distance between two solutions.
```
ChrisRackauckas: if you mean `u`, then use `u`. We shouldn't have one file in DiffEq use a different naming convention from all of the others.

src/equilibrate_utils.jl
```diff
@@ -0,0 +1,726 @@
...
+
+
+
+### Structures holding information about bifurcation diagrams. ###
+
+### Bifurcation Diagrams Structures###
+
+#Contains information from a single bifurcation line. A bifurcation plot would typically contain several such lines. These are the results from a method which tracks a solution through parameter space.
+struct BifurcationPath
+    p_vals::Vector{Float64}
```
ChrisRackauckas: Does it all need to be Float64s?

src/equilibrate_utils.jl
isaacsas: That is exactly what I had in mind when I suggested using `add_constraints` instead of `fix_concentrations` for the macro name. Right now constraints are only added into the equilibration data structures, but ultimately we should extend this to allow the generation of DAE models (or at least allow them when using min reaction networks where ODE construction is deferred).

Project.toml
TorkelE: This is possibly me working with things I am not expert on, but as we require Plots in the tests, but no where else, I figure it belonged here.

src/DiffEqBiological.jl
TorkelE: It has to do with the polyvars being created in the scope in which the macro is called, and some irritating issues that appeared were polyvars could have same names, but be different.

However, after your comment I had a look at this again. Due to the revised approach were I do a lot of stuff in the EquilibrateContent constructor I was able to move were the polyvars were created and could remove this reexport.

src/reaction_network.jl
TorkelE: You are ofcourse right. Will avoid this in the future.

test/equilibrate_test_4.jl
TorkelE: Basically these were to try the plot recipes, ensuring these  generated no errors. Could remove them if they are unnecessary though.

test/equilibrate_test_4.jl
TorkelE: @korsbo suggested a method were one gave a network topology type to designate which one to use, in the same spirit the solvers are given. I thought this would generate some messy since each method have requires different types of input

We have:
bifurcations requires `AbstractReactionNetwork`, `Vector{Float64}`, `Symbol`, `Tuple{Number,Number}` (technically the last one can be anything, but something like this which contains two numbers)
bifurcation_grid requires `AbstractReactionNetwork`, `Vector{Float64}`, `Symbol`, `AbstractRange`
bifurcation_grid_2d requires `AbstractReactionNetwork`, `Vector{Float64}`, `Symbol`, `AbstractRange`, `Symbol`, `AbstractRange`
bifurcation_diagram_grid requires `AbstractReactionNetwork`, `Vector{Float64}`, `Symbol`, `AbstractRange`, `Symbol`, `Tuple{Number,Number}`

I thought it would be clearer to have a separate function name for each, but it would be possible, and maybe a bit simpler, to have a single function , `bifurcations`, and depending on what form the input have it can have a dispatch for the corresponding type of diagram.

src/equilibrate_utils.jl
TorkelE: Good catch, moved.

src/equilibrate_utils.jl
TorkelE: I have removed all mentions of x and replaced them so that only u should be used.

src/equilibrate_utils.jl
TorkelE: `internal___polyvar___x[1:n]` are the polyvars corresponding to the n reactants of the system- I've changed it so `u` is used instead of `x` to designate polyvars.

src/equilibrate_utils.jl
TorkelE: Fixed.

src/equilibrate_utils.jl
TorkelE: Yes, I have shifted this.

src/equilibrate_utils.jl
TorkelE: Currently there only exists a single solver to solve steady states. It was written like this to prepare for if/when more was added. But yes, in the meantime we could remove this and have only a single function.

src/equilibrate_utils.jl
TorkelE: Yes, I've went over the files to ensure there is not more x used.

src/equilibrate_utils.jl
TorkelE: Probably me being overzealous defining types more narrowly then they should. used `::Vector` would be better?

korsbo: By not restricting it at all you increase the users ability to draw on the rest of Julias ecosystem. You don't have to sacrifice performance if you do something like:

```julia
struct BifurcationPath{T} where T
    p_vals::T
    ...
```

However, just using `::Vector` would be bad for performance since the compiler cannot figure out what you are giving it.

src/equilibrate_utils.jl
TorkelE: I have never really been involved with DAEs, but do we really get one? The simplest examplee would be X <--> Y, with the equations
dX/dt = Y - X
dY/dt = X - Y
Adding the information X+Y=1 doesn't really make the system different, since it is already defined by one differential equation for each variable, and an initial condition.

However, when we want to find steady states we get the system
Y - X = 0
X -Y = 0
which have an infinite amount of solutions, we need some additional information. Here a human "knows" the relation X+Y=1 (or = some parameter).`@add_constraint` adds this information, making the system
Y - X = 0
X -Y = 0
X+Y-1 =0
which we can solve. 

It is likely that one can, given some initial condition, compute this constraint for a reaction network. Adding such an algorithm would save one from having from defining the constraints (although one would still have to supply some initial condition, or name for a parameter designating the combined concentration).

All that said, adding a mass matrix to the ODE problem would still make sense (I presume this could increase efficiency?). Or would creating a DAE problem from it also be a way to solve the system, given this additional information, in a more efficient way?

src/equilibrate_utils.jl
```diff
@@ -7,7 +7,7 @@ mutable struct EquilibrateContent
     make_polynomial::Function
     constraints::Vector{Polynomial{true,Float64}}
     homotopy_continuation_templates::Vector{NamedTuple{(:p, :sol),Tuple{Vector{Complex{Float64}},Vector{Vector{Complex{Float64}}}}}}
-    equilibrium_polynomial::Union{Vector{Polynomial{true,Float64}},Vector{Polynomial{true,Int64}},Nothing}
+    equilibrium_polynomial#::Union{Vector{Polynomial{true,Float64}},Vector{Polynomial{true,Int64}},Nothing}
```
ChrisRackauckas: The issue is `Int64`. You should almost never use `Int64`, always `Int`.

ChrisRackauckas:
Change that last thing to Int and it's good to go. Docs should follow up.

ChrisRackauckas:
Let's continue to improve, but well done.

isaacsas:
@TorkelE @korsbo Awesome job on this one! Thanks!

PR #99, opened 2019-03-06
return nothing in ODE/SDE functions

isaacsas:
`btime` was reporting a small allocation in the ODE rhs function, even for a simple two-variable system. This seems to get rid of it (at least for small systems, on the larger BCR network I'm still seeing allocations).


isaacsas:
@TorkelE This look ok to you? (Just want to double check I put the `nothing` in the right place in the expression list...)

TorkelE:
Looking good, yes, this should ensure that the functions return nothing.

PR #101, opened 2019-03-07
Add `@inbounds` to generated functions

ChrisRackauckas:
I was wondering whether it should be an argument to enable/disable it, but since we are generating the function to match the reaction size, it's fine that it would segfault when used improperly?

### DIFF
--- a/src/reaction_network.jl
+++ b/src/reaction_network.jl
@@ -365,7 +365,7 @@ function make_func(func_expr::Vector{Expr},reactants::OrderedDict{Symbol,Int}, p
         push!(system.args, recursive_replace!(func_line, (reactants,:internal_var___u), (parameters, :internal_var___p)))
     end
     push!(system.args, :(nothing))
-    return :((internal_var___du,internal_var___u,internal_var___p,t) -> $system)
+    return :((internal_var___du,internal_var___u,internal_var___p,t) -> @inbounds $system)
 end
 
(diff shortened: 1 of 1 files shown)


src/reaction_network.jl
```diff
@@ -365,7 +365,7 @@ function make_func(func_expr::Vector{Expr},reactants::OrderedDict{Symbol,Int}, p
         push!(system.args, recursive_replace!(func_line, (reactants,:internal_var___u), (parameters, :internal_var___p)))
     end
     push!(system.args, :(nothing))
-    return :((internal_var___du,internal_var___u,internal_var___p,t) -> $system)
+    return :((internal_var___du,internal_var___u,internal_var___p,t) -> @inbounds $system)
```
ChrisRackauckas: `$system` is already a `begin ... end`, so this should just work.

ChrisRackauckas:
```julia
system = Expr(:block)
push!(system.args,:(x[2]+x[5]))
:(f(x) -> @inbounds $system)
```

```julia
:(f(x)->begin
          #= D:\OneDrive\Computer\Desktop\test.jl:32 =#
          #= D:\OneDrive\Computer\Desktop\test.jl:32 =# @inbounds begin
                  x[2] + x[5]
              end
      end)
```

shows that it works.

isaacsas:
Yeah, I'm fine with it always being on. We can always change this later if needed, but we shouldn't be generating indices outside of the allowable range...

PR #103, opened 2019-03-09
fix +0s and -1* and add inbounds for jumps

isaacsas:
This should (ultimately)
1. Fix the "+0" terms appearing in ODEs and other generated expressions.
2. Fix the "-1*" terms appearing in ODEs and other generated expressions. Here I just use the `clean_subtractions` function @korsbo wrote in Latexify which he said we should feel free to copy over.
3. Add `@inbounds` for the jump expressions it was still missing on.

@ChrisRackauckas or @TorkelE, would one of you mind taking a look that this seems reasonable? I tested on a few ODE and jump examples and didn't see issues with the generated expressions or solutions, but it would be good to have someone more comfortable working with expressions double check what I've done. If what I'm doing looks ok I'll update the Jacobian and noise generation functions too before merging.

### DIFF
--- a/src/maketype.jl
+++ b/src/maketype.jl
@@ -13,11 +13,10 @@ function maketype(abstracttype,
                   syms,
                   scale_noise;
                   params = Symbol[],
-                  pfuncs=Vector{Expr}(undef,0),
                   jac = nothing,
                   paramjac = nothing,
                   jac_prototype = nothing,
-                  symjac=Matrix{Expr}(undef,0,0),
+                  symjac=Matrix{Union{Expr,Symbol,Number}}(undef,0,0),
                   reactions=Vector{ReactionStruct}(undef,0),
                   syms_to_ints = OrderedDict{Symbol,Int}(),
                   params_to_ints = OrderedDict{Symbol,Int}(),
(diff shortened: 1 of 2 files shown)


isaacsas:
On the BCR network, `CVODE_BDF` with `abstol=reltol=1e-8` is about 15% faster on my laptop with these changes.

TorkelE:
After Monday I should start having some time again. Hopefully I will be able to have a look at this tomorrow though.

isaacsas:
Sounds good. One question for you: why do you force the `symjac` entries to be expressions? Why not just allow them to be of type `Any`? By using the `:(1*0)` trick the generated Jacobian evaluation function ends up having extra instructions, and it also causes Latexify to print a rather ugly Jacobian. 

Would it be ok for me to change `symjac` to be of type `Any`?

TorkelE:
In principle I have been as limiting when designating types as possible (in the uncertainty of what effect this actually have on performance).

Ye we can change it to `Any` (or `Union{Expr,Symbol,Number}`. We could probably also change a lot of `Any` in the recursive functions to `Union{Expr,Symbol,Number}` as well, but I have no idea if there is any gain to be done from this).

ChrisRackauckas:
Julia will optimize a union of 3 things just fine, so that's okay.

isaacsas:
Ok, I think this should be good to go, and as I said I am seeing a moderate performance improvement with CVODE_BDF and no Jacobians. I'd expect Jacobians to be faster too, as the generated code for them was indeed doing extraneous multiplies by 1. I've also reordered the generated Jacobian functions to traverse them in memory ordering (or at least that was my intent).

In a future PR we should get rid of the remaining `Any`'s in the reaction/reactant handling. Probably we should just make a type alias like
```
const MExpr = Union{Expr,Symbol,Number}
```
and use that everywhere. I'm going to save that for a future PR and instead I'll open an issue when this is done.

isaacsas:
Actually, I'll get rid of the `Any`s in the recursive functions now. Missed that.

isaacsas:
OK, now this should be done.

isaacsas:
I'll plan to merge then once tests pass.

Unfortunately, while the typing gives a speed up, the cleaning of "a + -1*b" to "a-b" gives a hit, so not sure we'll gain anything for generating ODEs. (But the ODE solves are faster.)

isaacsas:
I suspect the `clean_subtractions` code, which I just took from Latexify, could be sped up. Right now it is using splatting of arrays, which I'm guessing gives a performance hit.

PR #105, opened 2019-03-11
Query interface updates

isaacsas:
This adds more exported accessors for querying network properties, along with several functions to calculate dependency graphs. 

As we fill this in users should be able to rely less on directly accessing the reaction_network fields, which will be helpful if and when we convert the backend to using ModelingToolkit. (We can just update these functions as appropriate.)

This won't pass tests till a new DiffEqJump release incorporating [67](https://github.com/JuliaDiffEq/DiffEqJump.jl/pull/67). In either case I'll leave it open for a day in case anyone wants to comment on the accessor names.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
 #  - osx
 julia:
-  - 1.0
+  - 1.1
   - nightly
 matrix:
   allow_failures:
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
the names seem fine to me.

PR #106, opened 2019-03-11
- Replace several instances of "Any" with "ExprValues".

TorkelE:
Added `const ExprValues = Union{Expr,Symbol,Number}` and replaced it in all appropriate places. Should not be any `Any` left now.

### DIFF
--- a/src/DiffEqBiological.jl
+++ b/src/DiffEqBiological.jl
@@ -12,7 +12,7 @@ using Parameters
 @reexport using DiffEqBase, DiffEqJump
 using Compat
 
-const ExprValues = Union{Expr,Symbol,Float64,Int64}                   
+const ExprValues = Union{Expr,Symbol,Float64,Int}                   
 
 include("reaction_network.jl")
 include("maketype.jl")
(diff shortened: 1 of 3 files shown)


isaacsas:
Looks good, I only have two minor comments:
1. Maybe put the definition of `Exprvalues` in the DiffEqBiological.jl file before the includes?
2. I wonder if we should define `Exprvalues` to be
```
const ExprValues = Union{Expr,Symbol,Float64,Int}
```
since `Number` is an abstract type? I have no idea if that might give a performance gain from being more concretely typed.

TorkelE:
Sounds good, should it be `Union{Expr,Symbol,Float64,Int}` or `Union{Expr,Symbol,Float64,Int64}`?

isaacsas:
I’d use int since it is the native integer type on each platform.

isaacsas:
It looks like the Julia docs indicate up to four types should still be efficient. So I think we should be ok. 

You could look at the reported test timings before and after the change in type to make sure we don’t see any big slowdown.

TorkelE:
Good, changes are in.

isaacsas:
I’ll plan to merge when tests finish.

PR #110, opened 2019-03-13
Further network query functions

isaacsas:
This should finish off a pretty much complete set of functions for querying network properties.

Going forward, I'll update the docs to use these functions instead of `reaction_network` fields. Hopefully this allows us to transition to a different backend like ModelingToolkit down the line without any breaking changes.


isaacsas:
@TorkelE What do you think about my updating the docs to suggest using these accessors instead of reaction_network fields? There shouldn't be any real overhead, since they should get inlined, but it allows us to change around the internals of how we store the network and such while still maintaining a unified API. The main thing is switching any codes that depend on reaction_network fields to use these accessors going forward.

TorkelE:
I agree that this is a better way of doing it, as you say we could really use with some tidying up the internal parts. Telling people to use these functions in the docs would make us much more free to swap around however we want there.

PR #113, opened 2019-03-13
Incremental model construction

isaacsas:
This adds a set of functions for extending a `min_reaction_network` with more species/parameters/reactions, or creating an empty reaction network and then extending it. New functions:

* `@empty_reaction_network`
* `addspecies!`
* `addparam!`
* `addreaction!`
* `add_scale_noise_param!`

I think `addreaction!` could be made more efficient -- right now the user passes an `ExprValues` for the parameter and an `Expr` for the reaction. Basically the two components of one line of the normal macro. So the following two networks should have all of the `min_reaction_network` defined fields equal

```julia
minrn = @min_reaction_network begin
    (k*X*(X-1)*(X-2),Y * X), X + 2X ⟺ Y+ X
end k
```
and
```julia
emptyrn = @empty_reaction_network
addspecies!(emptyrn, :X)
addspecies!(emptyrn, "Y")
addparam!(emptyrn, :k)
addreaction!(emptyrn, :((k*X*(X-1)*(X-2),Y * X)), :(X + 2X ⟺ Y+ X))
```

The add* functions can also be used on a `min_reaction_network`, as long as `addodes!` and such hasn't already been called. (That might work too, but I haven't checked.) Once species, params and reactions have been added the addodes! and such functions can be called.

One caveat, any species used in the `min_reaction_network` should appear within the reactions and not just the rate `ExprValues` or they won't be added to the species list. i.e. don't do
```
minrn = @min_reaction_network begin
    k*X, Y --> 0
end k
```
If `X` is then added as a species later the dependents of the above reaction will be incorrect. (The general rule is species and params should be defined before any reaction using them.)

I'll again leave this open for a day for comments (and can leave it open longer if anyone wants more time to review it).

### DIFF
--- a/src/DiffEqBiological.jl
+++ b/src/DiffEqBiological.jl
@@ -21,14 +21,17 @@ include("massaction_jump_utils.jl")
 include("problem.jl")
 
 # reaction network macro
-export @reaction_network, @reaction_func, @min_reaction_network
+export @reaction_network, @reaction_func, @min_reaction_network, @empty_reaction_network
 
 # functions to query network properties
-export speciesmap, paramsmap, numspecies, numreactions, numparams
+export species, params, speciesmap, paramsmap, numspecies, numreactions, numparams
 export oderhsfun, jacfun, paramjacfun, odefun, noisefun, sdefun, jumps, regularjumps
 export odeexprs, jacobianexprs, noiseexprs, jumpexprs, rateexpr, oderatelawexpr, ssaratelawexpr
 export substratestoich, productstoich, netstoich, ismassaction, dependants, dependents, substrates, products
 export rxtospecies_depgraph, speciestorx_depgraph, rxtorx_depgraph
+
+# functions to extend empty_ and min_ reaction_networks
+export addspecies!, addparam!, add_scale_noise_param!, addreaction!
 
 # functions to add mathematical equations to the network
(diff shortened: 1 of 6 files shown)


isaacsas:
I meant to say, making `addreactions!` more efficient would, I think, be possible by having the user instead pass the substrate stoichiometry and product stoichiometry as vectors of tuples/pairs of symbols and integers. The "rate" would still need to be an `ExprValues` for now.

isaacsas:
I'll leave that as a future optimization.

isaacsas:
Almost have a second `addreactions!` interface that should be faster for at least mass action reactions.

isaacsas:
Ok, I've now added second `addreaction!` interface that is **much** faster (it can handle adding 250,000 reactions to the system in seconds). `addjumps!`, `addodes!` and such now become the blocking factor, but that is future work.

The basic interface is the following, for a reaction like `k*X, 2X+3Y -> 3W` one can now call 
```julia
addreaction!(rn, :(k*X), (:X=>2,:Y=>3), (:W=>3,))
```
If the rate expression is a number or symbol it should be fast. If it is an expression like `:(k*X)` above, then it will still be slow due to the `recursive_content` function call. In practice I would think this isn't an issue, since most really large networks are probably primarily composed of mass action reactions, but this is always something we can further optimize later on.

PR #123, opened 2019-04-18
Sparse Jacobian support

isaacsas:
~~This is pretty close to done modulo some performance work and cleanup. We could either close https://github.com/JuliaDiffEq/DiffEqBiological.jl/pull/120 and just use this, or merge https://github.com/JuliaDiffEq/DiffEqBiological.jl/pull/120 and then I can rebase. That would keep the sparse matrix specific commits separated from the general rewrite of mass action reaction expression handling in https://github.com/JuliaDiffEq/DiffEqBiological.jl/pull/120.~~ 

~~My preference would be to merge https://github.com/JuliaDiffEq/DiffEqBiological.jl/pull/120 so I can slim this down and work a bit more on optimizing the sparse Jacobian and a matrix free version.~~

Ok, the sparse Jacobian seems good to go (in that I get the same results when evaluating it as our dense Jacobian method gives in my testing). We should next get a sparse-oriented `linsolve` into OrdinaryDiffEq to use for further testing (of accuracy and efficiency). 

For both the sparse and dense generated Jacobians I'm finding issues with code compilation time on the BCR network (up to a factor of 20 slower than the time it takes to actually solve the network). Looking at some past Julia issues it seems like code generation time grows quadratically in the number of lines, so I think we'll need to start looking at creating sub-functions as followup work. That would interesting for playing around with threading too as @ChrisRackauckas had mentioned in the past.


src/reaction_network.jl
```diff
@@ -562,27 +577,89 @@ function calculate_symjac(reactions, reactants, parameters, syms)
...
+function calculate_jac(symjac::Matrix{ExprValues}, reactants::OrderedDict{Symbol,Int}, parameters::OrderedDict{Symbol,Int}, zeroout_jac=false)
     func_body = Expr(:block)
-    @inbounds for j = 1:size(symjac)[2], i = 1:size(symjac)[1]
-        ex =  (symjac[i,j] isa Number) ? symjac[i,j] : recursive_replace!(symjac[i,j],(reactants,:internal___var___u), (parameters, :internal___var___p))
-        push!(func_body.args, :(internal___var___J[$i,$j] = $(ex)))
+    zeroout_jac && push!(func_body.args, :(fill!(internal___var___J, zero(eltype(internal___var___J)))))
+    @inbounds for j = 1:size(symjac)[2], i = 1:size(symjac)[1]        
+        if symjac[i,j] isa Number
+            ex = symjac[i,j]
+            (!zeroout_jac || !iszero(ex)) && push!(func_body.args, :(internal___var___J[$i,$j] = $(ex)))
```
isaacsas: This maybe a bit subtle. The idea is that for large enough networks, like the BCR network, we can't generate a function with one line for each Jacobian entry (we would have to split the function into subfunctions). However, we would like a dense Jacobian as the matrix is not so large that they don't work well. (As I mentioned, once compiled it gives a faster solver with KenCarp4 than using finite differences.)

If we first zero out the matrix in the generated function we only need to then fill in non-zero Jacobian entries, which are usually sparse for larger systems. Setting `zeroout_jac=true` takes the latter approach. The default is `false` to keep the current DSL Jacobian behavior (where even zero entries in the Jacobian get a line in the generated function.)

isaacsas:
@ChrisRackauckas Let me know if you want to take a look, otherwise I'll plan to merge tomorrow night.

ChrisRackauckas:
>We should next get a sparse-oriented linsolve into OrdinaryDiffEq to use for further testing (of accuracy and efficiency).

`lu` on a sparse matrix automatically uses SuperLU which is a sparse LU

src/reaction_network.jl
```diff
@@ -129,7 +129,7 @@ function coordinate(name, ex::Expr, p, scale_noise)
     (reactions, reactants, parameters, syms, params) = get_minnetwork(ex, p)
 
     # expressions for ODEs
-    (f_expr, f, f_rhs, symjac, jac, paramjac, f_symfuncs) = genode_exprs(reactions, reactants, parameters, syms)
+    (f_expr, f, f_rhs, symjac, jac, paramjac, f_symfuncs, jac_prototype) = genode_exprs(reactions, reactants, parameters, syms)
     odefun = :(ODEFunction(f; jac=$jac, jac_prototype=nothing, paramjac=$paramjac, syms=$syms))
```
ChrisRackauckas: it looks like you're not specifying the sparsity pattern?

isaacsas: The default macro always generates a dense matrix. `addodes!` should be forwarding the prototype correctly.

isaacsas:
Do you know if the sparse `lu` from SuperLU also does reorderings to reduce fill? My guess is the B cell Jacobian would benefit from reordering.

ChrisRackauckas:
It does. SuperLU is the standard "state-of-the-art" sparse LU everyone uses.

isaacsas:
Great, I'm hoping to play around with it in the Julia native solvers this weekend. 

The bigger issue is how to speed up the Jacobian function compilation. Right now that is making it pretty slow unless one is looking to solve the ODE many times (at least on the B cell example).

ChrisRackauckas:
Yeah, I think other strategies likely should be used for J*v calculations, like seeded AD. I'm not sure the symbolic J*v will scale that well in terms of compile time, while the AD version will be quite comparable in speed.

isaacsas:
Interesting; I haven't played around with AD much. I guess I need to start reading up on it then.

PR #124, opened 2019-04-20
drop giving jacobian function an explicit name

isaacsas:


src/reaction_network.jl
```diff
@@ -130,11 +130,11 @@ function coordinate(name, ex::Expr, p, scale_noise)
 
     # expressions for ODEs
     (f_expr, f, f_rhs, symjac, jac, paramjac, f_symfuncs, jac_prototype) = genode_exprs(reactions, reactants, parameters, syms)
-    odefun = :(ODEFunction(f; jac=$jac, jac_prototype=nothing, paramjac=$paramjac, syms=$syms))
+    odefun = :(ODEFunction(f; jac=$jac, jac_prototype=$jac_prototype, paramjac=$paramjac, syms=$syms))
```
isaacsas: This changes nothing since the generated `jac_prototype=nothing` in this case, but now the decision on what to return is all handled in `genodeexprs`.

PR #128, opened 2019-06-14
add SparseArrays dependency

isaacsas:
and ignore Manifest.toml files in git.


ChrisRackauckas:
Does this need to get a tag so that way there's no warnings?

isaacsas:
I think so, but I actually still got warnings until I ran `resolve` too in the package manager. (But it didn't actually change the Project.toml any further...)

PR #133, opened 2019-07-17
- Remove some Int64

TorkelE:
Remove some uses of Int64  in equilibrate_utils.


src/equilibrate_utils.jl
```diff
@@ -357,7 +357,7 @@ struct BifurcationPath
     vals::Vector{Vector{Float64}}
     jac_eigenvals::Vector{Vector{ComplexF64}}
     stability_types::Vector{Int8}
-    length::Int64
+    length::Integer
```
ChrisRackauckas: Int

PR #141, opened 2019-07-23
Specifies the package before all instances of solve in the test files.

TorkelE:
Uses of solve should already be specifies in the src files. Here I simply add it to all instances in the test files, if this will make things go away.


test/test_gillespie.jl
```diff
@@ -5,14 +5,14 @@ end
 sir_prob = DiscreteProblem([999,1,0],(0.0,250.0))
 sir_jump_prob = JumpProblem(sir_prob,Direct(),sir_model)
 
-sir_sol = solve(sir_jump_prob,FunctionMap())
+sir_sol = DiffEqJump.DiffEqJump.solve(sir_jump_prob,FunctionMap())
```
isaacsas: This looks like a typo

isaacsas:
@TorkelE Why is this suddenly a problem? Is there a way we can import `solve` from the DiffEq* packages and HomotopyContinuation without having to qualify the package name? I guess I don't understand why this isn't getting resolve based on the types of variables being passed into solve.

ChrisRackauckas:
>I guess I don't understand why this isn't getting resolve based on the types of variables being passed into solve.

They are different `solve` functions

TorkelE:
I have gotten such errors when I have been working with both `HomotopyContinuation.jl` and `DifferentialEquations.jl` in the same scope. In practical use `HomotopyContinuation.solve()` is only ever used inside the `DiffEqBiological.jl` module, and this is of no worry to the user.

If the warning simply denotes that the module contains an instance were there is a `solve()` without module specified before, then this should fix it. However, if the warning simply tells us that this is something we should watch out for in general the we need to find a way to suppress it.

isaacsas:
So we could eliminate this issue if we instead don't `using HomotopyContinuation`, but explicitly qualified the calls to `HomotopyContinuation` functions? I guess we could `using HomotopyContinuation: fun1, fun2` the other functions we use, but not `solve`, and then internally write `HomotopyContinuation.solve`?

ChrisRackauckas:
The solution to this shouldn't be in the test files, but rather in how things are exported. Why not just `import HomotopyContinuation`?

TorkelE:
No reasons really. Never really thought of what the difference is and simply used `using`. It is an easy change though, will fix it. This was the only place were the package using `solve` was not specified, so I figured I'd target this. But if not I guess we can just discard this PR?

ChrisRackauckas:
Probably.

PR #142, opened 2019-07-29
Add ! after function names

TorkelE:
There were two functions (add_hc_template and make_hc_template) which modified their input arguments, but for which I forgot to add ! at the end. Now this is fixed.


TorkelE:
The one time I figure "No need to test before pushing, I most likely won't get any errors on such a small change".

One very quickly gets to appreciate test software...

PR #143, opened 2019-07-30
Remove solve warning

ChrisRackauckas:


TorkelE:
Ok, I understand how you mean now.

On line 671 it should be `HomotopyContinuation.iterator(` (iterator here is also a hc function). Otherwise it should be all good.

PR #149, opened 2019-08-15
Fix test dependencies and update Project.toml

devmotion:
I added some missing test dependencies. I'm not sure if there's any reason to keep the dependency on Compat (saw https://github.com/JuliaDiffEq/DiffEqBiological.jl/pull/130), since tests pass without it.


isaacsas:
@devmotion There was some problem that tests were missing that cropped up when @ChrisRackauckas removed Compat back in June. Unfortunately I've completely forgotten what it was... (I should have added an issue or test to cover...)

devmotion:
Meanwhile support for Julia 1.0 was dropped, so maybe it's not needed anymore?

isaacsas:
It could be. I think it might have been an `isnothing` issue, which shouldn't matter if support is >=1.1 now.

isaacsas:
I think that is what the problem was so I'll go ahead and merge.

PR #151, opened 2019-09-14
only require 1.0

ChrisRackauckas:


isaacsas:
Is this going to cause `isnothing` issues? That's 1.1 only right?

ChrisRackauckas:
I think we just need to add `Compat` for it. FWIW, this is now the last blocker for DifferentialEquations.jl to work on v1.1 again, so I'm going to try and find out what it needs and then we'll be fully compatible again.

ChrisRackauckas:
The /pr passes but /push didn't, so maybe something weird happened but it's good enough to know it's likely not a versioning thing.

PR #153, opened 2019-09-27
Fix dp errors

TorkelE:
While preparing a short tutorial for making bifurcation diagrams I actually discovered an error where some paths were not displayed correctly. It was basically due to be mixing up some parameters in the heuristics. Should be sorted now.

### DIFF
--- a/test/equilibrate_test_4.jl
+++ b/test/equilibrate_test_4.jl
@@ -48,7 +48,7 @@ ss_σ = steady_states(σ_network,σ_p)
 stabs_σ = stability(ss_σ,σ_network,σ_p)
 @test(sum(stabs_σ.==true)==2)
 bif_σ = bifurcations(σ_network,σ_p,:S,(0.,2.))
-if VERSION>=VersionNumber("1.1")
+if VERSION >= v"1.1"
     @test length(bif_σ.paths) == 3
 else
     @warn "Some test for bifurcation diagram path numbers are disabled for Julia 1.0 versions. Bifurcation diagrams might contain slight errors for tehse versions."
--- a/test/equilibrate_test_4.jl
+++ b/test/equilibrate_test_4.jl
@@ -27,7 +27,7 @@ brusselator_p = [1.,4.]
 @make_hc_template brusselator_network
 @add_hc_template brusselator_network
 bif_brusselator = bifurcations(brusselator_network,brusselator_p,:B,(1.,4.))
-if VERSION>=VersionNumber("1.1")
+if VERSION >=  v"1.1"
     @test length(bif_brusselator.paths) == 1
 else
     @warn "Some test for bifurcation diagram path numbers are disabled for Julia 1.0 versions. Bifurcation diagrams might contain slight errors for tehse versions."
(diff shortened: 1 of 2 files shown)


TorkelE:
Okay, I've investigated this error quite closely. Basically it works fine on Julia 1.1 and above, but errors on Julia 1.0. The reasons for this seems to sit quite deep, and partially due to behaviours in `HomotopyContinuation.jl`. It is likely possible to make it work, but will require a significant time investment, and some additional code which is unnecessary on newer julia versions.

Either we let this PR rest until we make DiffEqBiological require a later Julia version (1.1). However, I do think this change is somewhat important for bifurcation analysis, so the rest time should not get too long.

Or I make some slight modification to the test, making them a bit less strong, so the 1.0 errors slip by (they are not that serious. Basically some paths are added twice to the diagram. However, when they are plotted the identical paths overlap and the plot itself is not different. There might be a problem if someone uses the bifurcation diagram objects for non-visual purposes)

isaacsas:
I'm fine with requiring 1.1 and up, but I think @ChrisRackauckas was having problems somewhere else that had required us to still support 1.0. Maybe he can chime in.

We could make a major version number increment to indicate 1.1 is now required.

ChrisRackauckas:
It blocks the DifferentialEquations.jl tests. Maybe put an if statement on the bifurcation tests and only run them if >= v"1.1" and otherwise throw a warning that those are disabled on 1.0?

TorkelE:
> It blocks the DifferentialEquations.jl tests. Maybe put an if statement on the bifurcation tests and only run them if >= v"1.1" and otherwise throw a warning that those are disabled on 1.0?

That's sounds like a good solution.

test/equilibrate_test_4.jl
```diff
@@ -65,4 +70,22 @@ ss_cc = steady_states(cc_network,cc_p)
...
+end
+
+bs_network = @reaction_network begin
+    d,    (X,Y) → ∅
+    hillR(Y,v1,K1,n1), ∅ → X
+    hillR(X,v2,K2,n2), ∅ → Y
+end d v1 K1 n1 v2 K2 n2
+bs_p = [0.01, 1. , 30., 3, 1., 30, 3];
+bif_bs = bifurcations(bs_network, bs_p,:v1,(.1,10.))
+if VERSION>=VersionNumber("1.1")
```
ChrisRackauckas:
```suggestion
if VERSION >= v"1.1"
```

test/equilibrate_test_4.jl
```diff
@@ -45,9 +48,11 @@ ss_σ = steady_states(σ_network,σ_p)
 stabs_σ = stability(ss_σ,σ_network,σ_p)
 @test(sum(stabs_σ.==true)==2)
 bif_σ = bifurcations(σ_network,σ_p,:S,(0.,2.))
-@test length(bif_σ.paths) == 3
-
-[vcat(ss_σ...)...,vcat(ss_σ...)...]
+if VERSION>=VersionNumber("1.1")
```
ChrisRackauckas:
```suggestion
if VERSION >= v"1.1"
```

test/equilibrate_test_4.jl
```diff
@@ -24,11 +24,14 @@ brusselator_network = @reaction_network begin
     1, X → ∅
 end A B;
 brusselator_p = [1.,4.]
-brusselator_network.equilibrate_content
 @make_hc_template brusselator_network
 @add_hc_template brusselator_network
 bif_brusselator = bifurcations(brusselator_network,brusselator_p,:B,(1.,4.))
-@test length(bif_brusselator.paths) == 1
+if VERSION>=VersionNumber("1.1")
```
ChrisRackauckas:
```suggestion
if VERSION >=  v"1.1"
```

PR #155, opened 2019-10-01
Add Latexify.jl recipe.

korsbo:
This PR uses the new Latexify.jl recipe system to move Latexify's DiffEqBio support here rather than having it in Latexify.jl itself.

The PR also makes some improvements to the way that latexify deals with the noise terms and adds some customisation (https://github.com/korsbo/Latexify.jl/issues/50) . 

kwargs (= default):
- `noise = false` - add the noise terms to the deterministic equations.
- `noise_only = false` - replace the deterministic terms with the noise terms.
- `bracket = false` - surround variables with square brackets.
- `noise_var = :W` - choose a symbol to represent the noise process in the noise terms.

```julia
using Latexify
using DiffEqBiological

r = @reaction_network begin
    (hill(B, p_a, k, n), d_a), 0 ↔ A
    (p_b, d_b), 0 ↔ B
    (r_a, r_b), 3B ↔ A
end p_a k n d_a p_b d_b r_a r_b

latexify(r; noise=true)
```
\begin{align}
\frac{dA}{dt} =& \frac{p_{a} \cdot B^{n}}{k^{n} + B^{n}} - d_{a} \cdot A + \frac{r_{a}}{6} \cdot B^{3} - r_{b} \cdot A + \sqrt{\left\|\frac{p_{a} \cdot B^{n}}{k^{n} + B^{n}}\right\|} \cdot W_{1} - \sqrt{\left\|d_{a} \cdot A\right\|} \cdot W_{2} + \sqrt{\left\|\frac{r_{a}}{6} \cdot B^{3}\right\|} \cdot W_{5} - \sqrt{\left\|r_{b} \cdot A\right\|} \cdot W_{6} \\
\frac{dB}{dt} =& p_{b} - d_{b} \cdot B -3 \cdot \frac{r_{a}}{6} \cdot B^{3} + 3 \cdot r_{b} \cdot A + \sqrt{\left\|p_{b}\right\|} \cdot W_{3} - \sqrt{\left\|d_{b} \cdot B\right\|} \cdot W_{4} -3 \cdot \sqrt{\left\|\frac{r_{a}}{6} \cdot B^{3}\right\|} \cdot W_{5} + 3 \cdot \sqrt{\left\|r_{b} \cdot A\right\|} \cdot W_{6}
\end{align}

Status:
- [x] code
- [x] tests
- [ ] docs

I'm not sure where this is best documented. A main point of the Latexify recipe system is to change 'ownership' of package-specific code to the maintainers of those packages. As such, it would make more sense to have the specific documentation within those packages. 

Edit: removed the `symbolic` kwarg. It did not really make sense to keep it.


isaacsas:
Regarding docs; you could add a section to the DiffEqBio API in DiffEqDocs. Just add documentation to your new recipe function and then add a section pulling it in at:

https://github.com/JuliaDiffEq/DiffEqDocs.jl/blob/master/docs/src/apis/diffeqbio.md

korsbo:
Also, what is the preferred notation? 

dx/dt = p*x + sqrt(p)*W_1

dx = (p*x) dt + sqrt(p)*dW_1

or other? 

I'm currently assigning individual indices to the `W`'s of the different reactions (so that it is apparent when noise is coupled). 

These sort of things is pretty easy to change but I don't generally treat stochastic differential equations myself so I'm not sure what the standard notation is.

isaacsas:
I prefer the notation, 

dx_t = (p*x) dt + sqrt(p)*dW_1(t)

or 

dx(t) = (p*x) dt + sqrt(p)*dW_1(t)

korsbo:
image

isaacsas:
This looks good to me. With the new recipe is there still an option to drop the cdots?

korsbo:
image

Anything missing or mathematically incorrect?

korsbo:
I can't generate a meaningful docstring for the result of the latexrecipe (at least yet). I could put the documentation in https://github.com/JuliaDiffEq/DiffEqDocs.jl/blob/master/docs/src/models/biological.md instead.  

If so, then I don't think there is more to be done here and the PR can be merged as long as the CI passes.

korsbo:
I spotted a mistake. Hold up on merging.

isaacsas:
OK, I'll happily merge when you are done.

isaacsas:
For the docs you could also add a manual section at the end of the API documentation on Latexify and how to use it, with documentation on the DEBio specific kwargs. But putting it on the main page as you suggested is fine too.

korsbo:
It's ready as soon as AppVeyor passes. This does however seem to take an insane amount of time.

isaacsas:
I'll merge after it finishes. Probably tonight given how long it seems to take these days...

PR #156, opened 2019-10-01
- Fixes to some minor issues with bifurcation grid plotting.

TorkelE:
There were some minor issues with the plot receipts for some of the more obscure plots. These should now be fixed.


isaacsas:
Are you adding the V1.0 test check here or in the other PR? I'll go ahead and merge this one if the check to avoid test failures will be in the other (and would get rid of the failure here too).

Issue #161, opened 2019-10-17
Supporting decimal stoichiometric coefficients


jimmielin:
Hi,
I'm using DiffEqBiological.jl to generate ODEs for solving atmospheric chemistry reactions and some of the reactions in our chemical mechanisms are formulated with fractional stoichiometric coefficients, e.g.

```
{168} MVK + O3 = 0.064HO2 + 0.05RO2_R + 
       0.164OH + 0.05RCO_O2 + 0.475CO+
       0.1HCHO + 0.95MGLY + 0.351HCOOH : 		ARR(7.51e-16,1520.0,0.0);
```
(This is in [KPP](https://github.com/barronh/kpp) format which I am writing a parser for; ARR specifies the rate constant using the Arrhenius equation)

Specifying these kinds of coefficients directly into `addreaction!` result in a type mismatch. Minimal example:
```julia
using DiffEqBiological
testrn = @empty_reaction_network
addspecies!(testrn, :O2)
addspecies!(testrn, :O)
addreaction!(testrn, 2.643e-10, :(0.5O2 --> O))
```

```
MethodError: no method matching add_reactants!(::Symbol, ::Float64, ::Array{DiffEqBiological.ReactantStruct,1})
Closest candidates are:
  add_reactants!(::Union{Float64, Int64, Expr, Symbol}, !Matched::Int64, ::Array{DiffEqBiological.ReactantStruct,1}) at /.../DiffEqBiological/myMnx/src/reaction_network.jl:341
```

I have tried looking into the implementation of `add_reactants!`, but it seems like the code intrinsically relies on coefficients being `Int`: apart from the `ReactantStruct` specifying `Int64`, computing the reaction rates for the ODE using mass action law requires `factorial(coef)` (but I may be misunderstanding the code)

It would be good to have support for fractional coefficients as much of our data is in this format, however it would be possible to get rid of this in the pre-processing stage (multiplying eq. `*1000` and scaling the reaction rate constants) although this may introduce significant error to the reaction rate constants if we did `k^1000`. Hope to hear suggestions as to how this would be best approached. Thank you.

isaacsas:
Does the reactant stoichiometry always have integer coefficients?

ChrisRackauckas:
ODEs makes sense, but I am curious what this would mean in the context of jump problems: just have to require Float64 initial conditions and do discrete changes of specific floating point values? It seems odd but seems to make sense.

isaacsas:
[...]

PR #162, opened 2019-10-22
Supporting decimal stoichiometric coefficients

jimmielin:
This is a (very rough) Float-based implementation of: https://github.com/JuliaDiffEq/DiffEqBiological.jl/issues/161

It works for my particular ODE-based case but tests need to be run to ensure it does not break anything else. Alternate implementations using `Rational` or other methods are likely better suited to the problem, see #161 for discussion.


ChrisRackauckas:
We can keep the issue open, but the repo has changed so much that this should probably be closed.

PR #176, opened 2019-11-29
Fix latexification of derivatives.

korsbo:
I noticed that subscripts, e.g. `x_10`, weren't handled right and this was the answer.

### DIFF
--- a/src/latexify_recipes.jl
+++ b/src/latexify_recipes.jl
@@ -39,9 +39,9 @@
         end
     else
         if bracket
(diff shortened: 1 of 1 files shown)


korsbo:
Should I also bump the patch version so that we can tag immediately upon merge?

korsbo:
The test fails with 
```

```
is that the package manager can't find a suitable version of HomotopyContinuation.

korsbo:
This did not touch anything which is tested at all so I'll just merge it. The version bump will have to wait until we have sorted out #177.

PR #194, opened 2020-05-02
Transfer internals to ModelingToolkit

TorkelE:
I have started the work of re-making DiffEqBio to use MTK stuff.

For a starter, to keep things under control (since there is a lot of trashing old stuff going on), I have created a folder "MT_tmp" under the src folder. Everything else is untouched (except for DiffEqBio.jl, which now exports some additional stuff). In the "MT_tmp" folder I have started putting the "remade" version of files (currently "maketype.jl, "problem.jl", "reaction_network.jl").

In addition, several internal functions (within these new files) have a prefix `MT_` to keep them separate from old stuff. I also have a new macro 
```julia
macro MT_reaction_network(name, ex::Expr, parameters...
macro MT_reaction_network(ex::Expr, parameters...)
```
to use for the new version.
(when things are complete, we'll trash the old files and remove the prefixes)

Currently, there is only a very minimal version in place. The `@MT_reaction_network` macro creates a `reaction_network` with two fields only. 
```julia
reaction_system::ReactionSystem
ode_system::Union{ODESystem,Nothing}
```
The first is initialised with a reaction `ReactionSystem`, the second with `nothing`. The first time an `ODEProblem` is created, the `ode_system` field is filled out. Currently, a very basic syntax is now:
```julia
using DiffEqBiological, OrdinaryDiffEq, Plots
rn = @MT_reaction_network begin
    A, X → Y
    B, Y → X
end A B;
prob = MT_ODEProblem(rn,[1.,2.],(0.,5.),[10.,3.])
sol = solve(prob,Tsit5())
plot(sol)
```

Because the variables and parameters are created within the DiffEqBio structures, 'u0 = [X => 20.0, Y => 5.0]' doesn't work (instead you give a vector-like normal to the ODEProblem, and this is fixed there). We probably could make it work, but then reactants would become defined as variables in the main scope.

If you were to compare the new "reaction_network.jl" file to the last one it have been heavily reduced in size. I spent a considerable amount of time just deleting stuff in it. It looks good!


ChrisRackauckas:
>Because the variables and parameters are created within the DiffEqBio structures, 'u0 = [X => 20.0, Y => 5.0]' doesn't work (instead you give a vector-like normal to the ODEProblem, and this is fixed there). We probably could make it work, but then reactants would become defined as variables in the main scope.

A macro can be made that would scope those to `model.sys.states.X` etc. and reorders for the user.

>If you were to compare the new "reaction_network.jl" file to the last one it have been heavily reduced in size. I spent a considerable amount of time just deleting stuff in it. It looks good!

Awesome! For me, the validation will be running the BCR example with multitthreaded sparse matrices and hopefully seeing a speed demon there.

TorkelE:
> A macro can be made that would scope those to `model.sys.states.X` etc. and reorders for the user.

Good point, yes that shouldn't be hard. Just so I understand right, would you mean something like
```
rn = @reaction_network begin
... 
end p1 p2 p3
p = @make_paremeters rn [p2 => 2.5, p3 => 1.5, p1 => 3.]

```

TorkelE:
What is the correct notation for creating and solving a `SDEProblem` from a `ReactionSystem`? For `ODEProblem` it works well with either:
```julia
using DifferentialEquations, DiffEqBiological
rn = @MT_reaction_network begin
    A, X → Y
    B, Y → X
end A B
```
```
Oprob = ODEProblem(convert(ODESystem,rn.reaction_system),Pair.(rn.reaction_system.states,[10.,10.]),(0.,1.),Pair.(rn.reaction_system.ps,[10.,2.]))
sol = solve(Oprob)
Oprob = ODEProblem(ODEFunction(convert(ODESystem,rn.reaction_system)),[10.,10.],(0.,1.),[10.,2.])
sol = solve(Oprob)
```
When I try for SDEs, however,
```julia
Sprob = SDEProblem(convert(SDESystem,rn.reaction_system),Pair.(rn.reaction_system.states,[10.,10.]),(0.,1.),Pair.(rn.reaction_system.ps,[10.,2.]))
# sol = solve(Sprob)
```
generates a `MethodError: Cannot `convert` an object of type` and 
```julia
Sprob = SDEProblem(SDEFunction(convert(SDESystem,rn.reaction_system)),[10.,10.],(0.,1.),[10.,2.])
sol = solve(Sprob)
```
generates a `MethodError: no method matching similar(::Float64, ::Type{Float64})`.
(adding `noise_rate_prototype=...` doesn't change this)

Also, is there is no qualitative difference for which of the two options to use, right?

Finally, sorry if I am not all up to date, but the jump systems are not implemented yet (but is in the plan), right? Will there be a single `JumpSystem`, or will there be several versions for various types?

isaacsas:
> A macro can be made that would scope those to `model.sys.states.X` etc. and reorders for the user.
> 
Why a macro instead of just a function?

isaacsas:
@TorkelE Jumps are WIP for MT. I got it to the point that the rate and affect functions were generating and seemed to work, but then when I actually built a JumpProblem from them I was encountering errors where incorrect `affects!` seemed to be getting applied after a few jumps occurred. The end of the semester then hit here and I've been swamped with teaching the past couple of weeks... I should hopefully get back to it soon. That PR only handles constant rate and variable rate jumps -- mass action will be more work since they require things like (numerically) evaluating parameter expressions before constructing the system... 

The way that works, we'd need to determine and construct the jump types to pass into MT (i.e. MT analogues of constant, variable and mass action jumps). So we might still need some of the earlier parsing logic for deciding jump rate/affect functions and types.

isaacsas:
@TorkelE I'll look at the SDE issue later today and get back to you.

isaacsas:
These seem to work for me, modulo that there is a MT bug crashing the first currently (I'll put in a PR shortly):
```julia
rs = your reaction system
sprob = SDEProblem{true}(convert(SDESystem,rs),Pair.(states(rs),rand(2)),(0.,10.),Pair.(parameters(rs),rand(2)))
```
Notice, you need to specify if the problem is in or out of place, hence the `true`. It doesn't seem to work without that currently.

Alternatively you can use
```julia
sf = SDEFunction(convert(SDESystem,rs))
sprob = SDEProblem(sf,rand(2),(0.,10.),rand(2))
```

isaacsas:
OK, this should also now work with the latest MT master:
```julia
sprob = SDEProblem(convert(SDESystem,rs),Pair.(states(rs),rand(2)),(0.,10.),Pair.(parameters(rs),rand(2)))
```

isaacsas:
Well, once my PR is merged I mean.

TorkelE:
> > A macro can be made that would scope those to `model.sys.states.X` etc. and reorders for the user.
> 
> Why a macro instead of just a function?

I am not exactly sure of how the set up would work, so take this with a grain of salt, but if one wishes to use notation like `u0 = [X => 20.0, Y => 5.0]` it will generate an error (unless we make it so that `Y` and `X` is declared as variables in the scope where the reaction network is created). But maybe one could use a macro to do something hidden. If one used something like `u0 = some_function(rn[:X => 20.0, :Y => 5.0])` one might be able to avoid a macro. For many cases, I'd imagine that a normal vector `u0 = [...]` will be used, as have previously been the case.

Good to hear about the Jumps. I will keep things like they are for now, but it will be rather easy to add in whenever something is a pure mass action-reaction or not. But we will have the old code for it available at hand until we are there.

Still cannot get SDEProblem to work. Starting with
```julia
using DifferentialEquations, DiffEqBiological
rn = @MT_reaction_network begin
    A, X → Y
    B, Y → X
end A B
rs = rn.reaction_system
```
if I try
```julia
Sprob = SDEProblem{true}(convert(SDESystem,rs),Pair.(states(rs),rand(2)),(0.,10.),Pair.(parameters(rs),rand(2)))
sol = solve(Sprob)
```
or
```julia
Sprob = SDEProblem(convert(SDESystem,rs),Pair.(states(rs),rand(2)),(0.,10.),Pair.(parameters(rs),rand(2)))
sol = solve(Sprob)
```
then Julia crashes when I try too solve (citing `signal (11): Segmentation fault ...`. When I try 
```julia
sf = SDEFunction(convert(SDESystem,rs))
Sprob = SDEProblem(sf,rand(2),(0.,10.),rand(2))
sol = solve(Sprob)
```
then `solve` give me an error 
```
MethodError: no method matching similar(::Float64, ::Type{Float64})
Closest candidates are:
  similar(!Matched::JuliaInterpreter.Compiled, ::Any) at /home/torkelloman/.julia/packages/JuliaInterpreter/u3T76/src/types.jl:7
  similar(!Matched::Array{T,1}, ::Type) where T at array.jl:358
  similar(!Matched::Array{T,2}, ::Type) where T at array.jl:359
  ...
in top-level scope at playground.jl:33
...
```

isaacsas:
What if you directly pass the SDEFunction `f` and `g` fields to `SDEProblem`?

TorkelE:
Good idea, trying
```julia
using DifferentialEquations, DiffEqBiological
rn = @MT_reaction_network begin
    A, X → Y
    B, Y → X
end A B
rs = rn.reaction_system
sf = SDEFunction(convert(SDESystem,rs))
Sprob1 = SDEProblem(sf.f,sf.g,rand(2),(0.,10.),rand(2))
Sprob2 = SDEProblem(sf.f,sf.g,rand(2),(0.,10.),rand(2);noise_rate_prototype=zeros(2,2))
sol = solve(Sprob)
```
using `Sprob1` crashes things yet again, however `Sprob2` works fine. Going back to my earlier examples both crashes are fixed when I properly add `noise_rate_prototype=zeros(2,2)`
```julia
Sprob = SDEProblem{true}(convert(SDESystem,rs),Pair.(states(rs),rand(2)),(0.,10.),Pair.(parameters(rs),rand(2));noise_rate_prototype=zeros(2,2))
sol = solve(Sprob)
Sprob = SDEProblem(convert(SDESystem,rs),Pair.(states(rs),rand(2)),(0.,10.),Pair.(parameters(rs),rand(2));noise_rate_prototype=zeros(2,2))
sol = solve(Sprob)
```
however,
```julia
sf = SDEFunction(convert(SDESystem,rs))
Sprob = SDEProblem(sf,rand(2),(0.,10.),rand(2);noise_rate_prototype=zeros(2,2))
sol = solve(Sprob)
```
still yields a `MethodError: no method matching similar(::Float64, ::Type{Float64})` from the `solve()` function. But this is great help, I can continue with the SDE stuff now.

ChrisRackauckas:
The functions are made with `@inbounds` by default, so you might want to do `SDEFunction(convert(SDESystem,rs),checkbounds=true)` when testing. It sounds like we aren't setting the noise_rate_prototype by default, and that's an issue.

TorkelE:
Since the parameter and varriable fields in the MTK structures are of `Variable` type I thought we might as well include these as Symbol in the reaction network structure.

I also took the time to rename the variable one from `syms` to `vars`, thought that made more sense, but if I am missing something, feel free to correct me.

ChrisRackauckas:
`syms` is what's used in DiffEqBase for the plot naming, but if MTK is generating the functions then it should be fine.

TorkelE:
Added the ability to create `DiscreteProblem`s and `Jumpproblem`s. Example code:
```julia
using DiffEqJump, Plots, ModelingToolkit, DiffEqBiological

rn = @MT_reaction_network begin
    (A,B), X + Y ↔ XY
    mm(V,1,C), V → W
    D*X, W → V
end A B C D

u0 = [50, 50, 50, 50, 50]; p = [5.,3.,12.,2.]; tspan = (0.,10.)
dprob = MT_DiscreteProblem(rn, u0,tspan,p)
jprob = MT_JumpProblem(dprob, Direct(), rn)
using Plots; plot(sol)
```

isaacsas:
@TorkelE Do you think you'll be able to get back to this soon? Do you want some help if I have time this week?

FYI, @ChrisRackauckas  fixed the `noise_rate_prototype` issue, so you don't need to explicitly construct it for SDEs anymore.

isaacsas:
Also, might I suggest for `JumpProblem`s as long as we are breaking the interface that we make the `rn` the first argument? This is more consistent with all the other problems.

TorkelE:
I have had a look at PALC, hope to make a prototype of that during the week. Once that is done there should just be some shoring up to do. I'll write something here when I got the prototype/got stuck!

isaacsas:
For PALC support does it make more sense to add a `BifurcationSystem` that
works on top of an `ODESystem` in MT? If it doesn't exploit anything
special about the `ReactionNetwork` that would be a useful feature for
anyone using MT, and doesn't seem like it needs to be specified here. We
could then add a thin wrapper for that problem type that takes `rn`s here.

TorkelE:
Yes, there is an issue on where to put it. But a `BifurcationSystem` seems like a generally desirable thing to have.

ChrisRackauckas:
Is a BifurcationSystem needed, or is it just a function on an ODESystem which kicks out methods usable by PALC?

TorkelE:
Possibly not. Also, in that case, maybe a NonlinearSystem would be a more natural base than ODESystem? Technically we don't care about ODE stuff, just the system itself?

ChrisRackauckas:
Maybe. I am trying to think if there's something different about bifurcations that requires that you know it's dynamical. I think finding periodic orbits requires more than just having it as a f(x) = 0 system? Or the distinction might not be so big in this case.

TorkelE:
I think it is time to make significant changes here now. Unless someone protests during the day (@isaacsas @ChrisRackauckas ) I will start deleting files that don't make sense in DEB any more (like the problem ones, maketype will likely also go). I will also move things from the temporary files like "MT_tmp/reaction_network.jl" to reaction_network.jl (overwriting their previous content). The old `@reaction_network` macro will also have to go, being replaced by one that simply generates a `ReactionSystem`. 

Finally, a lot of tests will also have to be removed. We have quite a lot here which tests things like the solutions are correct for DEB generated networks, I will probably keep them here for now (although it might be more proper to move them to MTK, and simply make tests here that checks that the DSL generates `ReactionSystem`s that corresponds to what one generate in MTK).

ChrisRackauckas:
I think it's time.

TorkelE:
List of files that I believe might have to go:
- "equlibrate_utils.jl": We still haven't really figured out what to do with this, but probably don't belong here. Also, all the content will be out of date when the update hits anyway.
- "expression_utils.jl": Borderline. I am leaning towards keep, but with the update there will only be 3 functions, containing 2 lines of code each, here that is actually used. Depends on whenever there is something in "reaction_network.jl" which can sensibly be moved over here.
- "latexify_receipes.jl": This one should be moved over to MTK.
- "maketype.jl": Doesn't make sense anymore when we don't actually create a new type.
- "massaction_jump_utils.jl": Won't delete until @isaacsas confirms, but this one should belong in MTK, right?
- "network_properties.jl": Again something for MTK. Will go through it an ensure that nothing is forgotten in the move (I think most that make sense is already included).
- "plot_recipes.jl": Only for plotting bifurcation diagrams, so this one is out.
- "problem.jl": Everything except for `SteadyStateProblem` is already moved over.

This would leave 
- "DiffEqBiological.jl"
- "reaction_network.jl"
only.

isaacsas:
I think network_properties.jl should stay here and be updated as appropriate. It is the basic API, and I thought the conclusion was we should avoid polluting MT with such functions. We can add other features like merging networks, intersecting networks, etc over time there too.

isaacsas:
massaction_jump_utils.jl can go now.

isaacsas:
Maybe just merge the four functions to keep from maketype.jl into network_properties.jl. I can also do this when you are done with the other changes.

TorkelE:
> In maketype.jl I'd keep the `==`, `addreaction!`, `addspecies!`, and `addparam!` (but these need to be updated). I'm happy when you are done to take a pass at updating these along with the network_properties.jl methods.

Yes, these should probably be kept, but I think maybe make a new file for stuff relating to modifying networks (could also be a part of MTK's ReactionSystem functions, for modifying such a structure). I want to expand on this, e.g by adding a macro:
```julia
@add_reaction rn begin
    k, 0 --> X
end k
```
which allows for modifying a reaction network using the DSL. If we keep these functions they would probably belong together with such a macro.

> I think network_properties.jl should stay here and be updated as appropriate. It is the basic API, and I thought the conclusion was we should avoid polluting MT with such functions. We can add other features like merging networks, intersecting networks, etc over time there too.

I guess several of these functions would no longer fetch an actual field from the structure. However, they might yield properties which still are desirable, so could be nice to keep around as generators of whatever they used to fetch.

TorkelE:
For now, I haven't actually deleted any of the old files, but their content is no longer fetched and exported by the main "DiffEqBiological.jl" file.

TorkelE:
I have been revamping the tests, mostly making them more rigorous. Also tried order them up a bit., as well as creating a large file with 44 networks that can be used for miscellaneous testing. There are probably some errors still around, but will check tomorrow.

There are still a large number of old test files around (which are no longer called), they can be found under "test/inactive_files/". Most are either copies of other tests, but working on minimal reaction networks, or things relating to various types of SSA solving. (Minimal network tests should no longer be needed, since it will all become the same thing. But we will need tests that networks build in steps are identical to those declared in one go)

@isaacsas Could you have a look at the tests for JumpSystems? I implemented a few new ones, but it would be good if you could check if there is anything that I haven't covered, and if there is anything of the old tests that can be adapted to the update.

I have commented out the tests relating to latexify, since the package only works on the old `reaction_network`s, and not on `ReactionSystem`s. It's no big hurry as this is parallel to everything else, but at some time, could you have a look at it?

Finally, I am not fully sure I have gripped how the grouping of tests works, it would be good if someone could double-check that. (@ChrisRackauckas ?)

I think once we have settled on the tests we are almost done. I will only need to add the macro allowing one to expand on an already constructed system. The @isaacsas can have a look at the old `addspecies!` (and similar) functions, deciding what to keep, and how. Finally, we need to update network_properties file, with new query functions.

isaacsas:
Sounds great! I can take a look, hopefully over the weekend, and look at the grouping of tests too.

I think we'll need @korsbo to help with Latexify. That should probably get updated in MT though? I think there is an issue there about it. 

One way to handle extending a network would be to write a general `merge!(rs1, rs2)` routine, and then have the macro create a new ReactionSystem and merge it into the old one? I can see such a routine being useful independently, so was planning to add it to the API.

ChrisRackauckas:
>I think we'll need @korsbo to help with Latexify. That should probably get updated in MT though? I think there is an issue there about it.

Yes, it's free from MTK on the ODE/SDE part, but not directly on reactions yet.

ChrisRackauckas:
>One way to handle extending a network would be to write a general merge!(rs1, rs2) routine, and then have the macro create a new ReactionSystem and merge it into the old one? I can see such a routine being useful independently, so was planning to add it to the API.

Not sure of the utility of that if we have hierarchical systems.

TorkelE:
> I think we'll need @korsbo to help with Latexify. That should probably get updated in MT though? I think there is an issue there about it.


Was meaning to put a tag in, but seem I forgot!.

> One way to handle extending a network would be to write a general `merge!(rs1, rs2)` routine, and then have the macro create a new ReactionSystem and merge it into the old one? I can see such a routine being useful independently, so was planning to add it to the API.

> Not sure of the utility of that if we have hierarchical systems.

This might reveal my ignorance, but how do you mean hierarchical? Is that the ability to connect systems together? I still think a `merge!()` function would be a handy thing. Connections feel more like if I e.g. want to connect the networks from several neighbouring cells. But if I simply have a very large network and want to procedurally add all the degradation terms, merging them into the main network in a for loop, or similar, seems like the natural way? (while the same result, admittedly, could be made through connections). But maybe I am thinking about the wrong thing?

isaacsas:
@TorkelE i agree about merge. I think there is a reasonable case that someone may want to directly extend one network, and it is easy code to add. 

We probably need to update the MT code to handle the attached Systems correctly during conversions.

TorkelE:
@isaacsas 
Fixing with the tests, this piece of code
```julia
reaction_network = @reaction_network begin
    A, ∅ → X
    1, 2X + Y → 3X
    B, X → Y
    1, X → ∅
end A B;
u0 = rand(1:Int64(10),length(reaction_network.states))
p = factor*rand(length(reaction_network.ps))
prob1 = JumpProblem(reaction_network,DiscreteProblem(reaction_network,u0,(0.,1.),p),Direct())
```
generates a 
```
MethodError: no method matching get_variables(::Int64)
```
error. Similar with `(::Float64)` for `1, 2X + Y → 3X`. This seems like something that should not happen?

TorkelE:
In addition, these networks all generate a `type Operation has no field value` in the last line of the following code
```
reaction_network_weird = @reaction_network begin
    exp(p), ∅ → X
    d*exp(X), X → ∅
end p d

reaction_network_weird = @reaction_network begin
    k1, ∅ → X
    k2*log(12+X), X → Y
    k3*log(3+Y), Y → Z
    log(5,6+k4), Z → ∅
end k1 k2 k3 k4

reaction_network_weird = @reaction_network begin
    (k1*tanh(X1),k2), X1 ↔ 2X1
    (2+sin(X1),3+sin(k3)), X1 + X2 ↔ X3
    (k5,4+cos(X3+X1)+sin(X2)), X3 ↔ X2
end k1 k2 k3 k4 k5 k6

factor = [1e-1, 1e0, 1e1][1]
u0 = rand(1:Int64(factor*100),length(reaction_network_weird.states))
p = factor*rand(length(reaction_network_weird.ps))
dp = DiscreteProblem(reaction_network_weird,u0,(0.,1.),p)
prob = JumpProblem(reaction_network_weird,dp,Direct())
```
Admittedly the are all rather implausible, and if there are some more fundamental reason we should support them, then lets ignore these tests. 
```
dp = DiscreteProblem(reaction_network_weird,u0,(0.,1.),p)
```
works fine though.

isaacsas:
We’re probably not supporting integer/float rates correctly in MT currently. Our tests there are only for parameters and operations.

isaacsas:
Can you open a MT issue and paste in the error messages?

isaacsas:
Just a note, the Project.toml should be cut down to remove packages that are no longer needed / imported. Starting DEBio on this branch the first time is taking very long...

TorkelE:
> Just a note, the Project.toml should be cut down to remove packages that are no longer needed / imported. Starting DEBio on this branch the first time is taking very long...

Good point.

isaacsas:
@TorkelE I've got fixes for the issue you opened and will put them through shortly. The other problem, with `exp(p)` not being evaluated numerically I've opened an issue. It needs Chris or Shashi to take a look since it involves the substitution/simplification method returning an `Operation` when it should be numerically evaluating `exp(p)`.

TorkelE:
If the second issue takes time, or is not actually desired to have in specifications, we will simply put in a check which skips those tests.

isaacsas:
Maybe mark them as a broken test for now? `@test_broken`

TorkelE:
Cool macro! Didn't even know that one existed.

TorkelE:
Two changes that I made while I restructured everything:
It is now possible to create an empty system (with/without parameters) using
```julia
rn = @reaction_network 
rn = @reaction_network p1 p2 p3 p4 p5
```
The character `t` have previously not been allowed as a parameter or variable value. I expanded the concept by adding an array `forbidden_symbols`, containing a list of symbols which are not permitted as parameter/variable. Some of these might actually work now with MTK, but I decided to err on the side of caution. We can add/remove stuff if to/from it. Currently, this is the list
```julia
forbidden_symbols = [:t, :π, :pi, :ℯ, :im, :I, :nothing, :∅]
```

TorkelE:
Another thing I encountered. Unsure how to reproduce it without this branch though. All of these
```julia
using DiffEqBiological, DiffEqJump

test = @reaction_network begin
    p, ∅ → X1
    d, X1 ⟾ ∅
end p d
dprob = DiscreteProblem(test, [10], (0.0,1.0), [1.,1.])
jprob = JumpProblem(test, dprob, Direct())

test = @reaction_network begin
    p, ∅ → X1
    d, X1 ⟾ X2
end p d
dprob = DiscreteProblem(test, [10], (0.0,1.0), [1.,1.])
jprob = JumpProblem(test, dprob, Direct())

test = @reaction_network begin
    p, ∅ ⟾ X1
end p d
dprob = DiscreteProblem(test, [10], (0.0,1.0), [1.,1.])
jprob = JumpProblem(test, dprob, Direct())
```
generates a 
```
MethodError: no method matching (::JuliaVariables.var"#IS_SCOPED#64"{IdDict{Expr,Any}})(::JuliaVariables.State, ::JuliaVariables.SymRef)
```
error when creating the `JumpProblem`. It seems to happen when asking to ignore mass action, while having only a parameter in the rate. More peculiar, this works fine:
```julia
test = @reaction_network begin
    p, ∅ → X1
    1., X1 ⟾ ∅
end p d
dprob = DiscreteProblem(test, [10], (0.0,1.0), [1.,1.])
jprob = JumpProblem(test, dprob, Direct())
```
I have created an issue on it at MTK, although it required this branch to actually check.

test/runtests.jl
```diff
@@ -1,51 +1,50 @@
...
+  @time @safetestset "ODE System Solving" begin include("solve_ODEs.jl") end
+  @time @safetestset "Make Jacobian" begin include("make_jacobian.jl") end
 end
 
-if !is_APPVEYOR && (GROUP == "All" || GROUP == "EquibrilateII")
-  @time @safetestset "Equilibrate (3)" begin include("equilibrate_test_3.jl") end
-  @time @safetestset "Equilibrate (4)" begin include("equilibrate_test_4.jl") end
+# Tests related to solving Stochastic Differential Equations.
+if GROUP == "All" || GROUP == "SDE Problems"
+  @time @safetestset "SDE System Solving" begin include("solve_SDEs.jl") end
```
ChrisRackauckas, line 36: is this a major regression? Or was it taking like 30 minutes before?

TorkelE:
I'm not sure why, but Travis doesn't seem to run the actual tests, just checking that the build passed?

test/runtests.jl
TorkelE: Not sure what you mean? I scratched the equi tests since that feature is temporarily removed.

For the splitting SDE Problems I just divided things into a couple of natural groups. However, admittedly I don't really know what is the preferred way to group things! So things might seem weird to someone actually understanding it. Feel more than happy to make modifications as you seem fit.

None of the current tests should take anywhere close to 30min. Some of the solvers are not fast, but it would be very easy to remove a couple of repeats (e.g. at one point I create and run an ODEProblem over a single time unit, for 44 different networks, and does so 3 times each for random `u0`/`p`. This is probably overdoing it...)

isaacsas:
Travis has been messed up intermittently for several days now. You can always try closing and reopening the PR to see if that gets it to run properly. I've had a few times this week that it took a few tries to work.

test/runtests.jl
ChrisRackauckas: Oh I see. We usually split to hit timings on Travis, not "natural groups". The more you split the more you have to pay compile times. Normally this is only done to keep tests below 60 minutes.

TorkelE: I see, yeah, we don't have to worry about that kind of run times! I will run some timing tests and then re-group.

ChrisRackauckas:
This isn't a Travis issue. It's because new test groups were added to runtests.jl but new environment variables weren't added to .travis.yml. However, it's probably not a good idea to add those test groups anyways, so I'd revert the runtests.jl changes.

test/runtests.jl
ChrisRackauckas: IIRC they were already split at 30-40 minute groups.

TorkelE:
Sounds sensible, I'll just make it a normal set of tests for now.

TorkelE:
So it seems like the ODE solver (when solutions close to zero) occasionally runs into negative. If you then run something with non-integer hill functions you can get stuff like `(-0.0000001)^1.2`, which gives an error. Hopefully this should fix that.

TorkelE:
This seems to keep running into a `signal (11): Segmentation fault` error when I run it here. It is not at the first instance of solving ODEs, but a bit down that file.

When I run the same test on my home computer, I pass that section (line 88 to 99), but then run into a
```
julia: /buildworker/worker/package_linux64/build/src/codegen.cpp:3322: jl_cgval_t emit_invoke(jl_codectx_t&, jl_expr_t*, jl_value_t*): Assertion `(((jl_value_t*)(((jl_taggedvalue_t*)((char*)(mi) - sizeof(jl_taggedvalue_t)))->header & ~(uintptr_t)15))==(jl_value_t*)(jl_method_instance_type))' failed.

signal (6): Aborted
in expression starting 
```
somewhere one the section starting at line 103 in the ODE file. Are these issues/error messages recognised (both followed by a massive error message)?

isaacsas:
No clue. Can you narrow down which network(s) cause the error, and is it deterministic that they cause it?

TorkelE:
Not sure which networks does it. Making an update to learn now. The `signal (11): Segmentation fault` encountered by Travis is somewhere in:
```

for (i,networks) in enumerate(identical_networks_1)
    for factor in [1e-2, 1e-1, 1e0, 1e1, 1e2]
        u0 = factor*rand(length(networks[1].states))
        p = factor*rand(length(networks[1].ps))
        (i==3) && (p = min.(round.(p).+1,10))                      #If parameter in exponent, want to avoid possibility of (-small u)^(decimal). Also avoid large exponents.
        prob1 = ODEProblem(networks[1],u0,(0.,100.),p)
        sol1 = solve(prob1,Rosenbrock23(),saveat=1.)
        prob2 = ODEProblem(networks[2],u0,(0.,100.),p)
        sol2 = solve(prob1,Rosenbrock23(),saveat=1.)
        @test all(abs.(hcat((sol1.u .- sol2.u)...)) .< 100*eps())
    end
end
```
(not actually specified where).

The `signal (6): Aborted` on my machine occurs in:
```
for (i,reaction_network) in enumerate(reaction_networks_all)
    for factor in [1e-2, 1e-1, 1e0, 1e1]
        u0 = factor*rand(length(reaction_network.states))
        p = factor*rand(length(reaction_network.ps))
        in(i,[[11:20...]...,34,37]) && (p = min.(round.(p).+1,10))  #If parameter in exponent, want to avoid possibility of (-small u)^(decimal). Also avoid large exponents.
        prob = ODEProblem(reaction_network,u0,(0.,1.),p)
        solve(prob,Rosenbrock23())
    end
end
```
(the next set of tests in the same file for ODE tests)
I believe for this one that it is not always the same network which triggers it, different ones every time. But I am confirming right now.

TorkelE:
So for the thing that is happening on my computer.
`signal (6): Aborted` seems to happen when I run it on my computer, directly from the "test/solve_ODEs.jl" file in Juno. But if I run from the "test/runtests.jl" file, using the line
```
@time @safetestset "ODE System Solving" begin include("solve_ODEs.jl") end
```
I get something different:
```
ERROR: LoadError: UndefVarError: curly not defined
Stacktrace:
 [1] from_type at /home/travis/.julia/packages/GeneralizedGenerated/4p9fF/src/ngg/typeable.jl:43 [inlined]
 [2] from_type at ./none:0 [inlined]
 [3] interpret(::Type{TypeEncoding(fatal: error thrown and no exception handler available.
UndefVarError(var=:curly)
```
This one is copied from Travis, because the error message disappeared when I re-run things in Juno. 

Which brings us to the next observation, rerunning Travis with the checking, I actually passed the first problem (which generated a `signal (11): Segmentation fault`), and I reach that error above (this is not consistent, running Travis again I get stuck at a `signal (11): Segmentation fault` error). This error (the `curl` one), was generated at the same place on my machine and Travis, at network number 27, that is:
```
reaction_networks_constraint[7] = @reaction_network begin
    (k1,k2), X1 + X2 ↔ X3
    (mm(X3,v,K),d), ∅ ↔ X2
    (k3,k4), X2 ↔ X4
end k1 k2 k3 k4 v K d
```

Finally, making things more confusing, when I rerun the line
```
@time @safetestset "ODE System Solving" begin include("solve_ODEs.jl") end
```
but with a `(i<27) && continue`, skipping directly to the fault network. And then things go fine, the network is processed without error. As a last test, I extracted network 27, and run the code in a loop (30 times) without getting any error.

It should be pointed out that tests are partly random, as parameter values and initial conditions are randomized each time.

Now you probably are as confused as I, and maybe more, but I will need to go to bed...

TorkelE:
Some further testing. As stated previously, this section:
```julia
for (i,networks) in enumerate(identical_networks_1)
    for factor in [1e-2, 1e-1, 1e0, 1e1, 1e2]
        u0 = factor*rand(length(networks[1].states))
        p = factor*rand(length(networks[1].ps))
        (i==3) && (p = min.(round.(p).+1,10))                      #If parameter in exponent, want to avoid possibility of (-small u)^(decimal). Also avoid large exponents.
        prob1 = ODEProblem(networks[1],u0,(0.,100.),p)
        sol1 = solve(prob1,Rosenbrock23(),saveat=1.)
        prob2 = ODEProblem(networks[2],u0,(0.,100.),p)
        sol2 = solve(prob1,Rosenbrock23(),saveat=1.)
        @test all(abs.(hcat((sol1.u .- sol2.u)...)) .< 100*eps())
    end
end
```
in the test file "solve_ODEs.jl" gives an error along the lines of 
```
julia: /buildworker/worker/package_linux64/build/src/codegen.cpp:3322: jl_cgval_t emit_invoke(jl_codectx_t&, jl_expr_t*, jl_value_t*): Assertion `(((jl_value_t*)(((jl_taggedvalue_t*)((char*)(mi) - sizeof(jl_taggedvalue_t)))->header & ~(uintptr_t)15))==(jl_value_t*)(jl_method_instance_type))' failed.

signal (6): Aborted
```
when I run it on my computer. I added loads of printouts and ran the line:
```
@time @safetestset "ODE System Solving" begin include("solve_ODEs.jl") end
```
in the "runtests.jl" file. It gave the error three times out of three, on the same network (4), but different values of factor (1e1, 1e1, 1e2). The exact line which errored seemed to be
```
sol1 = solve(prob1,Rosenbrock23(),saveat=1.)
```
I also printed out the exact u0 and parameter values used when causing the error:
```

u0 = [8.099349422497452, 6.333100695616949, 1.4765544644975193]
p = [9.415655245389194, 8.307789433359286, 3.36644613194101, 8.437192310247383, 6.57817378448593, 6.956958189681471]

[3.96661099707847, 7.3033852728967075, 7.3116042406695865]
[1.1083132568332643, 2.0028556039724177, 0.63746817330391, 8.908852945702053, 9.321217578202397, 0.07881708444586799]

[40.84531801687139, 6.700986453154512, 59.56490413332472]
[41.71897919872871, 16.925918189369106, 93.65545167751763, 27.287325093047343, 68.9737023180762, 14.243489959501044]
```
for the three different times, respectively. I tried to rerun the erogenous section only, with those exact values:
```juliausing DiffEqBiological, OrdinaryDiffEq, Random, Test
include("test_networks.jl")

function real_functions_4(du,u,p,t)
    X1,X2,X3 = u
    k1,k2,k3,k4,k5,k6 = p
    du[1] = -k1*X1 + k2*X2 + k5*X3 - k6*X1
    du[2] = -k3*X2 + k4*X3 + k1*X1 - k2*X2
    du[3] = -k5*X3 + k6*X1 + k3*X2 - k4*X3
end
networks =  reaction_networks_constraint[1] => real_functions_4

u0 = [8.099349422497452, 6.333100695616949, 1.4765544644975193]
p = [9.415655245389194, 8.307789433359286, 3.36644613194101, 8.437192310247383, 6.57817378448593, 6.956958189681471]

prob1 = ODEProblem(networks[1],u0,(0.,100.),p)
sol1 = solve(prob1,Rosenbrock23(),saveat=1.)
prob2 = ODEProblem(networks[2],u0,(0.,100.),p)
sol2 = solve(prob1,Rosenbrock23(),saveat=1.)
```
and receive no error!

Finally, I added a line:
```
identical_networks_1[3] = reaction_networks_constraint[1] => real_functions_4
```
This should take the fourth network (which generated the error), and copy it into the third position of the loop as well (which previously was fine). Running this, again I received the error for the fourth network (while successively passing the third network, which now was identical to the fourth one).

All this suggests that it is not only the network itself that matters but something else. What? No idea.

isaacsas:
Does the error happen with a different solver? It would help to know more clearly where in the `solve` call it is crashing.

TorkelE:
Good point. I will try a variety.

TorkelE:
Does not seem to be a solver issue. I tried several (probably grossly violating what kind of solver would be appropriate for the situation.

`Rosenbrock23`, `Rodas4p`, and `KenCarp4` all generate the same error as described.
`TRBDF2`, `Rodas5`, and `Kvaerno5` generated test errors probably exceeding the set test tolerance).

ChrisRackauckas:
If you replace that line in MTK with `eval` it should work for testing.

TorkelE:
Unfortunately no success here either. I tried modifying the line to
```julia
f_oop,f_iip = eval(generate_function(sys, dvs, ps; expression=Val{false}, kwargs...))
```
, if that's the right one.

On the plus side, I think I have managed to construct a somewhat minimal example, to help to nail down where the trouble is.
```julia
using ModelingToolkit, OrdinaryDiffEq

@parameters t d
@variables X(t)
@derivatives D'~t
eqs = [D(X) ~ 1 - d*X]
sys = ODESystem(eqs)

for i = 1:500
    print("\n",i) #Prints the number of iteration. The crash should happen at the same point every time.
    prob = ODEProblem(sys,[X => 1.0],(0.,1.),[d => 2.0])
    sol = solve(prob,Rosenbrock23())
    #print("          ",sol.u[end]) #This prints the same value each time.
end
```
When I run that one I get a crash at iteration 217. The error goes something like what follows. I'll make a report at modelling toolkit as well.
```
ulia: /buildworker/worker/package_linux64/build/src/codegen.cpp:3322: jl_cgval_t emit_invoke(jl_codectx_t&, jl_expr_t*, jl_value_t*): Assertion `(((jl_value_t*)(((jl_taggedvalue_t*)((char*)(mi) - sizeof(jl_taggedvalue_t)))->header & ~(uintptr_t)15))==(jl_value_t*)(jl_method_instance_type))' failed.

signal (6): Aborted
...
```

isaacsas:
Sounds like you should mark these as test broken and move on. Thanks for suffering through it to narrow it down!

ChrisRackauckas:
yes, I would mark as broken and keep moving. We should find the full scale of what needs fixing.

TorkelE:
Sounds good. Curious see know what the issue ends up being. If I ever kept a list of weird bugs I've encountered, I might have added this one...

TorkelE:
Unfortunately, the issue seems rather endemic, seeming to appear whenever I create/solve a large number of differential equation problems. This includes all of the test files. More importantly, it amplifies as I run the various test files in succession. In the end, I have had to disable almost all of the tests, just to be able to run through the entire test set.

ChrisRackauckas:
okay so it's really all just one issue left, this crashing, but it shows up everywhere?

TorkelE:
That's a good summary, yes.

ChrisRackauckas:
The upstream fixes are tagged now. Rerunning CI

src/reaction_network.jl
```diff
@@ -57,287 +57,119 @@ See the [Chemical Reaction Model
...
-funcdict = Dict{Symbol, Function}()                             #Stores user defined functions.
-
-#Coordination function, actually does all the work of the macro.
-function coordinate(name, ex::Expr, p, scale_noise)
+pure_rate_arrows = Set{Symbol}([:⇐, :⟽, :⇒, :⟾, :⇔, :⟺])
 
-    # minimal reaction network components
-    (reactions, reactants, parameters, syms, params) = get_minnetwork(ex, p)
+# Declares symbols which may neither be used as paraemters not varriables.
+forbidden_symbols = [:t, :π, :pi, :ℯ, :im, :I, :nothing, :∅]
```
isaacsas, line 69: Maybe we should allow `I`? It is very common in SIR models (for example), which are already in much of the documentation.

TorkelE:
It could be this section
```julia
for (i,network) in enumerate(reaction_networks_all)
    for factor in [1e-2, 1e-1, 1e0]
        u0 = factor*rand(length(network.states))
        p = factor*rand(length(network.ps))
        in(i,[[11:20...]...,34,37]) && (p = min.(round.(p).+1,10))  #If parameter in exponent, want to avoid possibility of (-small u)^(decimal). Also avoid large exponents.
        prob = ODEProblem(network,u0,(0.,1.),p)
        solve(prob,Rosenbrock23())
    end
end
```
which runs for a while without producing output. On my computer the entire "solve_ODEs.jl" file runs in about 7 minutes, but I will try to modify into 
```julia
@test solve(prob1,SSAStepper()).retcode == :Success
```
and see if that helps.

TorkelE:
It could be this section
```julia
for (i,network) in enumerate(reaction_networks_all)
    for factor in [1e-2, 1e-1, 1e0]
        u0 = factor*rand(length(network.states))
        p = factor*rand(length(network.ps))
        in(i,[[11:20...]...,34,37]) && (p = min.(round.(p).+1,10))  #If parameter in exponent, want to avoid possibility of (-small u)^(decimal). Also avoid large exponents.
        prob = ODEProblem(network,u0,(0.,1.),p)
        solve(prob,Rosenbrock23())
    end
end
```
which runs for a while without producing output. On my computer the entire "solve_ODEs.jl" file runs in about 7 minutes, but I will try to modify into 
```julia
@test solve(prob1,SSAStepper()).retcode == :Success
```
and see if that helps.

ChrisRackauckas:
Just add some printing

src/reaction_network.jl
TorkelE: Good point. I just had some memory of this giving weird errors back in the beginning and added it to the lsit. But you're right, if there's no error to it then given its prominence in SIR (and the docs) we should allow it.

isaacsas: It will error if the `LinearAlgebra` package is loaded. What functionality from it are we using in the main package now? Can we just import what's needed?

isaacsas: If you want to leave this for now I can get it when I take over... You've already done plenty!

TorkelE: Don't know. Trying to remove it right now, to check what happens.

TorkelE: Seems fine actually. Likely we can just remove `LinearAlgebra`.

TorkelE:
Whenever there where networks with fixed concentrations, like `X <--> Y`, the steady state solver would occasionally give funny results. The tests now avoid such networks when testing steady state solving.

isaacsas:
Is that due to the Jacobian being singular?

TorkelE:
> Is that due to the Jacobian being singular?

No idea really.

.travis.yml
```diff
@@ -12,12 +12,12 @@ julia:
 matrix:
   allow_failures:
     - julia: nightly
-env:
-  - GROUP=Core
-  - GROUP=Min
-  - GROUP=EquibrilateI
-  - GROUP=EquibrilateII
-  - GROUP=Misc
+#env:
```
ChrisRackauckas, line 15: remember to put these back in place?

TorkelE: Do you mean the equilibrate tests or the groups? The former or not valid anymore since that doesn't work with the new MTK implementation. Or do we want to use groups again? The current time is about 40 minutes, which I guess is borderline.

ChrisRackauckas: oh, if the total time is now 40 minutes, then yeah we're good.

isaacsas:
@TorkelE is this done now?

Project.toml
```diff
@@ -7,30 +7,23 @@ Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
...
-HomotopyContinuation = "f213a82b-91d6-5c5d-acf7-10f1c761b327"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
-Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
+ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
-SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
-SymEngine = "123dc426-2d89-5057-bbad-38513e3affd8"
```
isaacsas, line 19: The single best line of this PR.

ChrisRackauckas: glorious

isaacsas:
@TorkelE awesome effort, and great job with the insanely comprehensive set of networks you’ve brought in for testing. Thanks!

ChrisRackauckas:
Yes, this is probably the best stress test of the jump system that we could ask for

PR #199, opened 2020-07-07
Update api for new MT-based DEBio

isaacsas:


isaacsas:
I'll drop them in the next PR, where I'll update the README too. I want to take one last pass for functionality to update after this gets merged.

ChrisRackauckas:
okay I'll take a deep look in a sec

src/networkapi.jl
```diff
@@ -0,0 +1,228 @@
...
+- *Does not* currently simplify rates, so a rate of `A^2+2*A+1` would be considered
+different than `(A+1)^2`.
+"""
+function (==)(rn1::ReactionSystem, rn2::ReactionSystem)
+    issetequal(species(rn1), species(rn2)) || return false
+    issetequal(params(rn1), params(rn2)) || return false
+    isequal(rn1.iv, rn2.iv) || return false
+    (numreactions(rn1) == numreactions(rn2)) || return false
+    issetequal(equations(rn1), equations(rn2))
+    for sys1 in rn1.systems, sys2 in rn2.systems
```
isaacsas, line 120: This loop should probably be changed to using `issetequal`. I can update that tonight.

src/networkapi.jl
```diff
@@ -0,0 +1,228 @@
...
+        (sy1 == sys2) || return false
```
ChrisRackauckas, line 121: and typo

ChrisRackauckas:
Looks great to me.

isaacsas:
I'll let tests finish and then merge. I had to use `issubset` in two directions instead of `issetequal` for some reason I don't understand, but otherwise this seems ok now.

src/networkapi.jl
```diff
@@ -0,0 +1,234 @@
...
+different than `(A+1)^2`.
+"""
+function (==)(rn1::ReactionSystem, rn2::ReactionSystem)
+    issetequal(species(rn1), species(rn2)) || return false
+    issetequal(params(rn1), params(rn2)) || return false
+    isequal(rn1.iv, rn2.iv) || return false
+    (numreactions(rn1) == numreactions(rn2)) || return false
+    
+    # the following fails for some reason, so need to use issubset
+    #issetequal(equations(rn1), equations(rn2)) || return false
```
isaacsas, line 122: This doesn't work for some reason, but testing with `issubset` in both directions seems ok.

ChrisRackauckas: odd. I've never used `issetequal` before (didn't know it existed before this PR haha) so I am not sure why it would act like that.

isaacsas: It’s generally a great method but something is going wrong here. I did learn it converts to sets, so maybe it is something with that.

TorkelE:
Looks good!

PR #200, opened 2020-07-10
Enable network expansion macros

TorkelE:
This builds on https://github.com/SciML/DiffEqBiological.jl/pull/199 by allowing you to reach some of the functionality using the macro. Now:
```julia
rn = @reaction_network
```
creates an empty reaction network.
```julia
rn = @reaction_network p1 p2 p3
```
creates an empty reaction network, and loads it with the parameters `p1`, `p2`, and `p3`.

Also, the macro `@add_reactions` is added, where
```julia
rn = @reaction_network begin
   (k1,k2), X ↔ Y
end k1 k2
@add_reactions rn begin
   (p,d), X ↔ ∅
end p d
```
is equivalent to
```julia
rn = @reaction_network begin
   (k1,k2), X ↔ Y
   (p,d), X ↔ ∅
end k1 k2 p d
```
parameters can be declared repeatedly without worry, however, whenever reactions are added, all parameters used have to be marked. E.g. this
```julia
rn = @reaction_network begin
   (k1,k2), X ↔ Y
end k1 k2 p d
@add_reactions rn begin
   (p,d), X ↔ ∅
end p d
```
is fine, while 
```julia
rn = @reaction_network begin
   (k1,k2), X ↔ Y
end k1 k2 p d
@add_reactions rn begin
   (p,d), X ↔ ∅
end 
```
is not.

### DIFF
--- a/src/DiffEqBiological.jl
+++ b/src/DiffEqBiological.jl
@@ -15,7 +15,7 @@ include("expression_utils.jl")
 include("reaction_network.jl")
 
 # reaction network macro
-export @reaction_network
+export @reaction_network, @add_reactions
 export @reaction_func
 
 
--- a/src/networkapi.jl
+++ b/src/networkapi.jl
@@ -86,10 +86,10 @@ end
 """
     ==(rn1::ModelingToolkit.Reaction, rn2::ModelingToolkit.Reaction)
 
-Tests whether two `ModelingToolkit.Reaction`s are identical. 
+Tests whether two `ModelingToolkit.Reaction`s are identical.
 
 Notes:
-- Ignores the order in which stoichiometry components are listed. 
+- Ignores the order in which stoichiometry components are listed.
 - *Does not* currently simplify rates, so a rate of `A^2+2*A+1` would be
 considered different than `(A+1)^2`.
 """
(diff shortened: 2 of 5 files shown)


isaacsas:
Looks good! Is it not possible to avoid having to list reused parameters again?

TorkelE:
Well, it _should_ be possible, but it might be difficult. Essentially, what `@add_reactions` currently does is to create a new reaction network of the declared reactions, and then call `merge!()` on the previous and the newly create reaction networks (the code is thus very simple).

Right now the macros need to know which symbols are parameters, this is required when converting the code into `ReactionSystem(...)`. Since the previous reaction network is inputted simply as a symbol (`rn`), it does not contain information of what parameters are contained.

To achieve this one would need to create some new merge function, taking a reaction network, and some kind of structure generated by the `@add_reactions` macro (which does not commit to what is a parameter and what is a variable).

An alternative would be to assume that all symbols used by the `@add_reactions` macro are parameter, unless explicitly used as a variable, but this felt dangerous.

So, I think this could be done, but it would require taking a proper dive into metaprogramming to figure out exactly how to do it.

ChrisRackauckas:
I think it's good enough as a start. We can keep fiddling with it though. It's quite a nice API for it!

isaacsas:
I agree! I was just curious (and figured it might be an issue as @TorkelE mentioned).

PR #202, opened 2020-07-12
update more API functions

isaacsas:
This should be the last bunch and then we're done modulo API docs, README and tutorials.


ChrisRackauckas:
@kanav99 this repo will need a doc deployment as well.

isaacsas:
Aren't we leaving the docs as part of DiffEq?

ChrisRackauckas:
I think we'll want multiple pages, right?

What we'll be doing is have a nice overarching page that brings all of the docs together, but that's not here yet. I think we might as well split it off and get this fully documented, and then solve the question of pulling together 20 complete docs all at once.

isaacsas:
Sure, that sounds good. So we'll move both the overall "Chemical Models" section here, along with the API?

ChrisRackauckas:
Yes, and probably ReactionSystem as well, though @shashi might want to do the pins/observables.

isaacsas:
What are the pins/observables? Should I make a PR to MT that removes RS and then add them here? I wanted to put through doc updates there anyways so now is a good time.

isaacsas:
@TorkelE I seem to be getting failure in steady-state tests?

TorkelE:
Will have a look. They caused some troubles before, thought it was fixed, but seems something slipped through.

TorkelE:
Seems one of the networks tested has oscillations for certain parameter values, which makes it a poor choice for the test. 

I can update after this PR is done, alternatively, you can just swap line 42 in  the "steady_state_problems.jl" file for
```julia
steady_state_test_networks = [reaction_networks_standard[8], reaction_networks_standard[10], reaction_networks_weird[1]]
```

isaacsas:
Thanks! I’ll add that.

isaacsas:
@ChrisRackauckas I updated the license to remove the old "biologicalmodels.jl" name, and just have no name listed (since it is part of the repository anyways). I also removed the quoting of the license, which should allow Github to pick up what type of license it is. Is this ok to you?

isaacsas:
OK, I updated the license to MT format. @ChrisRackauckas can you let me know if this is ok with you? I'll then merge.

isaacsas:
Yeah, I don't think that is a big deal. If we were to make any change I'd like to tweak it so it registers properly with Github as an MIT license...

PR #205, opened 2020-07-14
Time for Catalyst

isaacsas:


isaacsas:
@TorkelE @ChrisRackauckas Here is a first draft for an updated README. I've currently commented out the steady-state and bifurcation material, along with the `ReactionNetworkImporters` material. As these get updated we can add them back.

isaacsas:
One more read through here and the new docs in DiffEqDocs, and then I'm ok with making the switch. (Though it would be nice to have Latexify updated too.)

Both in the README and in the docs I've made the name change to Catalyst it remains to search and replace throughout the various Project tomls and src files, along with changing over the repo. I'm not sure how we want to order switching here, and then pulling Catalyst into the Docs (which I guess needs a Catalyst release?). We could add a disclaimer to the README that the docs are not yet updated while everything propagates. @ChrisRackauckas thoughts?

ChrisRackauckas:
We can just propagate it fast enough that the disclaimer isn't needed. I'd rather just get it all lined up and pull the trigger.

README.md
```diff
@@ -12,56 +12,81 @@
...
 
-More detailed documentation is available from:
-* Several DiffEqBiological tutorials are available as part of the
+Here is a simple example of generating and solving an SIR ODE model:
+```julia
+using Catalyst, DiffEqBase, OrdinaryDiffEq
+rn = @reaction_network begin
+    k1, S + I --> R
+    k2, I --> R
+    end k1 k2
```
TorkelE, line 41: There's a tab before the "end"

README.md
```diff
@@ -183,38 +245,60 @@ rs = @reaction_network begin
...
 # solve JumpProblem using Gillespie's Direct Method
 u0 = [5]
-dprob = DiscreteProblem(rs, u0, tspan, params)
+dprob = DiscreteProblem(rs, u0, tspan, p)
 jprob = JumpProblem(dprob, Direct(), rs)
 jsol = solve(jprob, SSAStepper())
 ```
 
-## Importing Predefined Networks 
+Finer control over the generation of such models can be obtained explicitly
```
TorkelE, line 271: ...obtained "by" explicitly... ?

isaacsas:
Should I go ahead and update DiffEqBiological -> Catalyst in the code and Project.toml too then? (Is just changing the package name in the Project.toml the right way to handle the switch or do we need to generate a new uuid?)

ChrisRackauckas:
Yes, then after merging we change the repo name and all should be good.

isaacsas:
Assuming tests pass ok I'm good we kicking this off.

isaacsas:
@ChrisRackauckas I leave the rest to you :)

isaacsas:
Oh, I missed a test failed... @TorkelE looks like another SS issue...

TorkelE:
There seem to be quite heavy restrictions on which kind of networks the steady-state solver works on. I will have a look at what's the trouble this time.

isaacsas:
The failures seem pretty inconsistent too. 🤷🏻‍♂️

TorkelE:
I think there's a general patter that occasionally when you have networks with constraints (e.g. with `X <--> Y`), the steady state solvers fails. It seems that the same error (sometimes) occurred in the network `(k5,k6), ∅ ↔ X3 + X4` (my guess is that it is all networks where the steady state depends on the initial condition, and even when the same `u0` is given as input to the steady state solver and the ODE, something happens).

I will have a replacement test soon.

TorkelE:
Take what starts at line 6 of the test file and replace with
```julia
### Compares to netowork with know steady state ###
steady_state_network_1 = @reaction_network begin
    (k1,k2), ∅ ↔ X1
    (k3,k4), ∅ ↔ 3X2
    (k5,k6), ∅ ↔ (X3,X4)
    (k7,k8), X3 + X4 ↔ X5
end k1 k2 k3 k4 k5 k6 k7 k8

for factor in [1e-1, 1e0, 1e1], repeat = 1:3
    u0 = factor*rand(length(steady_state_network_1.states));
    p = 0.01 .+ factor*rand(length(steady_state_network_1.ps))
    prob = SteadyStateProblem(steady_state_network_1,u0,p)
    sol = solve(prob,SSRootfind()).u
    (minimum(sol[1:1]) > 1e-2) && (@test abs.(sol[1] - p[1]/p[2]) < 0.01)
    (minimum(sol[2:2]) > 1e-2) && (@test abs.(sol[2]^3/factorial(3) - p[3]/p[4]) < 0.01)
    (minimum(sol[3:5]) > 1e-2) && (@test abs.(sol[3]*sol[4]/sol[5] - p[8]/p[7]) < 0.01)
end
```
(The last line of the reaction network is replaced with 2 new, u0[3] and u0[4] are no longer set to be equal, and the last test is different)

isaacsas:
Since it seems like it went through ok this time let's save that for a separate PR once this is merged.

isaacsas:
@ChrisRackauckas Now good to go.

PR #208, opened 2020-07-17
New UUID for name change

ChrisRackauckas:


isaacsas:
I think we need to take another look at our tests — they seem to be prone to intermittent failures for some reason. A mix of inconsistency with the results in obtaining the requested accuracy, and perhaps the shear memory size of the many networks that are loaded?

TorkelE:
Which one failed this time?

isaacsas:
The SDE tests.

isaacsas:
But there was no info on where the failure occurred in the file that I saw on quick glance. The crash was in general stdlib code supposedly.

isaacsas:
Actually it looks like it finished SDEs but crashed before printing anything for jumps.

TorkelE:
Ok, that's weird. Yes, those should be fine.

PR #212, opened 2020-07-18
Add docs

isaacsas:
I think this should be the basic doc structure. It works / looks ok locally. I just mimicked the `make.jl` for `DiffEqDocs` aside from not having the Google Analytics ID or URLs for deployment. So this should now be set to deploy by someone who knows how to do that.


isaacsas:
Once this is up on the web I'll put in a PR at DiffEqDocs to remove the DiffEqBio docs (or perhaps we should then just link here from them?).

isaacsas:
Merging as this didn't touch anything except the new docs folder...

PR #225, opened 2020-07-20
add back latexify and update docs

isaacsas:
I can't get it to work in the docs due to mhchem issues with Katex, but it works in Jupyter and at the commandline.


isaacsas:
Just to add, this is hacky -- I dispatched a couple internal Latexify functions. Once @korsbo is back he can hopefully help us clean this up.

PR #228, opened 2020-07-21
Generation of graphviz graphs

isaacsas:
Some caveats:
1. Adds Catlab dependency since `GraphViz.jl` is no longer maintained. This means users must separately install the `GraphViz` command line tools.
2. Of the Lightgraph plotting tools very few seem to be actively maintained. Maybe we could use `GraphPlot.jl` instead.
3. Like all the graph libraries it does not generate graph windows with figures from the REPL, only saveable text. Figures can be saved, or generated within Jupyter.
Ex:
```julia
using Catalyst, Catlab.Graphics.Graphviz
rn = @reaction_network begin
           α, S + I --> 2I
           β, I --> R
       end α β
Graph(rn)
```
in Jupyter gives:
myfile2

Here orange nodes denote reactions and the labels on edges give the stoichiometry. Connections coming through the `rate` won't appear (i.e. Hill functions or such). We should think about how to visualize such non-constant rate expressions.

To save a graph one can just say:
```julia
savegraph(Graph(rn), filename, format)
```
where format is something like "png" (the default).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -10,6 +10,7 @@ ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
 [compat]
+Catlab = "0.7.2"
 Latexify = "0.13.5"
 MacroTools = "0.5"
 ModelingToolkit = "3.14"
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "479239e8-5488-4da2-87a7-35f2df7eef83"
 version = "5.0.1"
 
 [deps]
+Catlab = "134e5e36-593f-5add-ad60-77f754baafbe"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
--- a/src/Catalyst.jl
+++ b/src/Catalyst.jl
@@ -25,4 +25,6 @@ export dependants, dependents
 # for Latex printing of ReactionSystems
 include("latexify_recipes.jl")
 
+include("graphs.jl")
+
 end # module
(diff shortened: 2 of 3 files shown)


isaacsas:
@TorkelE @ChrisRackauckas Any thoughts? Too hacky or reasonable to add for now, with an eye towards improving as we go?

ChrisRackauckas:
Ehh, I think it's a good feature to have with a nice API, so let's do it like this for now and we might switch out how `Graph` and `savegraph`  work in the future, but that's all under the hood stuff.

ChrisRackauckas:
@jfairbanks comments?

We should also talk about adding tensoral reaction systems for representing spatial networks.

isaacsas:
I'd really like to get an interface for spatial problems together (in fact this is part of our proposal...). It needs to flexibly handle structured grids and more graph-like unstructured grids, while allowing for identical reaction rates across all site and varying reaction rates...

ChrisRackauckas:
Let's get an issue open on it. I think for what you want a tensor product formulation against (sparse) matrices is enough, since then the graphs can be represented by the adjacency matrix and structured grids are just structured matrices and can be specialized on by dispatch.

jpfairbanks:
I think this looks good, as far as spatial problems, we should get together on integration and interfaces. For the Catlab implementation of GraphViz it is better for this purpose than the graphplot stuff because it is designed for these kinds of annotated diagrams in a mathematical modeling context rather than plotting in the graph analytics context of LG and GraphPlot.

isaacsas:
We also need some work on spatial interfaces for jump problems, since there are spatial specific SSAs and they'll be different for unstructured vs structure grids (not needing to store as many jumps for the former).

TorkelE:
I think this looks quite good!

I have thought a lot about how to properly represent reaction networks graphically, and I'm starting to think that there's no good solution. It might require a few different schemes, and people would select whichever best suits their network. But currently, this looks both good and useful.

How does it handle production/degradation?

isaacsas:
No idea :) Probably we should make a blank state for that.

isaacsas:
Actually it works fine. Phew.

isaacsas:
Here's an example:
myfile2

TorkelE:
Looks good!

isaacsas:
@TorkelE I agree -- there is no single best type of graph. For example, one might want various dependency graphs that just connect species to species or rxs to rxs, or one might want graphs that show the rate law dependency of each rx, etc...

PR #232, opened 2020-07-23
Add noise scaling tests.

TorkelE:
At the risk of overdoing the tests...

This basically tests the noise scaling functionality recently added in MTK, but "in action". That is, it simulates a network with different levels of noise scaling and tests that the variation of the distributions scales properly.

### DIFF
--- a/test/solve_SDEs.jl
+++ b/test/solve_SDEs.jl
@@ -89,11 +89,15 @@ end
 push!(identical_networks, reaction_networks_constraint[9] => (real_f_3,real_g_3,zeros(7,6)))
 
 for (i,networks) in enumerate(identical_networks)
+    println(i)
     for factor in [1e-1, 1e0, 1e1], repeat in 1:3
+        println(repeat, "   ", factor)
         u0 = 100. .+ factor*rand(length(networks[1].states))
         p = 0.01 .+ factor*rand(length(networks[1].ps))
         (i==2) && (u0[1] += 1000.)
         (i==3) ? (p[2:2:6] .*= 1000.; u0 .+= 1000) : (p[1] += 500.)
+        println(u0)
+        println(p)
         prob1 = SDEProblem(networks[1],u0,(0.,100.),p)
         sol1 = solve(prob1,ImplicitEM(),saveat=0.01,maxiters=1e7)
         prob2 = SDEProblem(networks[2][1],networks[2][2],u0,(0.,100.),p,noise_rate_prototype=networks[2][3])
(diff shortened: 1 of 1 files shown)


TorkelE:
On a side note, I have been trying to figure out why there's occasional crashes in the tests. I have been rerunning the tests in a long `for` loop until Julia suddenly just stops. 

My hypothesis have been that there's some obscure case in the SDE tests where the solution goes into negative and causes an error (this should have been accounted for when the test was constructed). However, when I run the SDE tests just a lot of time, and eventually get a crash, and investigate the parameters which yielded the crash, it is fine. When I got a crash in `runtests.jl` it was also kinda random. Still not sure what is happening.

ChrisRackauckas:
https://travis-ci.com/github/SciML/Catalyst.jl/builds/176849991#L526

Is that the "crash" you're talking about? Error reports should be more specific, because just saying crash could mean anything. If it's this, then yeah something went negative and the next MTK release should handle that.

TorkelE:
Ok, that travis report is more helpful, yes, that was what I guessed went wrong.

The reason I used the very ambiguous "crash" was that when I wrapped the runtests files, or the solve_SDE tests file, in a `for` loop, then after a while Juno had simply exited mid-run. There was no error message or anything similar to give any hint what had happened. But obviously, something had gone wrong.

isaacsas:
It might be good to pare down the shear number of networks we test some though. Test really take a long time currently, with some test files allocating > 50GB of memory.

ChrisRackauckas:
We can split tests too. But yes, using one network to test many features is usually a good idea unless we're testing network generation itself.

isaacsas:
Or maybe there is a way to setup a standard vs full set of tests, where standard is what runs normally and is more compact, but we can occasionally trigger full to run more networks.

ChrisRackauckas:
If we split test groups, then we could do it like OrdinaryDiffEq where the full tests are only ran on CI and a subset are ran locally.

TorkelE:
Trying to reduce the rather excessive number of tests would be good. They were quite useful when written since they caught a few things. But at this stage, they mostly extend the test time. A shame to remove them though might come in useful.

TorkelE:
At a closer look, it turned out that this error was actually in the new tests that I added, and that these had a small chance to go into negative. I have modified to ensure that the chance of this happening is practically zero.

TorkelE:
Ok, I think I might finally be picking up a trail on what is going on here. Hopefully, I will have it nailed down soon.

TorkelE:
Ok, I think I have narrowed it down to something odd going on for certain SDE solvers https://github.com/SciML/StochasticDiffEq.jl/issues/344

isaacsas:
@TorkelE Looks like tests have now passed. OK to merge?

ChrisRackauckas:
yeah it should be good.

PR #234, opened 2020-07-28
add a first tutorial

isaacsas:


docs/src/tutorials/using_catalyst.md
```diff
@@ -0,0 +1,327 @@
...
+In this tutorial we'll provide an introduction to using Catalyst to specify
+chemical reaction networks, and then to solve ODE, jump and SDE models generated
+from them. Let's start by using the Catalyst [`@reaction_network`](@ref) macro
+to specify a simply chemical reaction network; the well-known repressilator.
+
+We first import the basic packages we'll need:
+
+```julia
+# If not already installed, first hit "]" within a Julia REPL. Then type:
+# add DiffEqBase OrdinaryDiffEq StochasticDiffEq DiffEqJump ModelingToolkit Catalyst Plots Latexify
```
ChrisRackauckas, line 11: we should really just recommend DifferentialEquations, reexport MTK. This should just be:

```julia
using Catalyst, DifferentialEquations, Plots, Latexify
```

docs/src/tutorials/using_catalyst.md
```diff
@@ -0,0 +1,327 @@
...
+
+# solve and plot, tstops is used to specify enough points
+# that the plot looks well-resolved
+sol = solve(sprob, LambaEM(), tstops=range(0., step=4e-3, length=1001))
+plot(sol)
+```
+![CLE Solution](../assets/birthdeath_cle.svg)
+
+We again have complete freedom to select any of the
+StochasticDifferentialEquations.jl SDE solvers, see the
```
ChrisRackauckas, line 306:
```suggestion
StochasticDiffEq.jl SDE solvers, see the
```

docs/src/tutorials/using_catalyst.md
```diff
@@ -0,0 +1,327 @@
...
+```julia
+u₀map  = Pair.(species(repressilator), u₀)
+pmap   = Pair.(params(repressilator), p)
+oprob2 = ODEProblem(osys, u₀map, tspan, pmap)
+```
+`oprob` and `oprob2` are functionally equivalent, each representing the same
+underlying problem. 
+
+At this point we are all set to solve the ODEs. We can now use any ODE solver
+from within the [OrdinaryDiffEq.jl](https://github.com/SciML/OrdinaryDiffEq.jl)
```
ChrisRackauckas, line 211:
```suggestion
from within the [DifferentialEquations.jl](https://diffeq.sciml.ai/latest/solvers/ode_solve/)
```

Non OrdinaryDiffEq solvers work.

docs/src/tutorials/using_catalyst.md
isaacsas: It used to be that in the old tutorial, but I was finding `using DifferentialEquations` to have become very slow (so much that I haven't used it in quite a while, so perhaps that improved).
In real use I don't think one would generally want to use it would they?

ChrisRackauckas: on the other hand, saying we're going to `using` 8 packages in the basic tutorial is quite intimidating. But yeah.. FWIW using times are dropping heavily in v1.5 and v1.6.

isaacsas: I switched it. I was debating that same point myself but had leaned towards it being better to show what I'd think is the normal use case vs. the easier case. But it is fine either way.

isaacsas:
@ChrisRackauckas Let me know if you have any more comments. Otherwise we can merge.

ChrisRackauckas:
I think it's really good. Nice job.

isaacsas:
Thanks. As you said, having a good introductory tutorial is probably critical. We can refine / expand as we add more features (e.g. tau-leap and such).

PR #235, opened 2020-08-03
Add :⇌, :⇋ to double_arrows

zdroid:
This is how I write it on the paper. Maybe even `:⥋` and `:⥊` can be useful.

The idea comes from https://github.com/zdroid/ChemEquations.jl, as some people suggested me to try integrate it with reaction networks from this repository.


isaacsas:
Please feel free to get in touch or open issues if there is anything on our end that would be useful to getting your package working with `ReactionSystem`s.

PR #241, opened 2020-08-04
Add no parameter tests

TorkelE:
Adds a simple test each for ODE, SDE, Jump, and SteadyState Problems, testing that they work fro systems without parameters.

### DIFF
--- a/test/solve_ODEs.jl
+++ b/test/solve_ODEs.jl
@@ -121,5 +121,5 @@ for factor in [1e0, 1e1, 1e2]
     u0 = factor*rand(length(no_param_network.states))
     prob = ODEProblem(no_param_network,u0,(0.,1000.))
     sol = solve(prob,Rosenbrock23())
-    @test abs.(sol.u[end][1] - 1.5/2) < 1000000*eps()
+    @test abs.(sol.u[end][1] - 1.5/2) < 1e-8
 end
--- a/test/solve_ODEs.jl
+++ b/test/solve_ODEs.jl
@@ -110,3 +110,16 @@ for (i,network) in enumerate(reaction_networks_all)
         @test solve(prob,Rosenbrock23()).retcode == :Success
     end
 end
+
+
+### No parameter test ###
+
+no_param_network = @reaction_network begin
+    (1.5,2), ∅ ↔ X
+end
+for factor in [1e0, 1e1, 1e2]
+    u0 = factor*rand(length(no_param_network.states))
+    prob = ODEProblem(no_param_network,u0,(0.,1000.))
+    sol = solve(prob,Rosenbrock23())
+    @test abs.(sol.u[end][1] - 1.5/2) < 1000000*eps()
+end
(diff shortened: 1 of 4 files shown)


test/steady_state_problems.jl
```diff
@@ -48,3 +48,15 @@ for network in steady_state_test_networks, factor in [1e-1, 1e0, 1e1]
...
+
+### No parameter test ###
+
+no_param_network = @reaction_network begin
+    (0.6,3.2), ∅ ↔ X
+end
+for factor in [1e0, 1e1, 1e2]
+    u0 = factor*rand(length(no_param_network.states))
+    sol_ss = solve(SteadyStateProblem(no_param_network,u0),SSRootfind())
+    @test abs.(sol_ss.u[1]-0.6/3.2) < 10000000*eps()
```
ChrisRackauckas, line 61: specify a tolerance for the rootfind and make sure it is roughly achieved?

test/solve_ODEs.jl
```diff
@@ -110,3 +110,16 @@ for (i,network) in enumerate(reaction_networks_all)
...
+### No parameter test ###
+
+no_param_network = @reaction_network begin
+    (1.5,2), ∅ ↔ X
+end
+for factor in [1e0, 1e1, 1e2]
+    u0 = factor*rand(length(no_param_network.states))
+    prob = ODEProblem(no_param_network,u0,(0.,1000.))
+    sol = solve(prob,Rosenbrock23())
+    @test abs.(sol.u[end][1] - 1.5/2) < 1000000*eps()
```
ChrisRackauckas, line 124: It'd just say something like `<1e-8`

test/steady_state_problems.jl
TorkelE: That's a good point, didn't even think of something like that. Will modify accordingly.

test/solve_ODEs.jl
TorkelE: Yes, that's better.

isaacsas:
@TorkelE Apparently the signature for `ReactionSystem` changed with the addition of some new fields. This is crashing my PR too.

I think it is from this PR to ModelingToolkit:

https://github.com/SciML/ModelingToolkit.jl/pull/400

isaacsas:
@TorkelE I'm working on a fix in https://github.com/SciML/Catalyst.jl/pull/242 but it might not be merged till tomorrow. I'll ping you when it is in so you can rebase.

isaacsas:
@TorkelE You need to merge master into your branch I think.

TorkelE:
Good point

TorkelE:
So I think what's going wrong is that Travis doesn't use a version of MTK that includes https://github.com/SciML/ModelingToolkit.jl/pull/543 (but by no means to knowledgable about how this works, so might be more complicated). 

I do not get this error on my machine, but when I roll back my local MTK version to pre https://github.com/SciML/ModelingToolkit.jl/pull/543 I get the same error. Also, Travis seems to run MTK 3.14.2, and checking that, I do not think it includes https://github.com/SciML/ModelingToolkit.jl/pull/543.

If that's the case, we can just wait until the next release of MTK and then this should be fine.

isaacsas:
@ChrisRackauckas ok to make a new MT release? (Not sure the state of master there currently.)

ChrisRackauckas:
It's in the process now.

isaacsas:
@TorkelE Looks like it is having SS issues again. I'll try restarting.

isaacsas:
@TorkelE Finally passing... Good to merge?

PR #242, opened 2020-08-04
Fix Usage of ModelingToolkit Accessor Functions 

isaacsas:
Also fixed some test warnings on 1.5 due to scoping.


isaacsas:
Nope, they assume it has been installed by the user and is accessible through the commandline.

isaacsas:
Not ideal, but the earlier Graphviz Julia library appears to be unmaintained, so it was this or something else. We could look into Graphplot -- that may still be being maintained to some degree. Overall it appeared the graph plotting tooling is not being maintained much at this point.

PR #257, opened 2020-09-14
don't run Latexify tests

isaacsas:
Let's turn these off temporarily until we can figure out how to make them more robust to differences that arise between OSes...

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -29,7 +29,7 @@ using SafeTestsets
 
 # Miscellaneous tests
 #@time @safetestset "Basic Plotting" begin include("plotting.jl") end
-@time @safetestset "Latexify" begin include("latexify.jl") end
+#@time @safetestset "Latexify" begin include("latexify.jl") end
 
 # the following can't really be run until there is an artifact for Graphviz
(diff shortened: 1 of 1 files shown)


isaacsas:
@TorkelE more random test failures? Maybe we should drop using random inputs?

TorkelE:
That might be sensible. 

I'm mixed about it, on one hand I think it has caught a few random weird cases. And it would be weird to cut out tests that errors. On the other, right now it only seems to enforce some things that we already know, like that the steady state solver should not be trusted blindly for these systems, and the only practical effect is to be irritating.

isaacsas:
At a minimum it would be great to ensure we are only using parameters over regimes where the generated models make physical sense. Given how long the tests run, it would be nice to make sure we aren't crashing due to simply evaluating models with parameters where they are unphysical (i.e. having negative solutions and such).

ChrisRackauckas (approved):
Agreed. I tried and couldn't figure out how to fix it. It's some whitespace format issues or something...

TorkelE:
Yes.

I think I start would be to remove random parameters when solving SDEs. I have some heuristics in place that should ensure that they are far from negatives, but that's otherwise a place where one could expect weird stuff going on.

PR #272, opened 2020-11-02
fix graph plotting and Latexifying

isaacsas:
@shashi getting errors in Latexifying from this line:

https://github.com/SciML/Catalyst.jl/blob/4078eaf242ec8954183e4ebf2854a177634df8ba/src/latexify_recipes.jl#L66

giving the error:

```julia
julia> Expr(rx1)
ERROR: TypeError: in Expr, expected Symbol, got a value of type Term{Real}
Stacktrace:
 [1] Expr(::Any, ::Vararg{Any,N} where N) at ./boot.jl:224
 [2] top-level scope at REPL[33]:1

julia> @which Expr(rx1)
Expr(args...) in Core at boot.jl:224

julia> convert(Expr,rx1)
ERROR: MethodError: Cannot `convert` an object of type Term{Real} to an object of type Expr
Closest candidates are:
  convert(::Type{Expr}, ::Revise.RelocatableExpr) at /Users/isaacsas/.julia/packages/Revise/x84rs/src/relocatable_exprs.jl:22
  convert(::Type{T}, ::T) where T at essentials.jl:171
  Expr(::Any...) at boot.jl:224
Stacktrace:
 [1] top-level scope at REPL[35]:1
```

How should a term be converted to a Julia expression now?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Catalyst"
 uuid = "479239e8-5488-4da2-87a7-35f2df7eef83"
-version = "5.1.1"
+version = "6.0"
 
 [deps]
 Catlab = "134e5e36-593f-5add-ad60-77f754baafbe"
--- a/src/graphs.jl
+++ b/src/graphs.jl
@@ -7,8 +7,8 @@ edge_attrs  = Attributes(:splines=>"splines")
 
 function edgify(δ, i, reverse::Bool)
     attr = Attributes()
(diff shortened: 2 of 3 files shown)


isaacsas:
@shashi NV on the question. Think I got it figured out. If you could take a look that what I've done seems reasonable that would be great. Thanks!

shashi:
Looks good!

PR #289, opened 2020-12-30
update for MT fixes

isaacsas:


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Catalyst"
 uuid = "479239e8-5488-4da2-87a7-35f2df7eef83"
-version = "6.2.0"
+version = "6.2.1"
```
ChrisRackauckas, line 3: I think dependency changes require a minor bump.

ChrisRackauckas, line 3:
```suggestion
version = "6.3.0"
```

isaacsas: Yeah, I wasn't sure. When I googled about it the semver repo issue I looked at seemed to indicate it is ok to bump like I did if the API on our end hasn't changed, but it seemed a bit iffy there too...

PR #290, opened 2021-01-09
Updates for upstream breaking changes

shashi:
needs https://github.com/SciML/ModelingToolkit.jl/pull/720 and the latest release of SymbolicUtils.jl

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Catalyst"
 uuid = "479239e8-5488-4da2-87a7-35f2df7eef83"
-version = "6.3.1"
+version = "6.4.1"
 
 [deps]
 Catlab = "134e5e36-593f-5add-ad60-77f754baafbe"
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Catlab = "0.7.3, 0.8, 0.9, 0.10"
 Latexify = "0.13.5, 0.14"
 MacroTools = "0.5.5"
-ModelingToolkit = "4.4.1"
+ModelingToolkit = "5"
 Reexport = "0.2, 1.0"
 julia = "1.3"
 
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
Should do the compats too?

shashi:
latexify failed again?!

isaacsas:
Looks like you forgot to evaluate the second reaction network in the file before generating the third and fourth test?

Project.toml
```diff
@@ -13,7 +13,7 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Catlab = "0.7.3, 0.8, 0.9, 0.10"
 Latexify = "0.13.5, 0.14"
 MacroTools = "0.5.5"
-ModelingToolkit = "4.4.1"
+ModelingToolkit = "5"
```
isaacsas, line 16: Bump Catalyst too for a new release?

isaacsas:
If tests now pass I think this should be gtg.

isaacsas:
@shashi seems like I'm now seeing crashes in a non-Latexify test. (The SDE test.) Does that pass for you?

isaacsas:
@TorkelE We seem to have a reproducible test failure here with the latest ModelingToolkit and SymbolicUtils in the SDE tests. Could you take a look?

isaacsas:
FWIW, the test passes locally on my computer when I just include the SDE test file, but fails when I run `test` through the package manager.

TorkelE:
Weird, bussy today, but will hopefully be able to take a look tomorrow.

isaacsas:
@TorkelE I dropped checking the solution in those SDE tests, and instead switched to just testing if the generated `f` and `g` functions agree well. Can you take a look and let me know if these changes look ok to you? From what I could tell the test had a bug anyway and was not actually comparing the two different solutions (instead looking at the ratio of a solution to itself).

TorkelE:
I was a bit ambitious and figured that for a solution over a long-enough timeframe the means and stds would be similar enough, but if the f and g functions are identical there should be no problem (and if there is, it's unlikely to be in this package). Doing it like this is good to me, especially when it saves us some headache. And as you noted, the test was bugged anyway.

isaacsas:
@shashi finally done! Thanks for all your help migrating us over!

PR #296, opened 2021-02-02
Add bif tutorial

TorkelE:
Adds a tutorial for creating bifurcation diagrams to the docs.

### DIFF
--- a/docs/src/tutorials/bifurcation_diagram.md
+++ b/docs/src/tutorials/bifurcation_diagram.md
@@ -26,6 +26,10 @@ p_span = (0.1,20.)   # The parameter range for the bifurcation diagram.
 plot_var_idx = 1     # The index of the variable we plot in the bifurcation diagram.
 ```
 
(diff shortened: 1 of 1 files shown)


TorkelE:
Is there a way to preview the doc changes to see if they turned out as intended? Probably everything should be fine.

ChrisRackauckas:
>Is there a way to preview the doc changes to see if they turned out as intended? Probably everything should be fine.

We could install a bot for that, but 🤷 I usually do the janky method of try and repeat.

ChrisRackauckas:
There's no link to the page added in the docs in this PR, so it won't show up. But you can manually go to the page to see.

TorkelE:
> > Is there a way to preview the doc changes to see if they turned out as intended? Probably everything should be fine.
> 
> We could install a bot for that, but  I usually do the janky method of try and repeat.

Sometimes the more blunt solutions are the easier ones. Will fix the link though.

PR #299, opened 2021-02-06
Remove catlab

isaacsas:


ChrisRackauckas (approved):

isaacsas:
@ChrisRackauckas Once tests finish I'll make a new release.

isaacsas:
Looks like ModelingToolkit has changed the system interface again?

ChrisRackauckas:
>Looks like ModelingToolkit has changed the system interface again?

@YingboMa added a show method. Was it made too generic?

isaacsas:
I think it was that pins were dropped from the system interface. Hopefully that is the only breaking change to handle.

ChrisRackauckas:
Oh yes, pins were dropped.

isaacsas:
The Catalyst macro directly constructs a `ReactionSystem` structure, bypassing the simplified constructor call (to allow for a completely empty system). So whenever the `ReactionSystem` definition changes we need to update here too.

ChrisRackauckas:
FWIW, StructuralTransformations.jl is "complete", and it's starting to be tested on full scale models. If those dev teams don't find anything major missing, then MTK might be finally out of its major change phase.

ChrisRackauckas:
>The Catalyst macro directly constructs a ReactionSystem structure, bypassing the simplified constructor call (to allow for a completely empty system). So whenever the ReactionSystem definition changes we need to update here too.

My hope is that once the interface is set, that ReactionSystem can actually live here. It's just a bit in flux.

isaacsas:
Cool! I have a bunch of things I've been holding off on starting while those changes are going on. Once they settle down we can make sure Catalyst is all caught up to the new setup, and start on adding features again.

ChrisRackauckas:
@shashi @YingboMa

isaacsas:
Make sense, though I don't think where it lives is a big deal in the end since we'll still probably re-export MT here. It might be easier to keep it there so it gets new system features added to it over time.

PR #302, opened 2021-02-15
make test RNGs reproducible

isaacsas:


TorkelE:
This seems like a great way to solve this.

isaacsas:
I like that for local testing we can always change the seed to get new random numbers too.

Issue #300, opened 2021-02-06
Re-enable graphs


isaacsas:
Removed by https://github.com/SciML/Catalyst.jl/pull/299 to drop Catlab dependency.

Should just be a matter of copying some of their graph-related definitions into Catalyst. (Longer-term we should consider using GraphPlot when it becomes more mature.)

PR #303, opened 2021-02-15
Reenable graphs

isaacsas:
Closes https://github.com/SciML/Catalyst.jl/issues/300


isaacsas:
@ChrisRackauckas are you ok with the two dependencies this adds?

ChrisRackauckas:
yeah that should be fine.

Issue #292, opened 2021-01-22
Population Balance equations


yewalenikhil65:
https://en.m.wikipedia.org/wiki/Smoluchowski_coagulation_equation

Would love to see an example of this equation efficiently solved using Catalyst and/or NetworkDynamics packages

I found them easy to implement with simple MATLAB functions but am unable to think of using Catalyst here

yewalenikhil65:
I did a very naive implementation. But I did by writing a user-defined function in Julia as follows:
I suspect the equations would be very expensive when N -> Larger  (  N = 4;   # number of clusters of particles we allow in the code)
```julia
using OrdinaryDiffEq, DiffEqJump
using Plots
gr()

function PBE(du,u,p,t)

  K(x,y) = x + y;
    # additive-kernel ...could be any function in x and y
    # generic function for kernel operator which is equivalent to rate of reaction
    # some more types of  kernels are K=1; K = x*y..
    # in general K(x,y,t)

  N = 4;   # number of clusters..
 # u ,..not exactly species but cluster array and,
 # index of cluster array denotes the number of particles in cluster

  RHS_1 = zeros(N);          RHS_2 = zeros(N);
  for i ∈ 1:N
    for j ∈ 1:i-1
        RHS_1[i] = RHS_1[i] + K(i-j,j)*u[i-j]*u[j];
    end
    RHS_1[i] = (1/2)*RHS_1[i];

    for j ∈ 1:N
        RHS_2[i] = RHS_2[i] + K(i,j)*u[i]*u[j];
    end
    du[i] = RHS_1[i] - RHS_2[i];
  end

end

tspan = (0., 0.05)

# initially system consists of..
#  100 singlets, 0 doublets, 0 triplets and 0 quadralets(if that is even a term)
u0 = [100.; 0.; 0.; 0.]


# solve ODEProblem
oprob = ODEProblem(PBE, u0, tspan)
sol = solve(oprob, Tsit5())
plot(sol,lw=2)



```

yewalenikhil65:
Hi @isaacsas , You are right , it should be modelled as a discrete jump process to begin with.  I was able to do it in MATLAB by say simply treating singlets , doublets etc as some bins which will get filled based on Gillespie rule. But I had to implement original Gillespie for that and decide which collisions/reactions will occur . I wondered if it was doable in easier way using julia packages like catalysts, and diffeqjump ..

yewalenikhil65:
[...]

PR #307, opened 2021-02-26
Generating reactions programmatically using Reaction Systems

yewalenikhil65:
Tutorial for Implementation of (Smoluchowski coagulation equation) population balance equation- by programmatically generating the reactions using ReactionSystem
Based on discussion at https://github.com/SciML/Catalyst.jl/issues/292


isaacsas:
@yewalenikhil65 I'll go through tomorrow morning and give you some comments. Got busy with meetings and class stuff today.

yewalenikhil65:
> walenikhil65 I'll go through tomorrow morning and give you some comments. Got busy with meetings and class stuff today.

No problem ,Sir

isaacsas:
You should be able to `git pull` to sync my changes back to your local git repository.

docs/src/tutorials/generating_reactions_programmatically.md
```diff
@@ -0,0 +1,152 @@
...
+
+We begin by importing some necessary packages.
+```julia
+using ModelingToolkit, LinearAlgebra
+using DiffEqBase, DiffEqJump, OrdinaryDiffEq
+using LoopVectorization, Plots
+using BenchmarkTools
+using SpecialFunctions
+```
+Suppose the maximum cluster size is `N`. Lets initialize the system with some initial concentration `C`, initial number of singlets `uₒ` in the system. Since its a bimolecular chain of Reaction system(`nr` number of reactions), the bulk volume `V` of the system in which these binary collisions occur is important in the calculation of rate laws.
```
isaacsas, line 14: Should the initial concentration here be called `C` or `N_0`?

docs/src/tutorials/generating_reactions_programmatically.md
```diff
@@ -0,0 +1,152 @@
...
+  
+```julia
+## Parameter
+N = 10                       # maximum clusters size
+Vₒ = (4π/3)*(10e-06*100)^3   # volume of a singlet/monomer in cm³
```
isaacsas, line 19: Are singlets and monomers the same? If so, I suggest using monomer, dimer, trimer, etc throughout and not using signlet/doublet/triplet, etc.

docs/src/tutorials/generating_reactions_programmatically.md
yewalenikhil65: You are right..made a mistake while typing.. Initial concentration should be `Nₒ`

docs/src/tutorials/generating_reactions_programmatically.md
yewalenikhil65: > Are singlets and monomers the same?
>
Yes they are same..
 >If so, I suggest using monomer, dimer, trimer, etc throughout and not using signlet/doublet/triplet, etc.
>
I will make these changes

yewalenikhil65:
Hey @isaacsas   updated according to your suggestions..

isaacsas:
@yewalenikhil65 made some more updates. Slowly getting through it. (This week is pretty busy.)

I do have one question for you, see below.

isaacsas:
Actually never mind; I forgot that I had figured out what I was confused about!

yewalenikhil65:
Corrected a little mistake I spotted in writing `kv` for constant kernel case.. I like your changes in wording too

isaacsas:
OK, I updated it one more time. I'd be happy to merge it now. Some notes:
- You weren't actually using ModelingToolkit / Catalyst in your previous code. When you created the `MassActionJump` and passed it to `JumpProblem` you were actually just using `DiffEqJump` directly. 
- Along those lines, the `ReactionSystem` model used `2*k[n]` for reactions with involving reactants that had the same number of monomers in them, however, what you were plotting and matching to the theory was just using `k[n]` for such reactions! I switched the `ReactionSystem` to use `k[n]` too since that is what made it match the theoretical formulas you used. Please double check that this is all correct though! You can see the new figure I made in the assets folder.

Let me know after you've taken a look, and if you approve I'll merge.

yewalenikhil65:
> OK, I updated it one more time. I'd be happy to merge it now. Some notes:
> 
>     * You weren't actually using ModelingToolkit / Catalyst in your previous code. When you created the `MassActionJump` and passed it to `JumpProblem` you were actually just using `DiffEqJump` directly.
> 
Yes, you are right. But i guess pushing reactions into for loop uses ModelingToolkit .
>     * Along those lines, the `ReactionSystem` model used `2*k[n]` for reactions with involving reactants that had the same number of monomers in them, however, what you were plotting and matching to the theory was just using `k[n]` for such reactions! I switched the `ReactionSystem` to use `k[n]` too since that is what made it match the theoretical formulas you used. Please double check that this is all correct though! You can see the new figure I made in the assets folder.
> 
Again you are right.I plotted for just `k[n]` when I wrote the tutorial and uploaded that figure...But i have double checked using `2*k[n]` and `k[n]`also. In both cases, multiplicative kernel doesn't match with analytical ..I do not know why!
> 
> Let me know after you've taken a look, and if you approve I'll merge.
>
We can merge it now. Only bit I am unsure is how to reason no-match of multiplicative kernel case, so i skipped commenting on it.

isaacsas:
Hmm, perhaps we should just use the additive kernel then and remove the other two?

yewalenikhil65:
> Hmm, perhaps we should just use the additive kernel then and remove the other two?

Additive and constant kernels..Both are matching with analytical profiles.. 
I am not comfortable with removing multiplicative kernel even though its not matching. What do you think about calculation of the `bulk volume of the system` with which we are using in calculation of rate-constants ? Did we miss anything there ?

yewalenikhil65:
Hey @isaacsas ..just realised something
This non-matching behaviour for `2*k[n]` that you got is because of this https://github.com/SciML/Catalyst.jl/pull/307/commits/58a703e4db8c7e23c750293281eced3212ebf69a
I believe if you do add `stoich construction` as I did previously ,it matches in additive and constant kernel even if you use `2*k[n]` as i mentioned earlier.I am confused why it is such though.

isaacsas:
The `MassActionJump` you are building is using `k[n]` though not `2*k[n]` for those reactions, right? When you use the `mass_act_jump` none of the `ReactionSystem` code is even getting used so it is irrelevant what you set as the rate in the `Reaction`s.

isaacsas:
Unfortunately, I haven't read this exact solution paper / method, so can't comment on why the multiplicative case does not match it.

yewalenikhil65:
> The `MassActionJump` you are building is using `k[n]` though not `2*k[n]` for those reactions, right? When you use the `mass_act_jump` none of the `ReactionSystem` code is even getting used so it is irrelevant what you set as the rate in the `Reaction`s.

Ahh.. understood now.. In that case, https://github.com/SciML/Catalyst.jl/commit/58a703e4db8c7e23c750293281eced3212ebf69a seems correct and usage of `k[n]` seems more correct. I just `latexified` reactions into ODEs and checked if it matches with equation in wiki page.Usage of `k[n]` is appropriate one as you have done. 
I think we can leave multiplicative case for now then. and merge only additive and constant kernels case ?

isaacsas:
OK, can you remove the multiplicative case from the tutorial text and code? I'll then take a look and merge.

yewalenikhil65:
> OK, can you remove the multiplicative case from the tutorial text and code? I'll then take a look and merge.

@isaacsas ..Done!

isaacsas (approved):

isaacsas:
All merged! Thanks!

PR #309, opened 2021-03-03
fix depwarns

ChrisRackauckas:


isaacsas:
Thanks! That has been on my TODO list...

Issue #317, opened 2021-03-20
add back ability to name `ReactionSystem`s via the DSL


isaacsas:
i.e. something like
```julia
sirmodel = begin sir
  k, S --> I
end k
```
returning a `ReactionSystem` with the `name` field set to `sir`. This seems useful, and needed for allowing modular composition of systems.

PR #319, opened 2021-03-24
Allow sys names and other API fixes

isaacsas:
Closes https://github.com/SciML/Catalyst.jl/issues/317

Seems to work for non-empty networks. Empty networks would require a breaking change to drop allowing an empty network that has parameters defined (which seems reasonable to me since parameters can be subsequently added if needed). @TorkelE do you remember why we allowed `@reaction_network k1 k2 k3` to work and make an empty system with parameters `k1`, `k2`, `k3`? 

- [x] Needs doc updates.
- [x] Handle empty network.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Catalyst"
 uuid = "479239e8-5488-4da2-87a7-35f2df7eef83"
-version = "6.9.0"
+version = "6.10.0"
 
 [deps]
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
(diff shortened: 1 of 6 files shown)


TorkelE:
I think the reason was that if one (at the time of the creation of the empty network) knew all of the parameters that were to be included, it would make sense to allow to declare it with those, and then one could go on adding reactions without worrying about setting parameters. It would also make it easier to specify which order parameters were in, and give a nice place where one could find all.

But since this seems to prevent declaring names, changing it makes sense.

isaacsas:
Alright, then I think I'm going to switch it to only allow specification of a name. That seems more useful as we add composition of networks.

TorkelE:
agreed.

PR #320, opened 2021-03-24
Fixed error message for @reaction_network with no reactions

kaandocal:
Currently when using `@reaction_network` with no reactions and one of the parameters is reserved Julia throws an error before the desired error message is printed out. This is due to the code trying to access a nonexistent variable which has been removed in this version.

### DIFF
--- a/src/reaction_network.jl
+++ b/src/reaction_network.jl
@@ -87,7 +87,7 @@ end
 
 #Returns a empty network (with, or without, parameters declared)
 macro reaction_network(parameters...)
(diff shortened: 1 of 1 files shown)


isaacsas:
Thanks! Do you use this functionality? I was thinking of deprecating it to enable giving names to empty networks (which is needed for compositional modeling). i.e. you would only be able to say:
```julia
rn = @reaction_network 
```
or
```julia
rn = @reaction_network rn_name
```
One could still add parameters afterwards using `addparams!`.

In any case I'll merge this for now.

kaandocal:
I personally don't,I just came across this while studying the source code!

isaacsas:
Great, I doubt it is widely used, and the change makes more sense to me. Let us know if you find any other issues! We definitely have to fix up the API functions some to handle sub-systems properly, but that won't be done till I get systems working correctly in `ReactionSystem`s themselves (hopefully soon).

Issue #229, opened 2020-07-21
Representing Spatial Reaction Networks


isaacsas:
Issues to consider
- representing continuous domains and BCs
- Unstructured vs structured grids for spatial domains and graph geometries
- BCs
- simply reusing one reaction network everywhere vs allowing spatially varying rates
- species dependent transition rates, allowing them to be spatially varying?
- spatial reactions i.e. `A_i + B_j -> C_k + D_l` and such.

ludoro:
Talking about epidemiological modeling, it could be also nice to take into account the possibility of having some ways to specify compartments. Say we have a simple SIR model, with I = {I_1,I_2,...,I_n}, instead of writing all the reactions "by hand" (which in this case is doable), there could be the possibility to let the user specify all the "type" of reactions that he wants, which rates etc and then call the @reaction_network macro with this information to build the network.  
I am currently working on a proof of concept using catalyst for this, maybe I can provide some code in the near future to help brainstorming. As discussed with isaacsas on slack, this could be related to "spatial" information of the network.

isaacsas:
Some modalities to consider:
- Reactions on graphs or meshes; given an implicit or explicit connectivity (i.e. a graph / adj. matrix) allow users to specify several types of reactions on graphs:
  a. First order "spatial" transitions between nodes, i.e. `S_i -> S_j` where `i` and `j` denote different nodes / mesh cells / compartments. These are usually restricted to `i` and `j` that are neighbors.
  b. Non-spatial reactions that are replicated at each node, i.e. `A_i + B_i <-> C_i`, `A_i <-> B_i`, `0 <-> A_i`, etc, at all nodes, `i`. Here we should support isotropic rates and varying rates that depend on `i`.
  c. Spatial interactions `A_i + B_j <-> C_k` for different nodes, `i`, `j`, `k`. An important point here is that such reactions might be desired for sites that are not nearest neighbors in the spatial transition sense (a. above). i.e. we either want a separate graph for neighbors for these types of reactions, or to allow non-neighbor interactions.

TorkelE:
Would it make sense to allow an arbitrary number of graphs to be in the input, and then reactions could use "neighbours" through any select graph?

In the case of a system where one has reactions using neighbours and neighbours neighbours, there's one graph associated with each type. That way one could avoid having to handle various weird connection schemes, and leave it to the users to come up with a way of producing the appropriate graphs for their desired type?
[...]

PR #321, opened 2021-03-24
A simple interface for general spatial reaction networks

kaandocal:
**Motivation**

Many modellers work with spatial reaction networks, e.g. ones defined on a graph or on a spatial grid, which can be seen as a special type of graph. These can be represented straightforwardly by "normal" reaction networks where each species is split up into multiple sub-species corresponding to each node of the graph, but doing this by hand is very tedious an error-prone. This issue has been brought up previously in #229 and I have attempted to augment Catalyst.jl by extending the reaction network language.

**Implementation**

In order not to break anything I implemented this as an experimental macro whose output can be sent to @reaction_network; this should prevent any code-breaking changes and retain full backwards compatibility.

Species are indexed by vertices in arbitrary graphs, and multiple graphs for one system are supported. Since I am a relative newcomer to Julia I am not sure if there are any standard tools for dealing with graphs, so I've hacked my own for demonstration purposes.

Each species can have an index into a graph, denoted using indexing: `X` becomes `X[i]`, `Y` becomes `Y[j]`, etc. Every index is declared to belong to one graph as macro parameters of the form `G[i] H[j,k]`, which says that `i` is a vertex of `G` and `j`,`k` are vertices of `H`. The species `X[i]` is then expanded to a set of species `X_i` for every vertex `i` of `G`. Constants in reaction rates can be extended similarly, but they can depend on multiple indices, e.g. if a is a constant then `a[i,j]` mapping to `a_(i,j)` can be defined for each `i` in G, `j` in H. Every line using indexing is thus expanded into multiple lines for every valid combination of indices, e.g. 

`d, X[i] -> 0`
`c[i], 0 -> X[i]`
`D[j,k], Y[j] -> Y[k]`
`a, X[i] + Y[j] + Y[k] -> X[i]`

become

`d, X_i -> 0` (for each i in G)
`c_i, 0 -> X_i` (for each i in G)
`D_(j,k), Y_j -> Y_k` (for each edge j<->k in H)
`a, X_i + Y_j + Y_k -> Y_j` (for each i in G, edge j<->k in H)

**Example:**
```
# Cyclic graph on 4 elements
G = BasicGraph([ 1, 2, 3, 4 ], [ 1=>2, 2=>3, 3=>4, 4=>1 ]) 

@spatial_reaction_network begin
  sigma[i] --> X[i]
  D, X[i] --> X[j]
  delta, 2X[i]-->0
  a, 0 -> Z
  b, Z + X[i] -> 2X[i]
end G[i,j]
```

returns

```
quote                                               
    (sigma, 0 → X_1)                                
    (sigma, 0 → X_2)                                                                                     
    (sigma, 0 → X_3)                                
    (sigma, 0 → X_4)                                                                                     
    (D, X_2 → X_1)                                  
    (D, X_3 → X_2)                                  
    (D, X_4 → X_3)                                  
    (D, X_1 → X_4)                                  
    (D, X_1 → X_2)                                  
    (D, X_2 → X_3)                                                                                       
    (D, X_3 → X_4)                                                                                       
    (D, X_4 → X_1)                                  
    (delta, 2X_1 → 0)                                                                                    
    (delta, 2X_2 → 0)                               
    (delta, 2X_3 → 0)                               
    (delta, 2X_4 → 0)                               
    (a, 0 → Z)                                      
    (b, Z + X_1 → 2X_1)                             
    (b, Z + X_2 → 2X_2)                             
    (b, Z + X_3 → 2X_3)                             
    (b, Z + X_4 → 2X_4)                             
end
```

I haven't spent much time with either Julia or SciML, so I would greatly appreciate any feedback!

**Remarks:**
* If `n` indices belonging to a graph appear in a reaction, the reaction is expanded for every n-edge of the graph. If `n==1` this corresponds to every node of the graph, if `n==2` we get every edge and for `n>=3` you can define your own higher-order edges to model more complex interactions. 
* If two different graphs appear in a reaction we expand jointly over all valid permutations of indices of both graphs

**Todos:**
* UGLY: the macro needs the values of the graphs to generate its output, and for this I used `Base.MainInclude.eval()`. What's the proper way to do this?
* The generated variables have names of the form var"X_$i", where i ranges over the vertices of the graph. There is probably a better way to encode this information...
* Prevent one species from being indexed by multiple graphs to avoid name-conflicts?
* Do some testing
* Work on graph interface
* How best to combine this with arbitrary parameters?

### DIFF
--- a/src/spatial_network.jl
+++ b/src/spatial_network.jl
@@ -59,13 +59,13 @@ function get_edge_iter(i::Symbol, j::Symbol, dim::Int64, grid::PeriodicGrid)
                                for sgn in [-1,+1])
 end
 
-function graph_iterator(grid::PeriodicGrid, idcs::Set{Symbol})
+function graph_iterator(grid::PeriodicGrid{N}, idcs::Set{Symbol}) where N
     if length(idcs) == 1
         idx = first(idcs)
         return ( Dict(idx => coord_to_str(coords)) for coords in grid_coords(grid) ) 
     elseif length(idcs) == 2
         i, j = collect(idcs)
-        return Iterators.flatten(get_edge_iter(i, j, dim, graph) for dim in 1:N)
+        return Iterators.flatten(get_edge_iter(i, j, dim, grid) for dim in 1:N if grid.size[dim] != 1)
     end
     
     return nothing
@@ -80,7 +80,7 @@ function get_idx_dict(exprs)::Dict{Symbol,Any}
     for idx_map in exprs
         (isa(idx_map, Expr) && idx_map.head == :ref) || throw("malformed index assignment: $(idx_map)")
         graph_name = idx_map.args[1]
-        graph = Base.MainInclude.eval(graph_name)
+        graph = Base.eval(@__MODULE__, graph_name)
         idcs = idx_map.args[2:end]
         
         for idx in idcs
@@ -106,7 +106,7 @@ function iterate_idcs(idcs::Set{Symbol}, idx_dict::AbstractDict)
         push!(iterators, iterator)
     end
     
-    return (merge(dicts...) for dicts in Iterators.product(iterators...))
+    ( merge(dicts...) for dicts in Iterators.product(iterators...) )
 end
 
 # Get indices occurring in a reaction
@@ -204,7 +204,11 @@ function get_spatial_reactions(ex::Expr, idx_dict::AbstractDict)
     Expr(:block, reactions...)
 end
         
-macro spatial_reaction_network(ex::Expr, graphs...)
+function make_spatial_reaction_network(ex::Expr, graphs)
     idx_dict = get_idx_dict(graphs)
-    esc(get_spatial_reactions(MacroTools.striplines(ex), idx_dict))
+    get_spatial_reactions(MacroTools.striplines(ex), idx_dict)
+end
+
+macro spatial_reaction_network(ex::Expr, graphs...)
+    esc(make_spatial_reaction_network(ex, graphs))
 end
--- a/src/spatial_network.jl
+++ b/src/spatial_network.jl
@@ -1,0 +1,210 @@
+
+
+# This function takes a graph and a set of indices
+# and returns an iterator over assignments
+# i -> e[1], j -> e[2], k -> e[3], ...
+# for all n-edges of the graph, where n is the
+# number of indices. The assignments should be
+# of type Dict{Symbol,Symbol}.
+#
+# For an ordinary graph, the 1-edges are the vertices
+# and the 2-edges are the edges in the usual sense.
+#
+# Important: as the order of the indices is not
+# defined, this function should assign elements
+# to all permutations of the edges, ie. for
+# i => 4, j => 5 it should also return i => 5, j => 4
+function graph_iterator end
+
+struct BasicGraph{T}
+    verts::AbstractVector{T}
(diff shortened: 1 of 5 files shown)


src/spatial_network.jl
```diff
@@ -0,0 +1,210 @@
...
+# For an ordinary graph, the 1-edges are the vertices
+# and the 2-edges are the edges in the usual sense.
+#
+# Important: as the order of the indices is not
+# defined, this function should assign elements
+# to all permutations of the edges, ie. for
+# i => 4, j => 5 it should also return i => 5, j => 4
+function graph_iterator end
+
+struct BasicGraph{T}
```
ChrisRackauckas, line 19: We should really just build this on LightGraphs.jl instead of building out a whole new graph format.

isaacsas:
@kaandocal looks cool. See also https://github.com/SciML/DiffEqJump.jl/pull/131 where we started on a direct interface to the Gillespie methods in a similar way (giving a graph and flattening it to a giant set of reactions). 

I'll take a look tonight or tomorrow and give you some feedback (sorry, need to finish up lectures now!).

isaacsas:
@kaandocal haven't forgotten about this or your other PR; was just a busy week. Blocking out time tomorrow afternoon to go through them carefully.

kaandocal:
No need to hurry, while I appreciate any feedback this is not an urgent issue!

Vilin97:
Hi. This is currently being built. Check out https://vilin97.github.io/posts/post4/ where I describe how to set up a spatial jump problem. Or check out any of the spatial tests in `DiffEqJump`.

yewalenikhil65:
> Hi. This is currently being built. Check out https://vilin97.github.io/posts/post4/ where I describe how to set up a spatial jump problem. Or check out any of the spatial tests in `DiffEqJump`.

@Vilin97  i found this blog post really useful. I was curious whether we can define some say vector field(velocity) over the cartesian grid, and check out it's influence on the Jump process (where, say the rate laws is dependent on vector field)

Vilin97:
> I was curious whether we can define some say vector field(velocity) over the cartesian grid, and check out it's influence on the Jump process (where, say the rate laws is dependent on vector field)

You can define your own hopping rates. For example, your hopping rate to the nodes on the right might be 1, while the hopping rates to the nodes on the left might be 0.1. Thus your molecules will tend to drift right. Check out this test: https://github.com/SciML/DiffEqJump.jl/blob/master/test/spatial/diffusion.jl#L84 to see how to set this up.

yewalenikhil65:
> > I was curious whether we can define some say vector field(velocity) over the cartesian grid, and check out it's influence on the Jump process (where, say the rate laws is dependent on vector field)
> 
> You can define your own hopping rates. For example, your hopping rate to the nodes on the right might be 1, while the hopping rates to the nodes on the left might be 0.1. Thus your molecules will tend to drift right. Check out this test: https://github.com/SciML/DiffEqJump.jl/blob/master/test/spatial/diffusion.jl#L84. It sets up a similar thing: only hopping up and left is allowed.

@Vilin97 
This is awesome  work! Hope advection of molecules effects are also incorporated in near future

isaacsas:
Advection could be simulated right now. The user needs to specify the hopping rates currently, so they would need to calculate what the advective transition rates for their model are. Later we can hopefully move model specification to higher levels with something like this PR combined with automatic discretization of transport operators to transition rates. That's a bit off though.

PR #324, opened 2021-03-31
Netstoich api fun

isaacsas:


isaacsas:
@YingboMa Do you have any idea about this test failure I'm getting in the doc test that says it is occurring in Symbolics? Thanks!

YingboMa:
Umm, I am not sure. I cannot reproduce it locally.

ChrisRackauckas:
https://github.com/SciML/Catalyst.jl/blob/master/.github/workflows/Documentation.yml#L17 make that just 1

isaacsas:
Thanks! Wonder where I copied that from when I added it. Might need to update some other repositories too now...

isaacsas:
@yewalenikhil65 Just FYI!

PR #325, opened 2021-04-02
Update to SymbolicUtils v0.10

YingboMa:


isaacsas (approved):

isaacsas:
LGTM. How does one construct parameters or variables from strings/symbols now?

YingboMa:
Yeah... I will need to add that `@variables $x` thingy.

isaacsas:
OK, cool. I need that feature for `ReactionNetworkImporters` to generate variables/symbols from the network files it loads.

shashi:
You can call `toparam` on a variable.

Issue #318, opened 2021-03-23
Feature request for Representing reactions with focus on intermediate complexes


yewalenikhil65:
I think it would be nice to introduce another representation of chemical reactions in `Catalyst` which represents network with a focus on  intermediate complexes.. Such way is very useful in modelling reduced order methods. 

This shall enable in calculating certain `incidence matrices` of reaction graph, `Laplacian matrices` etc..This also helps in writing ODESystem in another alternate form that involves these matrices

On similar lines, it would be nice to visualize network with `Graph(rn)` representing  intermediate complexes as nodes (naturally stoichometry is also taken care in the nodes) and edges as reaction arrows. It could be less messy for bigger reaction networks to visualize that way.

isaacsas:
If you want to add new graph visualization options and/or a new system type with all the conversions it needs and such go ahead! PRs are welcome. I’d like to get a system type in, perhaps `StoichMatSystem`, that is the representation we’ve previously discussed, but right now I don’t have time to work on it. In ~ month classes should end, and I can circle back to this if no one else has picked it up.

yewalenikhil65:
> If you want to add new graph visualization options and/or a new system type with all the conversions it needs and such go ahead! PRs are welcome. I’d like to get a system type in, perhaps `StoichMatSystem`, that is the representation we’ve previously discussed, but right now I don’t have time to work on it. In ~ a month classes should end, and I can circle back to this if no one else has picked it up.

I will like to try on following basis when i begin.. noting here for future purpose
"A network dynamics approach to chemicalreaction networks" https://arxiv.org/pdf/1502.02247.pdf

yewalenikhil65:
@isaacsas , I would to know your thoughts on https://github.com/yewalenikhil65/Catalyst.jl/discussions/2
whenever you are free.. Its based on what i said in https://github.com/SciML/Catalyst.jl/issues/318#issue-838413865

yewalenikhil65:
clsoed with https://github.com/SciML/Catalyst.jl/pull/373

PR #326, opened 2021-04-09
some more networkapi functions ? 

yewalenikhil65:
Based on https://github.com/SciML/Catalyst.jl/issues/318 

for now, i have written this in a form of tutorial ! Should this be in tutorial form Or should we add some more `networkapi` functions for `incidence matrix of complexes` , `complex stoichiometric matrix` , `outgoing matrix`, `Laplacian matrix` and so on?


isaacsas:
@yewalenikhil65 very nice! I think it would make more sense though to add some of these intermediate matrix representations to the API, and then have the tutorial show something cool one can do using them. What do you think?

yewalenikhil65:
> @yewalenikhil65 very nice! I think it would make more sense though to add some of these intermediate matrix representations to the API, and then have the tutorial show something cool one can do using them. What do you think?

Yes, i think having these matrices as output of some `networkapi` functions makes sense.  I shall these matrices into `networkapi` after testing more tonight.. As for tutorial to do something nice using them, i will do that in few days soon.

yewalenikhil65:
@isaacsas , added a `complexstoichmat` function to `networkapi`. Please check for improvements if any in the loops. I would suggest to not quickly merge, as i plan to come soon with brief tutorial of application of this function to some problem.

src/networkapi.jl
```diff
@@ -183,6 +183,63 @@ function netstoichmat(rn; smap=speciesmap(rn))
...
+kd --> Array of rate of reactions
+complexes_mat --> unique intermediate complexes that define the system,
+
+We can define a laplacian matrix from these returned arrays from complexstoichmat as,
+KD = diagm((map(Num, kd)))
+L = B*KD*Δ', and rewrite the ODESystem as x' = -Z*L*exp.(Z'*log.(states(rn)))
+
+and/or also check that, netstoichmat(rn) == Z*B 
+"""
+function complexstoichmat(rn; smap=speciesmap(rn))
```
isaacsas, line 201: Would it make sense to split this into several functions that individually construct and return some of the other matrices? It seems like this function is doing a lot, and perhaps some of the intermediate matrices would be useful to calculate on their own? You could then make them optional keyword arguments the user can pass in to `complexstoichmat`, using the user's functions if passed in, or otherwise calling those helper functions here.

isaacsas (changes requested):

isaacsas:
Could you also add tests for some examples to make sure the returned matrices are all correct?

src/networkapi.jl
yewalenikhil65: > Would it make sense to split this into several functions that individually construct and return some of the other matrices? It seems like this function is doing a lot, and perhaps some of the intermediate matrices would be useful to calculate on their own? You could then make them optional keyword arguments the user can pass in to `complexstoichmat`, using the user's matrices if passed in, or otherwise calling those helper functions here.

Yes, that makes sense.. I will add individual functions for each entity needed

yewalenikhil65:
> Could you also add tests for some examples to make sure the returned matrices are all correct?

Yup,  .. any suggestion on what kind of reactions i should preferably use for tests ?

isaacsas:
I'd suggest systems with a variety of reactions, including non-mass action reactions if they make sense, and where you know the answer (either from a paper, or from working it out by hand).

isaacsas:
@yewalenikhil65 will take a look later in the week, I’m swamped through Thursday currently.

yewalenikhil65:
> @yewalenikhil65 will take a look later in the week, I’m swamped through Thursday currently.

No problem

yewalenikhil65:
Hi @isaacsas 
Didn't get time to follow up this later in past month(COVID issues at home) 

There example is missing a demonstration for the use of these functions. Will start to work on it this week.

isaacsas:
@yewalenikhil65 what is the status on this PR? Are you done with it? (Your last comment said there was more you planned to do.)

yewalenikhil65:
> @yewalenikhil65 what is the status on this PR? Are you done with it? (Your last comment said there was more you planned to do.)

hi @isaacsas wrapping it up here https://github.com/yewalenikhil65/Catalyst.jl/discussions/2#discussion-3308278
would you please take a look and suggest/improve it ?

yewalenikhil65:
@isaacsas  closing this PR , and will come up with new PR with functions + proposed tutorial

isaacsas:
@yewalenikhil65 sorry for my delays with this one. If you feel this is done and in good shape I can review; I just wasn't clear if you had further updates you intended to make or not! 

I'll look through your comment -- somehow I missed that in the previous discussion.

yewalenikhil65:
> @yewalenikhil65 sorry for my delays with this one. If you feel this is done and in good shape I can review; I just wasn't clear if you had further updates you intended to make or not!
> 
> I'll look through your comment -- somehow I missed that in the previous discussion.

@isaacsas  sure.. please do review https://github.com/yewalenikhil65/Catalyst.jl/discussions/2#discussion-3308278 , so that we can decide whether it goes well with the complex-stoichiomatrices matrices functions in networkapi.jl that i proposed earlier.
I will add a new PR for this one.

PR #334, opened 2021-05-24
use correct syntax for graphviz jll

isaacsas:
@giordano does this look right to you in the case that the Graphviz jll is installed? This is following up on the discussion in

https://github.com/JuliaPackaging/Yggdrasil/issues/1428

Thanks!


isaacsas:
Actually, @giordano this seems to still have issues with pdfs, see the test failure. Any suggestions?

isaacsas:
FWIW it seems to work locally on my mac using the jll in the testing I tried.

giordano:
Yes, the syntax looks ok

Issue #280, opened 2020-11-25
TODO per discussion of 11/25


isaacsas:
- [x] Eliminate `@reaction_func` and how Hill functions work. (Niklas)
- [x] Fix ReactionNetworkImporters. (Sam)
- [x] Fix BCR loading and update BCR benchmarks. (Sam)
- [x] Work towards previous BCR performance. (Sam)

Tutorials:

- [x] Bifurcations (adapting https://github.com/rveltz/BifurcationKit.jl/issues/27) (Torkel and Alex)
- [x] Adding forcing to a network. (Sam)
- [x] Parameter estimation. (Chris and Alex)
- [ ] Discovery of model components. (Chris and Alex)
- [ ] Discovery and elimination of conservation laws and such.
- [x] Diagram of the IR (Chris)
- [x] Diagram of the user features (Sam and Torkel)
- [ ] Structural transformations?

ChrisRackauckas:
[ ] Diagram of the IR
[ ] Diagram of the user features

isaacsas:
@ChrisRackauckas @TorkelE @korsbo The non-tutorial issues that we had listed here are now basically done aside from getting BCR back to where we were at before moving to MT. (@shashi and I have been working more on that, and will hopefully meet up to make some more progress this month.) 

I think with those tutorials we listed done we could use them in building the paper.

yewalenikhil65:
**Discovery and elimination of conservation laws and such.**
@ChrisRackauckas @isaacsas , I am thinking along following lines for the point in to-do list

The dynamics of most CRNs is written as `dx/dt= S. v(x(t))` , where S is a net-stoichiometric matrix, and `v(x(t))` is vector of `oderatelaws.(rn.eqs)`
If `κ  ∈ ker(Sᵀ)` ,We can see that `κᵀ. dx/dt == 0` 
i.e. `κᵀ.x = constant` is satisfied.
So `κ` could be easily found from `nullspace` of `Sᵀ` (scaled-to-integer version of columns)

Now to find out what is the conserved quantity, we can easily write this as,
`κᵀ.x = κᵀ.x(t=0)`  (conservation law, where `κᵀ.x(t=0)` is the conserved quantity )

isaacsas:
If you add the conservation law to the generated ODEsystem does structural simplification then eliminate one of the ODEs?

ChrisRackauckas:
[...]

PR #337, opened 2021-06-03
Adding support for conservation law

kaandocal:
Following onwards from [here](https://github.com/SciML/ModelingToolkit.jl/issues/965) I've added some code taken from [here](https://github.com/kaandocal/FiniteStateProjection.jl/) to detect conservation laws for a reaction system together with a rudimentary unit test for Catalyst. 

### Notes:
- There are multiple equivalent matrices of conservation laws for most systems, afaik there's no way to guarantee which one you're going to get
- It introduces a dependency on AbstractAlgebra.jl
- This will not be very efficient for large (>100 species) systems

Going forward it would be great if we could add functionality to simplify a system based on a conservation law. Basically I was thinking of something like a function `structsimplify(rn, species, cons)` that takes a reaction network and a list of species to be elided and returns an equivalent reaction network without these species. It should check whether these species can be removed simultaneously and replace their occurrences by the other species (using the conserved quantities `cons`). 

Example:
```julia
rn = @reaction_network begin
    ρ_u, G_u --> G_u + P
    ρ_b, G_b --> G_b + P
    d, P --> 0
    σ_u, G_b --> G_u + P
    σ_b, G_u + P --> G_b
end ρ_u ρ_b d σ_u σ_b
```

There is one conservation law `[G_u] + [G_b] == N`, and we can replace the above by the equivalent system
```julia
rn = @reaction_network begin
    ρ_u, G_u --> G_u + P
    ρ_b * (N - G_u), 0 --> P
    d, P --> 0
    σ_u * (N - G_u), 0 --> G_u + P
    σ_b, G_u + P --> 0
end ρ_u ρ_b d σ_u σ_b N
```

(thanks to @palmtree2013 for the example). I've implemented equivalent functionality in my FSP package but it might be nicer if we could just move this straight to Catalyst.jl.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 *.jl.*.cov
 *.jl.mem
 Manifest.toml
+
+# Vim files
+*.swp
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,7 @@ uuid = "479239e8-5488-4da2-87a7-35f2df7eef83"
 version = "6.12.1"
 
 [deps]
+AbstractAlgebra = "c3fe647b-3220-5bb0-a1ea-a7954cac585d"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
--- a/src/Catalyst.jl
+++ b/src/Catalyst.jl
@@ -12,6 +12,7 @@ import MacroTools
 import Base: (==), merge!, merge
 using Symbolics
 using Latexify, Requires
+using AbstractAlgebra
 
 # as used in Catlab
 const USE_GV_JLL = Ref(false)
@@ -42,6 +43,7 @@ include("networkapi.jl")
 export species, params, reactions, speciesmap, paramsmap, numspecies, numreactions, numparams
 export make_empty_network, addspecies!, addparam!, addreaction!
 export dependants, dependents, substoichmat, prodstoichmat, netstoichmat
+export conservationlaws, conservedquantities
 
 # for Latex printing of ReactionSystems
 include("latexify_recipes.jl")
--- a/src/networkapi.jl
+++ b/src/networkapi.jl
@@ -183,6 +183,47 @@ function netstoichmat(rn; smap=speciesmap(rn))
 end
 
 
+######################## conservation laws ###############################
+
+""" 
+    conservationlaws(netstoichmat::AbstractMatrix)::Matrix
+
+Given the net stoichiometry matrix of a reaction system, computes a matrix
+of conservation laws, each represented as a row in the output. 
+"""
+function conservationlaws(nsm::AbstractMatrix)::Matrix
+    n_reac, n_spec = size(nsm)
+    
+    # We basically have to compute the left null space of the matrix
+    # over the integers; this is best done using its Smith Normal Form.
+    nsm_conv = AbstractAlgebra.matrix(AbstractAlgebra.ZZ, nsm)
+    S, T, U = AbstractAlgebra.snf_with_transform(nsm_conv)
+    
+    # Zero columns of S (which occur after nonzero columns in SNF)
+    # correspond to conserved quantities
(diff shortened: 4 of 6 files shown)


TorkelE:
If you got to the test folder, and then the "test_netowkrs.jl" file, it declares a vector `reaction_networks_constraint` which contains 10 different reaction networks, all I think should be subject to having conservation laws. These might be useful for creating more tests.

Also, would it be possible/desirable, to have some way of knowing if a network had gone through this process, and have the parameter N be deduced from the initial conditions?

It would also be good if one could designate which parameters to be used for the conserved concentrations (either declaring new ones, or having ones already present in the system).

isaacsas:
@kaandocal is the Smith Normal form routine what limits the size that can effectively be used?

kaandocal:
> If you got to the test folder, and then the "test_netowkrs.jl" file, it declares a vector reaction_networks_constraint which contains 10 different reaction networks, all I think should be subject to having conservation laws. These might be useful for creating more tests.

Good point, I had a brief look earlier but didn't find any conservation laws for the first few. The code does return some for the later  ones.

> Also, would it be possible/desirable, to have some way of knowing if a network had gone through this process, and have the parameter N be deduced from the initial conditions?

I think you would have to tag the reduced system in some way, referring to the original system and the parameter(s) N. The function `conservedquantities` takes a state vector and deduces the parameters.

> It would also be good if one could designate which parameters to be used for the conserved concentrations (either declaring new ones, or having ones already present in the system).

I think we would have to use new parameters and name them consistently.

> @kaandocal is the Smith Normal form routine what limits the size that can effectively be used?

Exactly. There are probably better ways to find all conservation laws - the task boils down to finding the left null space of the net stoichiometry matrix over the integers

isaacsas:
This looks good to me, so I’ll merge for now, but we should work to make this usable for large systems and to reduce the network using the conservation laws similar to `structural_simplify`.

kaandocal:
I've just added more unit tests in accordance with @TorkelE's suggestion, they're in my branch

kaandocal:
This (old) [discussion](https://mathoverflow.net/questions/42637/exact-computation-of-the-null-space-basis-of-an-integer-matrix) might be useful. SNF roughly has polynomial computational complexity but I think it's worse than cubic. A sparse SNF implementation would be cool. I'll have a look at `structural_simplify` and how to extend it.

isaacsas:
Can you make a new PR with the tests?

isaacsas:
The idea would be to add and use an observed field for specifying the eliminated variables and the algebraic formula to calculate them, and then generate a new network with that field and a reduced set of ODEs. 

Making this work for large systems has to be a solved problem. This should be a standard calculation I’d think for chemical engineering problems too…

TorkelE:
> > If you got to the test folder, and then the "test_netowkrs.jl" file, it declares a vector reaction_networks_constraint which contains 10 different reaction networks, all I think should be subject to having conservation laws. These might be useful for creating more tests.
> 
> Good point, I had a brief look earlier but didn't find any conservation laws for the first few. The code does return some for the later ones.

If you check in the file (it is the networks 21 to 30 I think), then after the networks the conservation law(s) should be written out.

yewalenikhil65:
> Exactly. There are probably better ways to find all conservation laws - the task boils down to finding the left null space of the net stoichiometry matrix over the integers.

Hi @kaandocal 

I was trying this [here](https://github.com/SciML/Catalyst.jl/issues/280#issuecomment-804021518) sometime back but couldn't proceed forward. 
Glad to see this PR !

kaandocal:
@isaacsas is there any code in `structural_simplify` that should be used or shall I just have a look at writing structural simplification code from scratch?

isaacsas:
I'd suggest just taking a look at how it works and how it uses the observed field to handle eliminated variables. We probably want a different name since `structural_simplify` does something different. The idea here would be to give a set of equality constraints and use them to eliminate as many variables as possible (or I guess to allow the user to specify which variables to eliminate), and to then generate the corresponding reduced ODESystem. The elimination code can probably all work in ModelingToolkit at the level of ODESystems instead of ReactionSystems.

isaacsas:
Just to clarify, we can use the ReactionSystem representation to generate the equality constraints to apply, but the general simplification seems like it would make sense for general ODEs + equality constraints.

PR #340, opened 2021-06-03
Adding support for reduction of ReactionSystems

kaandocal:
This PR is for automatic simplification of ReactionSystems by using conservation laws to eliminate variables (see [here](https://github.com/SciML/Catalyst.jl/pull/337) and [here](https://github.com/SciML/Catalyst.jl/issues/280#issuecomment-804021518)). The main addition is the function `reducereacsys` which takes a reaction system and a list of species to eliminate, checks whether that is possible and if yes, returns a reduced system with fewer state variables and more parameters (including all conserved quantities).

I'm not sure how best to write unit tests for this so none are supplied at the moment, but it yields the correct answer e.g. for the example system in the first link above.

The naming conventions and interface are fully subject to change, happy about any suggestions :)


src/reduction.jl
```diff
@@ -0,0 +1,172 @@
...
+    syms_cons = conssyms(rn, C)
+    
+    states_new_idcs = filter(x -> !(x in elided_species), 1:size(C, 2))
+    states_new = map(idx -> rn.states[idx], states_new_idcs)
+    
+    ps_new = [ rn.ps; syms_cons ]
+    name_new = Symbol(rn.name, "_R")
+    
+    subs = getsubsdict(rn, C, elided_species, syms_cons)
+    reactions_new = reducedreactions(rn, C, elided_species, syms_cons;
```
ChrisRackauckas, line 168: you're going to want to add the removed species to observables with the reconstruction relations.

kaandocal: Aha, I didn't know that that field was for!

ChrisRackauckas: It makes so you can do things like `sol[x]`, even if `x` doesn't exist, to give the user a way to not worry about what is reduced out.

src/reduction.jl
```diff
@@ -0,0 +1,174 @@
...
+""" 
+    canelide(cons_laws::AbstractMatrix, species::AbstractMatrix)
+
+Check if the given species can be simultaneously elided using the given matrix of conservation laws.
+"""
+function canelide(C::AbstractMatrix, species::AbstractVector) 
+    # This amounts to checking that the transformation [ (kept) (elided) ] -> [ (kept) (conserved quantities) ]
+    # is injective, equivalently that the linear map (elided) -> (conserved) is injective,
+    # equivalently that it has full row rank
+    rank(@view C[:,species]) == length(species)
```
YingboMa, line 12: I am not sure about this. Are all the coefficients integers?

src/reduction.jl
```diff
@@ -0,0 +1,174 @@
...
+"""
+function getinvmap(C::AbstractMatrix, elided_species::AbstractVector)
+    @assert canelide(C, elided_species)
+    
+    n_elided = length(elided_species)
+    n_kept = size(C, 2) - n_elided
+    n_cons = size(C, 1)
+    kept_species = filter(x -> !(x in elided_species), 1:size(C, 2))
+    M_cons_float = pinv(C[:,elided_species])
+    M_cons = round.(Int, M_cons_float)
```
YingboMa, line 31: This is really not ideal for numerous reasons. If we want to handle this correctly, we need to use an integer algorithm where no round-off errors occur.

YingboMa (commented):
Computing the rank and pseudo-inverse of integer valued matrices is not ideal. Are there papers that actually suggest this? This is really bad from the perspective of numerical analysis. Better integer algorithms must exist, right? Computationally, if there are thousands of species, pseudo-inverse will just take too long to compute.

Also, we need tests on the functionalities.

@isaacsas what do you think of this PR?

src/reduction.jl
```diff
@@ -0,0 +1,174 @@
...
+
+"""
+    conssyms(rn::ReactionSystem, C::AbstractMatrix)
+
+Return a list of new SymbolicUtils symbols for the conserved valued determined by the
+conservation law matrix C. These become parameters for the reduced reaction system.
+
+TODO: Use gensymb to ensure uniqueness, or an alternative.
+"""
+conssyms(rn::ReactionSystem, C::AbstractMatrix) = map(id -> Sym{Number}(id), Symbol("_C", i) for i in 1:size(C,1))
```
YingboMa, line 47:
```suggestion
conssyms(rn::ReactionSystem, C::AbstractMatrix) = map(i -> Sym{Real}(Symbol("_C", i)), for i in 1:size(C,1))
```

src/reduction.jl
```diff
@@ -0,0 +1,174 @@
...
+
+"""
+    jumpratelaw(rx; rxvars=get_variables(rx.rate), combinatoric_ratelaw=true)
+
+Variant of `jumpratelaw` that ignores variables not in `rxvars`, for 
+partial creation of the full jump rate law. The original doesn't seem to use 
+that parameter at all.
+"""
+function newjumpratelaw(rx; rxvars=get_variables(rx.rate), combinatoric_ratelaw=true)
+    ModelingToolkit.@unpack rate, substrates, substoich, only_use_rate = rx
```
YingboMa, line 87: Parameters.jl is already available in Catalyst. Just writing `@unpack` would be fine.

src/reduction.jl
```diff
@@ -0,0 +1,174 @@
...
+    rl = rate
+    if !only_use_rate
+        coef = one(eltype(substoich))
+        for (i,stoich) in enumerate(substoich)
+            substrates[i] in rxvars || continue
+            
+            s   = substrates[i]
```
YingboMa, line 94: Inconsistent whitespace.

src/reduction.jl
```diff
@@ -0,0 +1,174 @@
...
+            rl *= s
+            isone(stoich) && continue
+            for i in one(stoich):(stoich-one(stoich))
```
YingboMa, line 97: Why `one(stoich)` instead of just `1: stoich-1`?

src/reduction.jl
```diff
@@ -0,0 +1,174 @@
...
+"""
+function getsubsdict(rn::ReactionSystem, C::AbstractMatrix, elided_species::AbstractVector, syms_cons)
+    M_kept, M_cons = getinvmap(C, elided_species)
+    ret = Dict{Any,Any}()
+    syms = states(rn)
+    kept_species_idcs = filter(x -> !(x in elided_species), 1:size(C, 2))
+    kept_species = map(idx -> syms[idx], kept_species_idcs)
+    for (i, species) in enumerate(elided_species)
+        sym = syms[species]
+        ret[sym] = sum(view(M_kept, i, :) .* kept_species) + sum(view(M_cons, i, :) .* syms_cons)
```
YingboMa, line 63: This could be made non-allocating by writing out a for-loop.

src/reduction.jl
```diff
@@ -0,0 +1,174 @@
...
+
+Return a dictionary of the form `elided_species => f(state_kept, cons)` of symbolic expressions
+expressing the elided species in terms of the kept species and the conserved quantities.
+"""
+function getsubsdict(rn::ReactionSystem, C::AbstractMatrix, elided_species::AbstractVector, syms_cons)
+    M_kept, M_cons = getinvmap(C, elided_species)
+    ret = Dict{Any,Any}()
+    syms = states(rn)
+    kept_species_idcs = filter(x -> !(x in elided_species), 1:size(C, 2))
+    kept_species = map(idx -> syms[idx], kept_species_idcs)
```
YingboMa, line 60: This could be a filter as `kept_species_idcs` is not used later.

src/reduction.jl
kaandocal: So far yes. In any case at this stage of the code it is assumed the species are indices into the `states` vector

src/reduction.jl
kaandocal: I'll see if I can use Smith Normal Form for this, again

src/reduction.jl
kaandocal: I just copied this code from my version of ModelingToolkit :)

kaandocal:
@YingboMa It was already pointed out in #337 that the current approach of obtaining conservation laws by using the Smith Normal Form of the net stoichiometry matrix is likely not scalable. Finding the conservation laws is equivalent to computing the left null space of the matrix, and finding the transformation from kept species + conserved quantities to the removed species involves obtaining the left inverse of a submatrix, both operations being performed over integers. I'd be happy about any suggestions for speeding this up.

YingboMa:
In MTK, we use the Bareiss algorithm for finding the nullspace of an integer-valued matrix. I think with unit tests, we can merge this PR, and improve on the algorithms later.

kaandocal:
Do you by any chance have some more documentation on the MTK implementation? By the way I think the lines around `pinv` are wrong, I will have a closer look later on

YingboMa:
ModelingToolkit (MTK) implements the algorithm in section 5 of https://ep.liu.se/ecp/132/064/ecp17132565.pdf . Do you have some recommended papers on this subject?

src/reduction.jl
kaandocal: What would be the best way to determine the type of the sum?

kaandocal:
This is something I'm working on at the moment, I'll see if I can dig up some references. Currently it's all more or less on a case-by-case bases and dependent on specifics of the reaction system (quasiequilibrium or quasi-steady-state assumptions), at least for stochastic reaction networks. There's much more on systems of ODEs and probably SDEs that I'm not very familiar with...

isaacsas:
Thanks again for getting this started -- this PR was a big help in thinking about how to do this!

https://github.com/SciML/Catalyst.jl/pull/506 gets this together in a pure integer manner using ModelingToolkit's new nullspace method, and adds elimination for ODEs. I think it should be straightforward in a followup PR to add reduction for jump systems, though I must admit I'm not sure it is actually beneficial with regards to Gillespie simulations (since MassActionJumps will become ConstantRateJumps, which are generally slower...). But I plan to put it in for jumps and SDEs as an option in any case.

PR #343, opened 2021-06-21
Use explicit time dependence

lamorton:
ModelingToolkit now requires explicit time dependence for variables in time-dependent systems. See https://github.com/SciML/ModelingToolkit.jl/pull/1064


ChrisRackauckas (approved):

lamorton:
All Catalyst tests should pass now, but there are still instances where systems are constructed with variables that are not time-dependent but also not forwarded to ODESystems.  Maybe this is OK b/c those could be turned into time-independent systems.

ChrisRackauckas:
Thanks!

PR #352, opened 2021-07-26
hillr to hillR

thompsonmj:
`hillr` changed to `hillR` to avoid error `UndefVarError: hillr not defined`

### DIFF
--- a/docs/src/tutorials/using_catalyst.md
+++ b/docs/src/tutorials/using_catalyst.md
@@ -21,9 +21,9 @@ is used for zeroth order production and first order degradation reactions:
 
 ```julia
 repressilator = @reaction_network begin
-    hillr(P₃,α,K,n), ∅ --> m₁
-    hillr(P₁,α,K,n), ∅ --> m₂
-    hillr(P₂,α,K,n), ∅ --> m₃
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Why the capital @isaacsas ?

isaacsas:
IDK, blame @TorkelE :)

We could add a PR adding `hillr` and such too.

ChrisRackauckas:
`hillr` would match the style standards, so I would prefer it.

isaacsas:
So drop the current naming and have a breaking release? Or keep the docs as is but add back `hillr`?

isaacsas:
i.e. remove `hillR` and such?

TorkelE:
Yeah, seems `hillr` is not enabled. Totally could add that as well. Not sure why it didn't make it when I remade how the functions worked. Is there a reason we would want `hillR` actively removed? Otherwise I'll just make a PR add `hillr` in.

TorkelE:
see https://github.com/SciML/Catalyst.jl/pull/353

isaacsas:
@thompsonmj I'm going to close this since @TorkelE instead added in the lowercase function versions. Thanks for bringing this to our attention!

PR #353, opened 2021-07-26
Fix function names that are enabled

TorkelE:
With this pr these are possible:

### Michaelis-Menten
`mm`, `Mm`, `MM`

### Repressive Michaelis-Menten
`mmR`, `MmR`, `MMR`, `mmr`, `Mmr`, `MMr`

### Hill
`hill`, `Hill`

### Repressive Hill
`hillR`, `HillR`, `hillr`, `Hillr`

### Combined Activating/Repressive Hill
`hillC`, `hillAR`, `HillC`, `HillAR`, `hillc`, `hillar`, `Hillc`, `Hillar`

This is way more than what's actually needed, which is only really useful to cater to people with obscure preferences. If anyone has any opinions we could remove a couple.

### DIFF
--- a/src/registered_functions.jl
+++ b/src/registered_functions.jl
@@ -23,8 +23,11 @@ Symbolics.derivative(::typeof(MM), args::NTuple{3,Any}, ::Val{3}) = - args[2]*ar
 mmR(X,v,K) = v*K / (X + K)
 MmR(X,v,K) = v*K / (X + K)
 MMR(X,v,K) = v*K / (X + K)
+mmr(X,v,K) = v*K / (X + K)
+Mmr(X,v,K) = v*K / (X + K)
+MMr(X,v,K) = v*K / (X + K)
 
-@register mmR(X,v,K); @register MmR(X,v,K); @register MMR(X,v,K);
+@register mmR(X,v,K); @register MmR(X,v,K); @register MMR(X,v,K); @register mmr(X,v,K); @register Mmr(X,v,K); @register MMr(X,v,K);
 
 Symbolics.derivative(::typeof(mmR), args::NTuple{3,Any}, ::Val{1}) = - (args[2]*args[3]) / (args[1]+args[3])^2
 Symbolics.derivative(::typeof(mmR), args::NTuple{3,Any}, ::Val{2}) = args[3]/(args[1]+args[3])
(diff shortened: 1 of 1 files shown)


isaacsas:
I'd vote to keep backwards compatibility, but make sure the docs use the all lowercase versions.

isaacsas (approved):

thompsonmj:
This makes me squirm just a bit because it deviates from the Style Guide's [convention](https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/) for naming functions. 

My first preference would be to nix everything butt the `hillr` style options just to rip off the bandage.
Second preference would be preserve backwards compatibility and add only the `hillr` style options to keep close to the style guide (remove all the other iterations from PR).

ChrisRackauckas:
Removing functions has easy deprecations, so I'd rip off the bandaid. Having alternative namings for functions is a bad idea in general anyways, so we should really be picking one for each. Also, exporting anything like `mm` will surely cause problems in the future, so this might be good to move to a submodule so it can be separately `using`ed.

TorkelE:
Updated version with only small letters, leaving us with

### Michaelis-Menten
`mm`

### Repressive Michaelis-Menten
 `mmr`

### Hill
`hill`

### Repressive Hill
`hillr`

### Combined Activating/Repressive Hill
`hillar`,


Most of them I don't see people giving different names to, especially in a script where they do CRN modelling. The exception is if someone declares a `hill()` function with the same context, but different functionality. 

I know some functions in the Plots.jl package are fetched using e.g. `using Plots.Measures`. If we decide to move these separately, would it be possible to have them in something like `using Catalyst.Functions`?

ChrisRackauckas:
Yeah, you can have a submodule so that `using Catalyst.Functions` exports them all, but by default `mm` is contained.

TorkelE:
Figured it might be best to merge this to have it done, and then open an issue on moving stuff: https://github.com/SciML/Catalyst.jl/issues/355

Also, some residual changes for this are in https://github.com/SciML/Catalyst.jl/pull/354

yewalenikhil65:
@isaacsas  can we expect a newer release of catalyst soon with these updated function names ? wanted to include some test networks for renewed PR https://github.com/SciML/Catalyst.jl/pull/326 with `hillr` and `mmr` .  v6.14.0 seems to demand `hillR` and `mmR`

isaacsas:
I'll make a new release now.

PR #364, opened 2021-07-31
fix for new reaction sys constructor

isaacsas:
Needs a new ModelingToolkit release and then tests to pass before merging.


ChrisRackauckas (approved):

isaacsas:
@ChrisRackauckas is there any reason not to bump compat to the latest MT release (6.0) and Symbolics 2.0?

PR #370, opened 2021-08-10
Add reactionsystems

isaacsas:
Needs a ~~ModelingToolkit release~~, ~~doc updates~~, and tests to pass. Tests currently pass locally with https://github.com/SciML/ModelingToolkit.jl/pull/1187.

@TorkelE I had to remove the steady-state tests as they are failing consistently again. Everything else seems to pass ok though.


isaacsas:
We should probably switch those tests to use NonlinearProblems I guess.

PR #373, opened 2021-08-11
Reaction complexes v2

isaacsas:
@yewalenikhil65 so as not to pollute your original PR I created a branch off it and made updates. What do you think? The tests you added now all pass for me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ Manifest.toml
 
 # Vim files
 *.swp
+
+# vscode stuff
+.vscode
+.vscode/*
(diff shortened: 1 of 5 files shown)


isaacsas:
I think creating a `ReactionSystem` from Z,B matrices is for ReactionNetworkImporters. We can add a new importer for that type of matrix input. Feel free to open a PR there if you want. 

Adding visualization here makes sense.

isaacsas:
@yewalenikhil65 nice job on this one!

isaacsas:
It would also be good to add a new tutorial showing how to use these matrices to do something interesting (maybe once you add the visualization features).

yewalenikhil65:
@isaacsas Thanks sir for your inputs.

PR #377, opened 2021-08-14
ModelingToolkit unit validation logic

lamorton:
This will be necessary (in order to take advantage of unit validation) when [ModelingToolkit.jl#1156](https://github.com/SciML/ModelingToolkit.jl/pull/1156#) lands.

### DIFF
--- a/src/reactionsystem.jl
+++ b/src/reactionsystem.jl
@@ -174,10 +174,11 @@ function ReactionSystem(eqs, iv, species, params;
                         default_u0=Dict(),
                         default_p=Dict(),
                         defaults=_merge(Dict(default_u0), Dict(default_p)),
-                        connection_type=nothing)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

isaacsas:
Do we need to add the validation check in the ReactionSystem constructor too? (I noticed it in your MT PR, but that lives here now.)

lamorton:
> 
> 
> Do we need to add the validation check in the ReactionSystem constructor too? (I noticed it in your MT PR, but that lives here now.)

Yup, just pushed it.

isaacsas:
Could you propagate the checks keyword to the outer constructors too so users can use it easily?

isaacsas:
Looks great. Thanks for all the effort with adding in unit support!

PR #381, opened 2021-08-19
Unit validation and tests 

isaacsas:
I also added a `show` definition for `Reaction`s that prints out stuff like
```
Reaction{Any, Int64}: γ, A + B --> 2B
```
so a user could easily round-trip to/from the DSL. (I kept the summary with type info, but could drop that from `show` if anyone thinks that makes sense.)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Catalyst"
 uuid = "479239e8-5488-4da2-87a7-35f2df7eef83"
-version = "8.2"
+version = "8.2.0"
 
 [deps]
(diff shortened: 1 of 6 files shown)


isaacsas:
@lamorton could you check that I haven't messed anything up w/regards to the MT plans for validation? I changed how it works a bit to work at the level of a `Reaction`, and then added a version that works on the whole `ReactionSystem` (which is what is now checked in the `ReactionSystem` constructor).

lamorton:
> @lamorton could you check that I haven't messed anything up w/regards to the MT plans for validation?

It looks OK to me. I'm unsure exactly how things will evolve on the MT side, but it should be fine if you have specialized the validation to suite your needs here.

Issue #386, opened 2021-08-22
Sparse `stoichmat` dispatches?


yewalenikhil65:
Does it make sense to add Sparse `stoichmat` dispatches, now that we have many `stoich` functions ?
If yes, how do we decide for `sparse` or `dense` version as all these `stoichmat` functions take reaction system as argument ?
Maybe `numreactions(rn)` could be deciding factor ?(If yes, how much threshold ?)

yewalenikhil65:
Or else, we can also pass a `Bool` as flag for `sparsity` in the function. default set to `true` ?

isaacsas:
I guess we could make a three argument form for the various matrix functions. Maybe something like
```julia
function netstoichmat(::Type{SparseMatrixCSC}, rn::ReactionSystem; smap=speciesmap(rn))
...
end
function netstoichmat(::Type{Matrix}, rn::ReactionSystem; smap = speciesmap(rn))
# the existing function code
end
function netstoichmat(rn::ReactionSystem, smap=speciesmap(rn))
   netstoichmat(Matrix, rn; smap=smap)
end
```

yewalenikhil65:
> ```julia
> function netstoichmat(rn::ReactionSystem, smap=speciesmap(rn))
>    netstoichmat(Matrix, rn; smap=smap)
> end
> ```
This part I didn't understand properly. 

But
I was thinking along following lines
```
function netstoichmat(rn::ReactionSystem, sparsity::Bool;smap=species(rn)) 

if sparsity == true
  #sparse code
else
  # existing function code
end


end
```

Also we should not use `spzeros` way in `# sparse code` part., but try constructing using `sparse(Is, Js, Vs) ` way as it's more effective `@btime` wise

yewalenikhil65:
hey @isaacsas , did a little benchmarking on large reaction mentioned here https://github.com/SciML/Catalyst.jl/issues/378#issue-973096530
using `spzeros` vs `sparse(I,J,K)`
```julia
function substoichmat(rn::ReactionSystem; smap=speciesmap(rn))
    smat = spzeros(Int,numreactions(rn),numspecies(rn))
    for (k,rx) in enumerate(reactions(rn))
        stoich = rx.substoich
        for (i,sub) in enumerate(rx.substrates)
            smat[k,smap[sub]] = stoich[i]
        end
    end
    smat
end

julia> @benchmark substoichmat(rn)
BenchmarkTools.Trial: 5530 samples with 1 evaluation.
[...]

PR #387, opened 2021-08-26
Sparse representation of `stoichiometric matrices` functions and matrices in complex-based functions

yewalenikhil65:
from https://github.com/SciML/Catalyst.jl/issues/386#issuecomment-904876595

example. We use a second  function argument of type `Bool` to switch `sparsity` on or off. Default is dense representation as earlier, i.e.`sparsity=false`
```julia
rn = @reaction_network begin
    k₁, 2A --> B
    k₂, A --> C
    k₃, C --> D
    k₄, B + D --> E
end k₁ k₂ k₃ k₄

julia> substoichmat(rn)   # usual function or substoichmat(rn,false)
4×5 Matrix{Int64}:
 2  0  0  0  0
 1  0  0  0  0
 0  0  1  0  0
 0  1  0  1  0

julia> substoichmat(rn,true)     # sparse representation
4×5 SparseMatrixCSC{Int64, Int64} with 5 stored entries:
 2  ⋅  ⋅  ⋅  ⋅
 1  ⋅  ⋅  ⋅  ⋅
 ⋅  ⋅  1  ⋅  ⋅
 ⋅  1  ⋅  1  ⋅
```

### DIFF
--- a/src/networkapi.jl
+++ b/src/networkapi.jl
@@ -465,11 +465,12 @@ Given the net stoichiometry matrix of a reaction system, computes a matrix
 of conservation laws, each represented as a row in the output. 
 """
 function conservationlaws(nsm::AbstractMatrix)::Matrix
-    n_reac, n_spec = size(nsm)
+    n_spec,n_reac = size(nsm)
     
     # We basically have to compute the left null space of the matrix
     # over the integers; this is best done using its Smith Normal Form.
-    nsm_conv = AbstractAlgebra.matrix(AbstractAlgebra.ZZ, nsm)
+    # note, we transpose as this was written when netstoichmat was reac by spec
+    nsm_conv = AbstractAlgebra.matrix(AbstractAlgebra.ZZ, nsm')
     S, T, U = AbstractAlgebra.snf_with_transform(nsm_conv)
     
     # Zero columns of S (which occur after nonzero columns in SNF)
--- a/src/networkapi.jl
+++ b/src/networkapi.jl
@@ -135,105 +135,108 @@ end
 
 
 """
-    substoichmat(rn,sparsity=false; smap=speciesmap(rn))
+    substoichmat(rn; sparse=false, smap=speciesmap(rn))
 
-Returns the substrate stoichiometry matrix
+Returns the substrate stoichiometry matrix, S, with Sᵢⱼ the stoichiometric
+coefficient of the ith substrate within the jth reaction.
 
 Note:
-- Set sparsity=true for sparse representation
+- Set sparse=true for a sparse matrix representation
 """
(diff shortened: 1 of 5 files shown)


isaacsas:
@yewalenikhil65 take a look and let me know your thoughts on the changes. Mainly I made `sparse=true` a keyword to use for picking a sparse matrix. This is consistent with some of the usage in ModelingToolkit functions.

isaacsas:
I also should have made all matrices have columns = reactions now.

yewalenikhil65:
> @yewalenikhil65 take a look and let me know your thoughts on the changes. Mainly I made `sparse=true` a keyword to use for picking a sparse matrix. This is consistent with some of the usage in ModelingToolkit functions.

This looks nice, exactly how i imagined.. Thanks!
I was also little concerned of    https://github.com/yewalenikhil65/Catalyst.jl/blob/6a2b79eafac8376586c5483491d8980d5da7dd17/src/networkapi.jl#L473
does transposing here add any considerable computational cost ? I remember conversation thread regarding SNF to be made optimized later or searching for alternative LinearAlgebra based methods

src/networkapi.jl
```diff
@@ -179,7 +179,7 @@ coefficient of the ith product within the jth reaction.
 Note:
 - Set sparse=true for a sparse matrix representation
 """
-function prodstoichmat(::Type{SparseMatrixCSC{Int}}, rn::ReactionSystem; smap=speciesmap(rn))
+function prodstoichmat(::Type{SparseMatrixCSC{Int,Int}}, rn::ReactionSystem; smap=speciesmap(rn))
```
yewalenikhil65, line 182: Is it good idea to be using `SparseMatrixCSC{Int,Int}` entries ? i also remember some conversation thread in Catalyst issues to leave scope for non-integer stoichiometries later  probablt in furture

isaacsas:
> > @yewalenikhil65 take a look and let me know your thoughts on the changes. Mainly I made `sparse=true` a keyword to use for picking a sparse matrix. This is consistent with some of the usage in ModelingToolkit functions.
> 
> This looks nice, exactly how i imagined.. Thanks!
> I was also little concerned of https://github.com/yewalenikhil65/Catalyst.jl/blob/6a2b79eafac8376586c5483491d8980d5da7dd17/src/networkapi.jl#L473
> does transposing here add any considerable computational cost ? I remember conversation thread regarding SNF to be made optimized later or searching for alternative LinearAlgebra based methods
> 
> Anyway, i will get back to fixing PR In ReactionNetworkImporters after you release new Catalyst version

I don't think this should be a big cost relative to computing the SNF, so using the transpose shouldn't be an issue. I think what @YingboMa was maybe suggesting is that we should implement and use the Bareiss method instead of SNF for finding the nullspace basis.

Issue #390, opened 2021-08-29
`Linkage classes`, `deficiency`, Part 1


yewalenikhil65:
[These](https://github.com/JuliaGraphs/LightGraphs.jl/blob/master/src/connectivity.jl) are a nice set of methods that could be useful in evaluating

-  [linkage classes](https://reaction-networks.net/wiki/Reaction_graph#Linkage_classes) in the network 
- [reversibility(weakly or strongly reversible)](https://reaction-networks.net/wiki/Reaction_graph#Reversibility)
- [deficiency](https://reaction-networks.net/wiki/Reaction_graph#Deficiency) of the reaction network(once we know the number of linkage classes)

But would mean dependency on `LightGraphs.jl` (we already have such dependency in `ModelingToolkit` i believe). Is it advisable to pursue this with a view of PR for `Catalyst`? Or is there existing functionality related to this in `Graphviz` ?

yewalenikhil65:
Rough idea for `Linkage class`
```julia
using LightGraphs
using Catalyst, ModelingToolkit

function create_graph(incidence::Matrix{Int64})
           n = size(incidence)[1]
           graph = LightGraphs.DiGraph(n)    # find nodes where edge goes out and edge goes

           v_out = findall(x -> x == -1, incidence) #  edge goes out of vertex
           v_in = findall(x -> x == 1, incidence)  # edge goes in of vertex
           # match the Cartesian index between the two lists, add edge
           for i in v_in
               for o in v_out
                   if o[2] == i[2]
                       add_edge!(graph, o[1], i[1])
                   end
               end
           end
           return graph
       end


rn = @reaction_network begin
    k₁, 2A --> B
    k₂, A --> C
    k₃, C --> D
    k₄, B + D --> E
end k₁ k₂ k₃ ### k₄
B = reactioncomplexes(rn)[2]
g = create_graph(B)
julia> connected_components(G)
3-element Vector{Vector{Int64}}:
 [1, 2]
 [3, 4, 5]
 [6, 7]
```
where `connected_components` return the vectors of indices of `complexes`  `[`2A`, `B`, `A`,`C`,`D`,`B+D`,`E`]`

yewalenikhil65:
To avoid dependency on `LightGraphs.jl` , we could referr to some code from https://github.com/JuliaGraphs/LightGraphs.jl/blob/master/src/SimpleGraphs/simpledigraph.jl

isaacsas:
[...]

PR #394, opened 2021-09-02
[WIP] `linkageclasses` and `deficiency` of reaction network

yewalenikhil65:
from https://github.com/SciML/Catalyst.jl/issues/390

based on defintions of [linkageclasses](https://reaction-networks.net/wiki/Reaction_graph#Linkage_classes)
and [deficiency](https://reaction-networks.net/wiki/Reaction_graph#Deficiency)


src/networkapi.jl
```diff
@@ -465,6 +465,68 @@ function complexoutgoingmat(rn::ReactionSystem; sparse=false, B=reactioncomplexe
...
+      else
+         add_edge!(graph, row[2], row[1])
+      end
+   end
+   return graph
+end
+
+
+"""
+Given a directed simple graph ,created from incidence matrix
```
isaacsas, line 507: Start the docstring with
```
     linkageclasses(incidencegraph)
```

src/networkapi.jl
```diff
@@ -465,6 +465,68 @@ function complexoutgoingmat(rn::ReactionSystem; sparse=false, B=reactioncomplexe
...
+   LightGraphs.connected_components(graph)
+end
+
+
+"""
+Given a ReactionSystem, returns deficiency of the reaction network
+defined as,
+deficiency = number of reaction complexes - number of linkage classes - dimension of stochiometric subspace
+"""
+function deficiency(rs::ReactionSystem;sparse::Bool=false)
```
isaacsas, line 522:
```suggestion
function deficiency(rs::ReactionSystem, ns, ig, lc=linkageclasses(ig))
```
where `ns` is the netstoichiometry, `ig` is the incidence graph, and `lc` are the linkage classes. You can get the number of complexes from the number of nodes in `ig` right? (There should be a Lightgraphs function for that.)
```

File: src/networkapi.jl [hunk 3]
```diff
@@ -465,6 +465,68 @@ function complexoutgoingmat(rn::ReactionSystem; sparse=false, B=reactioncomplexe
...
+end
+
+
+"""
+Given a directed simple graph ,created from incidence matrix
+of reaction network, returns Vector of indices of directly connected
+reaction-complexes(more precisely indices of reactioncomplexes(network)[1])
+in reactions
+"""
+function linkageclasses(graph::SimpleDiGraph{Int64})
```
isaacsas, line 512:
```suggestion
function linkageclasses(incidencegraph)
```

src/networkapi.jl
```diff
@@ -465,6 +465,68 @@ function complexoutgoingmat(rn::ReactionSystem; sparse=false, B=reactioncomplexe
...
+Given an incidence matrix of reaction-network in dense or sparse matrix form,
+writtens a {n ,m} directed simple graph,
+where n is number of nodes(reaction-complexes)
+and m is number of edges(reactions)
+"""
+function incidencematgraph(incidencemat::Matrix{Int})
+   @assert all(∈([-1,0,1]) ,incidencemat)
+   n = size(incidencemat,1)  # no. of nodes/complexes
+   graph = LightGraphs.DiGraph(n)
+   for col in eachcol(incidencemat)
```
isaacsas, line 481:
```suggestion
   for col in eachcol(incidencemat)
       src = 0; dst = 0;
       for i in eachindex(col)
              (col[i] == -1) && (src = i)
              (col[i] == 1) && (dst = i)
              (src != 0) && (dst != 0) && break
       end
       add_edge!(graph, src, dest)
   end
```

isaacsas (changes requested):

src/networkapi.jl
yewalenikhil65: @isaacsas  yes, number of vetrtices cn be found using `nv` function in LightGraphs
how about following , as per other functions we have written previously in API ? 

```julia
function deficiency(rs::ReactionSystem;sparse::Bool=false,ns=netstoichmat(rs;sparse),
                        ig = incidencematgraph(reactioncomplexes(rs;sparse=sparse)[2]),
                        lc=linkageclasses(ig))

   LightGraphs.nv(ig) - length(lc) - AbstractAlgebra.rank(AbstractAlgebra.matrix(AbstractAlgebra.ZZ,ns))
end

```

src/networkapi.jl
```diff
@@ -465,6 +465,84 @@ function complexoutgoingmat(rn::ReactionSystem; sparse=false, B=reactioncomplexe
...
+      end
+   end
+   return graph
+end
+
+
+"""
+      linkageclasses(incidencegraph)
+
+Given a directed simple graph ,created from incidence matrix
```
isaacsas, line 515:
```suggestion
Given a directed simple graph created from the incidence matrix
```

yewalenikhil65:
Thanks @isaacsas 
Looks good now. 
Next stop of network analysis should be about ` weakly reversible networks`, and `summarize` functions that you were talking about. And somewhere we could fit deficiency theorems  in all this.

src/networkapi.jl
yewalenikhil65: > @isaacsas yes, number of vertices can be found using `nv` function in LightGraphs
> how about following , as per other functions we have written previously in API ?
> We need `reactionSystem` to be only input,. rest all as optional kwargs supplied as below or user can pass them in case they have been pre-calculated
> 
> ```julia
> function deficiency(rs::ReactionSystem ; sparse::Bool=false, ns::AbstractMatrix = netstoichmat(rs;sparse),
>                         ig::SimpleDiGraph{Int64} = incidencematgraph(reactioncomplexes(rs;sparse=sparse)[2]),
>                         lc::Vector{Vector{Int64}} = linkageclasses(ig))
> 
>    LightGraphs.nv(ig) - length(lc) - AbstractAlgebra.rank(AbstractAlgebra.matrix(AbstractAlgebra.ZZ,ns))
> end
> ```

@isaacsas , i myself am now in doubt of this..  if `ns`, `ig` and `lc` are precalculated and supplied by user , `rs::ReactionSystem` remains unused !

isaacsas (approved):

isaacsas:
@yewalenikhil65 made some modifications. In particular, I think `deficiency` requires enough inputs that it is better to expect users to pass the correct stuff, but make sure the docs show how to do this for an example (which I added). Having `sparse` in it just felt wrong, and it was getting too complicated in all the stuff it was sequentially calculating as kwargs.

Issue #395, opened 2021-09-06
[WIP]- `subnetworks`, `linkagedeficiency` 


yewalenikhil65:
Creating here separate issue to keep track of  stuff in addition to https://github.com/SciML/Catalyst.jl/pull/394,
I think it's useful to know which reactions form `subnetworks` in the reaction network based on linkage class.

```julia
"""
   Given a `ReactionSystem` , returns a Vector of  Vector of Reactions that form   
subnetworks based on linkage class
"""
function subnetworks( rs::ReactionSystem;sparse::Bool=false,
               rmap::Vector{Vector{Pair{Int64, Int64}}} = collect(values(reactioncomplexmap(rs))),
               r::Vector{Reaction} = reactions(rs),
               ig::SimpleDiGraph{Int64} = incidencematgraph(reactioncomplexes(rs;sparse=sparse)[2]),
               lc::Vector{Vector{Int64}} = linkageclasses(ig) )

   subrs = Vector{Vector{Reaction}}()
   for i in 1:length(lc)
      rxind = unique(vcat([map(first, rmap[lc[i]][j]) for j in 1:length(lc[i])]...))
      push!(subrs , r[rxind])
   end
   subrs
end

rn = @reaction_network begin
    k₁, 2A --> B
    k₂, A --> C
    k₃, C --> D
    k₄, B + D --> E
end k₁ k₂ k₃ k₄

julia> subnetworks(rn)
3-element Vector{Vector{Reaction}}:
 [Reaction{Any, Int64}: k₁, 2A --> B]
 [Reaction{Any, Int64}: k₂, A --> C, Reaction{Any, Int64}: k₃, C --> D]
 [Reaction{Any, Int64}: k₄, B + D --> E]
```
Implying there are three  `subnetworks` that do not share reaction-complexes.. This is also evident from 
```julia
complexgraph(rn)
```
![Screenshot from 2021-09-06 05-57-56](https://user-images.githubusercontent.com/26343789/132145721-498afab1-db38-49af-af37-6dedaaa915b7.png)

yewalenikhil65:
This should form a basis for finding `deficiency` of each linkage class, (PR https://github.com/SciML/Catalyst.jl/pull/394 involves deficiency of whole network).

These studies could create a basis for future work in steady-state analysis of certain reaction network.

yewalenikhil65:
"""
   linkagedeficiency(network ;sparse ,subrn,ig ,lc )

[...]

PR #396, opened 2021-09-07
`subnetworks` and `linkagedeficieny`

yewalenikhil65:
from https://github.com/SciML/Catalyst.jl/issues/395


isaacsas:
@yewalenikhil65 since systems are the fundamental unit in ModelingToolkit, what about having `subnetworks` generate a vector of new `ReactionSystem`s each containing the appropriate species, parameters and reactiosn? Then users can just reuse the functionality from your previous PR.

Also, perhaps it makes sense to add an intermediary function that takes `reactioncomplexesmap` and inverts it, i.e. builds a mapping (vector of vectors?) from each reaction to the complexes that participate in it? Then that can just be passed into `subnetworks` as an optional kwarg.

yewalenikhil65:
> @yewalenikhil65 since systems are the fundamental unit in ModelingToolkit, what about having `subnetworks` generate a vector of new `ReactionSystem`s each containing the appropriate species, parameters and reactiosn? Then users can just reuse the functionality from your previous PR.
> 

I did think about this earlier but am confused over parameters for the sub -`reactionSystems`, as we push each reaction into vector. Rates for each reaction is a field of reaction but `parameter` doesnt seem to be the field

> Also, perhaps it makes sense to add an intermediary function that takes `reactioncomplexesmap` and inverts it, i.e. builds a mapping (vector of vectors?) from each reaction to the complexes that participate in it? Then that can just be passed into `subnetworks` as an optional kwarg.

This i think is doable and will try !

isaacsas:
You can get the parameters within the rate as follows:
```julia
ps = parameters(network)
newps = Set{eltype(ps)}()
for rx in linkageclass   # pseudocode...
    Symbolics.get_variables!(newps, rx.rate, ps)
end
# newps will now have all the parameters used in reaction rates in the reactions of the linkclass
```

yewalenikhil65:
> You can get the parameters within the rate as follows:
> 
> ```julia
> ps = parameters(network)
> newps = Set{eltype(ps)}()
> for rx in linkageclass   # pseudocode...
>     Symbolics.get_variables!(newps, rx.rate, ps)
> end
> # newps will now have all the parameters used in reaction rates in the reactions of the linkclass
> ```

Cool :grinning:  I did find `get_variables` , but i was trying to use one from `ModelingToolkit` and i was using it wrong way. Thanks for this hint ! I will update these functions today

isaacsas:
Yeah, there are a bunch of these types of useful helper functions but they aren't always prominent in the docs, and they tend to be split across MT, Symbolics, and SymbolicUtils...

yewalenikhil65:
I decided to call the intermediary function inside the `subnetworks` function because it is defined for an individual linkage class, and not as `kwarg` for `subnetworks`

src/networkapi.jl
```diff
@@ -574,6 +574,63 @@ function deficiency(ns, ig, lc)
...
+   subnetworks(network, linkage_classes ; r = reactions(network),
+                     rmap = collect(values(reactioncomplexmap(network))),
+                     ps = parameters(network))
+
+Find subnetworks corresponding to the each linkage class of reaction network
+"""
+function subnetworks(rs::ReactionSystem, lc::Vector{Vector{Int64}};
+                  r::Vector{Reaction} = reactions(rs),
+                  rmap::Vector{Vector{Pair{Int64, Int64}}} = collect(values(reactioncomplexmap(rs))),
+                  ps = parameters(rs))
```
isaacsas, line 604: `p` is probably more standard across SciML libraries.

src/networkapi.jl
```diff
@@ -574,6 +574,63 @@ function deficiency(ns, ig, lc)
...
+
+   t = (@variables t)[1]
```
isaacsas, line 606:
```suggestion
   t = get_iv(rs)
```

src/networkapi.jl
```diff
@@ -574,6 +574,63 @@ function deficiency(ns, ig, lc)
...
+   subreac = Vector{ReactionSystem}()
+   for i in 1:length(lc)
+      reacs, specs, newps = subnetworkmapping(lc[i], r, rmap[lc[i]] ,ps)
```
isaacsas, line 609: spacing...
```suggestion
      reacs, specs, newps = subnetworkmapping(lc[i], r, rmap[lc[i]], ps)
```

src/networkapi.jl
```diff
@@ -574,6 +574,63 @@ function deficiency(ns, ig, lc)
...
+      push!(subreac , ReactionSystem(reacs, t, specs, newps;name = gensym(:ReactionSystem)))
```
isaacsas, line 610: Reuse the old system name but add an "_i" maybe? 
```suggestion
      newname = Symbol(nameof(rs), "_", i)
      push!(subreac , ReactionSystem(reacs, t, specs, newps; name = newname))
```

src/networkapi.jl
```diff
@@ -574,6 +574,63 @@ function deficiency(ns, ig, lc)
...
+	
+
+"""
+	`subnetworkmapping` is an intermediary function
+	required in `subnetworks(network, linkage_classes)`
+"""
+# intermediary function for subnetwork of a particular linkage class
+function subnetworkmapping(linkageclass::Vector{Int64}, r, reacmap, pars)
+   rxind = unique(vcat([map(first, reacmap[j]) for j in 1:length(linkageclass)]...))
+   specs = unique(vcat([[r[rxind][j].substrates ; r[rxind][j].products] for j in 1:length(r[rxind])]...))
```
isaacsas, line 586: Similar comment to above. I'd try to simplify this down. Maybe use a `Set` again and then convert to a vector afterwards? (You wouldn't need to sort this one.)

src/networkapi.jl
```diff
@@ -574,6 +574,63 @@ function deficiency(ns, ig, lc)
...
+      reacs, specs, newps = subnetworkmapping(lc[i], r, rmap[lc[i]] ,ps)
+      push!(subreac , ReactionSystem(reacs, t, specs, newps;name = gensym(:ReactionSystem)))
+   end
+   subreac
+end
+
+
+
+"""
+   linkagedeficiency(subnetworks ,linkage_classes )
```
isaacsas, line 618:
```suggestion
   linkagedeficiency(subnetworks, linkage_classes)
```
what those spaces before commas instead of after...

src/networkapi.jl
```diff
@@ -574,6 +574,63 @@ function deficiency(ns, ig, lc)
...
+
+Calculates deficiency of each linkage class of reaction network,
+
+deficiency = (number of reaction complexes in linkage class - 1 -
+                  dim. of stochiometric subspace of linkage class)
+"""
+function linkagedeficiency(subnets::Vector{ReactionSystem} ,lc::Vector{Vector{Int64}})
```
isaacsas, line 625:
```suggestion
function linkagedeficiency(subnets, lcs)
```
The documentation should show how to use with an example, and indicate what the parameters are.

test/api.jl
```diff
@@ -209,7 +225,10 @@ B = [-1 0 0 1;
       0 0 1 0]
 Δ = [-1 0 0 0; 0 0 -1 -1; 0 -1 0 0; 0 0 0 0; 0 0 0 0]
 lcs = [[1,2,5],[3,4]]
-testnetwork(rns[5], B, Z, Δ, lcs, 1)
+r = reactions(rns[5])
+subrn = [[r[1],r[4],r[3]], [r[2]]  ]
```
isaacsas, line 229: Just to double check; you calculated these by hand right, and are not just reusing what your code gives as the correct answer?

isaacsas (changes requested):

isaacsas:
Also, don't forget to add the two new functions to be exported, and mention them in the `History.md` file.

src/networkapi.jl
```diff
@@ -574,6 +574,63 @@ function deficiency(ns, ig, lc)
...
+	
+"""
+   subnetworks(network, linkage_classes ; r = reactions(network),
+                     rmap = collect(values(reactioncomplexmap(network))),
+                     ps = parameters(network))
+
+Find subnetworks corresponding to the each linkage class of reaction network
+"""
+function subnetworks(rs::ReactionSystem, lc::Vector{Vector{Int64}};
+                  r::Vector{Reaction} = reactions(rs),
```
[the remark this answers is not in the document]
isaacsas: Maybe here keeping the type on `rs` makes sense since this is a common enough function other packages may have a similarly named function?

isaacsas:
@yewalenikhil65 with the new subnetwork functionality we can now determine if a network is weakly reversible right? (Doesn't this just correspond to checking if each of the subnetworks is strongly connected?)

yewalenikhil65:
> @yewalenikhil65 with the new subnetwork functionality we can now determine if a network is weakly reversible right? (Doesn't this just correspond to checking if each of the subnetworks is strongly connected?)
> 
> So we just generate the subnetwork graph, and then call `LightGraphs.is_strongly_connected` on it.

This one is on my mind and have to check it out. (Bdw, weakly reversible also may involve some subnetwork being cyclic so it wouldn't return true from `is_strongly_connected` . We may have to use `is_cyclic` type of function also to check this) 

I will check out this `LightGraphs` function today. Bdw, Sorry for the delay in updating this PR as I am little busy with family stuff, so have only phone nearby. Hopefully today I will be able to do it.

isaacsas:
No worries; take your time!

This reference says weakly reversible is equivalent to each linkage class being strongly connected: https://core.ac.uk/download/pdf/191323175.pdf

src/networkapi.jl
```diff
@@ -574,6 +574,63 @@ function deficiency(ns, ig, lc)
...
+                  rmap::Vector{Vector{Pair{Int64, Int64}}} = collect(values(reactioncomplexmap(rs))),
```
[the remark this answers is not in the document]
yewalenikhil65: this way would create some problemlike following
```julia
rn = @reaction_network begin
    k₁, 2A --> B
    k₂, A --> C
    k₃, C --> D
    k₄, B + D --> E
end k₁ k₂ k₃ k₄
julia> collect(values(reactioncomplexmap(rns[1])))  # this is good to go..
7-element Vector{Vector{Pair{Int64, Int64}}}:
 [1 => -1]
 [1 => 1]
 [2 => -1]
 [2 => 1, 3 => -1]
 [3 => 1]
 [4 => -1]
 [4 => 1]


julia> map(first, values(reactioncomplexmap(rns[1])))  
7-element Vector{Pair{Int64, Int64}}:
 1 => -1
 1 => 1
 2 => -1
 2 => 1   #  3 => -1 is missing
 3 => 1
 4 => -1
 4 => 1
```

src/networkapi.jl
yewalenikhil65: > Similar comment to above. I'd try to simplify this down. Maybe use a `Set` again and then convert to a vector afterwards? (You wouldn't need to sort this one.)

like this ?
```julia
rxind = collect(Set(vcat([map(first, reacmap[j]) for j in 1:length(linkageclass)]...)))
   specs = collect(Set(vcat([[r[rxind][j].substrates ; r[rxind][j].products] for j in 1:length(r[rxind])]...)))

```

src/networkapi.jl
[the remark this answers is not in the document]
isaacsas: Make a function to calculate it then and call that?

src/networkapi.jl
isaacsas: The point is to avoid all the allocations of intermediary vectors. Just directly insert into the Set, you can always use loops if needed…

test/api.jl
yewalenikhil65: > Just to double check; you calculated these by hand right, and are not just reusing what your code gives as the correct answer?

@isaacsas  i had checked them visually with the output of `complexgraph()` ,and see that the reaction tests match. Now with  https://github.com/SciML/Catalyst.jl/pull/396#discussion_r706517402 , since we are using `Set` , i will adjust `subrn` in `tests/api.jl` to be `Vector of Set of Reactions`.

src/networkapi.jl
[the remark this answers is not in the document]
yewalenikhil65: i have kept type of `rs` and `lcs` . removed all enforced typed on `kwargs`

src/networkapi.jl
yewalenikhil65: > The point is to avoid all the allocations of intermediary vectors. Just directly insert into the Set, you can always use loops if needed…

i wll hv to think more about this. for now have made updates that were easy to work with

isaacsas:
@yewalenikhil65 I'll tweak as I was suggesting since it is confusing.

src/networkapi.jl
```diff
@@ -574,6 +574,74 @@ function deficiency(ns, ig, lc)
...
 
+	
+
+"""
+	`subnetworkmapping` is an intermediary function
+	required in `subnetworks(network, linkage_classes)`
+"""
+# intermediary function for subnetwork of a particular linkage class
+function subnetworkmapping(linkageclass, r, reacmap, pars)
+   rxind = collect(Set(vcat([map(first, reacmap[j]) for j in 1:length(linkageclass)]...)))
```
isaacsas, line 585: @yewalenikhil65 I still don't understand why this isn't
```julia
j in linkageclass
```
Shouldn't `j` be the complex index, not the position in the linkageclass vector?

Let me know about this and I'll handle changing if needed.

isaacsas:
@yewalenikhil65 please make sure in the future you use the standard indentation of 4 characters. Looking at the PR locally it seems you indented the new functions by 3...

isaacsas (approved):

isaacsas:
LMK your thoughts on how I've modified it. This seems a reasonable tradeoff between performance vs. minimizing allocations I think.

For the future, please try to remember about using a default indentation spacing of 4 characters, or preserving what is used in the file, and try to write `f(a, b)` and not `f(a ,b)`. In most editors you can configure the default tab-width to four characters. Thanks!

src/networkapi.jl
```diff
@@ -603,19 +603,19 @@ For example, continuing the example from [`deficiency`](@ref)
...
-   for i in 1:length(lcs)
-      reacs, specs, newps = subnetworkmapping(lcs[i], rxs, complextorxmap[lcs[i]] ,p)
-      newname = Symbol(nameof(rs), "_", i)
-      push!(subreac , ReactionSystem(reacs, t, specs, newps;name = newname))
-   end
-   subreac
+    t = get_iv(rs)
+    subreac = Vector{ReactionSystem}()
+    for i in 1:length(lcs)
+        reacs, specs, newps = subnetworkmapping(lcs[i], rxs, complextorxmap, p)
```
yewalenikhil65: shouldn't we just pass `complextorxmap[lcs[i]]` ? i.e. reaction indices corresponding to complexes in linkageclass `lcs[i]`

yewalenikhil65:
> For the future, please try to remember about using a default indentation spacing of 4 characters, or preserving what is used in the file, and try to write `f(a, b)` and not `f(a ,b)`. In most editors you can configure the default tab-width to four characters. Thanks!

Sorry for this really and getting back late  ! I guess I should learn some good PR practices.

src/networkapi.jl
yewalenikhil65: > shouldn't we just pass `complextorxmap[lcs[i]]` ? i.e. reaction indices corresponding to complexes in linkageclass `lcs[i]`

I take this back! your way is efficient, just checked with `@time` n `@btime`.

isaacsas: Yeah, I was trying to avoid allocations of intermediary vectors.

yewalenikhil65:
I think this is good to go now. Lesson learnt that using loops could be more efficient sometimes than readymade functions like `vcat` and `unique` as i was doing
Thanks sir :grinning:

isaacsas:
Thanks again for putting this forward. We’re well on our way to a nice library of network analysis codes now I’d say.

yewalenikhil65:
> Thanks again for putting this forward. We’re well on our way to a nice library of network analysis codes now I’d say.

i will get back with `is_reversible` and `is_weakly_reversible`. I am still thinking that i read somewhere, that not strongly connected graphs but cyclics can be treated as `is_weakly_reversible`. WIll come back with a source for it

isaacsas:
Isn’t a cycle strongly connected since any node in it has a path to each other node?

isaacsas:
(A directed cycle…)

yewalenikhil65:
> Isn’t a cycle strongly connected since any node in it has a path to each other node?

shifting discussion here https://github.com/SciML/Catalyst.jl/issues/401

PR #397, opened 2021-09-10
Use Flint via Nemo for integer matrix calculations

isaacsas:
Also switch LightGraphs to be import only due to warnings about `degree` conflicting with Symbolics...


yewalenikhil65:
This will need to be carried forward in another PR that is currently open(using rank from nemo) . Will do it after u merge this one

isaacsas:
All merged.

Issue #398, opened 2021-09-12
Constraints in for weak acids and pH equilibrium


gagnonaa10:
Transferred from the [Julia Discourse](https://discourse.julialang.org/t/constraints-in-catalyst-jl-for-weak-acids-and-ph-equilibrium/68037)

Hello Everyone,

I was hoping for some guidance on how to incorporate constraints for pH equilibriums constants into a disinfection model I am building in Julia.

A simplified version of the reaction network is here:

```
using OrdinaryDiffEq
using Catalyst
using DifferentialEquations
using Plots
using Latexify

disinfectionmodel = @reaction_network begin
    k1, HOCl + NH3 --> NH2Cl
    k2, NH2Cl --> HOCl + NH3
    k3, HOCl + NH2Cl --> NHCl2
    k4, NHCl2 --> HOCl + NH2Cl
    k5, HOCl + H + Br --> BrCl
    (kf, kf/10^-14), H + OH ↔ H2O
    (kf, kf/10^-9.25), NH3 + H ↔ NH4
    (kf, kf/10^-7.53), OCl + H ↔ HOCl
end 
```

What I am having trouble with now is the constraints associated with the pH dependent equilibrium relationships of weak acids and of pH itself. The full reaction system includes multiple reactions in which H or OH are the products or the reactant. I have done a bit of sleuthing and found Issue #137 which discussed adding constraints.

However, when I tried to utilize the @add_constraints macro mentioned in the [Issue #137](https://github.com/SciML/Catalyst.jl/issues/137) above as shown below, I receive the following an error documented below:

```
@add_constraints disinfectionmodel begin
    log10(H) + log10(OH) = -14
    log10((NH3)*(H)/(NH4)) = -9.25
    log10((OCl)*(H)/(HOCl)) = - 7.53
            end

```

```
LoadError: UndefVarError: @add_constraints not defined
in expression starting at In[4]:1
```

I looked through the Catalyst resource material and was unsuccessful in finding anything that would guide me further but will acknowledge that I am still a bit of a novice when it comes to Julia and programming in general along with interpreting source material. Being able to add these constraints prior to utilizing the differential equation solver will be really helpful!

Thank you for any help anyone can provide!

isaacsas:
[...]

PR #404, opened 2021-09-14
Add subsystem support to enable constraints and composition of ODE/Nonlinear sys

isaacsas:
Enables https://github.com/SciML/Catalyst.jl/issues/398 at the system level.

Note, this should simply get subsystems working properly. Constraints can just be a nonlinear or ODE subsystem, using the new `ParentScope` functionality to escape up a level. Or users can add explicit coupling equations. 

We could next update the DSL to allow `@add_constraints` on an existing system. 

To summarize:
1. Composition of `ReactionSystem`s should now work.
2. Conversion of composed systems to `ODESystem`s and `NonlinearSystem`s should work.
3. `ReactionSystem`s can have `NonlinearSystem` or `ODESystem` subsystems and should still be convertible to each of those types respectively. Note having `NonlinearSystem`s as subsystems means you can only convert to `NonlinearSystem` (with a similar constraint for `ODESystem`s). If MT adds more conversions between systems in the future we can try to support them. I think right now MT only supports conversion from nonlinear to ODE via `convert_system`. Having a separate call here makes sense due to the need for potentially adding a time variable, but it would be nice to also have `convert(::NonlinearSystem, ODESystem)` which shouldn't need any additional parameters.

One thing I noticed is our `NonlinearSystem`s keep the `t` dependence (i.e. `A(t)` vs `A` as a variable), but this doesn't seem to cause any problems as far as I can tell.


isaacsas:
@YingboMa do you see any issues with how I handled name spacing for reaction systems here or how I changed `equations`?

HISTORY.md
```diff
@@ -1,13 +1,21 @@
 # Breaking updates and feature summaries across releases
 
 ## Catalyst unreleased (master branch) 
-- **BREAKING:** added a dependency on Nemo.jl for integer matrix linear algebra methods.
-- Added `incidencematgraph`, `linkageclasses`, `deficiency`, `subnetworks` and `linkagedeficiency` API functions.
+- **BREAKING:** added a dependency on Nemo.jl for integer matrix linear algebra
```
ChrisRackauckas, line 4: This is GPL and adds a huge dependency tree IIRC?

ChrisRackauckas:
>One thing I noticed is our NonlinearSystems keep the t dependence (i.e. A(t) vs A as a variable), but this doesn't seem to cause any problems as far as I can tell.

That should be fine. Is there a reason `promote_subsystem` wasn't used? I thought @YingboMa created a whole subsystem promotion system.

HISTORY.md
isaacsas: Should we drop it then? This says LGPL 2.1, but it does seem Nemo has a big mix of stuff so it could be hard to know if one is going into GPL code: https://github.com/Nemocas/Nemo.jl/blob/master/LICENSE.md

Unfortunately AbstractAlgebra doesn’t have exactly the same functionality.

isaacsas: In terms of the dependency tree I found it was only twice as slow as AbstractAlgebra at loading (I.e. calling `using`), and that wasn’t anywhere near how long MT takes so not really noticeable.

isaacsas:
> > One thing I noticed is our NonlinearSystems keep the t dependence (i.e. A(t) vs A as a variable), but this doesn't seem to cause any problems as far as I can tell.
> 
> That should be fine. Is there a reason `promote_subsystem` wasn't used? I thought @YingboMa created a whole subsystem promotion system.

Do you mean `convert_system`? It looked like it was still WIP since the only conversion currently is NonlinearSystem to ODESystem from what I saw.

HISTORY.md
isaacsas: Actually, even AbstractAlgebra’s license statement is confusing, saying it may link to GPL only packages but their source is BSD-2.

Not sure what we should do here. I guess we could remove all the network analysis functionality which relies on integer LA and put it in a separate CatalystNetworkAnalysis package?

isaacsas:
@ChrisRackauckas I switched back to AbstractAlgebra. LMK if you feel their license statement is ok (it seems to indicate the only GPL components would be from Julia itself?).

ChrisRackauckas:
Yes, AbstractAlgebra.jl is fine and might already be in the indirect deps.

PR #407, opened 2021-09-16
Update API and docs for subsystems

isaacsas:
Follow up to https://github.com/SciML/Catalyst.jl/pull/404 to make as many API functions work with subsystems, and give an error message for those that don't support a `ReactionSystem` with subsystems. 

Main changes:
1. `species`, `reactionparams`, and `reactions` now collect the states, parameters, and `Reaction`s from only those subsystems that are `ReactionSystem`s. This leaves the normal MT API (`states`, `parameters`, `equations`) to return larger sets that may include components from non-`ReactionSystem` subsystems. This seemed like a reasonable division to me, ~~but we'll need further doc updates to make the distinction clear~~ .
2. Depreciated with a warning `params` and `numparm`s since we now have either `parameters` or `reactionparams` depending on what a user wants.
3. Depreciated `merge` since we now have `MT.extend`.


isaacsas:
This should be good to go. Next would be adding `@add_constraints`.

ChrisRackauckas (approved):
lgtm. `isequal_without_names` is an odd name but it makes sense.

isaacsas:
I’ll merge but change it if I come up with something better…

PR #416, opened 2021-09-21
update MTK and Symbolics compat

anandijain:


anandijain:
actually is this not needed? it looks like Symbolics = 3.0 still allows for 3.4, likewise with MTK on 6.3 -> 6.5

ChrisRackauckas:
These are lower bounds

PR #417, opened 2021-09-21
Chemical Master Equation lowering

anandijain:
ported from my MTK PR

I'd like to know the best way to integrate this into the lowering code that exists.
Should I add a kwarg in the `convert(<:ODESystem, rn)` code that defaults to the current ode rate law?

### DIFF
--- a/src/cme.jl
+++ b/src/cme.jl
@@ -57,10 +57,9 @@ function ODESystem_cme(rs, u0; truncation=nothing, kwargs...)
 	rxns = reactions(rs)
 	M = numreactions(rs)
 	jump_rates = jumpratelaw.(rxns)
-    nsm = netstoichmat(rs)
-    species = states(rs)
+	nsm = netstoichmat(rs)
+	species = states(rs)
 	
-	# eqs = Equation[]
 	eqs = Vector{Equation}(undef, N)
 	for (i, x) in enumerate(state_space)
 		eq = Num(0)
@@ -71,7 +70,6 @@ function ODESystem_cme(rs, u0; truncation=nothing, kwargs...)
 			t2 = substitute(jump_rates[j], Dict(species .=> x)) * u[d[x]]
 			eq += t1 - t2
 		end 
-		# push!(eqs, D(u[d[x]]) ~ eq)
 		eqs[i] = D(u[d[x]]) ~ eq
 	end
 	ODESystem(eqs, t, u, parameters(rs); name=nameof(rs), kwargs...)
--- a/src/cme.jl
+++ b/src/cme.jl
@@ -1,0 +1,78 @@
+
+"""
+Performs a breadth-first graph traversal of a  to enumerate all possible states.
+
+The truncation kwarg should be given as a vector of Int, denoting the maximum species' counts.
+If, upon taking traversing a reaction the amount of any species is greater than the value in the truncation vector, 
+the graph traversal will not add the state to the list of possible states.
+"""
+function traverse_reactionsystem(rs, u0; truncation=nothing)
+    num_molecules = sum(u0)
+    num_species = length(u0)
+    state_space = []
+    Q = Queue{typeof(u0)}()
+    enqueue!(Q, u0)
+    push!(state_space, u0)
+    nsm = netstoichmat(rs)
+    edges = eachcol(nsm)
+    while !isempty(Q)
+        w = dequeue!(Q)
+        for e in edges
+            x = w .+ e
+            if x ∉ state_space && all(y >= 0 for y in x)
+                if truncation === nothing 
+                    if sum(x) > num_molecules 
+                        error("need to provide truncation for infinite reaction system")
+                    else 
+                        push!(state_space, x)
+                        enqueue!(Q, x)
+                    end
+                else 
+                    if all(x[i] <= truncation[i] for i in 1:num_species)
+                        push!(state_space, x)
+                        enqueue!(Q, x)
+                    end
+                end
+            end
+        end
+    end
+    return state_space
+end
+
+function ODESystem_cme(rs, u0; truncation=nothing, kwargs...)
+	state_space = traverse_reactionsystem(rs, u0; truncation=truncation)
+
+	N = length(state_space)
+	d = Dict(state_space .=> 1:N)
+
+    # initializes the state corresponding to u0 to have probability 1
+	defs = zeros(N)
+	start_state = d[u0]
+	defs[start_state] = 1.
+
+
+	@variables t u[1:N](t) = defs
+	D = Differential(t)
+
+	rxns = reactions(rs)
+	M = numreactions(rs)
+	jump_rates = jumpratelaw.(rxns)
+    nsm = netstoichmat(rs)
+    species = states(rs)
+	
+	# eqs = Equation[]
+	eqs = Vector{Equation}(undef, N)
+	for (i, x) in enumerate(state_space)
+		eq = Num(0)
+		for j in 1:M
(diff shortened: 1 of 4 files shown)


src/cme.jl
```diff
@@ -0,0 +1,76 @@
...
+	N = length(state_space)
+	d = Dict(state_space .=> 1:N)
+
+    # initializes the state corresponding to u0 to have probability 1
+	defs = zeros(N)
+	start_state = d[u0]
+	defs[start_state] = 1.
+
+
+	@variables t u[1:N](t) = defs
```
ChrisRackauckas, line 54: should these states be named in a clever way to make them easy to interpret?

anandijain: i was thinking maybe P_i or something, meaning probability of being in state i.
theres probably a better way than that though

anandijain: i guess the real problem is that you don't know what state the variable actually corresponds to. that should definitely be resolved

isaacsas:
@anandijain I'll try to review carefully tomorrow. A few high level comments for now:

1. I'd suggest something like `truncated_cme` as the name since this isn't actually providing a representation for the (possibly infinite) full CME in general (and there are other approximations one might want to use). 
2. Have you looked at https://github.com/kaandocal/FiniteStateProjection.jl? Is there anything from there that would be helpful to use / can be reused to avoid redundant code? (I don't think we should add a dependency on it, but perhaps @kaandocal would have some ideas on components that make sense for handling CMEs in Catalyst.)
3. Could you add some tests where you actually write out the truncated CME ODEs for one or more simple systems, hitting the major reaction types (zero-order, first-order, and second-order), and check they match what is generated from your code (or check that the corresponding ODE rhs functions have the same value for a given `u`). We want to make sure the generated equations are correct.

Thanks!

src/cme.jl
```diff
@@ -0,0 +1,76 @@
+
+"""
+Performs a breadth-first graph traversal of a  to enumerate all possible states.
+
+The truncation kwarg should be given as a vector of Int, denoting the maximum species' counts.
+If, upon taking traversing a reaction the amount of any species is greater than the value in the truncation vector, 
+the graph traversal will not add the state to the list of possible states.
+"""
+function traverse_reactionsystem(rs, u0; truncation=nothing)
```
isaacsas, line 9: Maybe something like `reachable_states` would be a better name?

src/cme.jl
```diff
@@ -0,0 +1,76 @@
...
+                        enqueue!(Q, x)
+                    end
+                end
+            end
+        end
+    end
+    return state_space
+end
+
+function ODESystem_cme(rs, u0; truncation=nothing, kwargs...)
```
isaacsas, line 42: If this approach is only going to generate the full CME for a specific, deterministic initial condition, that needs to be made clear in the documentation. I think in general a user would expect to get truncated equations which could be used for any initial condition (which is normally given by a vector of probabilities and does not require a deterministic number of each molecule initially).

src/cme.jl
```diff
@@ -0,0 +1,76 @@
...
+    push!(state_space, u0)
+    nsm = netstoichmat(rs)
+    edges = eachcol(nsm)
+    while !isempty(Q)
+        w = dequeue!(Q)
+        for e in edges
+            x = w .+ e
+            if x ∉ state_space && all(y >= 0 for y in x)
+                if truncation === nothing 
+                    if sum(x) > num_molecules
```
isaacsas, line 24: Do we want the threshold this tight? What if we have the simple system `A <--> 2B`? If `A --> 2B` is ordered first, wouldn't we immediately violate this condition?

isaacsas: Hmm, if we know a network is conservative can we give a rigorous upper bound on the amount of molecules we'll ever see? I don't know the theory here, but if there are any results we could add a method to check for conservativeness and use the related bound.

src/cme.jl
isaacsas: Return `d` too?

src/cme.jl
```diff
@@ -0,0 +1,76 @@
...
+
+
+	@variables t u[1:N](t) = defs
+	D = Differential(t)
+
+	rxns = reactions(rs)
+	M = numreactions(rs)
+	jump_rates = jumpratelaw.(rxns)
+	nsm = netstoichmat(rs)
+	species = states(rs)
```
isaacsas, line 61: Better to use `species` as `states` and `species` may be different now... (Though in the practical case where they differ this wouldn't work anyways.)

src/cme.jl
```diff
@@ -0,0 +1,76 @@
...
+	
+	eqs = Vector{Equation}(undef, N)
+	for (i, x) in enumerate(state_space)
+		eq = Num(0)
+		for j in 1:M
+			x′ = x .- @view(nsm[:, j]) # traverse a reaction and get the new counts
+			p_x′ = haskey(d, x′) ? u[d[x′]] : 0 # if the new state is not in `state_space`, set probability to 0
```
isaacsas, line 68:
```suggestion
                       t1 = haskey(d,x') ? substitute(jump_rates[j], Dict(species .=> x′)) * u[d[x']] : 0			
```

src/cme.jl
```diff
@@ -0,0 +1,76 @@
...
+			t1 = substitute(jump_rates[j], Dict(species .=> x′)) * p_x′
+			t2 = substitute(jump_rates[j], Dict(species .=> x)) * u[d[x]]
```
isaacsas, line 70: Don't you need to check if `x .+ @view(nsm[:, j])` is an allowable state to include this term?

isaacsas (changes requested):

kaandocal:
This seems similar to what I've been trying to do in [FiniteStateProjection.jl](https://github.com/kaandocal/FiniteStateProjection.jl), namely converting a reaction network into a (truncated) version of the CME. I did not opt to use `ODESystem`s for this. mainly because I sometimes deal with very state spaces (e.g. 10k states) and thought this might be inefficient, but I never did any benchmarks. The package can convert RNs into Julia functions computing the CME directly using loops, or into matrices that represent the right-hand side of the CME `dP/dt = A * P`. I think it would be cool to incorporate this as well, ie. add support for conversion to `ODESystem`s along the lines of this PR :)

isaacsas:
> This seems similar to what I've been trying to do in [FiniteStateProjection.jl](https://github.com/kaandocal/FiniteStateProjection.jl), namely converting a reaction network into a (truncated) version of the CME. I did not opt to use `ODESystem`s for this, mainly because I sometimes deal with very state spaces (e.g. 10k states) and thought this might be inefficient, but I never did any benchmarks. The package can convert RNs into Julia functions computing the CME directly using loops, or into matrices that represent the right-hand side of the CME `dP/dt = A * P`. I think it would be cool to incorporate this as well, ie. add support for conversion to `ODESystem`s along the lines of this PR :)

My guess is that with 10K states using loops or matrix-vector form will work better as we may run up against compiler limitations for the size of the MT generated rhs function. This is more what I had in mind; representing the CME as a (possibly infinite) differential-difference equation in many discrete variables. i.e. it is symbolically represented by one equation for an abstract differential-difference equation in the species population variables, and then we can custom generate appropriate functions from that depending on the desired approximation. It would be cool if there was a way to get the symbolic code generated into loops. I guess we need a version of `build_function` that is designed for difference-type equations/operators. @ChrisRackauckas is there something along those lines used by the PDE conversion tooling?

src/cme.jl
isaacsas: Just return `d` too?

isaacsas:
This is too far out of date to work at this point I think, but having a CME system representation we could convert to and apply transformations too would be awesome at some point.

PR #423, opened 2021-09-27
Constraints, flattening, and extending for ReactionSystems

isaacsas:
Enables `extend`ing `ReactionSystem`s with algebraic and ODE constraints. While we currently can have sub-systems of algebraic (or even ODE) constraints, this is annoying for specifying algebraic equations to couple systems (since everything needs to get parent scoped). i.e. imagine wanting to hook three systems representing genes together, conceptually this is not necessarily a hierarchical system, but more a collection of parallel systems with coupling equations one level up. (This wouldn't necessarily be an issue if we could make the top-level system a `NonlinearSystem`, but currently MT requires sub-systems to have the same type, so top-level systems must be `ReactionSystem`s.)

This PR allows attaching to a `ReactionSystem` an internal constraint system, which is considered at the same scope as the `ReactionSystem`. In this way we can logically `extend` `ReactionSystem`s with equations and/or other `ReactionSystem`s (we already support `compose`). For a simple `alias_elimination` of basic `A ~ B` type equations we could then `flatten` the full hierarchy of systems into one `ReactionSystem` and one internal `sys.constraint` system, and then apply the elimination.

~~Still TODO:~~
- [x] `get_constraints(network)` and support a parallel constraint subsystem
- [x] `flatten`
- [x] `extend`
- [x] tests
- [x] Allow / use `ODESystem`s as constraints, and error if then try to convert to NonlinearSystem.
- [x] Allow interpolation of variables storing network names in the DSL.

### DIFF
--- a/src/reactionsystem.jl
+++ b/src/reactionsystem.jl
@@ -232,23 +232,43 @@ function ReactionSystem(iv; kwargs...)
     ReactionSystem(Reaction[], iv, [], []; kwargs...)
 end
 
-for prop in (:constraint_eqs, :constraint_states, :constraint_ps)
-    fname1 = Symbol(:get_, prop)
-    fname2 = Symbol(:has_, prop)
-    @eval begin
-        $fname1(sys::ReactionSystem) = getfield(sys, $(QuoteNode(prop)))
-        $fname2(sys::ReactionSystem) = isdefined(sys, $(QuoteNode(prop)))
-    end
+
+####################### ModelingToolkit inherited accessors #############################
+
+get_constraint_eqs(sys::ReactionSystem)    = getfield(sys, :constraint_eqs)
+get_constraint_states(sys::ReactionSystem) = getfield(sys, :constraint_states)
+get_constraint_ps(sys::ReactionSystem)     = getfield(sys, :constraint_ps)
+has_constraint_eqs(sys::ReactionSystem)    = isdefined(sys, :constraint_eqs)
+has_constraint_states(sys::ReactionSystem) = isdefined(sys, :constraint_states)
+has_constraint_ps(sys::ReactionSystem)     = isdefined(sys, :constraint_ps)
+
+function ModelingToolkit.states(sys::ReactionSystem)
+    csts    = get_constraint_states(sys)
+    sts     = isempty(csts) ? get_states(sys) : [get_states(sys); get_constraint_states(sys)]
+    systems = get_systems(sys)
+    unique(isempty(systems) ? sts : [sts; reduce(vcat,namespace_variables.(systems))])
+end
+
+function ModelingToolkit.parameters(sys::ReactionSystem)
+    cps     = get_constraint_ps(sys)
+    ps      = isempty(cps) ? get_ps(sys) : [get_ps(sys); get_constraint_ps(sys)]
+    systems = get_systems(sys)
+    unique(isempty(systems) ? ps : [ps; reduce(vcat,namespace_parameters.(systems))])
 end
 
 function ModelingToolkit.equations(sys::ReactionSystem)
-    eqs = get_eqs(sys)
+    eqs     = get_eqs(sys)
+    ceqs    = get_constraint_eqs(sys)
     systems = get_systems(sys)
-    if !isempty(systems)
-        eqs = Any[eqs; reduce(vcat, MT.namespace_equations.(systems); init=Any[])]
+    if isempty(systems)
+        alleqs = isempty(ceqs) ? eqs : Any[eqs; ceqs]
+    else
+        alleqs = Any[eqs; ceqs; reduce(vcat, MT.namespace_equations.(systems); init=Any[])]
     end
-    eqs
+    alleqs
 end
+
+######################## Conversion to ODEs/SDEs/jump, etc ##############################
 
 """
     oderatelaw(rx; combinatoric_ratelaw=true)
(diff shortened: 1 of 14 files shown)


src/reactionsystem.jl
```diff
@@ -642,4 +699,117 @@ function ModelingToolkit.modified_states!(mstates, rx::Reaction, sts::Set)
...
+    csys = get_constraints(rs)
+    csys2 = get_constraints(sys)
+    if csys === nothing        
+        newcsys = csys2
+    else
+        newcsys = (csys2 === nothing) ? csys : extend(csys2, csys, name)
+    end
+    
+    ReactionSystem(eqs, get_iv(rs), sts, ps; observed = obs, name = name, 
+                    systems = syss, defaults = defs, checks=false, constraints=csys)
```
isaacsas, line 814:
```suggestion
                    systems = syss, defaults = defs, checks=false, constraints=newcsys)

```

isaacsas:
Some simple examples
```julia
rn = @reaction_network rn begin
    (k1,k2), A <--> B
    end k1 k2
@parameters a,b
@variables t,A(t),C(t)
D = Differential(t)
eqs = [D(C) ~ -b*C + a*A]
@named osys = ODESystem(eqs,t,[A,C],[a,b])
rn2 = extend(osys,rn)
rnodes = convert(ODESystem,rn2)

# if there are actually ODEs in the ODESystem we error
@test_throws ErrorException convert(NonlinearSystem,rn2)

# no ODEs below, so can represent as ODE or Nonlinear systems and convert to either
eqs = [0 ~ -a*A + C, 0 ~ -b*C + a*A]
@named nlsys = NonlinearSystem(eqs,[A,C],[a,b])
rn2 = extend(nlsys,rn)
rnodes = convert(ODESystem,rn2)
rnnlsys = convert(NonlinearSystem,rn2)

@named nlsys = ODESystem(eqs,t,[A,C],[a,b])
rn2 = extend(nlsys,rn)
rnodes = convert(ODESystem,rn2)
rnnlsys = convert(NonlinearSystem,rn2)
```

isaacsas:
Perhaps a better example, which I plan to put in a tutorial:
```julia
function repressed_gene(; name)
  @parameters t α₀ α K n δ β μ
  @variables m(t) P(t) R(t)
  rxs = [
         Reaction(α₀, nothing, [m]),
         Reaction(α / (1 + (R/K)^n), nothing, [m]),
         Reaction(δ, [m], nothing),
         Reaction(β, [m], [m,P]),
         Reaction(μ, [P], nothing)
        ]
  
  specs = [m,P,R]
  pars  = [α₀,α,K,n,δ,β,μ]
  ReactionSystem(rxs, t, specs, pars; name=name)
end
@named sys₁ = repressed_gene()
@named sys₂ = repressed_gene()
@named sys₃ = repressed_gene()
connections = [sys₁.R ~ sys₃.P,
               sys₂.R ~ sys₁.P,
               sys₃.R ~ sys₂.P]
@named csys = NonlinearSystem(connections, [sys₁.R,sys₃.P,sys₂.R,sys₁.P,sys₃.R,sys₂.P],[])
@named repressilator = ReactionSystem(t; constraints=csys, systems=[sys₁,sys₂,sys₃])

# or 
@named repressilator = ReactionSystem(t; systems=[sys₁,sys₂,sys₃])
@named repressilator = extend(csys, repressilator)
```

TorkelE:
The repressilator example is really elegant! Have been lagging a bit on developments in this direction, but looking forward to playing around with it.

isaacsas:
> The repressilator example is really elegant! Have been lagging a bit on developments in this direction, but looking forward to playing around with it.

Thanks! BTW, do you know if there is a way we could make the macro work in such a function? i.e. something like
```julia
function repressed_gene(; name)
    @reaction_network $name begin
    ...
    end
end
```
It would be cool if we could use the DSL to generate the core gene module in this way.

TorkelE:
I _think_ it should be possible, but uncertain how to get it working. E.g. I tried
```julia
name = :repressed_gene
rg =  @reaction_network :($(name)) begin
    (p,d), 0 <--> X
end p d
rg.name
```
but got an error
```
LoadError: AssertionError: @parameters expects a tuple of expressions or an expression of a tuple (`@parameters x y z(t) v[1:3] w[1:2,1:4]` or `@parameters x y z(t) v[1:3] w[1:2,1:4] k=1.0`)
in expression starting at /home/torkelloman/.julia/packages/Catalyst/RTdmC/src/reaction_network.jl:178

Stacktrace:
 [1] _parse_vars(macroname::Symbol, type::Type, x::Tuple{Symbol, Expr, Symbol, Symbol}, transform::Function)
   @ Symbolics ~/.julia/packages/Symbolics/sITWZ/src/variable.jl:134
 [2] var"@parameters"(__source__::LineNumberNode, __module__::Module, xs::Vararg{Any, N} where N)
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/Mo4gw/src/parameters.jl:30
 [3] eval
   @ ./boot.jl:360 [inlined]
 [4] include_string(mapexpr::typeof(REPL.softscope), mod::Module, code::String, filename::String)
   @ Base ./loading.jl:1094
```
there are some meta-programming stuff going on here which I don't fully remember how to deal with. It would probably be possible to make a function `remake(::ReactionSystem;kwargs...)` where one could set `remake(repressed_gene_base,name=my_name)`, but that doesn't really solve the initial problem.

isaacsas:
@TorkelE I think I just added that feature in. Could you take a look at the changes in `reaction_network.jl`? This seems to handle it and I think I haven't broken anything after trying the various permutations of with/without names that we allow.

TorkelE:
I _think_ that looks good. Admittedly I haven't done metaprogramming in a while, so a bit rusty when it comes to these things. I'll play around with things a bit when it is merged as well in case I were to find anything, but I'm not worried.

isaacsas:
Once this gets merged I'm planning to add a tutorial using all this stuff.

Issue #425, opened 2021-09-29
simple constructor `ReactionSystem(rxns)`


anandijain:
maybe there is not an analog but MTK has a really nice `ODESystem(eqs)` constructor that finds all the parameters, states, etc.

I think it'd be nice to have this for ReactionSystem.

rough sketch
```julia 
species(rxn::Reaction) = unique(reduce(vcat, (rxn.substrates, rxn.products)))
species(rxns::Vector{<:Reaction}) = unique(reduce(vcat, map(rxnspecies, rxns)))

ReactionSystem(rxn::Reaction; kwargs...) = ReactionSystem(Reaction[rxn]; kwargs...)

function ReactionSystem(rxns::Vector{<:Reaction}; kwargs...)
    ReactionSystem(rxns, Catalyst.DEFAULT_IV, species(rxns), []; kwargs...)
end
```

but I didn't know how to find all the parameters for a reaction. I think we disallow symbolic params in stoich coefficients, right? 

That means I'd just need to find them in the `rate` part, I think.

This isn't urgent but I bumped up against it so I wanted to document this

isaacsas:
Hmm, parameters now have meta-data tagging them right? So I guess we could walk all reaction rate expressions and identify if each term is a parameter, and if not declare it a species? (i.e. we need to look at the `rx.rate`s too, not just the `rx.substrates` and `rx.products` -- this is actually an issue with the DSL that needs to get fixed too.) 

For now the stoichiometry only allows species, but I'd like to get more general stoichiometry in (the big hangup is the issues I mentioned in https://github.com/SciML/ModelingToolkit.jl/issues/1246).

isaacsas:
This would be a great add though; not having to list the parameters, `iv` and species would be a nice improvement!

isaacsas:
Once https://github.com/SciML/Catalyst.jl/pull/423 is merged I have a bit more to do and then I’ll make a new breaking release. Might be a week or so.

anandijain:
is it breaking? i don't think so, right

isaacsas:
This isn’t, but other unreleased changes are I think.

PR #426, opened 2021-09-29
Simplified ReactionSystem constructor

isaacsas:
Closes https://github.com/SciML/Catalyst.jl/issues/425.
CC @anandijain

### DIFF
--- a/src/reactionsystem.jl
+++ b/src/reactionsystem.jl
@@ -240,7 +240,6 @@ function ReactionSystem(rxs::Vector{<:Reaction}, iv; kwargs...)
     end
 
     ReactionSystem(rxs, t, collect(sts), collect(ps); skipvalue=true, kwargs...)
-
 end
 
 
--- a/src/reactionsystem.jl
+++ b/src/reactionsystem.jl
@@ -142,6 +142,8 @@ function get_netstoich(subs, prods, sstoich, pstoich)
     ns
 end
 
+arrtype(v::Symbolics.Arr{U}) where {U} = U
+
 """
 $(TYPEDEF)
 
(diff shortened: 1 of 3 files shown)


isaacsas:
@anandijain I did need to require passing the `iv` too as I didn't want to try to have to make a robust way to detect it...

anandijain:
awesome! thanks a bunch

out of curiosity, what's a practical case where `t` is not the `iv`?

isaacsas:
Don't know, but also don't want to assume... It's better to make users be explicit on what they used in constructing everything.

isaacsas:
I have to read up a bit on how symbolics arrays work, they're breaking my conversions from `Num`s...

src/reactionsystem.jl
```diff
@@ -207,17 +207,42 @@ function ReactionSystem(eqs, iv, species, ps;
...
+            if MT.isparameter(var) 
+                push!(ps, var)
+            else
+                push!(sts, var)
+            end
+        end
+        empty!(vars)
+    end
+
+    ReactionSystem(rxs, value(iv), collect(sts), collect(ps); skipvalue=true, kwargs...)
```
isaacsas, line 242:
```suggestion
    ReactionSystem(rxs, t, collect(sts), collect(ps); skipvalue=true, kwargs...)

```

src/reactionsystem.jl
```diff
@@ -207,17 +207,43 @@ function ReactionSystem(eqs, iv, species, ps;
...
+                push!(ps, var)
+            else
+                push!(sts, var)
+            end
+        end
+        empty!(vars)
+    end
+
+    ReactionSystem(rxs, t, collect(sts), collect(ps); skipvalue=true, kwargs...)
+
```
isaacsas, line 243:
```suggestion
```

PR #430, opened 2021-10-08
add smap as a karg for convert to ODESystem

LauraBMo:
Allow picking species' order when generating an ODESystem from a ReactionSystem.
Now calls like
```
   convert(ODESystem, rn; smap=my_smap)
```
work. I needed it for my research, mainly for exporting capabilities (I could elaborate on that).

If you like the idea and the implementation, I could implement the remaining `convert`'s.

An MWE to see the difference:
```
using Catalyst

rn = @reaction_network begin
    k1,  X2 → X1
end k1

odefun = ODEFunction(convert(ODESystem, rn))

odefun([2,0], [1], 0) # gives [-2, 2]

my_smap = Dict(species(rn) .=> [2,1])
odefun = ODEFunction(convert(ODESystem, rn; smap=my_smap))

odefun([2,0], [1], 0) # gives [0,0] 
```

### DIFF
--- a/src/networkapi.jl
+++ b/src/networkapi.jl
@@ -913,6 +913,9 @@ end
     reorder_states!(rn, neworder)
 
 Given a [`ReactionSystem`](@ref) and a vector `neworder`, orders the states of `rn` accordingly to `neworder`.
+
+Notes:
+- Currently only supports `ReactionSystem`s without constraints or subsystems.
 """
 function reorder_states!(rn, neworder) 
    (get_constraints(rn) === nothing) && isempty(get_systems(rn)) || 


isaacsas:
Hi @LauraBMo, thanks for the PR! Can you explain a bit about why you want to do this?

This certainly seems like it could be useful. The main question is where it should be housed (i.e. maybe it makes more sense to support general reordering of equations and states at a lower level like `build_function`). @ChrisRackauckas any thoughts? 

@LauraBMo, if after we discuss we decide to proceed could you rebase/merge master into your PR so it is up to date with master? That will make it easier to review as right now it is out of sync with recent changes.

isaacsas:
@LauraBMo are you all done now? If so I'll merge once tests finish.

Thanks!

PR #432, opened 2021-10-09
allow interpolation of species and parameters in the DSL

isaacsas:
This should allow interpolation of species variables and parameters into the DSL, which is useful for compositional modeling. i.e.
```julia
function repressedgene(; t=(@variables t)[1], R=(@variables R(t))[1],  P=(@variables P(t))[1], name)
     @reaction_network $name begin
          (k1,k2), $R + G <--> RG
          k3, G --> $P
     end k1 k2 k3
end
```
Now we can compose several genes together fairly easily (without algebraic equations!)
```julia
@variables t, G2₊P(t)
@named G1 = repressedgene(; R=ParentScope(G2₊P))
@named G2 = repressedgene(; R=ParentScope(G1.P))
@named rn = ReactionSystem(t; systems=[G1,G2])
```
giving
```julia
Model rn with 6 equations
States (6):
  G2₊P(t)
  G1₊P(t)
  G1₊G(t)
  G1₊RG(t)
  G2₊G(t)
  G2₊RG(t)
Parameters (6):
  G1₊k2
  G1₊k1
  G1₊k3
  G2₊k2
  G2₊k1
  G2₊k3

julia> equations(rn)
6-element Vector{Any}:
 G1₊k1, G2₊P + G1₊G --> G1₊RG
 G1₊k2, G1₊RG --> G2₊P + G1₊G
 G1₊k3, G1₊G --> G1₊P
 G2₊k1, G1₊P + G2₊G --> G2₊RG
 G2₊k2, G2₊RG --> G1₊P + G2₊G
 G2₊k3, G2₊G --> G2₊P
```

~I could also look into adding this for parameters if we think that might be useful.~ Should work for parameters too now.

Also fixes a bug in `show` for `Reaction`s.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -36,7 +36,7 @@
   `ODESystem` within the `ReactionSystem`, and accessible via
   `get_constraints(reactionnetwork)`.
 - Added `Catalyst.flatten(rn)` to allow flattening of a `ReactionSystem` with
-  sub-systems into one `ReactionSystem`. Non-`ReactionSystem` subsystems a
+  sub-systems into one `ReactionSystem`. Non-`ReactionSystem` subsystems are
   merged into the constraints of the flattened `ReactionSystem`, and accessible
   via `get_constraints`.
 - **BREAKING:** `ReactionSystem`s are now always flattened when calling


isaacsas:
OK, this should finally be good to go. We need to address https://github.com/SciML/Catalyst.jl/issues/433 at some point to make such internal changes easier to handle...

PR #452, opened 2021-12-03
Fix parameter ordering

TorkelE:
Currently, if you declare a reaction network using the DSL:
```julia
rn = @reaction_network begin
       (p,d), 0 <--> X
end d p
```
the order of the parameters in `rn.ps` is according to the order which they appeared in the reactions, not the order they had at the end (so in this case, the order is `[p,d]`, not `[d,p]`. This means that if one does:
```julia
X₀ = 1.0
u₀_vals = [X₀]
d = 0.1; p = 1.0; 
p_vals = [d,p]
u₀map = species(rn) .=> u₀_vals
pmap  = parameters(rn) .=> p_vals
tspan = (0., 10000.)
oprob = ODEProblem(rn, u₀map, tspan, pmap)
```
one gets the wrong order.

In many cases, the order will probably be the same, but it seems dangerous to risk this when this is not the case. In this simple example, it is easy to manually check the order, but for larger networks, it is not very ideal.

I went through how the macros do it, and now we just take all the species and parameters and add them to the initial `ReactionSystem` call (I believe they were previously auto-extracted from the reactions?). I haven't worked with the macro code over the last few MTK updates, so probably good if someone has a 2nd look, but it's only 2 added lines, so not very much changed.

The only thing which doesn't work is this test:
```julia
@parameters k, k1, k2
@variables t, A(t), B(t), C(t), D(t)
AA = A
AAA = A^2 + B
rn = @reaction_network rn begin
    k*$AAA, C --> D    
end k
rn2 = ReactionSystem([Reaction(k*AAA, [C], [D])], t, [A,B,C,D],[k]; name=:rn)
#@test rn == rn2
```
where the species `A` and `B` aren't detected. Just writing them in directly, e.g.
```julia
rn = @reaction_network rn begin
    k*(A^2 + B), C --> D    
end k
```
is fine, there is something metaprogramming wise I don't really understand here.


TorkelE:
Actually, the best way to fix the issue with
```julia
@parameters k, k1, k2
@variables t, A(t), B(t), C(t), D(t)
AA = A
AAA = A^2 + B
rn = @reaction_network rn begin
    k*$AAA, C --> D    
end k
rn2 = ReactionSystem([Reaction(k*AAA, [C], [D])], t, [A,B,C,D],[k]; name=:rn)
#@test rn == rn2
```
might be to simply add another methods,
```julia
function ReactionSystem(rxs::Vector{<:Reaction}, iv,states::Nothing,ps; kwargs...)
```
which takes the parameters as the p's provided, but extracts the species from the reactions.

TorkelE:
The error seems to be
```
ReactionSystem Components Test: Error During Test at /home/runner/.julia/packages/SafeTestsets/A83XK/src/SafeTestsets.jl:25
  Got exception outside of a @test
  LoadError: MethodError: no method matching extend(::ModelingToolkit.ODESystem, ::Nothing)
  Closest candidates are:
    extend(::Union{ModelingToolkit.NonlinearSystem, ModelingToolkit.ODESystem}, ::Catalyst.ReactionSystem; name) at ~/work/Catalyst.jl/Catalyst.jl/src/reactionsystem.jl:879
    extend(::ModelingToolkit.AbstractSystem, ::ModelingToolkit.AbstractSystem; name) at ~/.julia/packages/ModelingToolkit/r5ZaU/src/systems/abstractsystem.jl:998
  Stacktrace:
    [1] top-level scope
      @ ~/work/Catalyst.jl/Catalyst.jl/test/reactionsystem_components.jl:220
    [2] include(mod::Module, _path::String)
      @ Base ./Base.jl:418
    [3] include(x::String)
      @ Main.##275 ~/.julia/packages/SafeTestsets/A83XK/src/SafeTestsets.jl:23
    [4] macro expansion
      @ ~/work/Catalyst.jl/Catalyst.jl/test/runtests.jl:9 [inlined]
    [5] macro expansion
      @ /opt/hostedtoolcache/julia/1.7.0/x64/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
    [6] top-level scope
      @ ~/work/Catalyst.jl/Catalyst.jl/test/runtests.jl:9
    [7] eval(m::Module, e::Any)
      @ Core ./boot.jl:373
    [8] macro expansion
      @ ~/.julia/packages/SafeTestsets/A83XK/src/SafeTestsets.jl:23 [inlined]
    [9] macro expansion
      @ timing.jl:220 [inlined]
   [10] macro expansion
      @ ~/work/Catalyst.jl/Catalyst.jl/test/runtests.jl:9 [inlined]
   [11] top-level scope
      @ timing.jl:220
   [12] include(fname::String)
      @ Base.MainInclude ./client.jl:451
   [13] top-level scope
      @ none:6
   [14] eval
      @ ./boot.jl:373 [inlined]
   [15] exec_options(opts::Base.JLOptions)
      @ Base ./client.jl:268
   [16] _start()
      @ Base ./client.jl:495
  in expression starting at /home/runner/work/Catalyst.jl/Catalyst.jl/test/reactionsystem_components.jl:220
```

It is caused by the last line here:
```julia
rn = @reaction_network rn begin
    (k1,k2), A <--> B
    end k1 k2
@parameters a,b
@variables t,A(t),C(t)
D = Differential(t)
eqs = [D(A) ~ -a*A + C, D(C) ~ -b*C + a*A]
@named osys = ODESystem(eqs,t,[A,C],[a,b])
rn2 = extend(osys,rn)
```

I am uncertain, because I cannot reproduce that error on my local machine, using the Catalyst modification (or without it).

isaacsas:
Does this break something like:
```julia
@parameters k
@variables t, A(t), B(t), C(t), D(t)
AAA = k*A^2 + B
rn = @reaction_network rn begin
           $AAA, C --> D    
       end 
```
i.e. will it now require one to include `k` after `end`, thereby preventing the discovery of parameters from interpolated expressions?

My understanding, and @ChrisRackauckas can hopefully clarify this, is that in MT we really don't want users to assume specific orderings of parameters or species. 

The problem you are running into, which I think has several open MT issues, is there are issues with building parameter and initial condition maps, accessing solution components, and ploting solution components, when one only has access to a generated system (i.e. one did not manually create the symbolic parameters and variables in their current scope). For small systems or a few symbols one can always use
```
@unpack k = rn.k
```
or 
```
@nonamespace rn.k
```
However if one wants to bring all parameters and/or species into the current workspace there is no functionality to do so, and this makes the construction of parameter / initial condition mappings more tedious.

So this really seems like an issue with making ModelingToolkit either
1. Consistently work with `sys.k` for mappings, solution indexing, and plotting.
2. Provide a means to load all parameter syms and/or state syms into the current workspace (maybe something like `@unpack_states` and `@unpack_ps`?).
3. Allow parameter / state mappings that use regular Julia symbols, i.e. `u0 = [:A => 1.0, :B => 2.0,...]`.

@ChrisRackauckas any thoughts on how to handle this issue, or should we just provide a guaranteed ordering of unknowns via the DSL as we did in the past?

TorkelE:
So in your example, because `k` is declared as a parameter previously, the macro auto-detect that it should be a parameter in the macro? I didn't realise that we had a way to distinguish parameters from species, except for which were mentioned at the end, but that is kind of clever.

I think it is a good idea, generally, to have a system where the order is not important, and I don't think we should do anything where we lose any functionality. However, especially for a smaller, static, model, I think this is quite useful when just quickly checking out some circuits. Right now it is especially messy since we don't have a good solution to how to actually input the parameters.

I think your alternative (3) would be the easiest one, but not sure if it would cause other issues. Ideally, I think having (2) would be good, I'm not really the expert here, but having (2) as the default options seem good. If there were some way to have the initial order being as given, I think that would be useful to a certain group of modellers though (myself included), especially until we figure out how to do things. It shouldn't be that hard to initialise the parameters as those provided, but still add any extra ones encountered? I will have a look at it.

TorkelE:
I remade the change so that it should handle 
```julia
@parameters k
@variables t, A(t), B(t), C(t), D(t)
AAA = k*A^2 + B
rn = @reaction_network rn begin
    $AAA, C --> D    
end
```
types of networks as well. It should now be identical to the previous one, but if any parameters are given at the end, that order is used (and any possible other parameters are appended at the end). The changes are rather small:

Instead of starting with
```julia
rxexprs = :(ReactionSystem([],t,; name=$(name)))    
```
we now have
```julia
rxexprs = :(make_ReactionSystem_internal([],t,nothing,[]; name=$(name)))    
foreach(parameter -> push!(rxexprs.args[6].args,parameter), parameters)
```
essentially just providing the information of the parameter list. `make_ReactionSystem_internal` is not exactly an elegant name, and there should be neater ways of doing it with instead calling `ReactionSystem` and multiple dispatch, however, at this stage I just wanted to have something that 100% would cause any other changes.

Next, 
```julia
make_ReactionSystem_internal(rxs::Vector{<:Reaction}, iv, no_sps::Nothing, ps_in; kwargs...)
```
which we uses now is basically
```julia
ReactionSystem(rxs::Vector{<:Reaction}, iv; kwargs...)
```
which we used previously,, but with the input `ps_in` used as a base for `ps`:
```julia
ps   = OrderedSet{Any}(ps_in)
```
(`{Any}` is needed to be added, else it becomes an `Num` set, while this way it is kept as an `Any` set, which is what we had previously)

Also, we no longer parse the parameters and add leftover ones (as they are there in the beginning):
```julia
quote
        _ps = $pexprs
        $sexprs
        $rxexprs
    end
```

isaacsas:
@TorkelE will look through carefully tomorrow and merge if I don't see anything, sorry for the delay. Thanks for updating to cover that case!

src/reactionsystem.jl
```diff
@@ -279,6 +279,29 @@ function ReactionSystem(rxs::Vector{<:Reaction}, iv; kwargs...)
 end
 
 
+# Only used internally by the @reaction_network macro. Permits giving an initial order to the parameters, and then adds additional ones found in the reaction. Name could be changed.
+function make_ReactionSystem_internal(rxs::Vector{<:Reaction}, iv, no_sps::Nothing, ps_in; kwargs...)
```
isaacsas, line 283: Given the code redundancy, could you update the two argument constructor above to just call this too with an empty parameter vector?

isaacsas (commented):
Aside from that one comment looks good!

ChrisRackauckas:
Orders can change in MTK so I wouldn't rely on them. In fact, I would like to get rid of that parameter list at the bottom and use auto-detection if we could.

isaacsas:
This looks fine to me so I'll merge. As Chris said though, assuming an ordering is at one's own risk and could change in non-breaking releases. We've already announced that in the history.md as the policy going foward.

@ChrisRackauckas we need some way to label parameters, maybe something like this could be made to work though:
```julia
@parameters k, b
rn = @reaction_network rn begin
          k*A, A --> B
       end
```
where parameters are pre-declared using MT and species are inferred to be anything that is not yet defined? I'm not quite sure how to modify the macro to handle this, but it seems like it should be doable.

ChrisRackauckas:
Could they be deduced automatically from the DSL as the terms which are not functions and not states?

isaacsas:
The DSL has to deduce both states and parameters though, and there could be states that one wants to only appear in a rate expression. (For example, in linking component-based systems.) 

We could potentially change the default to make unknown symbols in a rate expression parameters and unknown symbols in a `A + B --> C` type expression species, with adding a parameter list optional to override this behavior (or pre-declaring one of them).

TorkelE:
Having something like 
```julia
rn = @reaction_network rn begin
    k*A, A --> B
end
u0 = [A => 1.0, B => 1.0]
p = [k=> 2.0]
tspan = (0.,10.)
oprob = ODEProblem(rn,u0,tspan,p)
```
with not giving any parameters at the end (i.e. this functionality disabled) sounds good to me. Auto-detecting parameters and species as @isaacsas suggests seems good and doable. Then, if one wishes to have species in the reaction rates we would require them to be pre-declared like:
```julia
@variables K(t)
rn = @reaction_network rn begin
    K*A, A --> B
end
u0 = [A => 1.0, B => 1.0, K=> 2.0]
p = []
tspan = (0.,10.)
oprob = ODEProblem(rn,u0,tspan,p)
```

The alternative is to simply require all components to be predeclared 
```julia
@parameters k
@variables A(t) B(t)
rn = @reaction_network rn begin
    k*A, A --> B
end
u0 = [A => 1.0, B => 1.0, K=> 2.0]
p = []
tspan = (0.,10.)
oprob = ODEProblem(rn,u0,tspan,p)
```
I'd prefer the former alternative, since it allows to make simpler code. However, I can recognise that their might be potential advantages with enforcing rigourity.

isaacsas:
The drawback is that the behavior of the macro now changes depending on context, but I guess it isn't too bad to say everything undefined in a reaction rate expression becomes a parameter, and everything undefined in a stoichiometry equation becomes a species, and if you want different behavior you need to pre-declare the associated species/parameter. As long as we can do this and still support interpolation I think it would be reasonable to change.

TorkelE:
Maybe there's a way to mark a species occurring in rates somehow? Is this something we expect to occur often at all? The functionality should definitely be available, but if it is very weird maybe we could add something for it, ensuring the macro is unambiguous.

TorkelE:
E.g. with us removing the parameter list at the end, one could have that instead be a list of stuff which appears in rates, but should be considered a species and not a parameter. That is a bit awkward, but leaves the option, and if we don't really expect it to happen it doesn't matter.
(although making this specific switch directly, without lots of printed warnings, will probably cause loads of confusion with those that don't read the update docs...)

isaacsas:
That wouldn't work if we later want to allow parameters in stoichiometric coefficients right? 

I think if we are making this switch, I would prefer to just have that unknown species appearing in rates and unknown parameters appearing in reactions require being pre-declared using the normal MT macros. We would also need to impose a convention on which of the two takes priority (i.e. if a symbol appears in both it is treated as a species unless pre-declared?).

TorkelE:
So, I was _thinking_ we might be able to distinguish something appearing as stochiometry from a species, but am getting more uncertain. It depends on if we can enforce order:
```
N(A+B) -->ANBN
```
Typically the stochiometry is written before, here we could assume N is the parameter. However, it means things can't be written as:
```
(A+B)N --> ANBN
``` 
Here, (A+B) becomes the sum of two parameters, and N a species. Convention says the stoichiometric constant is always first, i.e. no one writes "H*2+O +H_{2}O". Depending on how one mixes things it might become less clear.
```
(N1+N2)(A+B) ->AN1N2BN1N2
```
N1 and N2 still parameters, quite clear. What happens if there are three terms in the product (who sensibly uses that)
```
(N1+N2)*(H)(A+B) -->
```
N1 and N2 parameters, A and B species. What is H? Species*Species doesn't make much sense, so a parameter I guess? It is a little bit sketchy, so not sure whenever we actually want to make these kinds of assumtpions.

ChrisRackauckas:
The alternative is to make macros for generating `Reaction`s easier, not the full system. 

```julia
r1 = @reaction K*A, A --> B
ReactionSystem(r1,states,params)
```

which automates a lot of the mass action handling and all of that still, but does it in a form closer to the programmable style so that it can split apart more easily.

isaacsas:
Irrespective of changes to the `@reaction_network` DSL, I like the idea of adding such a macro for use in more programmatic contexts. It would also be nice in essentially allowing round-tripping of reactions. i.e. one could write a set of reactions to a file via the current `show` function, and load them back in by appending `@reaction` in front of each.

TorkelE:
Having a 
```julia
r1 = @reaction K*A, A --> B
```
macro sounds good. Happy to fix it once we have figured out how we want things

Issue #355, opened 2021-07-26
Moving functions to sub package


TorkelE:
The package  currently exports these function names:
`mm` , `mmr` , `hill` , `hillr` , `hillar`.

To avoid these potentially overwriting other stuff people have decalred, they could be moved to some other sub package. They would then be imported using e.g.
````
using Catalyst.Functions

isaacsas:
Going to close since I think we decided this isn't need in https://github.com/SciML/Catalyst.jl/pull/458.

PR #458, opened 2021-12-23
submodule for regsitered functions in registered_functions.jl

yewalenikhil65:
This is about https://github.com/SciML/Catalyst.jl/issues/355

Functions 
`mm` , `mmr` , `hill` , `hillr` , `hillar`
 can imported with this PR as
```
using Catalyst.Functions
```

### DIFF
--- a/src/Catalyst.jl
+++ b/src/Catalyst.jl
@@ -52,8 +52,11 @@ include("reaction_network.jl")
 export @reaction_network, @add_reactions
 
 # registers CRN specific functions using Symbolics.jl
-include("registered_functions.jl")
-export mm, mmr, hill, hillr, hillar
+module Functions
+    using Symbolics
+    include("registered_functions.jl")
+    export mm, mmr, hill, hillr, hillar
+end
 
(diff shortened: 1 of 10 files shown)


yewalenikhil65:
@TorkelE / @isaacsas  Is this submodule for rate-functions still in need ? I realised it was an old issue

isaacsas:
I can't speak for @TorkelE, but I'm generally not a big fan of submodules -- I'd prefer making a separate library like CatalystFunctionLibrary.jl. Right now we have a pretty small amount of these specialized functions, and I don't see us adding more than a couple more (maybe Goldbeter-Koshland functions?), so I don't think it is needed at this time. If it really becomes an issue we can just stop exporting them and tell users to explicitly import the ones they need, and/or move them to a separate package.

yewalenikhil65:
> 

Okay ! If the general consensus is same, i will close the PR.

yewalenikhil65:
> > 
> 
> Okay ! If the general consensus is same, i will close the PR.

closed for now !

PR #460, opened 2022-01-03
Update reaction_network.jl

rpmuller:
Fixing formatting of the first example system the way I believe CR originally intended.


isaacsas:
Thanks!

Issue #461, opened 2022-01-04
Variables added with `addparam!` or `addspecies!` are not accessible with `getproperty`


lgrozinger:
After adding parameters or species to a `ReactionSystem` using `addparam!` or `addspecies!` I would expect them to be accessible (as all other variables are) using `getproperty`, but doing so causes `ModelingToolkit.jl` to throw an `ArgumentError`. E.g.

```
using Catalyst
using ModelingToolkit

system = @reaction_network
@variables p
addparam!(system, p)
system.p
```

Throws:

```
ERROR: ArgumentError: System ##ReactionSystem#277: variable p does not exist   
Stacktrace:
 [1] getvar(sys::ReactionSystem{Nothing}, name::Symbol; namespace::Bool)
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/r5ZaU/src/systems/abstractsystem.jl:320
 [2] getproperty(sys::ReactionSystem{Nothing}, name::Symbol; namespace::
Bool)                                                                  
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/r5ZaU/src/systems/abstractsystem.jl:276
 [3] getproperty(sys::ReactionSystem{Nothing}, name::Symbol)
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/r5ZaU/src/systems/abstractsystem.jl:276
 [4] top-level scope
   @ REPL[6]:1
```

Even though `parameters` works as expected:
```
julia> parameters(system)
1-element Vector{Any}:
 p
```

lgrozinger:
This happens because `addparams!` and `addspecies!` does not update the `ReactionSystem`'s `var_to_name` dictionary as part of adding the variable. `ModelingToolkit.getvar` uses `var_to_name` to find variables.

isaacsas:
Thanks! Sometimes it is hard to keep up with ModelingToolkit's changes under the hood (which we need to parallel here). I'll try to fix this in the next few days (or feel free to put in a PR since it seems you see what needs to be updated).

lgrozinger:
No problem, I find Catalyst.jl to be a great package.
Yes, I am prepping a pull request at the moment.

PR #462, opened 2022-01-04
Fix access to variables added with addparam! or addspecies!

lgrozinger:
Variables added with addparam! or addspecies! can now be accessed
using Base.getproperty.

The addparams! and addspecies! methods now use
ModelingToolkit.process_variables! to update the var_to_name
dictionary when variables are added.


lgrozinger:
Fixes https://github.com/SciML/Catalyst.jl/issues/461

isaacsas:
Thanks!

PR #463, opened 2022-01-12
Update bifurcation_diagram.md

rveltz:
Minor change to respect BifurcationKit interface

### DIFF
--- a/docs/src/tutorials/bifurcation_diagram.md
+++ b/docs/src/tutorials/bifurcation_diagram.md
@@ -41,7 +41,7 @@ opts = ContinuationPar( dsmax = 0.05,        # Maximum arclength value of the ps
                         detectBifurcation=3, # Value in {0,1,2,3} determening to what extent bofurcation points are detected (0 means nothing is done, 3 both them and there localisation are detected).
                         newtonOptions = NewtonPar(tol = 1e-9, verbose = false, maxIter = 15)) #Parameters to the newton solver (when finding fixed points) see BifurcationKit documentation.
                         
-DO = DeflationOperator( 2.0,    # Algorithm parameter required when using deflated continuation, see BifurcationKit documentation.
+DO = DeflationOperator( 2,      # Algorithm parameter required when using deflated continuation, see BifurcationKit documentation.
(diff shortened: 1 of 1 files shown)


isaacsas:
Thanks! Just to double check, this runs on the latest release with these changes?

rveltz:
yes

rveltz:
This example is tough! The equations are badly conditioned. Where did you get it from?

isaacsas:
@TorkelE came up with it, so perhaps he can give a citation.

isaacsas:
Looks like some kind of gene regulatory model...

TorkelE:
It is a generalised version of a gene regulation model, I made it some time ago. I include it here because you could have a Hopf and a Fold bifurcation in the same diagram. If you have any comments on it I'd be interested to hear though!

rveltz:
The normal form of the Hopf has tiny coefficients, the period of the cycle is large (~900). Direct simulation near the hopf hints at a slow-fast system, see the homoclinic like orbit.... this is tough for BifurcationKit
Screen Shot 2022-01-12 at 21 27 29
t

rveltz:
> If you have any comments on it I'd be interested to hear though!

I think to study it, you have to "unfold" the parameters, like perform continuation of the fold and Hopf. Do you have a paper where the parameters where taken?

rveltz:
I can find a bogdanov-takens point nearby. I guess the homoclinic orbit comes from there.

TorkelE:
Practically, with regards to the example in the docs, do you think it might be better to replace it with a normal bistable switch?

I have to admit I am a bit out of my depth here, I am neither familiar with homoclinic orbits nor bogdanov-takens points. Also, with "unfold" the parameters, do you mean that I take another parameter and track how the position of the bifurcation points change as I change that parameter? I have to admit, I was unaware of this technique until I got more familiar with BifurcationKit and its examples. I have never really seen it done in analysis, so have never really thought of doing it myself.

The parameter set is nothing (intentionally) special. It is simply one where I could generate both bifurcation points (and a bit similar to a set I had used for a nondimensionalised version of the system, with noise added).

PR #465, opened 2022-01-16
Interpolate make_ReactionSystem_internal

andreasnoack:
...into `Expr` in `make_reaction_system` to ensure that it is properly qualified when `make_reaction_system` is used
from other modules.

This makes

```julia
julia> ex = quote
           (Ka, Depot --> Central)
           (CL / Vc, Central --> 0)
       end;

julia> Base.remove_linenums!(ex);

julia> eval(Catalyst.make_reaction_system(ex, (:CL, :Vc, :Ka)))
Model ##ReactionSystem#274 with 2 equations
States (2):
  Depot(t)
  Central(t)
Parameters (3):
  CL
  Vc
  Ka
```
work.

(I've also removed some white space while here, so please click the first commit when reviewing the diff).

### DIFF
--- a/src/reaction_network.jl
+++ b/src/reaction_network.jl
@@ -225,7 +225,7 @@ function make_reaction_system(ex::Expr, parameters; name=:(gensym(:ReactionSyste
     foreach(species -> (species isa Symbol) && push!(sexprs.args, Expr(:call,species,:t)), allspecies)
 
     # ReactionSystem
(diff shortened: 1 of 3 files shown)


isaacsas:
Thanks!

isaacsas (approved):

andreasnoack:
I forgot one thing. I'd like to add the example here as a test but which file would be the best fit?

isaacsas:
Probably dsl.jl

Issue #466, opened 2022-01-17
`extend` should add variables from constraint system to the `ReactionSystem`


lgrozinger:
The behaviour of `ModelingToolkit.extend` is that the variables from all systems are accessible in the new system. E.g.

```
using ModelingToolkit

@parameters a b
@variables t x(t) y(t)
@named sys1 = NonlinearSystem([x ~ a], [x], [a])
@named sys2 = NonlinearSystem([y ~ b], [y], [b])
sys = extend(sys1, sys2)
```
After which I can do:
```
julia> sys.a
sys1₊a

julia> sys.b
sys1₊b
```

This is not what happens when extending a `ReactionSystem` with a constraint system. E.g.
```

network = @reaction_network
@parameters a
@variables t x(t)
@named constraints = NonlinearSystem([x ~ a], [x], [a])
network = extend(constraints, network)
```
After which accessing the variables from `constraints` throws an `ArgumentError`:
```
julia> network.x
ERROR: ArgumentError: System constraints: variable x does not exist
Stacktrace:
 [1] getvar(sys::ReactionSystem{NonlinearSystem}, name::Symbol; namespace::Bool)
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/r5ZaU/src/systems/abstractsystem.jl:320
 [2] getproperty(sys::ReactionSystem{NonlinearSystem}, name::Symbol; namespace::Bool)
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/r5ZaU/src/systems/abstractsystem.jl:276
 [3] getproperty(sys::ReactionSystem{NonlinearSystem}, name::Symbol)
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/r5ZaU/src/systems/abstractsystem.jl:276
 [4] top-level scope
   @ REPL[5]:1
```

I would have expected `extend` to behave similarly in both cases.

lgrozinger:
#467

PR #467, opened 2022-01-17
extend adds variables from constraint system to the ReactionSystem

lgrozinger:
In order to address https://github.com/SciML/Catalyst.jl/issues/466


isaacsas:
Thanks! This makes sense at first glance, but before merging I want to think a little later today about whether this has any unforeseen consequences vs. making them accessible via the constraint system.

lgrozinger:
> Thanks! This makes sense at first glance, but before merging I want to think a little later today about whether this has any unforeseen consequences vs. making them accessible via the constraint system.

Yes, I must admit I am not 100% certain using the `var_to_names` will not have knock-on effects.

I will say however, that accessing via the constraint system itself is not always as useful as accessing directly from the network. E.g.
```
network = @reaction_network
@parameters a
@variables t x(t)
@named constraints = NonlinearSystem([x ~ a], [x], [a])
network = extend(constraints, network, name=nameof(network))
```
Will work fine but results in potential namespace troubles because:

```
julia> convert(ODESystem, network).x
var"##ReactionSystem#275₊x"(t)

julia> network.constraints.x
constraints₊x(t)
```

This has come up in a lot of the stuff I've been using Catalyst.jl for when composing stuff together, along the lines of:
```

network = @reaction_network
@parameters a
@variables t x(t)
@named constraints = NonlinearSystem([x ~ a], [x], [a])
network = extend(constraints, network, name=nameof(network))

@variables y(t)
@named moreconstraints = NonlinearSystem([y ~ network.constraints.x], [y], [])
network = extend(moreconstraints, network, name=nameof(network))
```

With the problem that the resulting system then ends up with more `x`s than there should be:
```
julia> equations(network)
2-element Vector{Any}:
 0 ~ a - x(t)
 0 ~ constraints₊x(t) - y(t)
```

Which I would guess is unexpected, though I may be entirely wrong.

lgrozinger:
An alternative option I can think of is to add a method `ModelingToolkit.getvar(::ReactionSystem, ::Symbol)` which checks in the constraint systems and handles the namespacing stuff

isaacsas:
Constraint variables should live at the same namespace level as the reaction system `states` field, so I think this makes sense. A variable `x` in both the `ReactionSystem`'s `state`s field and the constraint system's `state`s field should be considered the same variable. 

I'm not sure if this approach or modifying `getvar` makes more sense. @YingboMa do you have a recommendation for which makes more sense here?

src/reactionsystem.jl
```diff
@@ -252,8 +252,13 @@ function ReactionSystem(eqs, iv, states, ps;
     MT.process_variables!(var_to_name, defaults, ps′)
     MT.collect_var_to_name!(var_to_name, (eq.lhs for eq in observed))
 
-    ReactionSystem(eqs′, iv′, states′, ps′, var_to_name, observed, name, systems,
-                   defaults, connection_type, constraints;
+    if constraints != nothing
```
isaacsas, line 255:
```suggestion
    if constraints !== nothing

```

src/reactionsystem.jl
```diff
@@ -252,8 +252,14 @@ function ReactionSystem(eqs, iv, states, ps;
     MT.process_variables!(var_to_name, defaults, ps′)
     MT.collect_var_to_name!(var_to_name, (eq.lhs for eq in observed))
 
-    ReactionSystem(eqs′, iv′, states′, ps′, var_to_name, observed, name, systems,
-                   defaults, connection_type, constraints;
+    if constraints !== nothing
+
```
isaacsas, line 256:
```suggestion
```

isaacsas:
With this tweak I'll plan to merge once tests pass. We can update later on if it turns out it would be better to modify `getvars`.

isaacsas:
@lgrozinger can you add observables from the constraint system too for completeness? Like at: 
https://github.com/SciML/Catalyst.jl/blob/6f391eb194ae4aefde1ce851c94391c274c3b021/src/reactionsystem.jl#L253

YingboMa (approved):
This looks correct to me. `extend` should have the same effect as just copy-pasting everything and add a few new equations or variables.

lgrozinger:
> @lgrozinger can you add observables from the constraint system too for completeness? Like at:
> 
> https://github.com/SciML/Catalyst.jl/blob/6f391eb194ae4aefde1ce851c94391c274c3b021/src/reactionsystem.jl#L253

can do, will take a look later

lgrozinger:
Looking at this again, I have realised there are some details that it might be appropriate to address while we are at this.

The handling/propogation of default values and observed equations is a little wonky as is. I'll see if I can propose some fixing changes soon.

lgrozinger:
@isaacsas The latest commit adds a couple of lines and looks after defaults and the observed quantities.

src/reactionsystem.jl
```diff
@@ -250,10 +250,17 @@ function ReactionSystem(eqs, iv, states, ps;
     var_to_name = Dict()
     MT.process_variables!(var_to_name, defaults, states′)
     MT.process_variables!(var_to_name, defaults, ps′)
-    MT.collect_var_to_name!(var_to_name, (eq.lhs for eq in observed))
+    MT.collect_var_to_name!(var_to_name, eq.lhs for eq in observed)
 
-    ReactionSystem(eqs′, iv′, states′, ps′, var_to_name, observed, name, systems,
-                   defaults, connection_type, constraints;
+    if constraints !== nothing
+        MT.process_variables!(var_to_name, defaults, get_ps(constraints))
```
isaacsas, line 256: I think this is a bit trickier / more work than I initially realized -- my apologies for not realizing this sooner!

If we want to maintain consistency with how states, parameters and equations are handled, then `rn.defaults` should not store default values for states/parameters that live in the constraint system (these should be stored in the constraint system). So the `ReactionSystem.defaults` field should only give defaults that actually live within `rn`. To get all the `defaults` we should overload `ModelingToolkit.defaults(...)` to merge in the constraint system's defaults, similar to how we overload `states`, `parameters` and `equations` at:
https://github.com/SciML/Catalyst.jl/blob/6f391eb194ae4aefde1ce851c94391c274c3b021/src/reactionsystem.jl#L305

It probably makes sense to treat `observed` similarly. 

I'm now thinking it might be better to just overload `getproperty` for `ReactionSystem`s to look first in the `ReactionSystem`, and then in the constraint system for symbols. That avoids having a weird hybrid where `var_to_name` merges the `ReactionSystem` and the constraint system together, but all other fields are kept separate between the two systems.

The alternative is to truly treat the constraint subsystem as part of the `ReactionSystem`, but in that case it seems like there is no purpose to having it as its own system (i.e. we should just store all its info in the `ReactionSystem`s fields). I had avoided this approach with the idea that the `ReactionSystem` fields correspond to species and reactions, with other variables and equations stored in the constraint system. It seemed like this would make subsequent analysis tooling easier to handle.

If you want to tackle this please feel free, but otherwise if you let me know I can try to handle this over the next several days. And sorry again for going in a circle on this!

isaacsas (changes requested):

src/reactionsystem.jl
isaacsas: Actually, I think your original suggestion of overloading `getvar` is more appropriate to avoid an `ArgumentError` being thrown otherwise.

lgrozinger: @isaacsas I would be happy to do some more work on this. As you say, it is probably best to simply define some MT methods for ReactionSystem s so that the conceptual separation between the reaction network and the constraints system is maintained.

I will work on those methods presently.

> And sorry again for going in a circle on this!

No worries!

> The alternative is to truly treat the constraint subsystem as part of the ReactionSystem

That is interesting. Might it not be the case that a lot of code could be simplified or removed if that was how it was done? In that case, a `ReactionSystem` might simply be an `AbstractSystem` which can have both `Equation`s and `Reaction`s in the `eqs` field. The trade off would be extra logic in the `convert` functions, I would guess, but on balance it might work out to a leaner codebase.

isaacsas: Sounds good.

Merging would reduce having to overload ModelingToolkit functions, but not eliminate it (for example, some of the ModelingToolkit functions assume concrete types for the equation array but we'd have to use `Any`). It also means we no longer know more concretely what are species vs. algebraic variables, or know without searching reactions vs. equations (though this could be handled with a special array type at system construction).

lgrozinger: Could we not also add some metadata to variables that are introduced in `Reaction`s, to mark them as species? I will at some point explore this option, but for now I am just working on definitions for `MT.defaults(::ReactionSystem)` and `MT.observed(::ReactionSystem)` as a quick solution.

isaacsas: Sure, but wouldn't we then have to ask users to manually add that metadata if they are defining `ReactionSystem`s directly and not by the DSL?

In any case, for now I agree that it would be best to stick with the current approach and just add the two needed functions.

lgrozinger: > In any case, for now I agree that it would be best to stick with the current approach and just add the two needed functions.

Yes this is easier. Should be working now.

> but wouldn't we then have to ask users to manually add that metadata if they are defining ReactionSystems directly and not by the DSL?

I'm not sure about this, but my initial thought was to handle that transparently in the constructor for `Reaction`s. Unless there is some corner cases that go around the `Reaction` constructor but yet still introduce new species, that might work.

And/or as you suggest, a `@species` macro.

src/reactionsystem.jl
```diff
@@ -323,6 +323,62 @@ function MT.equations(sys::ReactionSystem)
...
+function MT.defaults(sys::ReactionSystem)
+    constraints = get_constraints(sys)
+    defs = get_defaults(sys)
+    if constraints !== nothing
+        defs = merge(defs, MT.defaults(rename(constraints, nameof(sys))))
+    end
+    return defs
+end
+
+function MT.observed(sys::ReactionSystem)
```
isaacsas, line 335: If you use `get_observed` on the constraint system they won't be namespaced, so can just be directly merged. I don't think we want to allow subsystems on the constraint system, so I don't think we need to worry about that.

For handling subsystems too, can you just mimic what is done at  

https://github.com/SciML/ModelingToolkit.jl/blob/ccf4af6336fb1fbabc51e960ea7b488cb054c47c/src/systems/abstractsystem.jl#L473-L480

src/reactionsystem.jl
```diff
@@ -323,6 +323,62 @@ function MT.equations(sys::ReactionSystem)
     return eqs
 end
 
+function MT.defaults(sys::ReactionSystem)
```
isaacsas, line 326: Similar comments to what I wrote for handling `observed`. Can't you just use `get_defaults` for the constraint system and avoid the `rename`? However, you need to handle subsystems too as at:

https://github.com/SciML/ModelingToolkit.jl/blob/ccf4af6336fb1fbabc51e960ea7b488cb054c47c/src/systems/abstractsystem.jl#L493-L503

isaacsas (changes requested):

isaacsas (approved):

isaacsas:
@lgrozinger looks good. There are probably some tweaks we can make down the line, but I'm happy to merge and make a new release as is. Are you done?

lgrozinger:
@isaacsas yes I think this issue can be closed.

Issue #464, opened 2022-01-13
provide defaults using @reaction_network


Sov-trotter:
`ReactionSystem` has a `defaults` field to allow for "initial conditions and/or parameter values" . 

But when I create a `ReactionSystem` using the `@reaction_network` macro, it's not currently possible to provide default values and `Catalyst.get_defaults(rn)` returns an empty `Dict`.

isaacsas:
Yeah, I don't think there is a nice way to do or add this with the current macro.

What we were discussing is making an intermediate level `rx = @reaction k, A+B --> C` macro that is more explicit and could hopefully keep the compact notation of the current macro, while allow the flexibility of using the underlying ModelingToolkit `@parameters` and `@variables` macros to species parameters and species (and default values).

isaacsas:
@Sov-trotter this should work in the next release (or using master):
```julia
rn = @reaction_network begin
    α, S + I --> 2I
    β, I --> R
end α β
defs = [:S => 999.0, :I => 1.0, :R => 0.0, :α => 1e-4, :β => .01]
setdefaults!(rn, defs)
op = ODEProblem(rn, [], tspan, [])
sol = solve(op, Tsit5())
```

PR #468, opened 2022-01-18
fix DSL annoyances

isaacsas:
Closes https://github.com/SciML/Catalyst.jl/issues/464.

When using the DSL, not having variables and parameters defined in the current context makes it annoying to construct parameter / initial condition mappings. This PR adds helpers to make this easier. 
```julia
rn = @reaction_network begin
    α, S + I --> 2I
    β, I --> R
end α β
defs = [:S => 999.0, :I => 1.0, :R => 0.0, :α => 1e-4, :β => .01]
setdefaults!(rn, defs)
op = ODEProblem(rn, [], tspan, [])
sol = solve(op, Tsit5())
```

The docs make clear that `setdefaults!` only acts on states and parameters at the level of a given system, so needs to be manually called on sub-systems in hierarchical models, or used after flattening.

Also added the ability to unpack all variables and parameters into the current context, e.g.
```julia
rn = @reaction_network begin
    α, S + I --> 2I
    β, I --> R
end α β
Catalyst.@unpacksys rn

# now this works
u₀ = [S => 999.0, I => 1.0, R => 0.0]
p = [α => 1e-4, β => .01]
op = ODEProblem(rn, u₀, (0.0, 250.0), p)
solve(op, Tsit5())    
```
This again does not recurse into subsystems, and is not currently exported or in the docs since it is implicitly `eval`ing into the user's calling module... (But this is useful in working with Catalyst for developing I've found.)

Finally, I added a way to specify initial condition and parameter maps using symbols:
```julia
sir = @reaction_network sir begin
    β, S + I --> 2I
    ν, I --> R
end β ν
subsys = @reaction_network subsys begin
    k, A --> B
end k
@named sys = compose(sir, [subsys])
symmap = [:S => 1.0, :I => 1.0, :R => 1.0, :subsys₊A => 1.0, :subsys₊B => 1.0]
u0map  = symmap_to_varmap(sys, symmap)
pmap   = symmap_to_varmap(sys, [:β => 1.0, :ν => 1.0, :subsys₊k => 1.0])
tspan = (0.0, 250.0)
oprob = ODEProblem(sys, u0map, tspan, pmap)  # this should now work
```

### DIFF
--- a/docs/src/api/catalyst_api.md
+++ b/docs/src/api/catalyst_api.md
@@ -172,7 +172,7 @@ netstoichmat
 reactionrates
 ```
 
-## Functions to Extend a Network
+## Functions to Extend or Modify a Network
 `ReactionSystem`s can be programmatically extended using [`addspecies!`](@ref),
 [`addparam!`](@ref), [`addreaction!`](@ref), [`@add_reactions`](@ref), or
 composed using [`ModelingToolkit.extend`](@ref) and
@@ -184,6 +184,7 @@ addspecies!
 reorder_states!
 addparam!
 addreaction!
+setdefaults!
 ModelingToolkit.extend
 ModelingToolkit.compose
 Catalyst.flatten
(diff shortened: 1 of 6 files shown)


isaacsas:
CC @rveltz I recall this was bothering you too, so this may be of interest.

PR #470, opened 2022-01-24
Drop system structure methods

isaacsas:


isaacsas:
@YingboMa let me know if this also isn't considered a reasonable function for a system type to use. I could just copy this over to Catalyst, but that seems silly since it is already defined in ModelingToolkit...

YingboMa:
Yeah, they are reasonable functions to rely on. Note that `get_eqs(csys)` only returns the equations of the currently level. Is that deliberate?

isaacsas:
Yes, it is deliberate. I wasn’t intending to allow constraint systems to have subsystems since they are parallel to the ReactionSystem that stores them, but I guess it can’t hurt here to be more general. 

I’ll make a new release once this is merged.

YingboMa (approved):

PR #482, opened 2022-02-21
Allow float stoichiometry

isaacsas:
There is a decision to be made on the default conventions. 

Prior to this PR we always scaled ODE and jump rates by a combinatorial factor, treating the rate constants as microscopic rate constants. In this way our ODE models are the mean-field of our jump models.

However, such a scaling doesn't seem to me to make sense when having non-integer stoichiometry coefficients (unless we want to restrict non-integer stoichiometry to products and not allow for reactants, or scale by a gamma function which seems weird). 

In this PR I set the conventions as:
1. All the reactant and product stoichiometry of a given reaction are promoted to the same type.
2. If this type is an integer we rescale by default as we currently do.
3. If this type is not integer we error if the user does not explicitly turn off rescaling.

We could alternatively just rescale silently for the rate law of an individual reaction that involves non-integer stoichiometry. The problem is then users will get a mix of rescaled and not rescaled rate laws, which seems worse than giving them a useful error and forcing them to understand that they must turn off rescaling.

edit: This will also only work for ODEs currently, but I think that is fine as it is unclear to me what a non-integer stoichiometry even means in the stochastic case (i.e. what are the jump model's rate laws).

### DIFF
--- a/src/reaction_network.jl
+++ b/src/reaction_network.jl
@@ -398,7 +398,7 @@ function push_reactions!(reactions::Vector{ReactionStruct}, sub_line::ExprValues
 end
 
 #Recursive function that loops through the reaction line and finds the reactants and their stoichiometry. Recursion makes it able to handle weird cases like 2(X+Y+3(Z+XY)).
-function recursive_find_reactants!(ex::ExprValues, mult::Int, reactants::Vector{ReactantStruct})
+function recursive_find_reactants!(ex::ExprValues, mult::Number, reactants::Vector{ReactantStruct})
     if typeof(ex)!=Expr || (ex.head == :escape)
         (ex == 0 || in(ex,empty_set)) && (return reactants)
         if in(ex, getfield.(reactants,:reactant))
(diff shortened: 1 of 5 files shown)


isaacsas:
Note the promotion of type process probably will need to be updated to handle non-numeric types (i.e. Symbolics expressions). I'll save that for a future PR though.

isaacsas:
@ChrisRackauckas this actually seems to generate reasonable ODEs in limited local testing, so if you agree with the conventions I choose I'll add some tests next.

ChrisRackauckas (approved):
This makes a lot of sense to me.

Floating point stoichiometries in the stochastic case seem fine too though? You'd just have the effect do `integrator.u[i] += 1/3` etc. I mean, model-wise it's unclear what kind of model would use it, but mathematically nothing seems too bad to me at face value?

isaacsas:
Since earlier I had been worried this would have performance implications I `@benchmark`'ed before and after this PR with loading BCR and converting it to ODEs, I find:
```julia
prnbng = loadrxnetwork(BNGNetwork(),fname)
```
went from a median time of 436.518 ms to 437.124 ms
```julia
bosys = convert(ODESystem, rnbng)
```
from 2.151 s to 2.126 s
```julia
boprob = ODEProblem(bosys, Float64[], (0.,tf), Float64[])
```
from 1.028 s to 1.027 s.

So this doesn't look to have had any impact on the time to load and generate an ODE model.

PR #486, opened 2022-02-25
bidirectional reactions need tuple of coefs

anandijain:


isaacsas:
I’ll merge for now but if this used to work and broke recently we should update the dsl code.

isaacsas:
Nevermind my comments -- I now see we give an error explicitly stating this was not supported. Thanks again!

PR #498, opened 2022-04-08
Update Bifurcation Tutorial

TorkelE:
Update the tutorial according to discussion in https://discourse.julialang.org/t/error-with-catalyst-jl-bifurcation-diagram-tutorial/79202/7


isaacsas:
I guess BifurcationKit had some breaking releases that weren't tagged as such?

It might be good to indicate in the tutorial which version of BifurcationKit it was developed for in case more breaking changes occur in the future.

TorkelE:
Good point, added a note.

isaacsas:
Looks good to me! (You can merge it right?)

PR #503, opened 2022-05-05
Tutorial for Network Representation using complexes-based representation

yewalenikhil65:
Couple of more things to add (conservation laws etc example)


yewalenikhil65:
@isaacsas Docs test is passed, but is the tutorial generated ? or the svgs are not being read ?

isaacsas:
Looks like the doc build had a bad link? https://github.com/SciML/Catalyst.jl/runs/6312982902?check_suite_focus=true#step:5:149

Anyways, let me know when you think this is done.

yewalenikhil65:
> Looks like the doc build had a bad link? https://github.com/SciML/Catalyst.jl/runs/6312982902?check_suite_focus=true#step:5:149
> 
> Anyways, let me know when you think this is done.

Yeah, i think its done from me ! Let me know if you think something is missing and I will add it ..

yewalenikhil65:
corrected the link
```
┌ Info: Deployment criteria for deploying preview build from GitHub Actions:
│ - ✔ ENV["GITHUB_REPOSITORY"]="SciML/Catalyst.jl" occurs in repo="github.com/SciML/Catalyst.jl.git"
│ - ✔ ENV["GITHUB_REF"] corresponds to a PR number
│ - ✘ PR originates from the same repository
│ - ✔ `push_preview` keyword argument to deploydocs is `true`
│ - ✔ ENV["GITHUB_ACTOR"] exists and is non-empty
│ - ✔ ENV["GITHUB_TOKEN"] exists and is non-empty
└ Deploying: ✘
```

docs/Project.toml
```diff
@@ -6,6 +6,6 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
-
+Graphviz_jll = "3c863552-8265-54e4-a6dc-903eb78fde85"
```
isaacsas, line 9: When you update this, please take out Graphviz as a required dependency. The jll is still too wonky (for example, saving doesn't work for svgs or pdfs on aarch64 Macs currently).

yewalenikhil65: @isaacsas  Sorry, got caught up in other work and didn't notice this untill now. What are your views on the rest of tutorial ? Do we need to add anything ?

yewalenikhil65: It looks nice.. Especially now that arguments of these functions take only `ReactionSystem` as input.
I did not understand though why the output of `complexgraph(rn)` mentioned in this [section](https://github.com/yewalenikhil65/Catalyst.jl/blob/upstreamCopy/docs/src/tutorials/reaction_network_representation.md#reaction-complex-representation) shows red arrow from `2*A + 3*B --> A + 2*C + D` ? I mean we had created such arrow only if there are multiple reactions in same substrates and products with different reaction rates, if i remember correctly.

isaacsas: So black arrow should be mass action.

yewalenikhil65:
@isaacsas  Should i be writing some example under Deficiency zero theorem ? I can do that i couple of days

isaacsas:
I'm planning to finish it this weekend. Just been busy with finishing a paper this week.

isaacsas:
@yewalenikhil65 I plan to merge this version once tests pass. Let me know if you see any issues and I can fix them in a followup PR if needed.

isaacsas:
@ChrisRackauckas any idea about this test failure? https://github.com/SciML/Catalyst.jl/runs/6851564267?check_suite_focus=true

We are just passing an `SDESystem` into `SDEFunction{false}` and then this failure occurs. This worked until the most recent release of SciMLBase. Is auto-generation of oop for an SDESystem no longer done in this manner?

yewalenikhil65:
> @yewalenikhil65 I plan to merge this version once tests pass. Let me know if you see any issues and I can fix them in a followup PR if needed.

I really liked deficiency zero theorem section. Couldn't have written it any better myself .. This sort of completes our matrix representations, doesn't it ? There are some similar matrix representations for stochastic systems of reactions as well.. I think at some point we should bring that in it too just for the sake of completeness  1

isaacsas:
I wouldn't say we have finished everything. There are lots more network analysis results we could add over time. for example results on deficiency one networks, complex balancing, etc... As you say there are also results for stochastic models.

isaacsas:
I don't think we need to include `complexoutgoingmat` explicitly, unless you have a useful application of it specifically (i.e. right now it is implicitly called by other functions, but a user might never need to use it).

ChrisRackauckas:
> @ChrisRackauckas any idea about this test failure? https://github.com/SciML/Catalyst.jl/runs/6851564267?check_suite_focus=true

Incoming patch: https://github.com/SciML/SciMLBase.jl/pull/184

isaacsas:
Ah great, thanks! I'll just hold off on merging this till there is a new release (and drop the temp SciMLBase cap I added for Master).

PR #506, opened 2022-05-18
Elimination of conservation laws

isaacsas:
`ReactionSystem`s store a new (internal) field, `networkproperties`, that caches all the needed info for the various `convert` methods to eliminate out conservation laws (along with also being used to cache various network properties in the API). I left it this way as it didn't really seem to make sense to try to eliminate species at the `ReactionSystem` level (i.e. reactions would then contain algebraic expressions for the eliminated species as their substrates). Happy to consider alternative designs if there are better / more standard approaches.

TODO:
- [x] Add `NetworkProperties` structure for storing cached properties
- [x] Update various API functions to use this structure consistently.
- [x] drop AbstractAlgebra dependency. 
- [x] Calculate symbolic equations and substitutions for conservation laws.
- [x] Update `convert`s with appropriate kwarg that generates reduced systems.
- [x] Add tests
- [x] Update API tests 

```julia
rn = @reaction_network begin
       k, A + B --> C
       k2, C --> A + B
       end k k2
conservationlaws(rn)
get_networkproperties(rn)
```
giving
```
Conserved Equations: 
B(t) ~ A(t) + _ConLaw[1]
C(t) ~ _ConLaw[2] - A(t)

Net stoichiometry matrix: 
[-1 1; -1 1; 1 -1]
```

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,50 @@
 # Breaking updates and feature summaries across releases
 
-## Catalyst unreleased (master branch) 
+## Catalyst unreleased (master branch)
+- **BREAKING:** Added the ability to eliminate conserved species when generating
+  ODEs, nonlinear problems, and steady-state problems via the
+  `remove_conserved=true` keyword that can be passed to `convert` or to
+  `ODEProblem`, `NonlinearProblem` or `SteadyStateProblem` when called with a
+  `ReactionSystem`. For example,
+  ```julia
+  rn = @reaction_network begin
+     k, A + B --> C
+     k2, C --> A + B
+     end k k2
+  osys = convert(ODESystem, rn; remove_conserved=true)
+  equations(osys)
+  ```
+  gives
+  ```
+  Differential(t)(A(t)) ~ k2*(_ConLaw[2] - A(t)) - k*(A(t) + _ConLaw[1])*A(t)
+  ```
+  Initial conditions should still be specified for all the species in `rn`, and
+  the conserved constants will then be calculated automatically. Eliminated
+  species are stored as observables in `osys` and still accessible via solution
+  objects. Breaking as this required modifications to the `ReactionSystem` type
+  signature.
+- **BREAKING:** Added an internal cache in `ReactionSystem`s for network properties, and
+  revamped many of the network analysis functions to use this cache (so just a
+  `ReactionSystem` can be passed in). Most of these functions will now only
+  calculate the chosen property the first time they are called, and in
+  subsequent calls will simply returned that cached value. Call
+  `reset_networkproperties!` to clear the cache and allow properties to be
+  recalculated. The new signatures for `rn` a `ReactionSystem` are
+  ```julia
+  reactioncomplexmap(rn)
+  reactioncomplexes(rn)
+  complexstoichmat(rn)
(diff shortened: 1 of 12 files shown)


isaacsas:
```julia
rn = @reaction_network begin
    (k1,k2), A + B <--> C
    (m1,m2), D <--> E
    b12, F1 --> F2
    b23, F2 --> F3
    b31, F3 --> F1
  end k1 k2 m1 m2 b12 b23 b31
osys = convert(ODESystem, rn; remove_conserved=true)
```
```
Model ##ReactionSystem#347 with 4 equations
States (4):  
A(t)
D(t)
F1(t)
F2(t)
Parameters (11):
  k1
  k2
  m1
  m2
  b12
  b23  
  b31
  _ConLaw[1] [defaults to D(t) + E(t)]
  _ConLaw[2] [defaults to A(t) + C(t)]  
  _ConLaw[3] [defaults to B(t) - A(t)]  
  _ConLaw[4] [defaults to F1(t) + F2(t) + F3(t)]
julia> equations(osys)
4-element Vector{Equation}: 
Differential(t)(A(t)) ~ k2*(_ConLaw[2] - A(t)) - k1*(A(t) + _ConLaw[3])*A(t) 
Differential(t)(D(t)) ~ m2*(_ConLaw[1] - D(t)) - m1*D(t) 
Differential(t)(F1(t)) ~ b31*(_ConLaw[4] - F1(t) - F2(t)) - b12*F1(t)
Differential(t)(F2(t)) ~ b12*F1(t) - b23*F2(t)
```

isaacsas:
@YingboMa is the way I'm setting the values for the constants in the conservation laws reasonable? They are determined by an algebraic equation from the initial conditions for the independent and dependent species. I'm just setting them to be the associated algebraic formula in the appropriate (time-dependent) states, and in my limited testing they get the right value when building `ODEProblem`s.

So basically, should I expect the following to keep working as a way to set the value of `__ConLaw[1]` using the `t=0` values of `D` and `E`?
```
  _ConLaw[1] [defaults to D(t) + E(t)]
```

Thanks!

YingboMa:
> So basically, should I expect the following to keep working as a way to set the value of __ConLaw[1] using the t=0 values of D and E?

Yeah, as long as `D` and `E` don't get reduced away. We should have better initialization support.

isaacsas:
One won't be a state (the eliminated variable), but it will be an observable, and users will be expected to provide initial conditions for all of them (i.e. even if moved to an observable).

isaacsas:
Most of the network analysis API function signatures now are much cleaner:
```julia
rn::ReactionSystem
reactioncomplexmap(rn)
reactioncomplexes(rn)
complexstoichmat(rn)
complexoutgoingmat(rn)
incidencematgraph(rn)
linkageclasses(rn)
deficiency(rn)
sns = subnetworks(rn)
linkagedeficiencies(rn)
isreversible(rn)
isweaklyreversible(rn, sns)
```

isaacsas:
I wasn't sure if they should get `!` on their names now or not, but figured to leave off since it is modifying an internal cache structure, and only modifies the first time it is called. I could change it to have them all end in `!` if preferred by anyone though.

PR #514, opened 2022-05-30
Constant and bc species for SBMLToolkit

isaacsas:
TODO
- [x] convert for jumps
- [x] tests for constant and bc species
- [x] ismassaction for jumps fixes

Also added `continuous_events` for ODEs (the only system type that currently supports them in ModelingToolkit as far as I am aware).


isaacsas:
@paulflang with this one we should have bc and constant species, along with continuous events propagating when used in an ODESystem constraint. You can look at the test I added to see how to create the constant and bc species.

A few caveats:
1. Constraints are only supported when converting to ODESystems (allowing ODESystem or NonlinearSystem constraints), or NonlinearSystems (allowing NonlinearSystem constraints). So BC species are not useful for SDE or jump models currently. 
2. Only an ODESystem constraints can have continuous events. No other systems types in ModelingToolkit support events (and none support discrete events currently). As ModelingToolkit adds support for them we can expand what is supported here.

If you have SBML examples we can test this with please do give it and try and let me know if there are issues!

isaacsas:
So hopefully now you don't need to manually convert `ReactionSystem`s to `ODESystem`s in SBMLToolkit just to then add in continuous events -- just use a constraint ODESystem.

isaacsas:
If tests pass I'll merge, and then I think we can make a new Catalyst (breaking) release.

isaacsas:
@ChrisRackauckas @TorkelE this should be good to go. Will give it a day for anyone to comment who might want to take a look, and will then merge and make a new (breaking) Catalyst release.

TorkelE:
Looks good to me.

src/reactionsystem.jl
```diff
@@ -891,19 +955,23 @@ end
...
-    csys = get_constraints(rs)
-    nps  = get_networkproperties(rs)
+# merge constraint components with the ReactionSystem components
+# also handles removing BC and constant species
+function addconstraints!(eqs, rs::ReactionSystem, ists; remove_conserved=false)
+    # if there are BC species, put them after the independent species
+    rssts = get_states(rs)
+    sts = any(isbc, rssts) ? vcat(ists, filter(isbc, rssts)) : ists
+
+    # if there are constant species, make them parameters
```
paulflang, line 965: @isaacsas I am wondering if it would be a good idea to make BC species that do not appear in any constraintsys to parameters, too? Otherwise we might end up with more variables than equations, as no ODEs are created from their participation in reactions.

isaacsas: I think a BC species is a contract that somewhere you will add an equation for the species, either in a constraint or after conversion by extending the converted model with more equations. We can't really detect the latter.

Otherwise if the species is really constant it should be declared as constant from the beginning.

paulflang: > I think a BC species is a contract that somewhere you will add an equation for the species

I wish you were correct, but if I understand the specs correctly, it only means that you must not allow reactions change its concentration.

> In SBML, it is possible to indicate that a given species’ amount is not determined by the set of reactions even when that species occurs as a product or reactant; i.e., the species is on the boundary of the reaction system, and its amount is not determined by the reactions. The boolean attribute boundaryCondition with value “true” can be used to indicate this. ... In practice, a boundaryCondition value of “true” means a differential equation derived from the reaction definitions should not be generated for the species. However, the species’ amount _may_ still be changed by AssignmentRule, RateRule, AlgebraicRule, Event, and InitialAssignment constructs if its constant attribute is “false”.

Importantly, it is a _may_, not a _must_. From that it would follow, that dSpecies/dt = 0, unless otherwise specified in a rule (or event).

> We can't really detect the latter.

In SBMLToolkit we can. But creating a parameter instead of a species is not really an option, as this would screw up generating mass action rate laws downstream. Alternatively we could add the equation `species(t) = species(0)` (in fact, this is what we currently do IIRC). Works for the most part, except when `structural_simplify` kicks out such a species and the solver later on tries to use or assign it in an Event. But for the moment, I suggest let's not worry about that and proceed as suggested below.

> Otherwise if the species is really constant it should be declared as constant from the beginning.

Agreed. But it does not seem to be a requirement. So I assume there will be a test suite case that does not set such a species as constant. But for the time being, let's work under the assumption of the contrary, i.e. that all test cases would set such a species as constant. And if I come across a case that does not, I'll ping you. How does that sound?

isaacsas: In SBMLToolkit, why can't you convert it to a constant species if it is actually constant?

isaacsas: That shouldn't mess up rate law generation (otherwise this PR needs updating!).

paulflang: IIUC setting these species to `constant` in SBMLtk would remove them in `convert`. This would indeed avoid having more species than equations. But then these species would be gone, i.e. could no longer be assigned by an event.

I believe the solution is to do it the other way round, i.e. increase the number of equations rather then reduce the number of species. So I can do both, create an equation `Species(t) ~ Species(0)` and set the Species as `input`.

isaacsas: OK, I thought you were talking about BC species that really are constant (i.e. not changed in events either). Setting them to constant species would indeed then make them a parameter during `convert`.

Ultimately this is a ModelingToolkit issue, in that it should support creating events that modify parameters. The event functionality still needs some development.

Have you tried not including an equation for it, but adding the `input` metadata? That is, are you sure it gets removed as a state even without an equation with that flag?

@ChrisRackauckas it looks like `fillzeros` no longer is used in Symbolics `build_function`: https://github.com/JuliaSymbolics/Symbolics.jl/blob/80f8217f48849b783d8b87d49dc0d7de603abd75/src/build_function.jl#L226

ModelingToolkit indicates the default is it is being used. Is that a known issue?

isaacsas: @paulflang I just checked, it won't be eliminated as Chris said if you use the input metadata:
```julia
using ModelingToolkit
@variables t A(t) B(t) [input=true]
@named osys = ODESystem([D(A) ~ -A],t,[A,B],[])
osys2 = structural_simplify(osys)
```
gives
```julia
julia> states(osys)
2-element Vector{Term{Real, Base.ImmutableDict{DataType, Any}}}:
 A(t)
 B(t)
```

isaacsas: I was saying that they should ideally be moved to be parameters when they have no equations (except in events), but that isn't feasible currently (i.e. for now they should remain states in this situation).

Nevermind about my example. I'm not sure what is going on, but it seems in fact `osys2` does have `B` eliminated. So `structural_simplify` is not respecting `input`. Maybe that is a bug?

That said, why are you passing in numerical vectors for `u0` and `p`? Pass defaults to Catalyst and let it pass them through, or use maps, i.e. `[A => 1.0, B => 1.0]`, etc. You shouldn't assume an ordering of states or parameters from ModelingToolkit. 

If you don't call `structural_simplify` everything seems to work:
```julia
prob = ODEProblem(osys, [A => 1.0, B => 2.0], (0.0,10.0), check_length=false)
sol = solve(prob, Tsit5())
```

isaacsas: And finally, this works too:
```julia
prob = ODEProblem(osys2, [A => 1.0, B => 2.0], (0.0,10.0), check_length=false)
sol = solve(prob, Tsit5())
```
so even with `B` eliminated everything works if you pass your initial conditions and parameters as mappings.

PR #517, opened 2022-06-05
move usings to pages.jl

isaacsas:
@ChrisRackauckas maybe this will fix https://github.com/SciML/SciMLDocs/issues/5?


ChrisRackauckas:
I don't think this will make a difference. All of the libraries are already `using`'d and passed as modules to the `makedocs` call. ModelingToolkit and Catalyst are in the list, so it'll be there.

PR #519, opened 2022-06-13
fix graphs

isaacsas:
@yewalenikhil65 when you use `savegraph` you need to make sure you pass the format too when choosing anything but png, i.e. `savegraph(graph, "fname.svg", "svg")`. I had to regenerate all the graphs locally as they were showing up messed up on iOS devices. I'm guessing it is probably because you did `savegraph(graph, "fname.svg")` which was likely saving them as pngs with an svg extension (but maybe it was some other issue).


isaacsas:
Also, one way this could be seen without using iOS is to look at a Markdown preview in VSCode. The previous figures wouldn't appear in the preview, while the new ones show up just fine.

yewalenikhil65:
> @yewalenikhil65 when you use `savegraph` you need to make sure you pass the format too when choosing anything but png, i.e. `savegraph(graph, "fname.svg", "svg")`. I had to regenerate all the graphs locally as they were showing up messed up on iOS devices. I'm guessing it is probably because you did `savegraph(graph, "fname.svg")` which was likely saving them as pngs with an svg extension (but maybe it was some other issue).

Yes, you are right.. I followed `savegraph(graph, "fname.svg")` .. I did not know about `savegraph(graph, "fname.svg", "svg")`

PR #520, opened 2022-06-13
Handle constraint params

isaacsas:


paulflang:
great, thanks!

paulflang:
Hmm, as for case 305, `S2(t)` is now correctly removed from the states and only present in the parameters of the `ODESystem` `sys`. But `structural_simplify` stragely still complains:

```
julia> states(sys)
2-element Vector{Term{Real, Base.ImmutableDict{DataType, Any}}}:
 S1(t)
 S3(t)

julia> parameters(sys)
4-element Vector{SymbolicUtils.Symbolic{Real}}:
 k1
 k2
 compartment
 S2(t)

julia> equations(sys)
2-element Vector{Equation}:
 Differential(t)(S1(t)) ~ -k2*S1(t)
 S3(t) ~ k1*S2(t)

julia> structural_simplify(sys)
ERROR: ExtraVariablesSystemException: The system is unbalanced. There are 3 highest order derivative variables and 2 equations.
More variables than equations, here are the potential extra variable(s):
 S1(t)
 S3(t)
Stacktrace:
 [1] error_reporting(state::TearingState{ODESystem}, bad_idxs::Vector{Int64}, n_highest_vars::Int64, iseqs::Bool)
   @ ModelingToolkit.StructuralTransformations C:\Users\wolf5212\.julia\packages\ModelingToolkit\fp90z\src\structural_transformation\utils.jl:35
 [2] check_consistency(state::TearingState{ODESystem})
   @ ModelingToolkit.StructuralTransformations C:\Users\wolf5212\.julia\packages\ModelingToolkit\fp90z\src\structural_transformation\utils.jl:65
 [3] structural_simplify(sys::ODESystem; simplify::Bool, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ ModelingToolkit C:\Users\wolf5212\.julia\packages\ModelingToolkit\fp90z\src\systems\abstractsystem.jl:1010
 [4] structural_simplify(sys::ODESystem)
   @ ModelingToolkit C:\Users\wolf5212\.julia\packages\ModelingToolkit\fp90z\src\systems\abstractsystem.jl:1005
 [5] top-level scope
   @ REPL[15]:1
```

isaacsas:
Yes, the `structural_simplify` issue is still present for any example with constant species. That won't go away until the change to making constant species parameters is in.

isaacsas:
But, hopefully, your SBML tests should work if you don't call `structural_simplify` -- I don't think any of the main modeling packages like Bionetgen or Copasi even apply such transformations. So it should be good enough for now to just test the generated ODESystem (until the constant species = parameters support is in).

paulflang:
No, they don't. When creating an ODEProblem I get:
```
ERROR: The LHS cannot contain nondifferentiated variables. Please run `structural_simplify` or use the DAE form.
```

isaacsas:
Are you on the latest ModelingToolkit?

paulflang:
This is case 23, which I believe always worked without structural_simplify. What is not working is case 305 (same code, except for the 4th line).

paulflang:
Oh wait, you said with 305? That is strange indeed.

paulflang:
I am on MTK v8.14.0, which is the latest.

paulflang:
cc @ChrisRackauckas

PR #522, opened 2022-06-13
Constant species as parameters

isaacsas:
@paulflang if you want to try this out for the `structural_simplify` issues feedback would be good on possible bugs.


paulflang:
Thanks, I'll see if I get to it today.

paulflang:
@isaacsas : I thought this would require a minor version bump so that I can use the registry version of Catalyst, doesn't it?

isaacsas:
No, this will be breaking so will need a major version change once released. (Since we are changing how the constants / BC species work.)

PR #527, opened 2022-06-23
format SciML Style

ChrisRackauckas:


test/latexify.jl
```diff
@@ -23,56 +23,53 @@ using Catalyst, Latexify
 ### will cause issues with Travis.
```
isaacsas, line 23:
```suggestion
#! format: off
### will cause issues with Travis.
```
Let's just turn it off for the Latexify tests for now.

PR #540, opened 2022-08-02
adding bibtex for bioarxiv paper of Catalyst.jl

yewalenikhil65:
adding bibtex for bioarxiv paper of Catalyst.jl 
https://www.biorxiv.org/content/10.1101/2022.07.30.502135v1


isaacsas:
Thanks!

PR #541, opened 2022-08-02
Homotopy continuation tutorial

TorkelE:
Was showing how to do it for conservation laws via some old example I had, but you used some additional functions in addition to `conservationlaws` in your JuliaCon example @isaacsas, so I plan to redo that part using your technique tomorrow.


docs/src/tutorials/homotopy_continuation_tutorial.md
```diff
@@ -0,0 +1,44 @@
...
+    k2, 2X --> X + Y
+    k3, X + Y --> Y
+    k4, X --> 0
+end k1 k2 k3 k4
+p = [:k1 => 8.0, :k2 => 2.0, :k3 => 1.0, :k4 => 1.5]
+```
+Next, we will need to extract the actual equations from our model. In addition, we will substitute in our parameter values.
+```@example hc1
+ns = convert(NonlinearSystem,wilhelm_2009_model)
+subs = Dict(Pair.(wilhelm_2009_model.ps.value,last.(p)))
```
isaacsas, line 20: Maybe take a look at what I did in the workshop for handling the parameter conversion when I did the conservation law case. I think it is more in line with the public ModelingToolkit/Symbolics interface. Those fields you are accessing like `.ps` and `.ps.value` are considered private, so we should try to avoid using them in tutorials (or even within Catalyst).

TorkelE:
I am doing the conservation law stuff now. I got it to work by setting defaults, however, is there some way to actually access conservation law parameters that can be manipulated (without changing the default species concentrations)?

isaacsas:
Right now conservation law parameter values are automatically determined from the initial conditions of the conservation law (when `ODEProblem` is called), but you could override this I think by setting a value for the associated symbolic variable in the parameter mapping.

isaacsas:
Probably easier to just change the initial amount of one species to control the total though.

TorkelE:
Yeah,  I am doing something like that now, and then maybe we expand some stuff so that it can be done neater later on. In either case it is nice to have examples using `conservationlaw`.

isaacsas:
We probably need a better naming scheme for the constant than `__ConservationLaw`, but it was meant to be internal so as not to clash with user parameters. It messes up Latex display too though so should get changed. Maybe we can use something with unicode a user is unlikely to type.

TorkelE:
Yeah, let's discuss it through some time.

TorkelE:
Again, I'm happy with this, and will merge tomorrow morning unless someone has some additional input.

isaacsas:
Why do you add the conservation relation back into the system? It should be solvable without those equations.

TorkelE:
Yes, but by adding it in the solutions becomes the full state vector (`[X1,X2]`). When I just reduced the system I only got a single output. While I could then calculate the full state vector from that one, it felt more natural to get the steady state as output from the solver command.

isaacsas:
OK, I'll leave it to you. It probably doesn't matter for a small system like this, but for larger systems you are increase the number of equations to solve which seems undesirable. But it is correct that one would have to play around with using the observables or `substitute` to then get values for the dropped variables.

Issue #553, opened 2022-09-22
Non-integer reaction coefficients for sub/prod/netstoichmat


jchitpin:
Would it be possible to support fractional reaction coefficients with `netstoichmat`/`prodstoichmat`/`substoichmat`?
I want to include a biomass growth pseudoreaction in my reaction network and these often contain fractional coefficients.

Here's a MWE below:
```
using Catalyst, DifferentialEquations

toy = @reaction_network Toy begin
    k₁, ∅ --> m₁
    k₂, m₁ --> 0.5 * m₂
    k₃, m₂ --> ∅
end k₁ k₂ k₃

netstoichmat(toy) # InexactError: Int64(0.5)
```

isaacsas:
You should be able to generate ODE and SDE models from this, but you are right that many API functions won't work with non-integer coefficients. That said, the stoichiometry matrices should be relatively easy to address.

jchitpin:
Yup, I have no issues generating my ODE models.
At the moment, returning the stoichiometry matrix is more of a convenience thing. I look forward to an update though in the future whenever that may be :)

isaacsas:
I've been swamped with the start of classes this past month, but should have some time starting today to work on stuff. I'll try to get a PR for this in today or tomorrow since it should be relatively easy.

isaacsas:
@jchitpin latest release should fix this for `substoichmat`, `prodstoichmat`, and `netstoichmat`.

PR #554, opened 2022-09-22
Update api funcs

isaacsas:
Closes https://github.com/SciML/Catalyst.jl/issues/553


isaacsas:
@ChrisRackauckas is that what the invalidations script is supposed to do?

PR #557, opened 2022-09-27
allow multivariable states

isaacsas:
See https://github.com/SciML/Catalyst.jl/pull/557/files#diff-3790c7b56be464b7ef6ce09f21fe4016c36ef785961317751d9282e6ccca15ff for example usage.

This adds a separate `sivs` field, and associated `get_sivs(rn)` accessor, to store spatial independent variables, as we need to be able to distinguish the spatial and time variables internally (with time still stored in `iv`).


isaacsas:
Probably makes more sense to add an `ivs` field like `PDESystem`s have, and then just have only one of `iv` or `ivs` be allowed to be defined.

isaacsas:
@xtalax do you see anything clearly wrong in the pdes.md tutorial attached to this PR? It is crashing for me in the discretize call with the message:
```julia
Discretization failed, please post an issue on https://github.com/SciML/MethodOfLines.jl with the failing code and system at low point count.

ERROR: ArgumentError: Dict(kv): kv needs to be an iterator of tuples or pairs
Stacktrace:
 [1] Dict(kv::Vector{Any})
   @ Base ./dict.jl:132
 [2] symbolic_discretize(pdesys::PDESystem, discretization::MOLFiniteDifference{MethodOfLines.CenterAlignedGrid})
   @ MethodOfLines ~/.julia/packages/MethodOfLines/7YJ7n/src/MOL_discretization.jl:132
 [3] discretize(pdesys::PDESystem, discretization::MOLFiniteDifference{MethodOfLines.CenterAlignedGrid})
   @ MethodOfLines ~/.julia/packages/MethodOfLines/7YJ7n/src/MOL_discretization.jl:162
 [4] top-level scope
   @ ~/.julia/dev/Catalyst/docs/src/tutorials/pdes.md:91

caused by: BoundsError: attempt to access Num at index [2]
Stacktrace:
 [1] indexed_iterate(I::Num, i::Int64, state::Nothing)
   @ Base ./tuple.jl:97
 [2] grow_to!(dest::Dict{Num, Term{Real, Nothing}}, itr::Vector{Any}, st::Int64)
   @ Base ./dict.jl:153
 [3] grow_to!(dest::Dict{Any, Any}, itr::Vector{Any})
   @ Base ./dict.jl:145
 [4] dict_with_eltype
   @ ./abstractdict.jl:578 [inlined]
 [5] Dict(kv::Vector{Any})
   @ Base ./dict.jl:129
 [6] symbolic_discretize(pdesys::PDESystem, discretization::MOLFiniteDifference{MethodOfLines.CenterAlignedGrid})
   @ MethodOfLines ~/.julia/packages/MethodOfLines/7YJ7n/src/MOL_discretization.jl:132
 [7] discretize(pdesys::PDESystem, discretization::MOLFiniteDifference{MethodOfLines.CenterAlignedGrid})
   @ MethodOfLines ~/.julia/packages/MethodOfLines/7YJ7n/src/MOL_discretization.jl:162
 [8] top-level scope
   @ ~/.julia/dev/Catalyst/docs/src/tutorials/pdes.md:91
```

The PDE equations look OK to me visually, but perhaps I'm messing something up?

isaacsas:
Seems to be working now. pdes.md shows how one can currently create and solve a PDE system from a `ReactionSystem` (without adding spatial transport info into the `ReactionSystem` to enable a `covert` routine).

I got the BPM pattern formation model to work, though the time to get the solution is quite long (maybe 20 minutes? -- I assume this is from compilation taking a long time):

bpm

This just needs tests and should be good to go.

TorkelE:
I wonder if there is some good way to define spatial events that can both be used on a continuous space and on a discrete graph? In https://github.com/SciML/Catalyst.jl/pull/558 I try to define reactions just as normal, but with the extra that it involves substrates and products from two cells. 

First-order reactions seem easy, e.g. diffusion of a single component in continuous space, or a single molecule moving between two neighbouring containers. But not sure if it is easy to generalise higher order stuff like e.g.
`X[1] + Y[2] --> X[2] + Y[2]`
or
`2X[1] + Y[2] --> X2[2] + Y[2]`
(X[1] denotes an X molecule in container one).
to the continuous case.

99% of all cases will likely be single-molecule transport/diffusion though. 

If we could have something like this, it would be possible for someone to define a ReactionSystem, and a set of spatial events, and then from that either generate spatial PDEs, or something on a discrete grid.

isaacsas:
Well, the continuous space generalization for a bimolecular reaction is two molecules react with a probability per time based on their spatial positions. One of my longer-term goals is to get solvers into JumpProcesses for such reactions on lattices, which can be derived as discretizations of the continuous space model. So you’d replace the lattice indices by spatial positions, A(x,t) + B(y,t) —> C(z,t) with a rate k(x,y,z).

In contrast, thinking of spatial transport as a “reaction” is a bit less clear to me in the continuous case.

isaacsas:
If tests pass this should be good to go, and is the minimum perturbation to a `ReactionSystem` that supports symbolically building PDE models. Next we could consider something like @TorkelE did and make an analogous `SpatialReactionSystem` to encapsulate continuous space problems (with the associated converts to make a `PDESystem`).

isaacsas:
@ChrisRackauckas did you want to take a look? Otherwise I’ll merge. Thanks

PR #558, opened 2022-09-30
Lattice reaction network

TorkelE:
Creates a simple interface that distributes a reaction system across a connected graph, and allows reactions between connected nodes.

Currently added MetaGraphs to dependencies.
The plotting interface also required Plots and GraphRecipes.

Definitely things in the code that could be made smoother. Also plan to bring stuff more in line with the rest of MTK, e.g. make `LatticeReactionSystem <: MT.AbstractTimeDependentSystem`, and sort out `SpatialReaction` in some nice way. But I figured it would be good to have something up first.

Example of usage:
```julia
using Catalyst
using OrdinaryDiffEq
using Graphs, MetaGraphs
using Plots

# Make the base network (a brusselator).
rs = @reaction_network base_network begin
    A, ∅ → X
    1, 2X + Y → 3X
    B, X → Y
    1, X → ∅
end A B;

# Declare a set of spatial reactions.
# Rate followed by substrates (compartment1,comaprtment2) and products (compartment1,comaprtment2).
# Stoichiomtery currently given using species=>stoichiometry, but can use somethign better.
srs = [SpatialReaction((@parameters D)[1],([:X=>1],[]),([],[:X=>1]))]

# Creates a graph on which to simualte network (now just a grid)
lattice = MetaGraph(Graphs.SimpleGraphs.grid([20,20]))
# Meta graoh allows setting properties in each node. here :X denotes initial concentration of :X.
# If no :X is set here, whatever X value is used in ODEProblem input will be used.
foreach(v -> (set_prop!(lattice,v,:X,10*rand());set_prop!(lattice,v,:Y,10*rand())), vertices(lattice))

# Bundles everything into a LatticeReactionSystem.
lrs = LatticeReactionSystem(rs,srs,lattice)

# Creates an ODEProblem
u0 =[] #Could set X and Y value, but since all nodes have one, not needed,
tspan = (0.0,100.0)
p = [:D => 0.1, :A=>1.0, :B=>4.0]
oprob = ODEProblem(lrs,u0,tspan,p)

# Simulates the problem.
@time sol = solve(oprob,QNDF(),saveat=0.1)

# The "plot_spatial_sol" function allows us to make an animation.
@time anim = plot_spatial_sol(sol,lattice,2;samp_freq=10,nodesize=0.2)
@time gif(anim, "anim.gif", fps = 2)
```
anim


isaacsas:
I was traveling today, but will take a look tomorrow. Looks cool!

TorkelE:
Stochastic systems work as well (haven't tried SDE yet though):
```julia
dprob = DiscreteProblem(lrs,u0,tspan,p)
@time jprob = JumpProblem(lrs,dprob,RSSACR(),save_positions=(false,false))
@time sol = solve(jprob,SSAStepper(),saveat=0.1)

@time anim = plot_spatial_sol(sol,lattice,2;samp_freq=10,nodesize=0.2)
@time gif(anim, "anim.gif", fps = 10)
```
anim

src/spatial_reactionsystem.jl
```diff
@@ -0,0 +1,142 @@
...
+    u0_full = vcat(map(comp -> map(species -> sym_to_var(species,lrs.rs,comp,lrs.lattice) => get_val(comp,species,lrs.lattice,u0), state_names), vertices(lrs.lattice))...)
+    p_full = vcat(map(comp -> map(species -> sym_to_var(species,lrs.rs,comp,lrs.lattice) => get_val(comp,species,lrs.lattice,p), ps_names), vertices(lrs.lattice))...)
+
+    u0map = symmap_to_varmap(lrs.rs, u0_full)
+    pmap_lattice = symmap_to_varmap(lrs.rs, p_full)
+    pmap_global = symmap_to_varmap(lrs.rs,filter(p_pair->!in(p_pair[1],ps_names), p))
+
+    osys = convert(ODESystem, lrs.rs; name, combinatoric_ratelaws, include_zero_odes, checks,
+                   remove_conserved)
+    return ODEProblem(osys, u0map, tspan, [pmap_lattice;pmap_global], args...; check_length, kwargs...)
```
ChrisRackauckas, line 106: This is missing the connection to the PDE wrapping system. Please make this make use of the growing PDESolution standards.

TorkelE: Yes, for now, I just wanted to make something that works, and then try to integrate it better.

TorkelE: Also, since this is just an ODE/SDE/SSA over connected cells, are there ways to utilise PDE solvers more generally?
(in this example it is just a grid, but that will probably be an exception)

src/spatial_reactionsystem.jl
```diff
@@ -0,0 +1,142 @@
...
+    
+end
+
+### Lattice Reaction Network Structure ###
+# Couples:
+# A network of compartments (a meta graph that can contain some additional infro for each compartment).
+# A reaction network (that is simulated within each compartment).
+# A set of spatial reactions (denoting interaction between comaprtments).
+struct LatticeReactionSystem # <: MT.AbstractTimeDependentSystem # Adding this part messes up show, disabling me from creating LRSs
+    """The reaction system expanded onto a graph."""
```
isaacsas, line 32: Why expand? Why not handle expansions at conversion time if needed? Jumps shouldn't need expansion, and  at some point ModelingToolkit might get a graph ODE system type we could then exploint.

Project.toml
```diff
@@ -1,18 +1,21 @@
...
+GraphRecipes = "bd48cda9-67a9-57be-86fa-5b3c104eda73"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 JumpProcesses = "ccbc3e58-028d-4f4c-8cd5-9ae44345cda5"
 LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
+MetaGraphs = "626554b9-1ddb-594c-aa3c-2596fe9399a5"
 ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
isaacsas, line 18: I'd really prefer not to add dependencies on Plots or Metagraphs, especially Plots. If we want to have visualization utilities perhaps we should put them in a separate (sub?)library with its own dependencies? (Like CatalystPlotting?) (We could move the graph plotting stuff into this library too.)

src/spatial_reactionsystem.jl
```diff
@@ -0,0 +1,142 @@
...
+    rs::ReactionSystem
+    """The original (non-spatial) reaction system that was expanded onto a graph."""
+    rs_base::ReactionSystem
+    """The spatial reactions defined between individual nodes."""
+    spatial_reactions::Vector{SpatialReaction}
+    """The graph on which the lattice is defined."""
+    lattice::MetaGraph
```
isaacsas, line 39: Just use a regular graph? The metagraph is only used to store the initial values right? I think it would be better to use the normal Catalyst/ModelingToolkit mechanisms for storing initial values i.e. 
```julia
@variables t X(t)=[1,2,3]
# or
setdefaults!(rn, [:X => [1,2,3]])
```

 (i.e. just store the initial values vector, with the i'th entry corresponding to the value for the i'th node).

src/spatial_reactionsystem.jl
```diff
@@ -0,0 +1,142 @@
...
+    end
+
+    # Adds the spatial reactions to the initial network.
+    return extend(rs_combined,ReactionSystem(spatial_reactions,rs_combined.iv;name=rs_combined.name))
+end
+
+# Helper functions.
+get_name(comp::Int64,lattice::MetaGraph) = has_prop(lattice,comp,:name) ? get_prop(lattice,comp,:name) : Symbol("container___",comp)
+make_subsystem(rs::ReactionSystem,comp::Int64,lattice::MetaGraph) = @set! rs.name = get_name(comp,lattice)
+sym_to_var(sym,rs,comp,lattice) = rs.var_to_name[Symbol(get_name(comp,lattice),:₊,sym)]
```
isaacsas, line 80: Use `get_var_to_name` instead of direct field access as it is part of the API. Saying `sys.field` is discouraged since `sys.field` is now supposed to represent accessing a symbolic variable/parameter, and not accessing an internal field.

isaacsas (changes requested):
I like the high-level approach of wrapping a `ReactionSystem` with a `LatticeReactionSystem` that has its own `convert`s! I think a similar thing would work well for `SpatialReactionSystem`s or such too.

A few additional high level considerations:
1. Can this work if one makes the default value of a parameter a vector of values (i.e. like a different reaction rate constant at each node)?
2. Similar comment to allow edge-dependent transition rates in the `SpatialReaction`. This is exceptionally common, so we should definitely try to support it from the start.
3. We should keep in mind supporting reactions with non-nearest neighbors on the graph (or otherwise allowing reactions to attach a separate graph with the same nodes but different connectivity). This would be needed for reactions like `A_i + B_j --> C_k` where the interaction depends on a spatial distance or such instead of just being a nearest-neighbor.

Project.toml
TorkelE: I agree with this. For now I put it here so that we could have a tool to visualise the result, and check how it worked. But having these added dependencies for only this I agree is inconvenient.

src/spatial_reactionsystem.jl
TorkelE: Parameter values are also stored, and (and this is not super important) the names of the individual cells/compartments. If we can get away with your notation I agree this is better. We probably can. I want to have some chat to some people that actually work with spatial models and see how what is convenient in the set-up.

src/spatial_reactionsystem.jl
TorkelE: Good catch.

src/spatial_reactionsystem.jl
TorkelE: You mean when converting to ODE/SDE/Jump  systems? Yes, that might be better.

isaacsas: Yes. For example, with jumps we don't want to flatten. That is the whole point of the spatial jump solvers (which are much faster and more memory efficient than flattening).

src/spatial_reactionsystem.jl
isaacsas: I've publishing papers that simulate spatial stochastic models for ~20 years...

isaacsas:
So one thing we need to be consider with having new system types is that it means we'll have to reimplement any compositional modeling type functionality (subsystems, flatten, extend, compose, all the various querying functions like (states, species, reactions, etc)). They'd have to be generalized for spatial systems anyways, but it is something for us to keep in mind.

TorkelE:
Yes. I should be in Boston by the end of the month. Hopefully we could meet up some time to discuss through what's the best way of setting everything up.

TorkelE:
I wanted to have rates dependant on both edge and contabased parameters. The former should be quite easy (will add), but it was not obvious what would be the best way to write a general rate with a mix of parameters from two adjacent cells.

isaacsas:
Yeah, specifying rate expressions is tricky. One really wants a stencil type language to specify them, which I think is available in Symbolics but not really been advertised. In JumpProcesses we just allowed a few factored forms, like `D[s] * L[i,j]` where `s` is a species index and `L[i,j]` could be a discrete Laplacian, or more generally `L[s,i,j]`.

isaacsas:
(above `i` and `j` denote vertices)

TorkelE:
Update:
- MetaGraph Dependency removed.
- Initial conditions/parameters for individual compartments/connections are now designated by passing vectors in u0/p.
- Parameter can now be specified for individual connections. Parameters specified in the initial reaction system are expanded across all vertices, and parameters for the spatial reactions are expanded across all edges.
- Now takes a DiGraph as input, permitting different connections in both directions between cells (this also simplifies the internal generation of the network). Graphs that are inputted are converted to digraphs (additional support for this, e.g. when using graph having the same parameter in both directions) needs to be implemented.
- The expanded reaction network is now never saved, but only generated when an O/S/Discrete/JumpProblem is created.
- Can now designate `xs` and `ys` vectors with the position of all compartments in the plot.

New workflow:
```julia
# Fetch packages
using Catalyst
using OrdinaryDiffEq
using Graphs
using Plots

# Declare the base network (a brusselator), the lattice, and the spatial reactions..
rs = @reaction_network base_network begin
    A, ∅ → X
    1, 2X + Y → 3X
    B, X → Y
    1, X → ∅
end A B;
srs = [SpatialReaction((@parameters D)[1],([:X=>1],[]),([],[:X=>1]))]
lattice = Graphs.SimpleGraphs.grid([20,20])

# Bundle into a LatticeReactionSystem.
lrs = LatticeReactionSystem(rs,srs,lattice)

# Set initial conditions and parameter values.
u0 = [:X => 10*rand(length(vertices(lattice))), :Y => 10*rand(length(vertices(lattice)))]
p = [:A => 1.0, :B => 4.0, :D => 0.2]

# Simulate system.
@time oprob = ODEProblem(lrs,u0,(0.0,100.0),p)
@time sol = solve(oprob,QNDF();saveat=0.1)

# Plot animation.
@time anim = plot_spatial_sol(sol,lattice,2;samp_freq=4,nodesize=0.2)
@time gif(anim, "anim.gif", fps=15)
```
anim

Currently, the 4 runtimes are
```
5.148575 seconds (22.17 M allocations: 1.436 GiB, 5.01% gc time)
3.662192 seconds (2.31 k allocations: 24.032 MiB)
68.967249 seconds (1.17 G allocations: 29.839 GiB, 8.03% gc time, 0.19% compilation time)
2.301520 seconds (507 allocations: 34.781 KiB)
```
with the most intense part being the actual plotting. I think that maybe Makie might be better, since it allows updating the graph, and not replotting each frame in full. I will look into this.

TorkelE:
Given that we have https://github.com/SciML/Catalyst.jl/pull/571, I'm closing this.

PR #565, opened 2022-11-08
Add tutorial for Catalyst and Julia for users inexperienced with Julia

TorkelE:
Currently added as a separate section, to keep it apart from the other tutorials (potentially we could add more sections). However, we might want to add this as just as single page under tutorials.


ChrisRackauckas:
This is looking great.

Don't forget to run the formatter.

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+On the surface, Julia got many similarities to languages like MATLAB, Python, and R. 
+
+*Values* can be assigned to *variables* through the use of a `=` sign. Values (possibly stored in variables) can be used for most basic computations.
+```julia
+length = 2.0
+width = 4.0
+area = length*width
+```
+
+*Functions* takes some input (enclosed by `()`) and returns some output. E.g. the `min` function returns the minimum of two values
```
isaacsas, line 16:
```suggestion
*Functions* take one or more inputs (enclosed by `()`) and return some output. E.g. the `min` function returns the minimum of two values
```

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+Here, `Float64` denotes decimal-valued numbers. Integer-valued numbers instead are of the `Int64` type.
+```julia
+typeof(1)
+```
+
+Finally, we note that the first time some code is run in Julia, it has to be *compiled*. However, this is only required once per Julia session. Hence, the second time the same code is run, it runs much faster. E.g. try running this line of code first one time, and then one additional time. You will note that the second run is much faster.
+```julia
+(245^0.43) + exp(0.431) - log(124541) * tan(1.43) - sin(4.3);
```
isaacsas, line 38: This is pretty instantaneous for me on 1.8.2 even with compilation. Maybe there is another example that has a more clear pause you could use?

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+This is useful to know when you e.g. declare, simulate, or plot, a Catalyst model. The first time you run the command there might be a slight delay. Howegver, the second run will execute much quicker. This holds even if you do minor adjustments before the second run (such as changing simulation initial conditions).
```
isaacsas, line 41:
```suggestion
This is useful to know when you e.g. declare, simulate, or plot, a Catalyst model. The first time you run a command there might be a slight delay. However, subsequent runs will execute much quicker. This holds even if you do minor adjustments before the second run (such as changing simulation initial conditions).
```

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+
+The `@reaction_network` command is followed by the `begin` keyword, which is followed by one line for each *reaction* of the model. Each reaction consists of a *reaction rate*, followed by the reaction itself. The reaction itself contains a set of *substrates* and a set of *products* (what is consumed and produced by the reaction, respectively). These are separated by a `-->` arrow. Finally, the model ends with the `end` keyword, followed by a list of its *parameters*.
+
+Here, we create a simple *birth-death* model, where a single species (*X*) is created at rate *b*, and degraded at rate *d*. The model is stored in the variable `rn`.
+
+```julia
+rn = @reaction_network begin
+    b, 0 --> X
+    d, X --> 0
+end b d;
```
isaacsas, line 84: Drop the `;` to show the output?

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+For more information on how to use the Catalyst model creator, please read [the corresponding documentation](https://docs.sciml.ai/Catalyst/stable/tutorials/dsl/).
+
+Next, we wish to simulate our model. To do this, we need to provide some additional information to the simulator. This is
+* The initial condition. That is the concentration of each species at the start of the simulation.
```
isaacsas, line 90:
```suggestion
* The initial condition. That is the concentration or number of each species at the start of the simulation.
```

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+* The timespan. That is, over the timeframe over which we wish to run the simulation.
```
isaacsas, line 91:
```suggestion
* The timespan. That is, the timeframe over which we wish to run the simulation.
```

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+* The parameter values. That is, the values of the model's parameters for this simulation.
+
+The initial condition is given as a *Vector*. This is a type which combines several different values. When declared, the values are contained by `[]`, and separated by `,`. Since we only have one species, the vector holds a single element. In this element, we set the value of *X* using the `:X => 1.0` syntax. Here, we first denote the name of the species, with a `:` pre-appended. Next follows `=>` and then the value of *X*. Since we wish to simulate the *concentration* of X over time, we will let the initial condition be decimal valued.
```
isaacsas, line 94:
```suggestion
The initial condition is given as a *Vector*. This is a type which collects several different values. To declare a vector, the values are specific within brackets, `[]`, and separated by `,`. Since we only have one species, the vector holds a single element. In this element, we set the value of *X* using the `:X => 1.0` syntax. Here, we first denote the name of the species, with a `:` pre-appended. Next follows `=>` and then the value of *X*. Since we wish to simulate the *concentration* of X over time, we will let the initial condition be decimal valued.
```

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+```julia
+sol = solve(oprob);
+```
+
+Finally, we can plot the solution through the `plot` function.
+```julia
+plot(sol)
+```
+
+Here, the plot described the time evolution of the concentration of the species *X*, from its initial condition.
```
isaacsas, line 128:
```suggestion
Here, the plot shows the time evolution of the concentration of the species *X* from its initial condition. 
```

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+Finally, we can plot the solution through the `plot` function.
+```julia
+plot(sol)
+```
+
+Here, the plot described the time evolution of the concentration of the species *X*, from its initial condition. 
+
+For more information about the numerical simulation package, please see the [DifferentialEquation documentation](https://diffeq.sciml.ai/stable/). For more information about the plotting package, please see the [Plots documentation](https://docs.juliaplots.org/stable/).
```
isaacsas, line 130:
```suggestion
For more information about the numerical simulation package, please see the [DifferentialEquation documentation](https://docs.sciml.ai/DiffEqDocs/stable/). For more information about the plotting package, please see the [Plots documentation](https://docs.juliaplots.org/stable/). 
```

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+
+We declare the model using the `reaction_network` macro, and store it in the `sir_model` variable.
+```julia
+sir_model = @reaction_network begin
+    b, S + I --> 2I
+    k, I --> R 
+end b k;
+```
+
+Here we note that the first reaction contains two different substrates (separated by a `+` sign). While there is only a single product (*I*), two copies of *I* are produced. The *2* in front of the product *I* denotes this.
```
isaacsas, line 156:
```suggestion
Note that the first reaction contains two different substrates (separated by a `+` sign). While there is only a single product (*I*), two copies of *I* are produced. The *2* in front of the product *I* denotes this.
```

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+params = [:b => .2, :k => 1.0];
+```
+
+Previously we have bundled this information into an `ODEProblem` (denoting a deterministic *ordinary differential equation*). Now we wish to simulate our model as a jump process (where each reaction event denotes a single jump in the state of the system). We do this by first creating a `DiscreteProblem`, and then using this as an input to a `JumpProblem`.
+```julia
+dprob = DiscreteProblem(sir_model,u0,tspan,params)
+jprob = JumpProblem(sir_model,dprob,Direct());
+```
+
+Again, the order in which the inputs are given to the `DiscreteProblem` and the `JumpProblem` is important. The last argument to the `JumpProblem` (`Direct()`) denotes which simulation method we wish to use. For now, we recommend the user simply use the `Direct()` option, and then consider alternative ones when they are more familiar with modelling in Catalyst and Julia.
```
isaacsas, line 171:
```suggestion
Again, the order in which the inputs are given to the `DiscreteProblem` and the `JumpProblem` is important. The last argument to the `JumpProblem` (`Direct()`) denotes which simulation method we wish to use. For now, we recommend the user simply use the `Direct()` option, and then consider alternative ones, see the [JumpProcesses.jl docs](https://docs.sciml.ai/JumpProcesses/stable/), when they are more familiar with modelling in Catalyst and Julia.
```

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
...
+
+Finally, we can simulate our model using the `solve` function, and plot the solution using the `plot` function. Here, the `solve` function also has a second argument (`SSAStepper()`). This is a stepping algorithm. Again we recommend that you at this stage simply use this option, and then explore further exactly what this means at a later stage.
```
isaacsas, line 173:
```suggestion
Finally, we can simulate our model using the `solve` function, and plot the solution using the `plot` function. Here, the `solve` function also has a second argument (`SSAStepper()`). This is a time stepping algorithm that calls the `Direct` solver to advance a simulation. Again, we recommend at this stage you simply use this option, and then explore further exactly what this means at a later stage.
```

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
```diff
@@ -0,0 +1,182 @@
+# Introduction to Catalyst and Julia for new Julia users
+The Catalyst tool for the modelling of chemical reaction networks is based in the Julia programming language. While experience in Julia programming is advantageous for using Catalyst, it is not necessary for accessing some of its basic features. This tutorial serves as an introduction to Catalyst for those unfamiliar with Julia, also introducing some basic Julia concepts. Anyone who plans on using Catalyst extensively is recommended to familiarise oneself more thoroughly with the Julia programming language, e.g. by reading [its documentation](https://docs.julialang.org/en/v1/).
```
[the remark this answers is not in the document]
TorkelE: Yes, that one should be linked.

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
TorkelE: I was going to do something with big matrix multiplication, but I also wanted something where it was clear to the user what code was being run. I will have a think over the day, might do matrix multiplication, but point out that that is what is happening.

docs/src/catalyst_for_new_julia_users/intro_to_catalyst.md
TorkelE: For some reason I got an error when I tried to display the network on my machine, so added this one here. Generally, I had several `;` to avoid clogging up the docs. But adding this one back in makes sense, and maybe I should do that for more blocks.

TorkelE:
I'm probably missing something with the formatter. I opened a terminal on my computer, and then ran:
```julia
]add JuliaFormatter
using JuliaFormatter, Catalyst
format(joinpath(dirname(pathof(Catalyst)), ".."))
```
but nothing seemed to happen.

Next, I went to the Catalyst package, did some changes, didn't commit, re-run the formatted thing, and then committed. However, this didn't seem to change anything.

The package contains both a ".JuliaFormatter.toml" and a "FormatCheck.yml" file.

isaacsas:
Maybe you need to use `format_md`? See https://domluna.github.io/JuliaFormatter.jl/dev/api/#JuliaFormatter.format_md-Tuple{AbstractString}

TorkelE:
Tried various versions using that, and also `format_file`, running both the from a Julia terminal and VSCode, but nothing works yet. This is a standard workflow we use for all packages?

isaacsas:
You might try tweaking a `src/.jl` file to have something like `f(x,y)` instead of `f(x, y)` and see if it catches this. If it doesn't something is wrong with your setup.

ChrisRackauckas:
The .md stuff is just a keyword argument that is disabled by default because it's very broken. But because it's disabled it wouldn't cause a test failure.

isaacsas:
It was `pages.jl` that was not formatted. I just formatted it and pushed the change to this PR.

isaacsas:
@TorkelE feel free to merge when you are done given that tests are now passing.

PR #568, opened 2022-11-10
fix local doc builds, use joinpath, and tweak tutorial

isaacsas:
@ArnoStrouwen does this look reasonable to you? I was having issues building docs locally unless I was in a specific directory, and also noticed this was using hardcoded unix-style paths instead of `joinpath`.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,8 +1,10 @@
 using Documenter
 using Catalyst, ModelingToolkit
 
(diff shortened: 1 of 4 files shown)


ArnoStrouwen:
Yes, it outputs the correct link:
https://github.com/SciML/Catalyst.jl/blob/ef6a1e76cb20613ad999863c7f68bf01a14cd0d0/previews/PR568/index.html#L401

isaacsas:
Thanks for taking a look!

PR #572, opened 2022-12-15
Redo dsl syntax

TorkelE:
Start new syntax for DSL. Many tests and the docs still need fixing, and updating history files. Need to add some new tests. Allowing default values still to be added as well.

### Parameters are no longer listed at the end of the macro
These are inferred directly from the network. Anything appearing in a reaction (as substrate/product) is assumed to be a species, everything else a parameter.

### Species and parameters can now be listed at the beginning of the macro
using this syntax:
```julia
rn = @reaction_network begin
@species X Y
@parameters A
    A, ∅ → X
    1, 2X + Y → 3X
    B, X → Y
    1, X → ∅
end
```

Furthermore, default is planned like this:
```julia
@species X=1 Y=5.0 
@parameters A
    A, ∅ → X
    1, 2X + Y → 3X
    B, X → Y
    1, X → ∅
end
```


ChrisRackauckas:
Make sure to test sub-blocking syntax:

```julia
rn = @reaction_network begin
    @species begin
        X=1 
        Y=5.0 
    end
    @parameters begin
        A
    end
    A, ∅ → X
    1, 2X + Y → 3X
    B, X → Y
    1, X → ∅
end
```

isaacsas:
Can we make `@species` a standalone macro like `@variables` and `@parameters`? Then it should just be callable within the `@reaction_network` macro if this is setup correctly I think. It should really just be equivalent to calling `@variables` with the same exact expression, and then adding an `isspecies` metadata tag to each generated variable. Then one could say
```julia
@variables t
@species A(t) B(t)
@parameters k
rx = reaction(k, [A], [B])
...
```
(for example). Moreover, then the DSL should be able to just call this macro internally right? (i.e. not really need any special handling.)

I think the DSL should then be able to be reduced so that
```julia
sys = @reaction_network SysName begin
    @species A(t) B(t)
    @parameters k k2
    k, A --> B
    k2, B --> A
end
```
simply generates something equivalent to following as the macro output (maybe wrapped in begin/end or such):
```julia
@species A(t) B(t)
@parameters k k2
rxs = [(@reaction $k, $A --> $B), (@reaction $k2, $B --> $A)]
sys = ReactionSystem(rxs, t, [A,B], [k, k2]; name=:SysName)
```

isaacsas:
@ChrisRackauckas you beat me to that comment :)

ChrisRackauckas:
jinx poke you owe me a coke.

TorkelE:
Added check so that
```julia
rn = @reaction_network begin
    @species begin
        X=1 
        Y=5.0 
    end
    @parameters begin
        A
    end
    A, ∅ → X
    1, 2X + Y → 3X
    B, X → Y
    1, X → ∅
end
```
should be handled (but need to add some tests of it as well).

I have created a `@species` macro, while the metadata is not yet fixed, this is currently used by the DSL. We kind of still have the extract this information (because when we parse parameters, if these are not given by the users, we need it to determine what is a parameter). In the end we then just pluf the same information into the output (so we don't have to have special cases for whenever users give `@species` or no). But the result is the same.

test/api.jl
```diff
@@ -79,9 +79,10 @@ addparam!(rs, k1, disablechecks = true)
 @test numreactionparams(rs) == 3
 
 rnmat = @reaction_network begin
+@parameters α β
```
isaacsas, line 82: I thought that now 
```julia
rnmat = @reaction_network begin
    α, S + 2I --> 2I
    β, 3I --> 2R + S
end
```
should work too? Why is the `@parameters` line needed?

TorkelE: I had a couple of 100 of these to change. In almost all cases adding the @parameters thing shouldn't matter, but I didn't want to look closer on all, and still risk messing up. I just went through all and replaced with a similar @paramters option.

I am adding additional tests to show that not having these (which should be what most people do) also work. But yes, had I written the tests from scratch I would not have had the `@parameters` part.

isaacsas:
We should also make sure to add new tests for the basic form of the DSL, and for the purely symbolic form using the standalone `@species` macro. (We can build two reaction systems with the same name via DSL and symbolically, and just check for equality.) We should also check that a basic network like
```julia
rn = @reaction_network
    k*C, D*A  --> B
    k2,  g*C --> D
end
```
gives that A, B, C and D are all species and k, k2, and g are parameters.

Also, just a general comment. It is very helpful for reviewers to avoid moving around existing functions too much. There seem to be a bunch of helper functions that have moved in `reaction_network.jl`, which makes it tough to see if you deleted them and/or you modified them.

TorkelE:
> Also, just a general comment. It is very helpful for reviewers to avoid moving around existing functions too much. There seem to be a bunch of helper functions that have moved in reaction_network.jl, which makes it tough to see if you deleted them and/or you modified them.

That's a good point. Yes, I did some reordering when trying to figure out which thing would go were, but you'r right that might have created a mess.

I've added some tests of the macro with/without `@parameters`/`@species` to check, but might add some more. I will let the tests run now to see if there are any errors so far though.

TorkelE:
I have added the option to set defaults now as well (however, need to add tests where these are used in simulations).

isaacsas:
What do you mean by set defaults?

TorkelE:
like in 
```
rn = @reaction_network begin
    @species begin
        X=1 
        Y=5.0 
    end
    @parameters begin
        A
    end
    A, ∅ → X
    1, 2X + Y → 3X
    B, X → Y
    1, X → ∅
end
``` 
the default initial condition values of X is 1 and of Y is 5.0. Basically from the Slack discussion last week

isaacsas:
But if you don’t reuse/wrap `@variables` then you drop support for all the various features it supports, like metadata in the macro:
```julia
@variables t A(t)=1 [isspecies=true] B(t)=2 [isconstant=true]
```

TorkelE:
No, so we don't actually just copy past the `@paraneters` lines. E,g, if we have
```julia
@reaction_network begin
@parameters p=1.0
     p, 0 --> X
end
```
we extract that the parameters ar `p` from the first line, create a new line
```
@parameters p
```
and then create the reaction network
```
@species X
ReactionSystem(...)
```

In the case where we have e.g.
```julia
@reaction_network begin
@species X=1.0
     p, 0 --> X
end
```
it seems a bit silly that we don't just re-use the first line. However, we have to internally extract a vector with all the species anyway (when we figure out what should be a parameter and what not). And we need the code to create a new
```
@species X
```
for the case where the user don't declare it. So in the end we don't actually need to use the fact that
```
@species X
```
is a perfectly valid notation outside the macro as well. I was experimenting with something like:
```
@reaction_network begin
species: X
     p, 0 --> X
end
```
to make this more clear, but right now kept it like this.

TorkelE:
Ok, this should be done now, including doc changes.

There is one small detail in the /api.jl tests @isaacsas if you could check. I made changes to this:
```julia
# constant and BC species test
Z = [2 0 1 0 0 0 0;
0 1 0 0 0 1 0;
0 0 0 1 0 0 0;
0 0 0 0 1 1 0;
0 0 0 0 0 0 1;
0 0 0 0 0 0 0]
@parameters F [isconstantspecies = true]
crn = @reaction_network begin
    @species A B C D E F
    k₁, 2A --> B
    k₂, A --> C + $F
    k₃, C --> D
    k₄, B + D + $F --> E
end
testnetwork(crn, B, Z, Δ, lcs, 0, subrn, lcd; skiprxtest = true)
```
to prevent things to go error. Basically F was detected as a species a parameter on a species place, and an error was thrown. I had to make it as a species and change the `Z` matrix correspondingly.

Problem is that I am not fully familiar with how constant species are handled (nor how we want to handle them), so I am not really sure how this is actually intended to work.

TorkelE:
But when we think that is OK, things should be ready to go.

TorkelE:
The codecov looks bad, bu that's because I left a couple of old version functions as the end (of the reaction_network.jl file) that are not used anymore. I am happy to delete these.

isaacsas:
I’m not sure I understand the problem you had, but the point of interpolating external species, parameters, or expressions is to avoid having to declare anything in the macro. 

I’ll take a closer look and review Friday or this weekend; I’m a bit busy with the start of classes the next couple of days.

TorkelE:
So, the interpolation was find, there are several other cases where it worked. The problem was that you try to put in a `@parameters F [isconstantspecies = true]` in a reaction, and things got a bit confused about how to treat this parameter/species. Had `F` just been a normal species it would have been fine.

isaacsas:
This should work though at the moment, so it seems like something in the DSL changes are breaking it.

TorkelE:
Yes, I think it is, but while I think I know what `@parameters F [isconstantspecies = true]` accomplishes, I am not really sure. I am not really sure how it works in MTK either. Hence I am not really sure how to handle it.

If it is non-trivial we can just discuss it next time we meet.

isaacsas:
It creates a parameter with the metadata `itconstantspecies` set to `true`. Internally we then tag this as something that is a parameter, but should be treated liked a species when building rate laws for ODEs/SDEs and jumps. Take a look in `reactionsystem.jl` for where `isconstant(...)` is called to see how it is used. It is a feature of SBML models that one can have a species that is just a constant concentration, and valid to use in rate laws and reaction stoichiometry.

TorkelE:
Think I got it. Hopefully, this update should do it fix it.

isaacsas:
@TorkelE I just played around a little, but this seemed like it might allow us to just wrap `@variables` in a new `@species` maco and in that way maintain feature parity with it. Note that it should extract the symbols of all the variables in `syms` too as you mentioned you need. Maybe we could make an internal function that handles this expression manipulation, and then expose it via `@species` as both a standalone macro and in the DSL (the DSL, needing more info, can use the internal function that `@species` calls).
```julia
macro species(ex...)
    vars = Symbolics._parse_vars(:variables, Real, ex)
    syms = vars.args[end].args
    lastarg = vars.args[end]
    resize!(vars.args, length(vars.args)-1)
    for sym in syms
        push!(vars.args, :($sym = setmetadata($sym, ModelingToolkit.MTKConstantCtx, true)))
    end
    push!(vars.args, lastarg)
    esc(vars)
end
```

isaacsas:
Note that ModelingToolkit also calls `Symbolics._parse_vars`, so I think we can consider that a function that is part of the API (at a minimum if it changes ModelingToolkit would need changes too).

TorkelE:
I am not sure I fully understand what you mean. Would this be a change in the intended functionality, or just a redesign of how it is achieved? 

Also, would you have the opportunity to meet up sometime in the coming weeks?

isaacsas:
The goal would be for `@species` to work both in the DSL and outside the DSL, and to be functionally equivalent to `@variables` but just append extra metadata to the species. Right now, the version in the DSL only allows setting initial conditions right? You can't set metadata or use array variables, or any of the other features `@variables` supports (or have I misunderstood?).

TorkelE:
No, you cannot. Wasn't really aware of the additional features of `@variables` in addition to simply creating variables. But yes, getting this functionality seems useful. The `@species` macro should already be created (although it currently is identical to `@variables`, so that would need fixing). 

If there's some additional use of it, making the DSL simply run the input @parameters/:variables, then that is not hard to make.

isaacsas:
Yes, I think the DSL's `@parameters` and `@species` should just call the standalone versions to fully support their features and not have a split code-base. (With `@species` then wrapping `@variables` as discussed.)

Can you point me to where the standalone `@species` macro is defined? I keep missing it when searching for some reason...

TorkelE:
Ok, new update. No we simply repaste the `@species` and @parameters` input strings. For the `@parameters` one it is simple, for the `@species` one there are two caveats:
1. The `@species` macros does not work, unsure why. So currently `@species` is replaced with `@variables` (should be changed when we fix `@species`).
2. In the `@species` option, the time variable `t` is not used, so the code has to be modified by adding `t` back in.

Hence, `@species X Y=1.0` is transformed into `@species t X(t) Y(t)=1.0`

TorkelE:
I couldn't actually find it either, so presume that it was never added in the first place (I think I trialed it and probably took it out or something).

I have added it in at line 87 of the "reaction_network.jl" file:
```julia
### The @species macro, currently just passing to the @varriables macro. ###
# Currenrtly not working.
macro species(args...)
    return :(@variables $(args...))
end
```

TorkelE:
The main cases are fixed. I need to make changes so that we in e.g. 
```julia
rn = @reaction_network begin
    @species X(t) A(t)
    k*X, A + B --> 0
end
```
can infer that B is a parameter. This will require some additional changes though.

TorkelE:
Given that specie and parameters are handled very similarly now (they both can suck up all undeclared symbols, given a list of the other), the underlying functions changed a bit. These functions are now in the "delete" section:
`find_parameters_in_expr!`, `extract_species`, `extract_parameters`

isaacsas:
I've been wiffle-waffling on this since we spoke yesterday. Do you think it might alternatively make sense to always use the behavior when no `@parameters` or `@species` are specified for unidentified symbols? i.e. if a symbol is not declared in `@species` or `@parameters` and it appears in a rate it is a parameter, if it appears on the reaction part of the line it is a species. 

PS - What is the tie-breaking behavior if something appears in both places but is undeclared (say a rate in one reaction but a substrate in another)? I guess make it a species?

TorkelE:
Tie-break is making it a species. E.g. this becomes a species:
```julia
@reaction_network begin
    X, X --> 0
end
```

Actually, just an hour again I was thinking exactly the same (but ended up continuing with the current plan because I didn't want to mess with things when one version was almost completed). It would (if we are thinking of the same) allow you to do e.g.
```julia
@reaction_network
    @species P
    k1*P, 0 --> X1
    k2, 0 --> X2
    k3, 0 --> X3
    k4, 0 --> X4
    k5, 0 --> X5
    k6, 0 --> X6
end
```
without having to give the full set, e.g.
```julia
@reaction_network
    @species P X1 X2 X3 X4 X5 X6
    k1*P, 0 --> X1
    k2, 0 --> X2
    k3, 0 --> X3
    k4, 0 --> X4
    k5, 0 --> X5
    k6, 0 --> X6
end
```
which I think makes sense. One advantage of giving all was that then ordering for the full set is provided, but we have agreed that this is not important.

Currently, I am leaning towards agreeing with your (although we would need to ensure we mean the same thing!). I will ponder more over it over lunch.

isaacsas:
Happy to Zoom this afternoon if you want to make sure we're on the same page (I'm free till 4pm). Just hit me up on Slack.

TorkelE:
The question is whenever something like
```julia
@reaction_network begin
    @parameters p
    p*X, A --> 0
end
```
should have species `[X, A]` or `[A]`.

Yes, let's zoom.

src/reactionsystem.jl
```diff
@@ -606,8 +606,8 @@ function make_ReactionSystem_internal(rxs::Vector{<:Reaction}, iv, no_sps::Nothi
             push!(ivs, value(siv))
         end
     end
-    sts = OrderedSet()
-    ps = OrderedSet{Any}(ps_in)
+    sts = OrderedSet{eltype(sts_in)}(sts_in)
```
isaacsas, line 609: @torkel Does this seem reasonable to you instead of what you had with checking if sts_in is set equal to sts? It should still preserve the order of sts_in and seems simpler (and more in line with how we handle ps_in).

src/reaction_network.jl
```diff
@@ -203,144 +257,196 @@ struct ReactionStruct
...
+function extract_syms(ex::Expr, vartype::Symbol)
+    vars = Symbolics._parse_vars(vartype, Real, ex.args[3:end])
+    Vector{Union{Symbol, Expr}}(vars.args[end].args)
 end
 
-function find_species_in_rate!(sset, rateex::ExprValues, ps)
+# Function looping through all reactions, to find undeclared symbols (species or
+# parameters), and assign them to the right category.
+function extract_species_and_parameters!(reactions, excluded_syms)
+    species = OrderedSet{Union{Symbol, Expr}}()
```
isaacsas, line 376: I changed this and parameters below to `OrderedSet`s. While for small networks this will likely be a bit slower, the DSL is so fast it shouldn't be noticeable. In contrast, I think this should help performance for larger networks. 

LMK if this is ok with you, or you had a reason to switch the older code to `Vector`s.

isaacsas: See also the tweaks to `add_syms_from_expr!`.

src/reaction_network.jl
```diff
@@ -203,144 +257,196 @@ struct ReactionStruct
...
+                       option_lines))
+
+    # Parses reactions, species, and parameters.
+    reactions = get_reactions(reaction_lines)
+    species_declared = haskey(options, :species) ?
+                       extract_syms(options[:species], :species) : Union{Symbol,Expr}[]
+    parameters_declared = haskey(options, :parameters) ?
+                          extract_syms(options[:parameters], :parameters) :
+                          Union{Symbol,Expr}[]
+    declared_syms = Set(Iterators.flatten((parameters_declared, species_declared)))
```
isaacsas, line 291: Made this a `Set` since we are repeatedly checking if something is in it.

isaacsas (commented):
I'm done with the docs and library code changes now. I'll look over the tests tomorrow and then merge if they look good.

src/reaction_network.jl
TorkelE: makes sense

src/reaction_network.jl
TorkelE: Think this should be good. If I remember next time we meet I will have to ask you why this improves stuff and I might learn something.

isaacsas: To be honest, I would have preferred to use a `Set` but that is breaking lots of tests and I don't want to handle updating them in this PR. (`Set` at least used to have better performance than `OrderedSet`.)

src/reactionsystem.jl
TorkelE: I think this sounds good. The tests should check that it works as well.

isaacsas:
@TorkelE ok, all done. If tests pass feel free to merge!

isaacsas:
OK, now all done. We can tweak docs / history and such later.

isaacsas:
We also have to remember to update the README before making a new release.

chriselrod:
`add_reaction` no longer supports parameters arguments; what's the correct way to update?

TorkelE:
I *think* that parameters should automatically be detected according to rules (e.g. stuff that appears as substrate/product is inferred as a species, everything else a parameter).

If you have some other cause, e.g. where you want to have a parameter substrate (or more likely, a rate that is a species) you will have to specify that elsewhere, probably when you initially declare the network.

Might we have a problematic case here somewhere?

chriselrod:
I don't really know this code, so I may try and leave this to someone else to update.
I can't test simply dropping the parameters, because then I get
```julia
ERROR: LoadError: UndefVarError: `@species` not defined
```

TorkelE:
sorry, do you have an example? might be that I don't really know what is going on.

chriselrod:
Sorry, I just looked into it a bit further. I just needed to import `Catalyst.@species`.

PR #577, opened 2023-01-15
[skip ci] doc compats

ArnoStrouwen:
Why is this compat helper script much more complicated than e.g [LinearSolve.jl](https://github.com/SciML/LinearSolve.jl/blob/main/.github/workflows/CompatHelper.yml )?
Would any of the other SciML packages benefit from the extra complexity here?


ChrisRackauckas:
What do you mean?

ArnoStrouwen:
The LinearSolve CompatHelper script is 16 lines, the script here is 44 lines. I wonder if it is doing something extra?

PR #579, opened 2023-01-18
Reorganise Project.tonl files

TorkelE:
This changes to main Project.toml file (to not include test stuff).

I also adds a Project.toml file under /test/.


isaacsas:
Is it necessary to duplicate the Catalyst dependencies and compat, or can the test project.toml just include the test specific dependencies?

TorkelE:
Not sure to be honest. If you are not sure either then I can ask around. I agree this would make sense.

TorkelE:
Hopefully, we will be able to do this by 1.10
https://discourse.julialang.org/t/link-project-toml-files-or-make-test-prokect-toml-depend-on-main-package-one/93223/3

isaacsas:
Based on that thread it sounds like we should maybe hold off till 1.10 on making this change? Otherwise we will have to keep remembering to update both toml files anytime there is a dependency version update or change.

TorkelE:
Yeah, I think the advantage and effort are minor in all cases. But holding it avoids this possibility of messing stuff up.

PR #580, opened 2023-01-19
Enable `=>` and `<=` to ignore law of mass action in DSL

TorkelE:
Since these can be written on a normal keyboard, `=>` and `<=` are now enabled when ignoring the law of mass action and only using the rate. I've added tests and updated the docs, should all should be ready.

I would have added `<=>` as well, but unfortunately `X <=> X` is not a valid Julai expression. If we are really keen on this we could try to moot for this (like `<-->` and `<--`, although these were way more important additions).


isaacsas (approved):

isaacsas:
Good catch. We should have thought of this a while ago!

PR #585, opened 2023-01-27
add and use metadata for species

isaacsas:
This will require changing most `@variables` to `@species` across Catalyst.

TODO still:
- [x] Check `dependants` usage or if it needs updates now.
- [x] Check independent states calculation makes sense with mixed states and species still.
- [x] Double check all usage of `states` or `get_states` is still correct or should be changed to species versions.
- [x] Check rest of API usage.
- [x] Update docs.
- [x] reenable jump retcode tests when new JumpProcesses is available


isaacsas:
Macro is going to be broken for array variables until https://github.com/JuliaSymbolics/Symbolics.jl/issues/842 is fixed. So I can finish and merge, but if we release nobody will be able to use array variables in Catalyst...

PR #588, opened 2023-02-03
Starting documentation reordering

TorkelE:
Starting the doc reorganisation. No major rewriting stuff yet. Current changes:

New sections:
- Introduction to Catalyst (the previous intro for new users. Contain the new user intro and have moved "using Catalyst" doc here.
- Catalyst Functionality (Name pending, basically stuff implemented in Catalyst).
- Tutorials (same as before, using Catalyst in the context of other stuff).

Other changes:
- I have added a folder under docs "Unpublished", where I am putting all tutorials currently not shown in the docs (currently only pdes).
- have renamed several of the documentation files so that it is easier to correlate the file with the tutorial. E.f. "generating_reactions_programmatically" is now called "smoluchowski_coagulation_equation.md".


docs/pages.jl
```diff
@@ -1,15 +1,24 @@
...
+                            ],
+            "Catalyst Functionality" => Any[
+                                "catalyst_functionality/dsl.md",
+                                "catalyst_functionality/programmatic_CRN_construction.md",
+                                "catalyst_functionality/basic_CRN_examples.md",
+                                "catalyst_functionality/compositional_modeling.md",
+                                "catalyst_functionality/parametric_stoichiometry.md"
+                            ],
+            "Tutorials" => Any[
+                                "tutorials/network_analysis.md",
```
isaacsas, line 14: I think this should be in "Catalyst Functionality" since that is what it provides. 

Maybe call "Tutorials" something like "Advanced Tutorials" or "Catalyst Applications"? Just so users don't think it is beginner material?

TorkelE:
Currently, API is divided into subsections, but there's only one. Now I have removed this and it appears like the FAQ and Home.

Most changes I make here are independent, so if there's anything we don't like we could revert it.

TorkelE:
I have now added links for all main headers in each doc file, the reference id is the same as the file name. I have also found some references to `using Catalyst` and replaced these.

TorkelE:
This proposed change is slightly bigger, I think we discussed it briefly only.

It creates a new section for example networks, and moves to doc pages there:
```julia
            "Catalyst Applications" => Any[
                                "tutorials/network_analysis.md",
                                "tutorials/homotopy_continuation.md",
                                "tutorials/bifurcation_diagrams.md",
                                "tutorials/parameter_estimation.md",
            
            "Example Networks" => Any[
                                "example_networks/basic_CRN_examples.md",
                                "example_networks/smoluchowski_coagulation_equation.md"
                            ],
            "FAQs" => "faqs.md",
```

TorkelE:
I will try to keep this PR to just organisation and re-categorization stuff. Then I will create different ones for actually expanding the docs.

docs/src/intro_to_catalyst/using_catalyst.md
```diff
@@ -110,7 +110,7 @@ conditions. We can build such mappings in several ways. One is to use Julia
 ```@example tut1
 pmap  = (:α => .5, :K => 40, :n => 2, :δ => log(2)/120,
          :γ => 5e-3, :β => log(2)/6, :μ => log(2)/60)
-u₀map = [:m₁ => 0., :m₂ => 0., :m₃ => 0., :P₁ => 20., :P₂ => 0., :P₃ => 0.]
+u₀map = [:m₁ => 0., :m₂ => 0., :m₃ => 0., :P₁ => 20., :P₂ => 0., :P₃ => 0.];
```
isaacsas, line 113:
```suggestion
u₀map = [:m₁ => 0., :m₂ => 0., :m₃ => 0., :P₁ => 20., :P₂ => 0., :P₃ => 0.]
nothing   # hide
```
Instead of adding a `;` which will appear in the docs, I think you can do the above. Give it a try locally and check if that effectively suppresses the output.

isaacsas (dismissed):
@TorkelE this looks good, except please move "tutorials/network_analysis.md" to under "Catalyst Functionality" since it is part of the functionality we provide with Catalyst. I left you one minor comment that should allow us to avoid having semicolons in the documentation.

TorkelE:
Thanks,
```
nothing   # hide
```
works, I am using it now.

I think these are all the changes for this PR. However, I currently haven't gotten the docs to build properly, I think that, at least when I tried locally, it used version 12... of catalyst, not 13, causing stuff crash that depends on 13.

isaacsas:
Since you are doing this as a branch on Catalyst master instead of in your local fork, the docs should build as part of CI and you can then preview them from the deployment check.

TorkelE:
The error seems to be a 
```
ERROR: empty intersection between Catalyst@13.0.0-DEV and project compatibility 12.3.0-12
```
Should I just update the 
```
Catalyst = "12.3"
```
line under docs/Project.toml
to
```
Catalyst = "13"
```
?

However, this didn't seem to cause any error when we did the last update to change the dsl, not sure why it is causing problems now.

isaacsas:
Probably? Never actually thought about that issue before. Hopefully that fixes it. Might as well bump it now as we'll have to bump it when we make release 13 anyways.

TorkelE:
Tried locally as well, but still get the same problem. All doc boxes that depend on the 13 changes fail.

isaacsas:
I guess you could try changing 13.0.0-DEV to 13.0.0 and see if that helps. Otherwise I'm not sure.

TorkelE:
Ok, this seems to be working now.

Do you want to merge and continue doc changes from here?

isaacsas:
Yes, go ahead and merge if you are done!

PR #589, opened 2023-02-06
Add `form` option to latexify

TorkelE:
This add an option form to `latexify(rs::ReactionSystem)` so that
```julia
using Catalyst, Latexify

rn = @reaction_network begin 
    (p,d), 0 <--> X
    (kB,kD), 2X <--> X2
end

latexify(rn; form=:ode)
```
returns the ODE form and 
```julia
latexify(rn; form=:sde)
```
returns the side form.

This saves us from having to do 
```julia
latexify(convert(ODESystem,rn)
```
all the time (especially as the ODE form is probably the most common thing one wants to do when doing latexify).

This will also save us from having to do this in the doc, so we can move to introduce ODESystem forward (in case we want to).

### DIFF
--- a/src/latexify_recipes.jl
+++ b/src/latexify_recipes.jl
@@ -176,13 +176,19 @@ function chemical_arrows(rn::ReactionSystem; expand = true,
     return latexstr
 end
 
-@latexrecipe function f(sys::ReactionSystem)
-    # Set default option values.
-    env --> :chem
-    cdot --> false
-
-    return sys
+@latexrecipe function f(rs::ReactionSystem; form=:reactions)
+    if form==:reactions    # Returns chemical reaction network code.
+        cdot --> false
+        env --> :chem
+        return rs
+    elseif form==:ode      # Returns ODE system code.
+        return convert(ODESystem,rs)
+    elseif form==:sde      # Returns SDE system code.
+        return convert(SDESystem,rs)
+    end
+    error("Unrecognised form argument given: $form. This should be either reactions (default), :ode, or :sde.")
 end
+
 
 function Latexify.infer_output(env, rs::ReactionSystem, args...)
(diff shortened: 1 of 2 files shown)


TorkelE:
This is fine now, with the exception of documentation, which exhibits the same problem as in https://github.com/SciML/Catalyst.jl/pull/588

isaacsas:
Why did this PR get all these doc updates?

TorkelE:
Sorry, I thought that was what happened. The doc update got merged, so I figured I'd rebase this one so it surely contained all those changes as well. Is anything messed up?

isaacsas:
There isn't a mistake in the code, but I wouldn't think this PR should show those commits if they were previously merged into master and the PR was updated to master before merging.

TorkelE:
Got it, will think about this next time to see what is going on.

isaacsas:
When you say you rebased, did you rebase into this PR from the updated master branch after you merged the other PR? Or did you rebase from a different branch?

TorkelE:
After I merged the doc update PR into master, my intention was to rebase this PR onto the updated master, before merging this one. Although at this stage I don't dare promise that what I did was exactly what I intended...

isaacsas:
Fair enough! Next just try to make sure you merge master into the PR so the redundant commits don't show in the diff. It keeps things cleaner and easier for us to distinguish what the PR added when looking back.

TorkelE:
Yes. Thanks for calling these things out, I've gotten much better with Git of late, but still have a bit longer to go!

isaacsas:
My usual workflow for updating to master is to use GitKraken, and just drag master onto my branch and select the merge master into branch option.

TorkelE:
Started using GitKraken as well, really like it.

PR #590, opened 2023-02-06
Dummy PR

TorkelE:
This add an empty line. Not to be merged, I just want to figure out what the docs won't build


TorkelE:
There's something fishy going on @isaacsas . This PR doesn't do anything still the documentation fails to build.

Maybe this caused it https://github.com/SciML/Catalyst.jl/commit/28f6b891bcb2729ab37b6527f0046f081b969144 ? Should be fine, but cannot come up with anything else.

TorkelE:
Things sorted out, this is no longer needed.

PR #591, opened 2023-02-07
Add advanced simulation tutorial

TorkelE:
This adds a new tutorial to the docs. It goes through some advanced simulation options:
- Monte Carlo simulations using `EnsembleProblem`s.
- Creating events using callbacks.
- Plotting individual species or making plots in phase space.
In each case, we link to where a full tutorial of the feature can be found.

PR #592, opened 2023-02-07
Various docs fixes

TorkelE:
Not planning to make any major changes in the PR, just various misc fixes.


TorkelE:
The sub headers where different across the docs (using ## or ##, capitalising every word or just the firs). 

Currently, I set it to:
Each doc page start with a # header, followed by ## headers.
In the first title on each page every word is capitalised (although wouldn't mind changing to only the firs). Following headers, only the first word is capitalized.

TorkelE:
I added two more models to the basic CRN examples as that one felt a bit thin. I also set it so that each of those examples begins with
```
using Catalyst, DifferentialEquations, Plots
```
(rather than just the first one)

That one someone can copy whichever example they want and run it directly without adding the `using ...` 
(happy to revert this if you prefer the old version)

docs/src/catalyst_functionality/network_analysis.md
```diff
@@ -235,7 +235,7 @@ complexgraph(rn)
 
 ![network_1](../assets/complex_rn.svg)
 
-### Linkage classes and sub-networks of the reaction network
+## Linkage classes and sub-networks of the reaction network
```
isaacsas, line 238: This was intentionally meant to be a sub-subsection.

docs/src/catalyst_functionality/network_analysis.md
```diff
@@ -274,7 +274,7 @@ and,
 
 ![subnetwork_2](../assets/complex_subnets2.svg)
 
-### Deficiency of the network
+## Deficiency of the network
```
isaacsas, line 277: This was intentionally meant to be a sub-subsection.

docs/src/catalyst_functionality/network_analysis.md
```diff
@@ -326,7 +326,7 @@ Quoting Feinberg [^1]
 > stoichiometry vectors] are as independent as the partition of complexes into
 > linkage classes will allow.
 
-### Reversibility of the network
+## Reversibility of the network
```
isaacsas, line 329: This was intentionally meant to be a sub-subsection.

docs/src/catalyst_functionality/network_analysis.md
```diff
@@ -375,7 +375,7 @@ isweaklyreversible(rn, subnets)
 Every reversible network is also weakly reversible, but not every weakly
 reversible network is reversible.
 
-### Deficiency Zero Theorem
+## Deficiency Zero Theorem
```
isaacsas, line 378: This was intentionally meant to be a sub-subsection.

isaacsas:
Aside from the sub-subsection comments LGTM.

TorkelE:
What do you think of making the ### headers into #### headers? Right now it is slightly hard to distinguish the ### ones from the ## ones. If we make them #### it might be clearer that these are subsections?
(and I don't think we will ever subsection something using all three of ##, ###, and ####, so might as well use the edge cases)

isaacsas:
Sure, as long as it keeps them larger than the surrounding text.

TorkelE:
I think that the error is from this part:
```
` ` `@example tut2
rn = @reaction_network begin
  1.0, X --> ∅
  k*X, Y --> ∅
end
` ` `
corresponding to the ODE model
` ` `@example tut2
using Latexify
latexify(rn; form=:ode)
` ` `
```
Trying to render this is a latex expression gives an error:
```
┌ Error: mode == :single requires the Markdown string to parse into a single block
│   s =
│    L"$\begin{align}
│    \frac{\mathrm{d} X\left( t \right)}{\mathrm{d}t} =&  - X\left( t \right) \\
│    \frac{\mathrm{d} Y\left( t \right)}{\mathrm{d}t} =&  - k X\left( t \right) Y\left( t \right)
│    \end{align}
│    $"
│   md.content =
│    2-element Vector{Any}:
│     :\
│     Markdown.Paragraph(Any["begin{align} \\frac{\\mathrm{d} X\\left( t \\right)}{\\mathrm{d}t} =&  - X\\left( t \\right) ", "\\", "\n\\frac{\\mathrm{d} Y\\left( t \\right)}{\\mathrm{d}t} =&  - k X\\left( t \\right) Y\\left( t \\right) \\end{align} \$"])
└ @ Documenter.Utilities ~/.julia/packages/Documenter/H5y27/src/Utilities/Utilities.jl:663
```

I am able to build the docs locally without getting this error, however the latex does not seem to render properly:
image

isaacsas:
If you convert to an ODESystem and then display does it look ok?

TorkelE:
Yep it does.

(this is in fact what my hack solution does)

Not sure why the new `latexify` fails in the docs, ti works locally.

It would probably be good to understand this, but would probably need to talk to Niklas on what is actually going on in latexify, and I am not sure how much time he has for that package these days. Alternatively, we can add this as an issue and go with this/revert latexify changes.

isaacsas:
If they print correctly after conversion in the docs just change to that instead, and drop showing off the shorthand version for now (since it isn't really working consistently).

TorkelE:
I guess we could do that. But don't you think the `latexify(rn; form=:ode)` should be the main approach? It seems weird to use an alternative. Also, it seems to be working in non-documenter situations. I could try to dig in to what is going on.

TorkelE:
I've redone it using convert, and will raise an issue.

isaacsas:
Sounds good. We can switch back to your form once it is working.

TorkelE:
Bot this and https://github.com/SciML/Catalyst.jl/pull/591 seems to work now, should I try to merge them?

isaacsas:
I need to look this one over again, I’m still seeing new typos.

TorkelE:
Sounds good, I will fix them as you find them. Thanks for having a proper look at it.

README.md
```diff
@@ -47,44 +47,43 @@ Several Youtube video tutorials and overviews are also available:
...
-  [DifferentialEquations.jl](https://github.com/SciML/DifferentialEquations.jl)
+  [DifferentialEquations.jl](https://docs.sciml.ai/DiffEqDocs/stable/)
   ODE/SDE/jump solver, and can be used within `EnsembleProblem`s for carrying
   out parallelized parameter sweeps and statistical sampling. Plot recipes
   are available for visualizing the solutions.
-- [Symbolics.jl](https://github.com/JuliaSymbolics/Symbolics.jl) symbolic expressions
-  and Julia `Expr`s can be obtained for all rate laws and functions determining the
-  deterministic and stochastic terms within resulting ODE, SDE, or jump models.
-- [Latexify](https://github.com/korsbo/Latexify.jl) can be used to generate
+- Julia `Expr`s can be obtained for all rate laws and functions determining the
```
isaacsas, line 77: This should probably be something like "Symbolics.jl symbolic expressions can be obtained for all...". Getting `Expr`s is actually not what we return in the API, and takes some more work to get...

docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -86,24 +86,24 @@ Numeric parameter values do not need to be set when the model is created, i.e.
...
 end
 ```
 All symbols that do not appear as a substrate or product in a reaction are
 designated by Catalyst as a parameter (i.e. all symbols appearing only within
-rate expressions and/or as stoichiometric coefficients). In this example `X`
-appears as a substrate and product, but not `p` or `d`. Hence `p` and `d` are
-designated as parameters. It is possible to manually specify what should be
-considered a species or parameter. This is described later in the tutorial.
+rate expressions and/or as [stoichiometric coefficients](@ref parametric_stoichiometry)). In this example `X` and `Y`
+appears as a substrates and products, but neither `k1` nor `k2`. Hence `k1` and `k2` are
```
isaacsas, line 96:
```suggestion
appear as a substrates and products, but neither `k1` nor `k2`. Hence `k1` and `k2` are
```

docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -86,24 +86,24 @@ Numeric parameter values do not need to be set when the model is created, i.e.
...
+designated as parameters. Later in this tutorial, we will describe how to manually specify what should be
+considered a species or parameter.
 
-## Production, Destruction and Stoichiometry
+## Production, Destruction, and Stoichiometry
 Sometimes reactants are produced/destroyed from/to nothing. This can be
-designated using either `0` or `∅`:
+designated using `0` (`∅` can also be used as an alternative to `0`):
```
isaacsas, line 102: I prefer the old wording here, why single out one notation?

docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -250,7 +250,9 @@ end
 ```
 corresponding to the ODE model
 ```@example tut2
-convert(ODESystem, rn)
+using Latexify
+# latexify(rn; form=:ode) # hide
+latexify(convert(ODESystem,rn))
```
isaacsas, line 255: Why add the latexify call? The point is that it automatically displays the Latex just fine when converting as the default system display type, and this shows the user how to get the ODEs at the same time...

docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -262,7 +264,8 @@ rn = @reaction_network begin
 end
 ```
 ```@example tut2
-convert(ODESystem, rn)
```
isaacsas, line 265: Same comment.

isaacsas: Generally, I don't see why we in docs should use `latexify(rn; form=:ode)` given that it is internally doing a `convert`, and explicitly just doing convert both shows the equations and shows the user how to get the ODEs.

docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -407,7 +417,7 @@ plot(sol)
 ## Naming the generated `ReactionSystem`
 ModelingToolkit uses system names to allow for compositional and hierarchical
 models. To specify a name for the generated `ReactionSystem` via the
-`reaction_network` macro, just place the name before `begin`:
+`@reaction_network` macro, just place the name before `begin`:
```
isaacsas, line 420:
```suggestion
[`@reaction_network`](@ref) macro, just place the name before `begin`:
```

docs/src/catalyst_functionality/network_analysis.md
```diff
@@ -204,16 +203,16 @@ complexgraph(repressilator)
 
 ![Repressilator complex](../assets/repressilator_complexgraph.svg)
 
-Here ∅ represents the empty complex, black arrows show reactions converting
+Here ∅ represent the empty complex, black arrows show reactions converting
```
isaacsas, line 206: Hmm, to be honest "represents" sounds better/more correct to me.

docs/src/catalyst_functionality/parametric_stoichiometry.md
```diff
@@ -1,11 +1,11 @@
-# [Parametric Stoichiometry](@id parametric_stoichiometry)
+# [Symbolic Stochiometries](@id parametric_stoichiometry)
 Catalyst supports stoichiometric coefficients that involve parameters, species,
 or even general expressions. In this tutorial we show several examples of how to
 use parametric stoichiometry, and discuss several caveats to be aware of.
```
isaacsas, line 4:
```suggestion
use symbolic stoichiometries, and discuss several caveats to be aware of.
```

docs/src/catalyst_functionality/parametric_stoichiometry.md
```diff
@@ -16,10 +16,10 @@ revsys = @reaction_network revsys begin
 end
 reactions(revsys)
 ```
-Note, as always the `@reaction_network` macro sets all symbols not declared to
-be parameters to be species, so that in this example we have two species, `A`
-and `B`, and four parameters. In addition, the stoichiometry is applied to the
-right most symbol in a given term, i.e. in the first equation the substrate `A`
+Note, as always the `@reaction_network` macro defaults to setting all symbols neither used as a reaction substrate nor a product to
+be parameters. Hence, in this example we have two species (`A`
+and `B`) and four parameters (`₊`, `k₋`, `m`, and `n`). In addition, the stoichiometry is applied to the
```
isaacsas, line 21:
```suggestion
and `B`) and four parameters (`k₊`, `k₋`, `m`, and `n`). In addition, the stoichiometry is applied to the
```

docs/src/catalyst_functionality/programmatic_CRN_construction.md
```diff
@@ -165,11 +165,7 @@ Note, there are a few differences when using the `@reaction` macro to specify
 one reaction versus using the full `@reaction_network` macro to create a
 `ReactionSystem`. First, only one reaction (i.e. a single forward arrow type)
 can be used, i.e. reversible arrows like `<-->` will not work (since they define
-more than one reaction). Second, the `@reaction` macro must try to infer which
-symbols are species versus parameters, and uses the heuristic that anything
-appearing in the rate expression is a parameter. Coefficients in the reaction
-part are also inferred as parameters, while rightmost symbols (i.e. substrates
-and products) are inferred as species. As such, the following are equivalent
+more than one reaction). Second, the `@reaction` macro does not have an option for designating what should be considered a species or parameter, but have to use the heuristic where anything that appears as either a substrate or a product is a species, and everything else (including stoichiometric coefficients) are parameters. As such, the following are equivalent
```
isaacsas, line 168:
```suggestion
more than one reaction). Second, the `@reaction` macro does not have an option for designating what should be considered a species or parameter, and instead assumes that any symbol that appears as either a substrate or a product is a species, and everything else (including stoichiometric coefficients) are parameters. As such, the following are equivalent
```

docs/src/faqs.md
```diff
@@ -100,8 +100,10 @@ generating reaction rate laws, see also the [Reaction rate laws used in
 simulations](@ref) section. Leaving this keyword out for systems with floating
 point stoichiometry will give an error message.
 
+For a more extensive documentation of non-integer stoichiometric coefficients, pelase read the [Symbolic Stochiometries](@ref parametric_stoichiometry) section.
```
isaacsas, line 103:
```suggestion
For a more extensive documentation of using non-integer stoichiometric coefficients, please see the [Symbolic Stochiometries](@ref parametric_stoichiometry) section.
```

isaacsas (changes requested):

README.md
```diff
@@ -47,44 +47,43 @@ Several Youtube video tutorials and overviews are also available:
...
-  [Symbolics.jl](https://github.com/JuliaSymbolics/Symbolics.jl).
+  built on [ModelingToolkit.jl](https://docs.sciml.ai/ModelingToolkit/stable/) and
+  [Symbolics.jl](https://docs.sciml.ai/Symbolics/stable/).
 - Non-integer (e.g. `Float64`) stoichiometric coefficients are supported for generating
   ODE models, and symbolic expressions for stoichiometric coefficients are supported for
   all system types.
-- The [Catalyst.jl API](http://docs.sciml.ai/Catalyst/stable/api/catalyst_api) provides
-  functionality for extending networks, building networks programmatically,
-  network analysis, and for composing multiple networks together.
+- The [Catalyst.jl API](@ref) provides functionality for extending networks,
```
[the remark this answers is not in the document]
TorkelE: No, you are probably right about this. Will revert.

docs/src/catalyst_functionality/compositional_modeling.md
```diff
@@ -91,19 +91,20 @@ ModelingToolkit.get_systems(flatrn)
...
 repressed gene, taking the repressor as input
 ```@example ex1
 function repressed_gene(; R, name)
-    @reaction_network $name begin
-        hillr($R,α,K,n), ∅ --> m
-        (δ,γ), m <--> ∅
-        β, m --> m + P
-        μ, P --> ∅
-    end
+  @reaction_network $name begin
```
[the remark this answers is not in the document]
TorkelE: I am all for 4 spaces, but it a lot of places we use 2 for the macro.

The reason I changed this was that  there was like 6 or 8 spaces for the reactions, and then 4 spaces before the end statement. Mostly wanted to remove indentations for the ned statement. If you want to make it so that 4 spaces are used everwhere that is probably what I prefer as well, just presumed it was 2 spaces because it was os prevalent.

README.md
TorkelE: Changed to "Symbolic Julia `Expr`s (implemented through [Symbolics.jl](https://github.com/JuliaSymbolics/Symbolics.jl)) can be obtained for all rate laws and functions determining the
  deterministic and stochastic terms within resulting ODE, SDE or jump models."

docs/src/catalyst_functionality/dsl_description.md
TorkelE: I thought maybe we wanted to go for the less uni-code stuff. But I like the use of `∅` so will revert.

docs/src/catalyst_functionality/dsl_description.md
TorkelE: If we just want to display the ODE I think it makes more sense to only use `latexify`, no reason to show get ModelingToolkit stuff involved unless we want to. This was the one that got broken, but once we get it working I think we should use `latexify(rn; form=:ode)` rather than `latexify(convert(ODESystem,rn))`, unless we are in a section explicitly discussing conversion between modelingtoolkit systems.

docs/src/catalyst_functionality/dsl_description.md
TorkelE: I think we should use `latexify(rn; form=:ode)` rather than `latexify(convert(ODESystem,rn))` for the same reason we do
`ODEProblem(rn,...)` rather than `ODEProblem(convert(ODESystem,rn),...)`

(right now we might actually do some additional stuff in this example, but at some point `ODEProblem(rn,...)`and `ODEProblem(convert(ODESystem,rn),...)` would have been equivalent, and the first one was preferred, and in my mind the latexify case is very similar)

docs/src/catalyst_functionality/network_analysis.md
TorkelE: Ok. I will defer to you on matters of language.

TorkelE: Actually, mixed up what was new and old. I agree that it should be "represents", not sure why it was changed.

TorkelE:
Ok, fixed after your comments, in summary, what is left hanging:
- One case where I rephrased the sentence, if you want we can rephrase that though.
- Whenever we want to move to 4 spaces throughout the docs, I am for this, but had some memory of there was a reason for 2 spaces.
- I'm ok to use `latexif(convert(ODESystem, rn))` for now, but when it works I think we should move to `latexify(rn; form=:ode)` in most places. But unless we want to we do not need to settle that until that is actually possible.

isaacsas:
This LGTM.

isaacsas:
If tests pass and the doc build looks ok this should be GTG.

ChrisRackauckas:
Just delete the unused SVGs.

PR #594, opened 2023-02-10
Improve package tests

TorkelE:
The tests have evolved organically for a while, so I figured I could do a pass of them as well.

Now all tests are grouped in subfolders according to roughly what they test. I have also tried renaming test files and test header messages to better describe what they test.


test/dsl/dsl_model_construction.jl
```diff
@@ -367,3 +367,32 @@ rn = @reaction_network SIR1 begin
...
+    k2, C --> A + B
+    b1, B --> 0
+end
+
+ps = ones(5)
+u0 = [10.0, 20.0, 0.0]
+tspan = (0, 100.0)
+oprob1 = ODEProblem(rn1, u0, tspan, ps)
+oprob2 = ODEProblem(rn2, u0, tspan, ps)
+sol1 = solve(oprob1, Tsit5())
```
isaacsas, line 394: Is this not covered by other existing tests?  This seems like a pretty basic example.

Why not just test equality of the two networks? Just give them the same name and then use `==`. That avoids having to add another ODE solve.

test/runtests.jl
```diff
@@ -4,50 +4,51 @@ using SafeTestsets
...
     if !Sys.isapple()
-        @time @safetestset "Graphs" begin include("graphs.jl") end
+        @time @safetestset "Graphs" begin include("visualization/graphs.jl") end
     end
 
-    @time @safetestset "Units" begin include("units.jl") end
+
+    # Runs various miscellaneous tests.
+    @time @safetestset "API" begin include("miscellaneous_tests/api.jl") end
+    @time @safetestset "Symbolic Stoichiometry" begin include("miscellaneous_tests/symbolic_stoichiometry.jl") end
```
isaacsas, line 50: API, symbolic stoich, and units should all precede the network analysis tests as they test lower-level functionality.

isaacsas:
Probably the single best thing to do for the tests is to eliminate all dependencies on the order of symbols / species (for example, by dropping an purely numeric initial conditions and/or parameter lists).

TorkelE:
I plan to remove the `@parameters` stuff in many places as they were the simplest at the time. Regarding the ordering, it is quite a lot and I want to wait until we see what happens with the various SciML base revamps.

I will finish this up with the last few additions, I have mostly done already. Then we can merge after your update.

TorkelE:
This is replaced by https://github.com/SciML/Catalyst.jl/pull/622

Issue #597, opened 2023-02-14
Forbidden symbols check in DSL no longer working


isaacsas:
On master
```julia
rn = @reaction_network begin
    π, A --> B
end
```
gives
```
Model ##ReactionSystem#293
States (2):
  A(t)
  B(t)
Parameters (0):
```
So instead of an error the symbol just doesn't get added as a parameter, which is a silent bug.

This issues also persists with `@reaction π, A --> B` which does not error on this example.

isaacsas:
https://github.com/SciML/Catalyst.jl/pull/605

PR #598, opened 2023-02-15
split forbbiden symbols in those for errors and those for skipping

TorkelE:
Basically, we had one list of forbidden symbols. This included stuff like `t`, which was neither to be considered a parameter or species, but would have some other property. It also includes `pi` which would be able to work like that, but should give an error. This should not be fixed by providing two lists.


isaacsas:
Can you add some tests that we now properly error when using a forbidden symbol?

isaacsas:
Wait, does this prevent users from using $\pi$ or `pi` in rate formulas and such? If so I think we should allow them, but always interpret them as the Julia constants (i.e. not make them a parameter or variable/species).

TorkelE:
yeah, I was thinking about that one as well. Since I didn't have the ability to check extensively I went with what we currently had. Long term letting it be a constant (3.141...) I think is the way to go. I can try to add it to the list with t, and it might work without further stuff...

TorkelE:
I have now enabled π, it should be the same as setting a constant rate to 3.141... It means that https://github.com/SciML/Catalyst.jl/issues/597 will still be weird, but I think that will be working as intended from now on. I have also added tests for this.

TorkelE:
Ok, this got messed up.

I tried to rebase onto master, but now it seems to have merged instead...

TorkelE:
Rebase failure, starting again.

PR #599, opened 2023-02-16
Update to Symbolics 5

ChrisRackauckas:


isaacsas:
This will require internal updates I think since we use some of the changed interface / types.

My plan was to make this update as part of version 13 since master is currently in an intermediate, breaking, state, so we'd have to backport this to the last release if we want it in a Catalyst 12 version. After the `@species` PR finishes I'll tackle this as part of the V13 TODO. (That PR should be done soon as tests are now passing, so I just need to update the docs, though I'd prefer to wait on https://github.com/JuliaSymbolics/Symbolics.jl/pull/844 getting into a Symbolics release before we release V13 as otherwise array states will no longer work in Catalyst...)

isaacsas:
Handled in https://github.com/SciML/Catalyst.jl/pull/585, should hopefully be out by next week.

Issue #595, opened 2023-02-10
Parse error when defining reaction network in Jupyter notebook


jonathanfischer97:
![image](https://user-images.githubusercontent.com/69228453/218194834-0a05519b-d101-4c17-8104-db6a56f099da.png)

Running Julia 1.8.5 within VSCode Jupyter notebook, and always receive this parse error when defining a new reaction network. Anybody know what's going on?

isaacsas:
Yup, Katex and Mathjax are incompatible in how they handle the chemical notation package (mhchem). We currently output Latex that works with the latter as it also works with Documenter. 

If anyone knows how to add a path for detecting Latexify is being called within a Jupyter environment we can try to hardcode different Latex output in that case, but I don't know how to do this.

TorkelE:
Would it be possible to add an option to `latexify` changing the output between the two forms? That way people that uses the "wrong" environment (like me) could still access the feature, even though in a bit klanky way.

TorkelE:
So I seem to get error with both options?
![image](https://user-images.githubusercontent.com/18099310/218199509-8a24f36b-ce0d-4f4b-9d80-9545e8498df3.png)

isaacsas:
Ahh, this is even worse than I imagined. @jonathanfischer97 this is the same issue you posted on Discourse right?

https://discourse.julialang.org/t/testing-whether-a-function-is-executing-in-a-jupyter-notebook/222

As I said there, Latex display kinds works in Jupyter for me, but that was in the web-browser interface, which I guess uses Mathjax.

It seems VS Code uses an entirely different display stack, I guess Katex-based, and so it doesn't work at all. 

You'd need to somehow extend VSCode's Jupyter install to load the mhchem Katex extension, and then we'd need to add code here to output Katex-compatible Latex if we detect we are running in Jupyter within VSCode. Unfortunately it seem Microsoft is not interested in adding this compatibility itself:

https://github.com/microsoft/vscode/issues/131874

isaacsas:
Alternatively, we could try to detect if we are in Jupyter here and write a non-mhchem Latex path. @TorkelE I'd be fine with dropping mhchem for everything and just using traditional Latex reaction arrows with over and under bars instead of mhchem. mhchem was just what Latexify has always used, but it is kinda a pain on the web...

TorkelE:
[...]

PR #609, opened 2023-02-19
Drop mhchem

isaacsas:
Closes https://github.com/SciML/Catalyst.jl/issues/595

This drops the mhchem package specialized reaction commands, switching to extensible arrows. It now outputs displayable reactions across Documenter, Jupyter browser notebooks, VSCode Jupyter notebooks, and Pluto. 

Unfortunately, Pluto and Documenter with Mathjax don't seem to render "\xrightleftharpoons" with a reversible arrow that stretches. However, I could switch Documenter back to Katex since we dropped mhchem and then the command stretches properly. 

There isn't much we can do unless we want to switch away from harpoon arrows to something like "\xleftrightarrow" (but this loses the arrow directionality so seems like the worse choice of current options).

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -15,9 +15,11 @@ Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
 
 [compat]
 Catalyst = "13"
+DiffEqFlux = "1.53"
 DifferentialEquations = "7.7"
 Distributions = "0.25"
 Documenter = "0.27"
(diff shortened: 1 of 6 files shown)


isaacsas:
Examples:

Documenter: 

documenter1


documenter2

Jupyter web browser:

jupyter_web

Jupyter VSCode:

vscode

Pluto:

pluto

REPL:

repl

Issue #547, opened 2022-08-18
consider ordering Reaction components


isaacsas:
Two `Reaction`s with the same substrates and products and such can hash differently if there is an order difference between them in the species, products, or generated net stoichiometry (with the latter having an order based on Julia's default dict so not really deterministic across versions). This can cause problems with lists of `Reaction`s not being `issetequal`. We should consider enforcing a partial ordering of states within `Reaction`s or updating the hashing and equality testing to ensure they don't inadvertently depend on such orderings.

isaacsas:
Hashing should now be updated to be order independent.

Issue #516, opened 2022-05-31
disallow repeated substrates or products in reactions


isaacsas:
From https://discourse.julialang.org/t/read-in-network-does-not-match-hard-coded-network/81997/5
```julia
rx = @reaction k, A+A --> B
@parameters k
@variables t A(t) B(t)
rx2 = Reaction(k, [A,A], [B], [1,1], [1])
rx != rx2   # true
```
We need an error check here (perhaps with a kwarg to disable), or need to detect this and correct the substrates/products.

TorkelE:
While I agree 
```
rx = @reaction k, 2A --> B
```
is much better than
```
rx = @reaction k, A+A --> B
```
is there a strong reason to disallow the latter?

isaacsas:
We support both notations via the DSL. The issue is in symbolically building `Reaction`s -- they currently assume their inputs are unique, and I think they should stay that way for internal use. So we either need an error check in their constructor or we need to normalize them in the constructor by merging repeated species.

PR #616, opened 2023-02-21
Disallow repeated reactants and making hashing order independent 

isaacsas:
Closes https://github.com/SciML/Catalyst.jl/issues/547
Closes https://github.com/SciML/Catalyst.jl/issues/516

### DIFF
--- a/src/reactionsystem.jl
+++ b/src/reactionsystem.jl
@@ -155,7 +155,7 @@ function Reaction(rate, subs, prods, substoich, prodstoich;
         prods = value.(prods)
     end
     allunique(prods) ||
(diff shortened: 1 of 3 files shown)


src/reactionsystem.jl
```diff
@@ -152,6 +154,8 @@ function Reaction(rate, subs, prods, substoich, prodstoich;
     else
         prods = value.(prods)
     end
+    allunique(prods) ||
+        throw(ArgumentError("Products can not be repeated in the least provided to `Reaction`, please modify the stoichiometry for any repeated products instead."))
```
isaacsas, line 158:
```suggestion
        throw(ArgumentError("Products can not be repeated in the list provided to `Reaction`, please modify the stoichiometry for any repeated products instead."))
```

PR #622, opened 2023-02-27
Tests remake

TorkelE:
Redoing https://github.com/SciML/Catalyst.jl/pull/594 but in a new PR


isaacsas:
While are so many of the files showing as completely new files instead of just moved files? If you just move them then there will be no diff showing beyond the change of file location, and the git history will be preserved.

TorkelE:
I renamed several tests files to make it easier to see what they are actually about, hence they show up as new files and not moved ones.E.g. `make_model.jl` is now `programmatic_model_expansion.jl`. I agree that it would have been nicer without this, but I think long-term it will be good to have files properly named.

isaacsas:
For example, what you did for "custom_functions" properly retained the git history and only shows the new changes.

TorkelE:
I copied most files over from the old PR, that might have something to do with it

isaacsas:
Yeah, please fix that so the history is preserved and the diff of what actually changed in the file is correct instead of looking like a new file.

TorkelE:
There were a lot of test changes in the old PR (e.g. I had gone through all the macros and removed the `@parameters ...` option inputs). Checking the test changes in 13, it was easier to just transfer those to the old test update PR file, as opposed to doing it the other way around. Not sure if there is a good way of getting the files to appear in history without manually getting all the https://github.com/SciML/Catalyst.jl/pull/594 changes though?

TorkelE:
For some reason the latexify file refuses to register as moved, currently investigating

isaacsas:
My guess is if you first move/rename the files and make a commit, and then add in the changes in a second commit, git will catch that they are moved files and not a new.

TorkelE:
Ok, seems like the CI have stalled. Hopefully, it picks up soon.

The PR should have been reordered so that the changes are tracked with moving files, it worked out in the end. I agree it is much better this way. It also enables us to better check that indeed nothing was lost in the change.

test/network_analysis/network_properties.jl
```diff
@@ -127,3 +124,14 @@ cls = conservationlaws(rn3)
 #     end
 #     println("-----------")
 # end
+
+
+### Tests additional network properties ###
+rn4 = @reaction_network begin
```
isaacsas, line 130: These are really API tests. Aren't these covered there? (I'd at least put them in there instead, within a new `let` block.

test/reactionsystem_structure/reactionsystem.jl
```diff
@@ -700,3 +698,16 @@ let
     @test rx == rx2
     @test hash(rx) == hash(rx2)
 end
+
+### Additional Unsorted Tests ###
+
+rn = @reaction_network begin k, X --> 0 end
```
isaacsas, line 704: Put these is their own let block for scoping.

test/runtests.jl
```diff
@@ -4,50 +4,44 @@ using SafeTestsets
...
-
-    # Tests related to PDESystems
-    @time @safetestset "PDE Systems" begin include("pdes.jl") end
-
-    # Miscellaneous tests
-    #@time @safetestset "Basic Plotting" begin include("plotting.jl") end
-    @time @safetestset "Latexify" begin include("latexify.jl") end
-
+    ### Tests the properties of ReactionSystems. ###
+    @time @safetestset "ReactionSystem Structure" begin include("reactionsystem_structure/reactionsystem.jl") end
```
isaacsas, line 8: I think just `reactionsystem` for the folder is fine. I'm not really sure what `structure` refers to here...

test/runtests.jl
```diff
@@ -4,50 +4,44 @@ using SafeTestsets
...
+    @time @safetestset "Custom CRN Functions" begin include("dsl/custom_functions.jl") end
+    @time @safetestset "DSL Options" begin include("dsl/dsl_options.jl") end
+
+    ### Non-DSL model cration and modication. ###
+    @time @safetestset "ReactionSystem Components Based Creation" begin include("programmatic_model_creation/component_based_model_creation.jl") end
+    @time @safetestset "Programmatic Model Expansion" begin include("programmatic_model_creation/programmatic_model_expansion.jl") end
+
+    # Runs various miscellaneous tests.
+    @time @safetestset "API" begin include("miscellaneous_tests/api.jl") end
+    @time @safetestset "Symbolic Stoichiometry" begin include("miscellaneous_tests/symbolic_stoichiometry.jl") end
```
isaacsas, line 23: Call these higher level functionality instead of miscellaneous?

test/runtests.jl
```diff
@@ -4,50 +4,44 @@ using SafeTestsets
...
+    @time @safetestset "Events" begin include("miscellaneous_tests/events.jl") end
+    @time @safetestset "Units" begin include("miscellaneous_tests/units.jl") end
+
+    ### Tests ODE, SDE, PDE, and Gillespie Simulations. ###
+    @time @safetestset "ODE System Simulations" begin include("model_simulation/simulate_ODEs.jl") end
+    @time @safetestset "Automatic Jacobian Construction" begin include("model_simulation/make_jacobian.jl") end
+    @time @safetestset "U0 and Parameters Input Variants" begin include("model_simulation/u0_n_parameter_inputs.jl") end
+    @time @safetestset "DiffEq Steady State Solving" begin include("model_simulation/solve_steady_state_problems.jl") end
+    @time @safetestset "SDE System Simulations" begin include("model_simulation/simulate_SDEs.jl") end
+    @time @safetestset "PDE Systems Simulations" begin include("model_simulation/simulate_PDEs.jl") end
```
isaacsas, line 33: I'd order these ODEs, SDEs, jumps, steady-state, PDEs. The last two are the least used, and so should be later in the testing (just to make it a bit less of a wait to see if an ODE/SDE/jump test fails).

test/runtests.jl
```diff
@@ -4,50 +4,44 @@ using SafeTestsets
...
+    @time @safetestset "Jump System Simulations" begin include("model_simulation/simulate_jumps.jl") end
+
+    ### Reaction network analysis. ###
+    @time @safetestset "Conservation Laws" begin include("network_analysis/conservation_laws.jl") end
```
isaacsas, line 37: The network analysis tests should probably go before all the solving tests given they are primarily testing internal catalyst properties and not testing interfacing with lower-level packages.

isaacsas (changes requested):
Remove the _old.jl files, they are just adding noise to the PR.

Also, one thing I've been trying to do when adding new tests is to put them in their own `let` block so that all the variables are local to the test and we avoid using global variables defined much earlier (along with a comment for the purpose of the test). If you could do that for any new tests you are adding that would be great.

test/dsl/dsl_model_construction.jl
```diff
@@ -363,13 +351,49 @@ end
...
+rn2 = @reaction_network arrowtest begin
+    a1, C --> 0
+    a2, 0 --> C
+    k1, A + B --> C
+    k2, C --> A + B
+    b1, B --> 0
+end
+
+@test rn1 == rn2
+f1 = ODEFunction(convert(ODESystem, rn1), jac = true)
```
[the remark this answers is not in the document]
TorkelE: I've removed the additional checks.

test/latexify.jl
```diff
@@ -1,111 +0,0 @@
-#! format: off
```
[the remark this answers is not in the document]
TorkelE: Not sure, this does not show up on my computer? For some reason the main latexify file keeps getting changed to delted/created instead of moved, are you talking about the old latexify test file?

test/model_simulation/solve_steady_state_problems.jl
```diff
@@ -0,0 +1,65 @@
+### Fetch required packages and reaction networks ###
```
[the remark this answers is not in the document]
TorkelE: Previously all these were commented out. The problem was that they were rather rigorous, at the steady state solving method we tested just are not that good, so they kept erroring (hence we commented it out).

What I did was to add the simple test back in  (since we do allow creating `SteadyStateProblem`s). However, I do not recommend this (and I don't think DiffEq does either) so we could just remove `SteadyStateProblem` functionality all together...

test/model_simulation/u0_n_parameter_inputs.jl
```diff
@@ -0,0 +1,66 @@
...
+    p_1 = 0.01 .+ factor*rand(rng,length(get_ps(test_network)))
+    p_2 = [p1=>p_1[1], p2=>p_1[2], p3=>p_1[3], k1=>p_1[4], k2=>p_1[5], k3=>p_1[6],
+           v1=>p_1[7], K1=>p_1[8], d1=>p_1[9], d2=>p_1[10], d3=>p_1[11], d4=>p_1[12],
+           d5=>p_1[13]]
+    p_3 = [k2=>p_1[5], k3=>p_1[6], v1=>p_1[7], d5=>p_1[13], p2=>p_1[2], p1=>p_1[1],
+           d2=>p_1[10], K1=>p_1[8], d1=>p_1[9], d4=>p_1[12], d3=>p_1[11], p3=>p_1[3],
+           k1=>p_1[4]]
+
+    sols = []
+    push!(sols,solve(ODEProblem(test_network,u0_1,(0.,10.),p_1),Rosenbrock23()))
```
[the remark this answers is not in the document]
TorkelE: I don't mind keeping the former. I agree that it takes some time, but I'd rather be rigorous than fully trust all DiffEq stuff.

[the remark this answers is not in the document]
TorkelE: I can see your point with the later

test/network_analysis/network_properties.jl
TorkelE: I was unsure where to put them, but in the end did them here because the functions was not in the API file (but the rs file). Happy to move it.

test/runtests.jl
TorkelE: that `ReactyionSystem` is a structure, basically tests of that structure. If you still want to rename it back I can do so.

test/runtests.jl
TorkelE: sounds good, will change.

test/runtests.jl
TorkelE: sounds good, will change.

test/network_analysis/network_properties.jl
isaacsas: Ahh, then put them in `reactionsystem.jl` I think. That is where the basic `reactionsystem` functionality should go.

test/runtests.jl
TorkelE: Aren't there some other categories (like visualization) that could also be considered higher level functionality? I just wanted an all-encompassing category for stuff I didn't know where to put.

TorkelE: If you want I could change it to higher level though

TorkelE:
I wanted to keep the _old files for reference, but I've removed them now. using `let ... end` blocks sounds good. I will do a general pass and add what I can to such blocks, getting some more structure to the files.

test/reactionsystem_structure/reactionsystem.jl
TorkelE: done

test/network_analysis/network_properties.jl
TorkelE: sorry, mixed up what was were. No, this should be in the API file as you initially said.
(it s the thing in your next comment that is reactionsystem file stuff)

PR #623, opened 2023-02-28
Variables in DSL

isaacsas:
Closes https://github.com/SciML/Catalyst.jl/issues/615
Closes https://github.com/SciML/Catalyst.jl/issues/624

PR #625, opened 2023-02-28
add @variables option

TorkelE:
Should work now, but should probably add some tests as well (and suggestions?).

Will probably do after the test update though.


TorkelE:
We will use https://github.com/SciML/Catalyst.jl/pull/623 instead.

PR #626, opened 2023-03-01
Finalise test remake

TorkelE:
Continuation of https://github.com/SciML/Catalyst.jl/pull/622

### DIFF
--- a/test/reactionsystem_structure/reactionsystem.jl
+++ b/test/reactionsystem_structure/reactionsystem.jl
@@ -347,10 +347,8 @@ let
     @test isequal2(oderatelaw(rxs[1]; combinatoric_ratelaw = false), k1 * S * S^2 * I^3)
 
     @named rs2 = ReactionSystem(rxs, t, [S, I, R], [k1, k2]; combinatoric_ratelaws = false)
-end
 
-# Test ODE scaling:
-let
+    # Test ODE scaling:
     os = convert(ODESystem, rs)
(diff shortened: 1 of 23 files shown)


test/reactionsystem_structure/reactionsystem.jl
```diff
@@ -137,25 +137,28 @@ end
...
-G2 = sf.g(u, p, t)
-@test norm(G - G2) < 100 * eps()
-
-# Test conversion to NonlinearSystem.
-ns = convert(NonlinearSystem, rs)
-fnl = eval(generate_function(ns)[2])
-dunl = similar(du)
-fnl(dunl, u, p)
-@test norm(du - dunl) < 100 * eps()
+let
```
isaacsas, line 142: What about if you do 
```julia
let t = 0.0
```

TorkelE: That might be worth trying, will try tomorrow morning.

TorkelE: Ok, unfortunately this didn't work.

isaacsas: What's the error?

TorkelE: These two blocks could be put in a "let .. end" statement, however, this causes "fnl(dunl, u, p)" to throw a "MethodError: no method matching Float64(::Num)" error.

PR #629, opened 2023-03-02
add variables()

TorkelE:
add `variables`, which is identical to `nonspecies`

### DIFF
--- a/docs/src/api/catalyst_api.md
+++ b/docs/src/api/catalyst_api.md
@@ -155,6 +155,7 @@ accessor functions.
 ```@docs
 species
 nonspecies
+variables
 reactionparams
 reactions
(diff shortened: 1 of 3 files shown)


isaacsas:
I'm not in favor of this. A species is a variable in the ModelingToolkit language, it could be created with `@variables` if one wanted to manually add the metadata. (Whereas a general variable is not always a species.) So if anything, `variables(sys)` should be equivalent to `states(sys)`.

More than that though, we shouldn't be creating functions that overlap possible ModelingToolkit naming.

TorkelE:
Ok. I figured since to fetch our @spcies X(t) we used species() it would make sense to do the same when fetching @variables X(t). But you are probably right.

PR #630, opened 2023-03-09
allow array variables in DSL

isaacsas:


isaacsas:
@TorkelE this should allow array variables/parameters/species in the DSL. Do the generated expression changes look ok to you?

PR #655, opened 2023-07-19
WIP: add example to parameter estimation doc

tstrey:


docs/src/catalyst_applications/parameter_estimation.md
```diff
@@ -143,3 +143,70 @@ scatter!(sample_times,sample_vals'; color = [:blue :red],
...
+
+rxs = [(@reaction k_on, A --> B), (@reaction k_off, B --> A)]
+discrete_events = (t == switch_time) => [k_on ~ 0.0]
+
+u0 = [:A => 10.0, :B => 0.0]
+tspan = (0.0, 4.0)
+p_real = [k_on => 100.0, k_off => 10.0]
+@named osys = ReactionSystem(rxs, t, [A, B], [k_on, k_off]; discrete_events)
+```
+Note that here `switch_time` is hard coded. Again, we create sample data point by simulating our model and adding noise.
```
isaacsas, line 164:
```suggestion
Note that here `switch_time` is hard coded. Again, we create sample data points by simulating our model and adding noise. 
```

docs/src/catalyst_applications/parameter_estimation.md
```diff
@@ -143,3 +143,70 @@ scatter!(sample_times,sample_vals'; color = [:blue :red],
...
+oprob = ODEProblem(osys, u0, tspan, p_real)
+sol_real = solve(oprob, Tsit5(); tstops = sample_times)
+sample_vals = Array(sol_real(sample_times))
+sample_vals .*= (1 .+ .1 * rand(Float64, size(sample_vals)) .- .05)
+
+default(; lw = 3, framestyle = :box, size = (800, 400))
+plot(sol_real; legend = nothing, color = [:darkblue :darkred])
+scatter!(sample_times, sample_vals'; color = [:blue :red], legend = nothing)
+```
+Now we define a function to fit our model. In this case, instead of `ADAM`, we are using `NelderMead()` because our model has an undefined derivative at switch_time. Additionally, `Optimization.AutoZygote()` is not necessary for gradient free optimizations.
```
isaacsas, line 176:
```suggestion
Now we define a function to fit our model. In this case, instead of `ADAM`, we are using `NelderMead()` because our model has an undefined derivative at `switch_time`. Additionally, `Optimization.AutoZygote()` is not necessary for gradient free optimizations. 
```

docs/src/catalyst_applications/parameter_estimation.md
```diff
@@ -143,3 +143,70 @@ scatter!(sample_times,sample_vals'; color = [:blue :red],
...
+    sol = solve(optprob, Optim.NelderMead())
+
+    return sol.u
+end
+```
+First we fit our function on a time interval before, `switch_time`. 
+```@example pe1
+p_estimate = optimise_p([100.0, 10.0], 1.5)
+```
+Next, we use this parametere estimate as an input to our second iteration of fitting this time on the time interval `(0.0, 4.0)`.
```
isaacsas, line 203:
```suggestion
Next, we use this parameter estimate as an input to our second iteration of fitting, this time on the time interval `(0.0, 4.0)`.
```

isaacsas:
This looks good, but let’s discuss today expanding this to work with data closer to what one might have experimentally.

ChrisRackauckas:
Should this be done here, or in a separate example page? In all of the other docs we've been building out examples pages for this kind of stuff.

isaacsas:
Depends what the final form is. If it is relatively basic as in the current form it might actually, after some tweaking, make sense as the first parameter estimation example here. (i.e. before the more complicated example @TorkelE did.)

But I was thinking we could build it out to ultimately show how one can fit SPR data, in which case it might make sense to move it to a separate example.

@tstrey is working on generalizing it now, so we’ll probably move it to a separate file ultimately.

isaacsas:
@tstrey could you split this off into its own file, making it an application called something like "SPR Fitting"? You can list it after the "parameter estimation" tutorial then in the site docs. 

Once you are done I can punch up the bio explanation to more clearly make the connection to fitting SPR data.

Thanks!

PR #663, opened 2023-08-27
Spatial ssa support

TorkelE:
I have added support for spatial jumps. Creating and solving a problem is relatively easy:
```julia
using Catalyst, Graphs, JumpProcesses

# Make model.
SIR_system = @reaction_network begin
    α, S + I --> 2I
    β, I --> R
end
trS = @transport_reaction dS S
trI = @transport_reaction dI I
lattice = Graphs.grid([2, 2])
lrs = LatticeReactionSystem(SIR_system, [trS, trI], lattice)

# Parameter values and initial conditions.
u0 = [:S => [999, 1000, 1000, 1000], :I => [1, 0, 0, 0], :R => 0]
pV = [:α => 0.1 / 1000, :β => 0.01]
pE = [:dS => [0.1, 0.1, 0.1, 0.2], :dI => [0.01, 0.01, 0.01, 0.02]]

# Create the problem.
dprob = DiscreteProblem(lrs, u0, (0.0,100.0), (pV,pE))
jprob = JumpProblem(lrs, dprob, NSM())
```

Current limitations:
- Cartesian grids are not supported, only Graphs.
- Of the 5 forms of the hopping rates, we now convert to the most general one (D_{s,i,j}). Once this is merged, I will introduce a routine which detects which is the most compact form that can be provided, and uses that one.
- Currently, the constant rate jumps in each compartment must have the same parameter values. Once this is supported in JumpProcesses I will add support for it here as well.


test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
```diff
@@ -0,0 +1,309 @@
...
+            @test jprob.discrete_jump_aggregation.hop_rates.hop_const_cumulative_sums == true_hopping_rates
+        end
+        # Provides parameters as a combined vector.
+        for pD in [pD_1, pD_2]
+            dprob = DiscreteProblem(lrs, u0, (0.0, 100.0), [pD; pC])
+            jprob = JumpProblem(lrs, dprob, NSM())
+            @test jprob.discrete_jump_aggregation.hop_rates.hop_const_cumulative_sums == true_hopping_rates
+        end
+    end
+
```
Vilin97, line 150: I recommend taking out these `if else` codeblocks. There is no reason to check them in if they won't actually run. We'll add them in subsequent PRs to keep this one smaller.

```suggestion

```

test/spatial_test_networks.jl
```diff
@@ -0,0 +1,186 @@
...
+using Catalyst, Graphs
+
+### Helper Functions ###
+
+# Generates ranomised intiial condition or paraemter values.
+rand_v_vals(grid) = rand(nv(grid))
+rand_v_vals(grid, x::Number) = rand_v_vals(grid) * x
+rand_e_vals(grid) = rand(ne(grid))
+rand_e_vals(grid, x::Number) = rand_e_vals(grid) * x
+function make_u0_matrix(value_map, vals, symbols)
```
Vilin97, line 11: A good practice is to keep randomness in tests as minimal as possible, because randomness leads to flaky tests. Sometimes it is unavoidable because the thing being tested is random but we should not introduce new randomness. Randomness also makes the test easier to understand, which is a problem, because tests are themselves untested. 
If you feel that you have to generate the initial conditions randomly, set a seed to make the test reproducible.

test/spatial_test_networks.jl
Vilin97, line 1: This file is in the `test` folder but has no `@test` in it. It also does not appear in `runtests`. What is the point of this file? I would recommend throwing in a couple of `@test` or removing this file. 
If you just want to keep these scripts around, you can make a [gist](https://gist.github.com/) or make a `scripts` branch in your fork of `Catalyst` and keep it there.

src/lattice_reaction_system_diffusion.jl
```diff
@@ -0,0 +1,501 @@
...
+    all_diff_rates = [haskey(diffusion_rates_dict, s) ? diffusion_rates_dict[s] : [0.0]
+                      for s in species(lrs)]
+    if length.(all_diff_rates) == 1
+        return Catalyst.matrix_expand_component_values(all_diff_rates,
+                                                       length(vertices(lrs.lattice)))
+    else
+        hopping_constants = [Vector{Float64}() for i in 1:(lrs.nS), j in 1:(lrs.nC)]
+        for (e_idx, e) in enumerate(edges(lrs.lattice))
+            for s_idx in 1:(lrs.nS)
+                push!(hopping_constants[s_idx, e.src],
```
Vilin97, line 439: `push!` is slow, because it allocates memory several times. Since we know the length of the array (it's `lrs.nS`), we should pre-allocate it.

src/lattice_reaction_system_diffusion.jl
```diff
@@ -0,0 +1,501 @@
...
+                          findfirst(column_view .== get_index(edge.dst, s, lrs.nS)) - 1
+            J_nzval[val_idx_dst] += get_component_value(rates, e_idx)
+        end
+    end
+
+    return SparseMatrixCSC(lrs.nS * lrs.nC, lrs.nS * lrs.nC, J_colptr, J_rowval, J_nzval)
+end
+
+### JumpProblem ###
+
```
Vilin97, line 398: It feels like the `JumpProblem` codebloack below belongs in its own separate file. As well as the ODE codebloack above. The reason is that they do not interact. It leads to better maintanability. For a simple example, when you update the `JumpProcesses.JumpProblem` in subsequent PRs, it will be nicer to have that change made in a file that is 70 lines long.

You might also want to make a `src/spatial` folder and put the new files there.

src/lattice_reaction_system_diffusion.jl
```diff
@@ -0,0 +1,501 @@
...
+# Creates a vector of DiffusionReactions.
+function diffusion_reactions(diffusion_reactions)
+    [DiffusionReaction(dr[1], dr[2]) for dr in diffusion_reactions]
+end
+# Gets the parameters in a diffusion reaction.
+ModelingToolkit.parameters(dr::DiffusionReaction) = Symbolics.get_variables(dr.rate)
+
+### Lattice Reaction Network Structure ###
+# Desribes a spatial reaction network over a graph.
+struct LatticeReactionSystem # <: MT.AbstractTimeDependentSystem # Adding this part messes up show, disabling me from creating LRSs
```
Vilin97, line 56: I think the `LatticeReactionSystem` deserves its own file as well.

Your `src/spatial` folder might look like
- `lattice_eaction_system.jl`
- `ode_problem.jl`
- `jump_problem.jl`
- `utils.jl`
- `diffusion_reaction.jl`

The benefits of keeping these in one folder but separate files is easier review process in the future and easier code navigation (tree-like structure is faster to navigate than linear structure), just to name a few benefits. In tests, you can do the same thing and put spatial tests in the `tests/spatial` folder.

test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
```diff
@@ -0,0 +1,309 @@
...
+            jprob = JumpProblem(lrs, dprob, NSM())
+            @test jprob.prob.u0 == true_u0
+            @test jprob.discrete_jump_aggregation.hop_rates.hop_const_cumulative_sums == true_hopping_rates
+        end
+        # Provides parameters as a combined vector.
+        for pC in [pC_1], pD in [pD_1, pD_2]
+            dprob = DiscreteProblem(lrs, u0, (0.0, 100.0), [pD; pC])
+            jprob = JumpProblem(lrs, dprob, NSM())
+            @test jprob.prob.u0 == true_u0
+            @test jprob.discrete_jump_aggregation.hop_rates.hop_const_cumulative_sums == true_hopping_rates
```
Vilin97, line 112: Let's also test that the massaction jumps are correct, along with the hopping rates.

test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
Vilin97, line 1: I suggest adding an integration test as well. In the spatial ODE test file, you actually solve the problem and check the solution. Let's do something like that here as well. To do this, we can just take a test from `JumpProcesses`. For example, let's use the [`ABC` test](https://github.com/SciML/JumpProcesses.jl/blob/master/test/spatial/ABC.jl#L25). On lines [55](https://github.com/SciML/JumpProcesses.jl/blob/master/test/spatial/ABC.jl#L55C7-L55C18) and 59 we create `JumpProblem`s. Instead of creating them from [`MassActionJump`](https://github.com/SciML/JumpProcesses.jl/blob/master/test/spatial/ABC.jl#L25), we will create them from a `LatticeReactionSystem` using your new code. The [test](https://github.com/SciML/JumpProcesses.jl/blob/master/test/spatial/ABC.jl#L62) itself can be copy-pasted.

Vilin97 (changes requested):
Thank you, Torkel! This is a great PR! My main comments/suggestions are:
1. Remove the code in tests that is not run in `runtests.jl`.This includes the large `if false` codeblock, the ODE performance test, and `spatial_test_networks.jl`. We do not want to keep dead code around. You can convert those scripts to gists or put them in a `scripts` branch in your fork of `Catalyst`.
2. Make an integration test for `JumpProblem`. Just reuse the [`ABC.jl` test](https://github.com/SciML/JumpProcesses.jl/blob/master/test/spatial/ABC.jl#L62) from `JumpProcesses`.
3. Split the large file in `src` into several smaller files in `src/spatial` or whatever name you think is best. This will make the code more maintainable.

test/spatial_test_networks.jl
TorkelE: It is called from `lattice_reaction_systems_ODEs.jl` and `lattice_reaction_systems_jumps.jl`. Basically, it contains all the networks that are used for the tests in these files (and kept here so that they can be reused in several test files).

src/lattice_reaction_system_diffusion.jl
```diff
@@ -0,0 +1,501 @@
...
+        error("Currently, for lattice jump simulations, spatial non-diffusion parameters are not supported.")
+    ___dprob = DiscreteProblem(lrs.rs, reshape(dprob.u0, lrs.nS, lrs.nC), dprob.tspan,
+                               first.(dprob.p[1]))
+    ___jprob = JumpProblem(lrs.rs, ___dprob, aggregator;
+                           hopping_constants = hopping_constants,
+                           spatial_system = lrs.lattice,
+                           combinatoric_ratelaws = combinatoric_ratelaws, checks = checks)
+    (length(___jprob.variable_jumps) != 0) &&
+        error("Currently, for lattice jump simulations, variable rate jumps are not supported.")
+    return ___jprob.massaction_jump
```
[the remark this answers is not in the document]
TorkelE: I am not to worried about the performance cost, as we are only creating a non-spatial system, but agree that it is non-ideal. Presumably, there should be some way to create the `MassActionJump` directly from a `ReactonSystem`. I tried to follow how this was done for non-spatial systems, but lost track through the code. I was hoping you or @isaacsas could advice on the best approach (see comment in code).

src/lattice_reaction_system_diffusion.jl
TorkelE: Could point, will update.

test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
TorkelE: I was meaning to, but ran into some weird problem, will see if I can recreate what happened.

test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
TorkelE: Sounds good, yeah, will try getting these as well.

TorkelE:
Thanks for the comments! Will be unavailable today/tomorrow, but will get to it this weekend.

src/lattice_reaction_system_diffusion.jl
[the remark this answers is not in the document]
Vilin97:
```suggestion
function make_majumps(lrs::LatticeReactionSystem)
    jumps = assemble_jumps(lrs.rs)
    if !(jumps[end] isa MassActionJump)
        error("Only MassAction Jumps are currently allowed in spatial simulation.")
    return jumps
```

test/spatial_test_networks.jl
Vilin97: Okay, having a seed is good. I would still recommend using hard-coded values though. If you truly don't care which choice the rng makes, just make the choice yourself. Or even just hard-code the values that are produced by this seed. There are several benefits to this:
1. Test breakage. RNGs sometimes change implementation under the hood, which leads to the same seed producing different randomness. This might break the test or make people afraid it will break the test.
2. Test readability. Random initialization unnecessarily introduces additional cognitive load on the reader of the code. Code is read **at least** 10 times more often than it is written. We should not make the reader dig in and run this test themselves just to learn which values this seed generates.

src/lattice_reaction_system_diffusion.jl
Vilin97: See my comment below for a suggestion for how to split it. Just a suggestion though.

test/spatial_test_networks.jl
isaacsas: StableRngs are explicitly for uses like CI, and are promised to remain stable for a given seed across releases. So they will return the same sequence of random numbers every time. That’s why we use it for code that needs a `rng`. That said, to use the stable rng one has to call  always call`rand(rng)`, otherwise the Julia default rng is called, so the stable rng isn’t being used here. (Independently, as you say it isn’t clear if one really needs to use a random value in this case either.)

TorkelE:
I tried the version you provided:
```julia
function make_majumps(lrs::LatticeReactionSystem; combinatoric_ratelaws=get_combinatoric_ratelaws(lrs.rs))
    jumps = assemble_jumps(lrs.rs; combinatoric_ratelaws=combinatoric_ratelaws)
    (jumps[end] isa MassActionJump) || error("Only MassAction Jumps are currently allowed in spatial simulation.")
return jumps
end
```
Unfortunately it returns an error. I think it is the same reason which we now need to create a spatial jump system and use its `massaction_jump` to work. Still no clue why.

TorkelE:
This branch won't check now, unsure, but seems to be something with conflict, locally it is fine and give no indications of the problem, so not sure what the problem is. Will avoid touching . I am afraid that this depend on the spatial ODE PR, which might not have the right base since the balancing got merged (even though that should be independent of this). Trying to correct this one at this stage feels it might create a proper mess, so will wait a bit.

isaacsas:
You can probably merge master into the original PR to update it, then merge the original PR branch into this to update here.

src/compound.jl
Vilin97, line 1: You should drop these changes to resolve the merge conflict with the master branch. You only did some formatting here, so it's not a big deal to drop.

src/lattice_reaction_system_diffusion.jl
```diff
@@ -0,0 +1,507 @@
...
+    pC, pD = lattice_process_p(p_in, Symbol.(compartment_parameters(lrs)),
+                               Symbol.(diffusion_parameters(lrs)), lrs)
+    return DiscreteProblem(lrs.rs, u0, tspan, (pC, pD), args...; kwargs...)
+end
+
+# Builds a spatial JumpProblem from a DiscreteProblem containg a Lattice Reaction System.
+function JumpProcesses.JumpProblem(lrs::LatticeReactionSystem, dprob, aggregator, args...;
+                                   name = nameof(lrs.rs),
+                                   combinatoric_ratelaws = get_combinatoric_ratelaws(lrs.rs),
+                                   checks = false, kwargs...)
```
Vilin97, line 414: Instead of asking the user a question, let's help them understand what they did wrong and how to fix it.
```suggestion
        error("Parameters in input DiscreteProblem is of an unexpected type: $(typeof(dprob.p)). This might be caused by passing a `LatticeReactionProblem` into the `DiscreteProblem`. Pass a `ReactionProblem` instead.")
```

test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
```diff
@@ -0,0 +1,165 @@
...
+                        dprob = DiscreteProblem(lrs, u0, (0.0, 100.0), (pV, pE))
+                        jprob = JumpProblem(lrs, dprob, NSM())
+                        solve(jprob, SSAStepper())
+                    end
+                end
+            end
+        end
+    end
+end
+
```
Vilin97, line 62: There is no point in having this if there is no `@test` called. I suggest deleting it.
```suggestion

```

TorkelE: Basically, they are meant to check that the simulation doesn't error under a wide range of circumstances, but I will add a check at the end to check the retcode of the solve output indicates a successful simulation.

test/spatial_test_networks.jl
TorkelE: `rng` is nowed used properly.

test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
TorkelE: Dead code has been removed.

src/lattice_reaction_system_diffusion.jl
TorkelE: I have circumvented using `push!`

src/lattice_reaction_system_diffusion.jl
TorkelE: I have now separated main folder and tests into separate files.

test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
TorkelE: I have transferred the ABC test and compared the Catalyst version.

src/lattice_reaction_system_diffusion.jl
TorkelE: done

test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
TorkelE: Added tests for this as well

TorkelE:
Rebased on https://github.com/SciML/Catalyst.jl/pull/644

TorkelE:
Superseded by https://github.com/SciML/Catalyst.jl/pull/737

Issue #672, opened 2023-09-08
Separate inverse problem documentation section, and expanded documentation for these


TorkelE:
I have been disucssing with Sebastian at https://github.com/sebapersson/PEtab.jl/issues/61 regarding direct integration between Catalyst and PEtab.jl for parameter calibration. I think we quite soon should be able to provide a relatively straightforward interface to the user, and I want to write a tutorial of this.

If so, I was thinking it would make sense to split of the inverse problem stuff into its own section (We currently have a single page on this, under "Catalyst application"). With time, we could provide guidelines for how to combine with DiffEqParamEstim, SciML/Optimization, PEtab.jl, and Turing.

TorkelE:
The https://github.com/SciML/Catalyst.jl/pull/674 PR splits of an "Inverse Problem" documentation section.

Currently, this is my plan for documentation to add to it:

- [ ] Introduction to parameter calibration for ODEs. Basically goes through the options (PEtab, SciML/Optimization, Turing), and when to use which.
- [ ] Parameter calibration for ODEs using PEtab.jl.
- [ ] Parameter calibration for ODEs using SciML/Optimization.jl and DiffEqParamEstim.jl.
- [ ] Parameter calibration for ODEs using Turing.jl.
- [ ] Stochastic system parameter calibration using moment matching.
- [ ] Structural identifiability analysis.
- [ ] Practical identifiability analysis.
- [ ] Approximation reaction rates using Neural Networks.
- [ ] Other uses of SciML/Optimization.jl. E.g. specifying a system behaviour and finding values in parameter space optimizing that behaviour.

Maybe also something more general about parameter fitting for Jump or SDE systems? We should also incorporate https://github.com/SciML/Catalyst.jl/pull/655/files somewhere. I am also thinking of moving https://docs.sciml.ai/Catalyst/stable/catalyst_applications/parameter_estimation/ to "Example Networks" (or Exampel applications), and then simply linking that tutorial in one of the above mentioned ones).

TorkelE:
I've added sensitivity analysis. 

Depending on how PEtab.jl turns out, I'm thinking of making it the go-to recommendation for doing parameter calibration/fitting (since it has an easier to use and has a more complete interface than SciML/Optimization), with SciML/Optimization's advantage being some additional flexibility for advanced users (hence the order). PEtab.jl is also mainly an extension of SciML stuff already. Happy to change the order though.

[...]

PR #674, opened 2023-09-12
Split inverse problem documentation

TorkelE:
This creates a new Documentation tab, "Inverse Problems". Right now, the "Parameter Estimation" tutorial is moved from "Catalyst Application" to "Inverse Problems".

If this passes and people are happy, I am good to have this merged, and then I will start adding the documentation outlined in https://github.com/SciML/Catalyst.jl/issues/672 in separate PRs.


docs/pages.jl
```diff
@@ -10,8 +10,8 @@ pages = Any["Home" => "index.md",
             "Catalyst Applications" => Any["catalyst_applications/simulation_structure_interfacing.md",
                                            "catalyst_applications/advanced_simulations.md",
                                            "catalyst_applications/homotopy_continuation.md",
-                                           "catalyst_applications/bifurcation_diagrams.md",
-                                           "catalyst_applications/parameter_estimation.md"],
+                                           "catalyst_applications/bifurcation_diagrams.md"],
+            "Inverse Problems" => Any["inverse_problems/parameter_estimation.md"],
             "Example Networks" => Any["example_networks/basic_CRN_examples.md",
                                       "example_networks/hodgkin_huxley_equation.md",
                                       "example_networks/smoluchowski_coagulation_equation.md"],
```
isaacsas, line 17:
```suggestion
            "Example Networks" => Any["example_networks/basic_CRN_examples.md",
                                                             "example_networks/hodgkin_huxley_equation.md",
                                                             "example_networks/smoluchowski_coagulation_equation.md"],
             "Inverse Problems" => Any["inverse_problems/parameter_estimation.md"],
```
I think it makes more sense to show examples of building models before showing an analysis that uses models.

isaacsas: Otherwise looks good.

TorkelE: I don't think any of the models under "Inverse Problems" would depend on those in "Example Networks". My thought was that the "Inverse Problem" section is basically equivalent to "Catalyst Application" section, i.e. showing how to do certain analysis (but cut off into its own section to avoid creating a really long one).

isaacsas:
Maybe we should swap the Catalyst applications and example sections too. The latter is Catalyst-focused, showing how to build more advanced networks just within Catalyst, while the latter is focused on how Catalyst interfaces with other packages. So the new ordering would be

1. Examples
2. Applications
3. Parameter Estimation

I'd still prefer if we keep the first parameter estimation tutorial focused on SciML tooling, in part because such approaches can be used for the other model types we support too, and are the most flexible (but also do require more to be done by the user). Then in the special case of ODE models we can have a tutorial that introduces PETab and explains how it provides a simple interface and more automated way to fit such models.

TorkelE:
Ok, let's put the SciML optimization before the petab one. I'm still not convinced by the ordering (as I was envisioning tutorials in Application and Parameter estimation making references to Examples, and also generally consider Example the part people might read last/the part not actually needed to learn how do something). However, so we can move forward with stuff we can merge in that order for a starter.

isaacsas:
Sounds good.

TorkelE:
I have
- Moved parameter fitting to a new inverse problem section at the end.
- moved the network examples section to a subsection of Catalyst Functionality (called "Model creation examples")

TorkelE:
If this sounds good to you @isaacsas I will merge (once tests passes)

TorkelE:
Closed by https://github.com/SciML/Catalyst.jl/pull/701

PR #678, opened 2023-09-17
[v14 ready] Improve noise scaling implementation

TorkelE:
One of the more niche features is the ability to (linearly) scale the noise magnitude in the CEL: https://docs.sciml.ai/Catalyst/stable/catalyst_applications/advanced_simulations/#Scaling-the-noise-magnitude-in-the-chemical-Langevin-equations. The current interface was a bit "clanky". 

New new interface (as opposed to the new one from last year that was meant to override the old one).

Reactions can be given an `noise_scaling` metadata. This is any expression, which is multiplied by the noise terms generated by that reaction in the CLE. Any normal expression is valid. E.g
```julia
using Catalyst, StochasticDiffEq, Plots

# Note that that `0 <--> Y` is two reactions, so metadata have to be supplied to each.
rn = @reaction_network begin
    (p,d), 0 <--> X
    (p,d), 0 <--> Y, ([noise_scaling=0.0], [noise_scaling=0.0]) 
end

u0 = [:X => 1.0, :Y => 1.0]
tspan = (0.0, 10.0)
ps = [:p => 10.0, :d => 1.0]
sprob = SDEProblem(rn, u0, tspan, ps)
sol = solve(sprob, ImplicitEM())
plot(sol)
```
image

If you want to have a parameter as a noise scaling parameter (like previously), you have to create this separately:
```julia
rn = @reaction_network begin
    @parameters η
    (p,d), 0 <--> X
    (p,d), 0 <--> Y, ([noise_scaling=η], [noise_scaling=η]) 
end

u0 = [:X => 1.0, :Y => 1.0]
tspan = (0.0, 10.0)
ps = [:p => 10.0, :d => 1.0, :η => 0.1]
sprob = SDEProblem(rn, u0, tspan, ps)
sol = solve(sprob, ImplicitEM())
plot(sol)
```
image

For cases where you want the same noise scaling in all reaction, I have added a `@default_noise_scaling` option:
```julia
rn = @reaction_network begin
    @parameters η
    @default_noise_scaling η
    (p,d), 0 <--> X
end

u0 = [:X => 1.0]
tspan = (0.0, 10.0)
ps = [:p => 10.0, :d => 1.0, :η => 0.0]
sprob = SDEProblem(rn, u0, tspan, ps)
sol = solve(sprob, ImplicitEM())
plot(sol)
```
image

You can still supply `noise_scaling` metadata to individual reactions when using `@default_noise_scaling` (I'm which case the default value is overridden).

I considered to have a option `@default_noise_scaling_parameter η` that combines 
```
@parameters η
@default_noise_scaling η
```
but decided to keep it like this.

Since we are having a breaking release, and probably the last one in a while, I am deprecating the previous syntax fully (only an error message left).


TorkelE:
I haven't written tests. If this update looks good I will fix that and update the docs as well.

TorkelE:
If you prefer not to allow mixed symbols/symbolics I could change that. generally, I think I prefer using `:X`, rather than `rn.X`, mostly because
```julia
σV_network = @reaction_network begin
    v0 + hill(σᵛ,v,K,n), ∅ → (σᵛ+A)
    deg, (σᵛ,A,Aσᵛ) → ∅
    (kB,kD), A + σᵛ ↔ Aσᵛ
    L*kC*Aσᵛ, Aσᵛ ⟾ σᵛ
end
p = [σV_network.v0 = 0.1, σV_network.v = 2.5, σV_network.K = 60, n = 2, σV_network.kD = 5, σV_network.kB = 10, σV_network.kC = 0.05, σV_network.deg = 0.01; σV_network.L = 0.]
```
becomes really messy. Either way, that's a discussion we can take another time.

isaacsas:
Yeah, why don't you get the functionality in here without changing the symbol mappings. Then we can discuss ways to improve their construction as a separate issue / PR.

isaacsas:
Also, and I haven't looked at your code closely, but can you make sure to still support the old style noise scaling for now, but add a deprecation warning that it will be removed for this new version in a future release? We should give an error if someone uses both. I'd prefer not to have to make a breaking release.

TorkelE:
Currently, we don't change the previous way. The previous way is basically how it is handled internally anyway, always felt a bit hacky and incomplete. The new version uses the interface of the previous way.

isaacsas:
OK, but if someone passes it the previous way via a kwarg it should now give a deprecation warning but still work, unless they pass it both the new way and the old way, in which case an error should be printed. That way we aren't making a breaking change, but we are warning users the old interface will go away in the future.

src/reactionsystem.jl
```diff
@@ -1414,6 +1425,20 @@ function Base.convert(::Type{<:SDESystem}, rs::ReactionSystem;
               kwargs...)
 end
 
+# Extracts any noise scaling parameters from a reaction system.
+function get_noise_scaling(rs::ReactionSystem)
```
isaacsas, line 1429: Generally MTK get functions only return what is in the top level system by convention, i.e. `get_ps(rs)` vs. `parameters(rs)`. Maybe make two versions `get_noise_scalings` and `noise_scalings` that are analogous to `get_ps` and `parameters`. Also, to be consistent with other MTK functions I'd just return the vector from `filter` and not have this variable return type.

src/reactionsystem.jl
```diff
@@ -1495,7 +1520,7 @@ end
 # SDEProblem from AbstractReactionNetwork
 function DiffEqBase.SDEProblem(rs::ReactionSystem, u0, tspan,
                                p = DiffEqBase.NullParameters(), args...;
-                               noise_scaling = nothing, name = nameof(rs),
+                               noise_scaling = get_noise_scaling(rs), name = nameof(rs),
```
isaacsas, line 1523: I'd suggest we deprecate this, so by default it is `nothing`. If a user passes anything here and `get_noise_scaling` is empty we can use it, but give a warning that it is deprecated. Likewise, if a user already set `get_noise_scaling` and uses this we give an error that both can't be used. Then in a future breaking release we can drop this kwarg, which is non-standard for `SDEProblem` anyway.

isaacsas (changes requested):
We already have lots of solving tests, which are pretty expensive on CI time, and I thought we already had noise scaling tests? So here you can just test if the generated problem gets the right parameters in it, and avoid more solving tests.

src/reactionsystem.jl
TorkelE: Sounds good, yeah, I will do this.

src/reactionsystem.jl
isaacsas: Oh never mind. I missed that this is internal only right? So don't worry about this comment. Since it will only be called within `SDEProblem` this shouldn't be an issue.

TorkelE:
New update:
- Attempting to scale noise by providing a kwarg to `SDEProblem` now throws a deprecation warning (or an error, if one also defined some parameters as noise scaling paraemters.
- Support for mixed symbols/symbolics in input vectors dropped.
- The previous test was superfluous given the new one, so I dropped it. Of the 3 remaining tests, I changed it so that only one actually runs a simulation.

src/networkapi.jl
```diff
@@ -563,6 +565,17 @@ pmap   = symmap_to_varmap(sys, [:β => 1.0, :ν => 1.0, :subsys₊k => 1.0])
...
+```julia
+sir = @reaction_network sir begin
+    β, S + I --> 2I
+    ν, I --> R
+end
+@unpack S = sir
+mixed_u0map = [S => 1.0, :I => 1.0, :R => 1.0]
+u0map  = symmap_to_varmap(sir, mixed_u0map)
+```
+
```
isaacsas, line 578:
```suggestion
```

```

File: src/networkapi.jl [hunk 4]
```diff
@@ -504,7 +504,9 @@ end
 
 # convert symbol of the form :sys.a.b.c to a symbolic a.b.c
 function _symbol_to_var(sys, sym)
-    if hasproperty(sys, sym)
+    if sym isa Num
+        return sym
+    elseif hasproperty(sys, sym)
```
isaacsas, line 509:
```suggestion
```

src/reaction_network.jl
```diff
@@ -364,6 +364,9 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
     parameters_declared = extract_syms(options, :parameters)
     variables = extract_syms(options, :variables)
 
+    # Handles a (potential) noise scaling parameter.
+    #noise_scaling_p_args = handle_noise_scaling_ps!(parameters_declared, options)
+
```
isaacsas, line 369:
```suggestion
```

src/reaction_network.jl
```diff
@@ -404,6 +407,9 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
         push!(rxexprs.args, get_rxexprs(reaction))
     end
 
+    # Adds potentital noise scaling parameters.
```
isaacsas, line 410:
```suggestion
```

src/reaction_network.jl
```diff
@@ -531,8 +537,26 @@ function get_pexpr(parameters_extracted, options)
...
+           (ns_expr.args[idx] isa Expr) && (ns_expr.args[idx].head == :(=)) # Parameter on form η=0.1.
+            if idx < length(ns_expr.args) && (ns_expr.args[idx+1] isa Expr)  && (ns_expr.args[idx+1].head == :vect)
+                push!(ns_expr.args[idx+1].args,:(noisescalingparameter=true))
+            else
+                insert!(ns_expr.args, idx+1, :([noisescalingparameter=true]))
+            end
+        end
+    end
+    return ns_expr.args[3:end]
+end
```
isaacsas, line 559: Will this work if the user gives additional metadata supported by MTK? Why is all this special code being used? Can't we do something like we do for `@species` and just create it as a normal parameter, and then add one extra line of generated code to attach the metadata?

src/reactionsystem.jl
```diff
@@ -1500,6 +1526,10 @@ function DiffEqBase.SDEProblem(rs::ReactionSystem, u0, tspan,
                                include_zero_odes = true, checks = false,
                                check_length = false,
                                remove_conserved = false, kwargs...)
+    if !isnothing(noise_scaling) 
+        any(isnoisescalingparameter.(parameters(rs))) && error("You have declared some paraemters as noise scaling parameters, and also given a \"noise_scaling\" argument to SDEProblem. Please remove the \"noise_scaling\", as this way of scaling CLE noise is being depricated.")
```
isaacsas, line 1530: This check should be in `convert` I think, since people may call that manually.

isaacsas: In convert it if you add it after flattening on `flatrs` you can just do
```julia
any(isnoisescalingparameter, get_ps(flatrs))
```
to avoid any allocations.

src/reactionsystem.jl
```diff
@@ -54,6 +54,17 @@ end
 # variables
 drop_dynamics(s) = isconstant(s) || isbc(s) || (!isspecies(s))
 
+### Other Metadata Fields ###
+
+# Denotes that a parameter controls the scaling of noise in the CLE.
+struct NoiseScalingParameter end
+Symbolics.option_to_metadata_type(::Val{:noisescalingparameter}) = NoiseScalingParameter
+
+isnoisescalingparameter(s::Num) = isnoisescalingparameter(MT.value(s))
```
isaacsas, line 63:
```julia
isnoisescalingparameter
```
is sufficiently long that I'd suggest adding underscores. It is hard to read as is.

test/model_simulation/simulate_SDEs.jl
```diff
@@ -121,29 +121,59 @@ end
...
+        @noise_scaling_parameters η[1:2]
+        (k1, k2), X1 ↔ X2 
+    end
+    @unpack k1, k2, η = noise_scaling_network_2
+    sol_2_1 = solve(SDEProblem(noise_scaling_network_2, u0, (0.0, 1000.0), [k1 => 2.0, k2 => 0.66, η[1] => 2.0, η[2] => 2.0]), ImplicitEM(); saveat=1.0)
+    sol_2_2 = solve(SDEProblem(noise_scaling_network_2, u0, (0.0, 1000.0), [k1 => 2.0, k2 => 0.66, η[1] => 2.0, η[2] => 0.2]), ImplicitEM(); saveat=1.0)
+    sol_2_3 = solve(SDEProblem(noise_scaling_network_2, u0, (0.0, 1000.0), [k1 => 2.0, k2 => 0.66, η[1] => 0.2, η[2] => 0.2]), ImplicitEM(); saveat=1.0)
+    @test var(sol_2_1[1,:]) > var(sol_2_2[1,:]) > var(sol_2_3[1,:]) 
+end
+
```
isaacsas, line 146: What are these covering that wouldn't be covered by an existing test?

isaacsas (changes requested):

test/model_simulation/simulate_SDEs.jl
TorkelE: They cover that the amount of noise is actually changed when you change the noise scaling parameter. The second one checks that the feature works when symbolic are used rather than symbols. If you want I could remove the second case, although I figured it was better to have as well, but CI time hasn't really been a main concern, which might be unwise.

src/reactionsystem.jl
TorkelE: Just took the standard of other metadata, but agree. Should I change:
`isnoisescalingparameter` to `is_noise_scaling_parameter`
and
`noisescalingparameter` to `noise_scaling_parameter` (where this is used, e.g. when declaring metadata)
?

test/model_simulation/simulate_SDEs.jl
isaacsas: Our CI is taking 26 minutes, see below, so generally I think we need to be mindful of not adding things that increase time while not providing new coverage. Should we instead just update the older noise scaling tests to use your new interface, maybe keeping one of them with the old style kwarg? That should be sufficient that it is propagating the values correctly. We already have tests that pass parameters to ODEs/SDEs, so I don't think we need new tests for that.

Moreover, given that these are SDEs, are these tests guaranteed to always pass? We don't want tests that can randomly fail due to poor luck with random numbers (and at a minimum we should be using `StableRNGs` for any new tests involving stochastic solvers).

src/reactionsystem.jl
TorkelE: I was thinking that this wasn't possible for some reason, but you are totally right, this should be moved to `covert`.

src/reactionsystem.jl
isaacsas: Sure, sounds good.

src/reaction_network.jl
TorkelE: There are 3 cases for detecting parameters on these forms:
```
@noise_scaling_parameters p1 p2[1:2] p3=1.0
```

Yes, additional metadata is supported, see tests. E.g
```
@noise_scaling_parameters p1 [description="A parameter]
```
works.

Doing like for species might be a good idea, yeah, will check and try to copy that.

test/model_simulation/simulate_SDEs.jl
TorkelE: The number differences should be so huge that random fails shouldn't be possible but good point. I removed the previous noise scaling test as they were worse than the current one, and not testing anything different.

Good point on CI, yeah, I haven't really thought much about CI time, but should probably. If I have some extra time some time I will try to go through CI to figure out what is taking time, and if something can be cut.

src/reaction_network.jl
TorkelE: I think I could down the species path. However, in that case it seems I might have to create a separate `noise_ps_expr` and add at the end as:
```julia
# Returns the rephrased expression.
quote
    $ps
    $noise_ps_expr
    $ivexpr
    $vars
    $sps

    Catalyst.make_ReactionSystem_internal($rxexprs, $tiv, union($spssym, $varssym),
                                          $pssym; name = $name,
                                          spatial_ivs = $sivs)
end
```

isaacsas: I think that’s fine to add a line for it after the `ps`.

isaacsas: Reusing the existing MTK stuff and just adding the needed metadata as an extra step seems much more future proof to me.

TorkelE: Agreed, will modify the code.

TorkelE:
I'm sorry, but I am stuck. I've been trying all evening, but have not been able to reuse the species approach, having tried several versions on it. I think I am almost there, although it required a new `@noise_scaling_parameters` macro (similar to the `@species` macro) which is what is pasted at the end. However, the resulting output is not enough as the result is not considered a `parameter`.

Here is my macro:
```julia
macro noise_scaling_parameters(ex...)
    vars = Symbolics._parse_parameters(:parameters, Real, ex)

    # vector of symbols that get defined
    lastarg = vars.args[end]

    # start adding metadata statements where the vector of symbols was previously declared
    idx = length(vars.args)
    resize!(vars.args, idx + length(lastarg.args) + 1)
    for sym in lastarg.args
        vars.args[idx] = :($sym = ModelingToolkit.wrap(setmetadata(ModelingToolkit.value($sym),
                                                                    Catalyst.NoiseScalingParameter,
                                                                    true)))
        idx += 1
    end

    # check nothing was declared isconstantspecies
    ex = quote end
    vars.args[idx] = ex
    idx += 1

    # put back the vector of the new species symbols
    vars.args[idx] = lastarg

    esc(vars)
end
```
however
```julia
using Catalyst
@parameters k1 k2
@variables t
@species X1(t) X2(t)
reaction = Reaction(k1, [X1], [X2], [1], [1])
@named rs = ReactionSystem([reaction], t, [X1, X2], [k1, k2, η])
```
gives `ERROR: ArgumentError: η is not a parameter.`.

I try putting a `println(vars)` at the end of `@noise_scaling_parameters` and `@species` to see how they are different. Here `@noise_scaling_parameters η` gives
```
begin
    η = (identity)((Symbolics.wrap)((SymbolicUtils.setmetadata)((Sym){Real}(:η), Symbolics.VariableSource, (:parameters, :η))))
    η = ModelingToolkit.wrap(setmetadata(ModelingToolkit.value(η), Catalyst.NoiseScalingParameter, true))
    begin
        #= /home/torkelloman/.julia/dev/Catalyst/src/reaction_network.jl:696 =#
    end
    [η]
end
```
and `@species K(t)`:
```
begin
    K = (identity)((Symbolics.wrap)((SymbolicUtils.setmetadata)(((Sym){(SymbolicUtils.FnType){NTuple{1, Any}, Real}}(:K))((Symbolics.value)(t)), Symbolics.VariableSource, (:variables, :K))))
    K = ModelingToolkit.wrap(setmetadata(ModelingToolkit.value(K), Catalyst.VariableSpecies, true))
    begin
        #= /home/torkelloman/.julia/dev/Catalyst/src/reaction_network.jl:147 =#
        all(!(Catalyst.isconstant) ∘ ModelingToolkit.value, [K]) || throw(ArgumentError("isconstantspecies metadata can only be used with parameters."))
    end
    [K]
end
```
but I cannot interpret this very well. `Symbolics.VariableSource` could have something with variables to do, but I am unsure how to change this.

isaacsas:
We should make it a normal standalone macro too though. Maybe get that working first?

isaacsas:
I can take a closer look later today.

isaacsas:
@TorkelE I think the issue is you aren't calling `_parse_vars` correct for parameters. See how the `@parameters` macro is defined in MTK:

https://github.com/SciML/ModelingToolkit.jl/blob/1aa480673305854e889be2211eb5b0e4487038ae/src/parameters.jl#L58-L62

src/reaction_network.jl
```diff
@@ -407,12 +411,13 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
...
     quote
         $ps
+        $ns_ps
         $ivexpr
         $vars
         $sps
 
         Catalyst.make_ReactionSystem_internal($rxexprs, $tiv, union($spssym, $varssym),
-                                              $pssym; name = $name,
+                                              union($pssym, $ns_pssym); name = $name,
```
isaacsas, line 420: Can we avoid a union here in the very common case that no noise scaling is given? i.e. define a variable that is the `ps` if there is no noise scaling, and the union otherwise, and then pass it in here.

src/reactionsystem.jl
```diff
@@ -1371,7 +1382,7 @@ Notes:
   differential equations.
 """
 function Base.convert(::Type{<:SDESystem}, rs::ReactionSystem;
-                      noise_scaling = nothing, name = nameof(rs),
+                      noise_scaling = get_noise_scaling(rs), name = nameof(rs),
```
isaacsas, line 1385:
```suggestion
                      noise_scaling = nothing,
```
given that you later check if it is `nothing` and set it to this anyways?

src/reactionsystem.jl
```diff
@@ -1380,6 +1391,7 @@ function Base.convert(::Type{<:SDESystem}, rs::ReactionSystem;
     flatrs = Catalyst.flatten(rs)
     error_if_constraints(SDESystem, flatrs)
 
+    isnothing(noise_scaling) && (noise_scaling = get_noise_scaling(rs)) # Required until passing nosie into SDEProblem can be depricated.
```
isaacsas, line 1394: This needs to use `flatrs` not `rs`.

isaacsas: Also, make sure `flattening`, `merging`, and all the other various API functions properly copy over the noise scaling field from system to system. (i.e. `extend`, `compose`, etc) You probably need to go through this file and the network api file carefully to check you haven't missed anywhere. Finally, make sure to update equality testing in the network API file.

src/reactionsystem.jl
```diff
@@ -1414,6 +1426,20 @@ function Base.convert(::Type{<:SDESystem}, rs::ReactionSystem;
               kwargs...)
 end
 
+# Extracts any noise scaling parameters from a reaction system.
+function get_noise_scaling(rs::ReactionSystem)
+    ns_params = filter(p -> is_noise_scaling_parameter(p), parameters(rs))
+    if isempty(ns_params)
+        return nothing
+    elseif length(ns_params) == 1
+        return Num(ns_params[1])
```
isaacsas, line 1435: Don't use `Num`s. I hope you can please try to keep this in mind going forward, as this seems to be a recurring issue.

src/reactionsystem.jl
```diff
@@ -1500,6 +1526,12 @@ function DiffEqBase.SDEProblem(rs::ReactionSystem, u0, tspan,
                                include_zero_odes = true, checks = false,
                                check_length = false,
                                remove_conserved = false, kwargs...)
+
+    if !isnothing(noise_scaling)
+        !isnothing(get_noise_scaling(rs)) && error("You have declared some parameters as noise scaling parameters, and also given a \"noise_scaling\" argument to SDEProblem. Please remove the \"noise_scaling\", as this way of scaling CLE noise is being depricated.")
+        @warn "Passing noise scaling input into SDEProblem will be deprecated. New standard is to declare one (or several) paraemter as noise scaling parameters when the ReactionSystem is created. Please read https://docs.sciml.ai/Catalyst/stable/catalyst_applications/advanced_simulations/#Scaling-the-noise-magnitude-in-the-chemical-Langevin-equations." 
+    end
```
isaacsas, line 1533: Shouldn't this check be in the `convert` call?

isaacsas (changes requested):
- Tests for standalone `@noise_scaling_parameters` macro?
- Test that interpolation works with noise scaling? Both inside and outside the DSL? i.e.
```julia
a = :b
@noise_scaling_parameters $(a) 
# check that the symbolic b works / becomes available?
```

src/reactionsystem.jl
TorkelE: This basically has to do with the fact that we are proofing for a deprecated feature. In principle, the kwarg should be `noise_scaling = get_noise_scaling(rs)`, and `SDEProblem` should not submit any `noise_scaling` value. However, since we previously allowed `noise_scaling` in `SDEProblem`, we have to keep this functionality running. 

I let it be `noise_scaling = get_noise_scaling(rs)` since that is the future version, thinking it would make things easier when we deprecate things. However, I will change it to nothing and add a comment about it to the code.

src/reactionsystem.jl
TorkelE: Sorry, will fix. Think I made a couple of these when I didn't really understand the difference between `Num` and other stuff. Maybe next time we meet you can go through the various symbolics representations, and why we use which in various places, understanding that would probably help.

isaacsas: `Num` is a wrapper around all symbolics stuff to use in the REPL primarily. So a `Num` can wrap a single scalar symbolic, but also a symbolic expression. The rule in MTK/Catalyst is to always unwrap these via either `ModelingToolkit.unwrap(s::Num)` or `ModelingToolkit.value(s::Num)` to get the actual symbolic. There are some MTK/Symbolics functions that don't work with `Num`s, which will then cause issues when adding code if we store them. An example:
```julia
julia> using ModelingToolkit
@parameters k
julia> @parameters k
1-element Vector{Num}:
@v arik

julia> @variables t A(t)
2-element Vector{Num}:
    t
 A(t)

julia> using ModelingToolkit: value

julia> typeof(k),typeof(value(k))
(Num, SymbolicUtils.BasicSymbolic{Real})

julia> typeof(A),typeof(value(A))
(Num, SymbolicUtils.BasicSymbolic{Real})

julia> typeof(k*A),typeof(value(k*A))
(Num, SymbolicUtils.BasicSymbolic{Real})

julia> typeof(k*A),typeof(value(k*A^2))
(Num, SymbolicUtils.BasicSymbolic{Real})

julia> typeof(Num(1)),typeof(value(Num(1)))
(Num, Int64)

julia> typeof(k*A^2*sin(A)),typeof(value(k*A^2*sin(A)))
(Num, SymbolicUtils.BasicSymbolic{Real})
```
so it looks like everything is a `BasicSymbolic{Real}` now actually.

TorkelE:
> * Tests for standalone `@noise_scaling_parameters` macro?
> * Test that interpolation works with noise scaling? Both inside and outside the DSL? i.e.
> 
> ```julia
> a = :b
> @noise_scaling_parameters $(a) 
> # check that the symbolic b works / becomes available?
> ```

Wil lfix this

TorkelE:
Your comments should be fixed.

I have also updated to check that `@noise_scaling_parameters` work in normal code, and that this can be used to create a programmatic network with scaling. One of the previous tests have been updated to use this approach.

Regarding interpolation to make this work:
```julia
η_stored = :η
rn = @reaction_network begin
    @noise_scaling_parameters :(η_stored)
    (p, d), 0 <--> X
end
```
this is currently not possible due to https://github.com/SciML/Catalyst.jl/issues/692.

I also wanted to use interpolation for the `@noise_scaling_parameters` test. However, this is not possible:
```julia
η_stored = :η

@variables t
@species X1(t) X2(t)
@noise_scaling_parameters $(η_stored)=0.0
@parameters k1 k2
r1 = Reaction(k1,[X1],[X2],[1],[1])
r2 = Reaction(k2,[X2],[X1],[1],[1])
@named noise_scaling_network = ReactionSystem([r1, r2], t, [X1, X2], [k1, k2, η])
```
since simply running `η` in the main scope causes an error (preventing me to pass it into `noise_scaling_network`), again, see https://github.com/SciML/Catalyst.jl/issues/692.

I could add tests for these two cases and mark them as broken, but until we figure https://github.com/SciML/Catalyst.jl/issues/692 out I think that is as far as we can get.

isaacsas:
The second (non-DSL) example should work -- it works with `@parameters`. As I said in the issue the first is ok to not have working as it isn't how we set the DSL up to work with `@parameters` either.

TorkelE:
If it is desired behaviour I will do the second one with
```julia
η_stored = :η

@variables t
@species X1(t) X2(t)
ηs = @noise_scaling_parameters $(η_stored)=0.0
@parameters k1 k2
r1 = Reaction(k1,[X1],[X2],[1],[1])
r2 = Reaction(k2,[X2],[X1],[1],[1])
@named noise_scaling_network = ReactionSystem([r1, r2], t, [X1, X2], [k1, k2, ηs[1]])
```

TorkelE:
Since we are more or less set on syntax, I have rewritten the noise scaling documentation given this update. I have also:
- Updated the test as described.
- Added an additional getter; `noise_scaling_parameters` which returns a system's noise scaling parameters. The main reason is that when there are several reactions, the order is used to decide which parameter scales the noise of which reaction, so I wanted a good function allowing people to confirm this order. This method is not ideal (but the case is rather niche). At some point we might be able to add metadata for reactions or something. At least when creating reaction systems programmatically, it might be possible to make a nicer interface for how to do this, but I'd say that's beyond the scope of this PR.

For the second case, I tried to add API for `noise_scaling_parameters()` to the docs. I have never really done this properly before, so you might want to check it is as you want it to be.

docs/src/catalyst_applications/advanced_simulations.md
```diff
@@ -307,59 +307,89 @@ tspan = (0.0, 10.0)
 p_1 = [:k1 => 1.0, :k2 => 1.0]
 
 sprob_1 = SDEProblem(rn_1, u0, tspan, p_1)
-sol_1 = solve(sprob_1)
+sol_1 = solve(sprob_1, ImplicitEM())
```
isaacsas, line 310: Why switch from using the default solver?

docs/src/catalyst_applications/advanced_simulations.md
```diff
@@ -307,59 +307,89 @@ tspan = (0.0, 10.0)
...
 Setting *η<1.0* will reduce noise and *η>1.0* will increase noise. The syntax
 for setting a noise scaling parameter `η` is
 ```@example ex3
 rn_2 = @reaction_network begin
-    @parameters η
+    @noise_scaling_parameters η
     (k1,k2), X1 <--> X2
 end
+```
+The `@noise_scaling_parameter` option is equivalent to the `@parameters` option, but also designates subsequent parameter as a *noise scaling parameter*. *η* becomes a parameter of the system, and we can now modualte its value to scale simualtion noise:
```
isaacsas, line 327:
```suggestion
The `@noise_scaling_parameter` option creates one or more new parameters with additional metadata that allows Catalyst to know they represent a noise scaling. *η* becomes a parameter of the system, and we can now modulate its value to scale simulation noise:
```

docs/src/catalyst_applications/advanced_simulations.md
```diff
@@ -307,59 +307,89 @@ tspan = (0.0, 10.0)
...
+sprob_2 = SDEProblem(rn_2, u0, tspan, p_2)
+sol_2 = solve(sprob_2, ImplicitEM())
 plot(sol_2; idxs = :X1, ylimit = (0.0, 20.0))
 ```
 
-Finally, it is possible to set individual noise scaling parameters for each
-reaction of the system. Our model has two reactions (`X1 --> X2` and `X2 -->
-X1`) so we will use two noise scaling parameters, `η1` and `η2`. We use the
-following syntax:
+It is worth noting that in the CLE, nosie is tied to *reactions* (and not species, which is a common missperception). If only a single noise scaling parameter is given, it will scale the noise for all reaction. However, it is also possible to set several nosie scaling parameters, with each scaling the noise of a single reaction. Our model has two reactions (`X1 --> X2` and `X2 --> X1`) so we will use two noise scaling parameters (`η1` and `η2`):
```
isaacsas, line 338:
```suggestion
It is worth noting that in the CLE, nosie is tied to *reactions* (and not species, which is a common missperception). If only a single noise scaling parameter is given, it will scale the noise for all reactions. However, it is also possible to set several noise scaling parameters, with each scaling the noise of a single reaction. Our model has two reactions (`X1 --> X2` and `X2 --> X1`) so we will use two noise scaling parameters (`η1` and `η2`):
```

docs/src/catalyst_applications/advanced_simulations.md
```diff
@@ -307,59 +307,89 @@ tspan = (0.0, 10.0)
...
 u0 = [:X1 => 10.0, :X2 => 10.0]
 tspan = (0.0, 10.0)
 p_3 = [:k1 => 1.0, :k2 => 1.0, :η1 => 0.1, :η2 => 1.0]
 
-sprob_3 = SDEProblem(rn_3, u0, tspan, p_3; noise_scaling = @parameters η1 η2)
+sprob_3 = SDEProblem(rn_3, u0, tspan, p_3)
 ```
-plotting the results, we see that we have less fluctuation than for the first
-simulation, but more as compared to the second one (which is as expected):
+Both the noise scaling parameters and the reaction are ordered (these orders can be seen by calling `reactions(rn_3)` and `noise_scaling_parameters(rn_3)`, respectively). The i'th noise scaling parameter scales the noise of the i'th reaction. Plotting the results, we see that we have less fluctuation than for the first simulation, but more as compared to the second one (which is as expected):
```
isaacsas, line 350: Relying on an ordering seems clunky and very much not what one should do, though I know it is how we did it in the past. It seems likely to cause problems too if one is composing / extending systems together as the ordering may change.

isaacsas (commented):
It really feels like maybe we should consider an interface here in which a `Reaction` takes a noise scaling parameter as a field. Then one could control the scaling on a per-reaction basis. Or at least somehow we attach a mapping from `Reaction` to noise scaling parameter somewhere in the system.

If we do add it to reactions we'll have to make sure stuff like

https://github.com/SciML/Catalyst.jl/blob/046ce758e2fc88f9989502abf3e498f7830ab6f4/src/reactionsystem.jl#L231

gets updated appropriately.

src/reactionsystem.jl
```diff
@@ -945,6 +945,17 @@ function get_metadata(reaction::Reaction, md_key::Symbol)
...
+"""
+    get_noise_scaling(reaction::Reaction)
+
+Retrives the nosie scaling factor for the reaction's noise term (as generated by the reaction in an SDE). 
+If a specific noise scaling expression have been provided in the reactions "noise_scaling" metadata,
+returns that. If not, returns `1.0`.
+"""
+function get_noise_scaling(reaction::Reaction)
+    has_metadata(reaction, :noise_scaling) ?  get_metadata(reaction, :noise_scaling) : 1.0
+end
```
isaacsas, line 957: The approach should be to have a separate `has_noise_scaling` function and check that is made, and one then only calls this when a noise scaling exists (otherwise giving an error).

isaacsas: That is more consistent with how we and MTK handle other optional fields.

src/reactionsystem.jl
```diff
@@ -1064,7 +1075,7 @@ function assemble_diffusion(rs, sts, ispcs, noise_scaling; combinatoric_ratelaws
 
     for (j, rx) in enumerate(get_rxs(rs))
         rlsqrt = sqrt(abs(oderatelaw(rx; combinatoric_ratelaw = combinatoric_ratelaws)))
-        (noise_scaling !== nothing) && (rlsqrt *= noise_scaling[j])
+        rlsqrt *= get_noise_scaling(rx)
```
isaacsas, line 1078:
```suggestion
        has_noise_scaling(rx) && (rlsqrt *= get_noise_scaling(rx))
```

src/reactionsystem.jl
```diff
@@ -1474,50 +1485,30 @@ Notes:
...
 """
 function Base.convert(::Type{<:SDESystem}, rs::ReactionSystem;
-                      noise_scaling = nothing, name = nameof(rs),
-                      combinatoric_ratelaws = get_combinatoric_ratelaws(rs),
+                      name = nameof(rs), combinatoric_ratelaws = get_combinatoric_ratelaws(rs),
                       include_zero_odes = true, checks = false, remove_conserved = false,
                       default_u0 = Dict(), default_p = Dict(), defaults = _merge(Dict(default_u0), Dict(default_p)),
+                      noise_scaling = nothing, # To be removed (used for deprication message only).
                       kwargs...)
+    isnothing(noise_scaling) || error("Supplying a noise_scaling agument to SDESystems have been depricated. Please see the \"Scaling the noise magnitude in the chemical Langevin equations\" section of the \"Advanced Simulation Options\" page of the documentation for the new approach.")
```
isaacsas, line 1500: Just drop this since we are making a breaking release and add it to the HISTORY for V14 that it is no longer supported here.

src/reactionsystem.jl
```diff
@@ -1555,7 +1546,9 @@ function Base.convert(::Type{<:JumpSystem}, rs::ReactionSystem; name = nameof(rs
                       combinatoric_ratelaws = get_combinatoric_ratelaws(rs),
                       remove_conserved = nothing, checks = false,
                       default_u0 = Dict(), default_p = Dict(), defaults = _merge(Dict(default_u0), Dict(default_p)),
+                      noise_scaling = nothing, # To be removed (used for deprication message only).
                       kwargs...)
+    isnothing(noise_scaling) || error("Supplying a noise_scaling agument to SDESystems have been depricated. Please see the \"Scaling the noise magnitude in the chemical Langevin equations\" section of the \"Advanced Simulation Options\" page of the documentation for the new approach.")
```
isaacsas, line 1551: remove

isaacsas (changes requested):
Many of the tests seem pretty removed from actually testing that the scaling is working. Is there not something more precise that could be tested? For example, manually creating the same symbolic SDEs and checking if the generated noise functions evaluate to the same values on specific inputs (checking if the symbolic noise terms are the same would be even better I'd think)? That seems a more direct test than calling solve like is done here.

TorkelE:
> Many of the tests seem pretty removed from actually testing that the scaling is working. Is there not something more precise that could be tested? For example, manually creating the same symbolic SDEs and checking if the generated noise functions evaluate to the same values on specific inputs (checking if the symbolic noise terms are the same would be even better I'd think)? That seems a more direct test than calling solve like is done here.

Could add some direct tests as well. The idea was basically to check in real simulations that the level of noise is affected by the noise scaling. But I agree that direct test of the function would work as well.

HISTORY.md
```diff
@@ -22,8 +22,9 @@ rn = @reaction_network begin
     @parameters η
     k, 2X --> X2, [noise_scaling=η]
 end
-getnoisescaling(rn)
+get_noise_scaling(rn)
 ```
+- `SDEProblem` no longer takes the `noise_scaling` argument (she above for new approach to handle noise scaling).
```
isaacsas, line 27:
```suggestion
- `SDEProblem` no longer takes the `noise_scaling` argument (see above for new approach to handle noise scaling).
```

docs/src/catalyst_applications/advanced_simulations.md
```diff
@@ -292,8 +292,9 @@ plot(sol)
 
 ## Scaling the noise magnitude in the chemical Langevin equations
 When using the CLE to generate SDEs from a CRN, it can sometimes be desirable to
-scale the magnitude of the noise terms. This can be done by introducing a *noise
-scaling parameter*. First, we simulate a simple two-state CRN model using the
+scale the magnitude of the noise terms. Here, each reaction of the system generates a separate noise term in the CLE. If you require identical scaling for all reaction, the `@default_noise_scaling` option can be used. Else, you can supply a `noise_scaling` metadata for each individual reaction, describing how to scale the noise for that reaction.
```
isaacsas, line 295:
```suggestion
scale the magnitude of the noise terms. Here, each reaction of the system generates a separate noise term in the CLE. If you require identical scaling for all reactions, the `@default_noise_scaling` option can be used. Else, you can supply a `noise_scaling` metadata for each individual reaction, describing how to scale the noise for that reaction.
```

docs/src/catalyst_applications/advanced_simulations.md
```diff
@@ -306,58 +307,61 @@ tspan = (0.0, 10.0)
...
-
-Next, we wish to introduce a noise scaling parameter ,`η`. This will scale the
-noise magnitude so that for $η=0.0$ the system lacks noise (and its SDE
-simulations are identical to its ODE simulations) and for $η=1.0$ noise is not
-scaled (and SDE simulations are identical to as if no noise scaling was used).
-Setting $η<1.0$ will reduce noise and $η>1.0$ will increase noise. The syntax
-for setting a noise scaling parameter `η` is
+Here we can see that the $X$ concentration fluctuates around a steady state of $X≈10.0$.
+
+Next, we wish increase the amount of noise with by a factor 2. To do so, we use the `@default_noise_scaling` option, to which we provide the desired scaling
```
isaacsas, line 315:
```suggestion
Next, we wish increase the amount of noise by a factor 2. To do so, we use the `@default_noise_scaling` option, to which we provide the desired scaling 
```

src/reaction_network.jl
```diff
@@ -651,7 +658,8 @@ function push_reactions!(reactions::Vector{ReactionStruct}, sub_line::ExprValues
     for i in 1:maximum(lengs)                       
         # If the `only_use_rate` metadata was not provided, this has to be infered from the arrow used.
         metadata_i = get_tup_arg(metadata, i)
-        if all(arg.args[1] != :only_use_rate for arg in metadata_i.args)
+        merge_metadata!(metadata_i, default_reaction_metadata)
```
isaacsas, line 661: Having metadata for the same things at both the system and individual reaction level seems quite likely to lead to confusing behavior and bugs. Why have both? If one wants to give metadata for a reaction one should specify it at the reaction level. If one wants to apply metadata uniformly to all reactions, i.e. like a uniform noise scaling, that seems better implemented as a system transformation that just regenerates all the reactions with the desired scaling inplace. i.e. `scaledsys = scale_reaction_noise_by(scaling, sys)` or such, with this function adding the scaling into each reaction individually.

isaacsas: We can just add such a function to the API, and it will keep the DSL code and internal code simpler since we don't have to worry about having mutually interacting metadata in multiple places...

test/model_simulation/simulate_SDEs.jl
```diff
@@ -119,31 +119,122 @@ let
...
+# Tries to create a large number of problem, ensuring there are no errors (cannot solve as solution likely to go into negatives). 
+let
+    for reaction_network in reaction_networks_all
+        for factor in [1e-2, 1e-1, 1e0, 1e1]
+            u0 = factor * rand(rng, length(states(reaction_network)))
+            p = factor * rand(rng, length(parameters(reaction_network)))
+            prob = SDEProblem(reaction_network, u0, (0.0, 1.0), p)
+        end
+    end
+end
```
isaacsas, line 133: What is this testing that isn't already tested? Don't we have lots of tests that create SDEProblems already?

isaacsas: Also this isn't using the correct form of initial conditions and parameters, which need to be mappings going forward as we've discussed.

test/model_simulation/simulate_SDEs.jl
```diff
@@ -119,31 +119,122 @@ let
...
+        (p,d), 0 <--> X2, ([description="Y"],[description="Y"])
+        (p,d), 0 <--> X3
+    end
+    noise_scaling_network = complete(noise_scaling_network)
+    
+    u0 = [:X1 => 1000.0, :X2 => 1000.0, :X3 => 1000.0]
+    ps = [noise_scaling_network.p => 1000.0, noise_scaling_network.d => 1.0, noise_scaling_network.η1 => 1.0]
+    sol = solve(SDEProblem(noise_scaling_network, u0, (0.0, 1000.0), ps), ImplicitEM(); saveat=1.0)
+    @test var(sol[:X1]) < var(sol[:X2]) 
+    @test var(sol[:X1]) < var(sol[:X3])
```
isaacsas, line 213: Would it not be better to just write a function to extract the noise coefficient from each equation in the generated `SDESystem` and then check it is correct for each term? That seems more robust than the `sol` test here (which I don't really see much value to, and seems like it could randomly fail on us).

isaacsas (changes requested):
@TorkelE lots of typos, misspellings in the docs here. Please give a careful reread -- I stopped flagging them after a point.

In the docs it would probably be good to show the actual scaled mathematical equations of at least one example to make completely clear how the scaling comes into play.

Finally, I don't like this mixing of noise scaling being metadata at both system and reaction levels and the changes it then induces in the DSL code (making it more complicated for such a niche case). It seems like if one wants to apply a uniform scaling to all reactions and not specify it when creating reactions, this would be better to be implemented as a system transformation function in the API as I mention in my comments.

ChrisRackauckas:
> Finally, I don't like this mixing of noise scaling being metadata at both system and reaction levels and the changes it then induces in the DSL code (making it more complicated for such a niche case). It seems like if one wants to apply a uniform scaling to all reactions and not specify it when creating reactions, this would be better to be implemented as a system transformation function in the API as I mention in my comments.

Agreed here, the interpretation is not necessarily unique like this.

TorkelE:
Just want to point out that we never mix system and reaction level metadata, that is just a misunderstanding. The only case where we (currently) use reaction metadata is one where you (in almost all cases) want the same value across all reactions. Hence, I added an option
```
@default_noise_scaling η
```
which just sets that input (here `η`) as the noise scaling across all reactions. In case someone actually gives a `noise_scaling` metadata to one reaction, that will override this default. This is only something we have on system creation (and currently only for noise scaling). It is never anything that is stored on the system level.

(This is something I've used personally to avoid writing the same noise scaling for 30 reactions in a stress response model)

TorkelE:
See above responses. I will have another rearead through the doc thing. Generally, I didn't spend as much time on it as I should, because I had already had to re-write it once unnecessarily, and this is a part that I plan to rewrite once we redo parts of the docs (and I didn't want to spend too much effort on it until that is done).

isaacsas:
But why not add a default via a system transformation instead of a special DSL command. The former would have the benefit of preserving equivalence between DSL and symbolic construction, and work for both. The system transformation could be something like `newsys = scale_reaction_noise(sys, scaling; keep_default_scaling = true)` or such, and would then be usable on any `ReactionSystem` no matter how it is created.

isaacsas:
Even if we keep the DSL command, I'd advocate that it simply correspond to applying that transformation function after creating the system within the DSL instead of merging scalings via processing of the DSL expressions. It seems cleaner to me, and ensures consistency between DSL and symbolic approaches (plus it seems like a useful function to have in general).

TorkelE:
I just don't see why this specific thing should be a system transformation, and not:
- Adding metadata to a system.
- Adding default values to species/parameters
- Specifying that a parameter should be of the `Int64` type

and so on. Right now there is no part of model creation that is handled as a system transformation (if we ignore composition) rather than just being something you define at system creation. There might be some good reason why just noise scaling in SDE should be the exception, but I think in that case we'd have to discuss that in person.

We can add a default argument for noise scaling in the `ReactionSystem` if you want

isaacsas:
Just to address some of these points:

>  Adding metadata to a system.

We don't use system metadata? (I know MTK has a notion of it, but I don't think it is used by anything except JuliaHub internal stuff?)

> Adding default values to species/parameters

This was an MTK choice long ago, so not really something we control. I agree it is confusing to have three ways to specify the default value of a parameter or initial condition (i.e. via metadata to the symbolic, via defaults, or via the mappings). But that just supports the idea that we should only handle noise scaling in one way (i.e. as metadata within reactions), and not try to also handle it with, for example, a conflicting system property that complicates things more.

> Specifying that a parameter should be of the Int64 type

This is a new choice made by MTK, and not something we control in Catalyst, but I don't see how it relates to this PR at all. I personally would have preferred if this decision was only made at problem construction time, as it reduces flexibility in modeling as you've observed in some of the broken tests I think. But it is done so we need to deal with it.

Here is why I prefer making this a transformation:
1. It doesn't change the DSL setup you've created in that one can still have the default scaling command. It just eliminates  some messy expression hacking since one doesn't need to reconcile that command with scalings passed via reaction metadata. (i.e. instead at the end of the DSL we just conclude with generating a `sys = apply_default_scaling(sys, scaling)` or such command if needed.
2. It immediately makes for `ReactionSystem`s one is just given, including systems coming from SBML or that are symbolically created. This seems like a nice gain.
3. I think it is much easier to understand, reason, and in the future update, a simple function that loops over reactions and adds the appropriate metadata scaling to them vs. trying to read through and update DSL expression hacking code. I generally would strongly prefer that we minimize DSL updates that require expression hacking to keep the code as compact and simple as possible (within the constraint of now trying to maintain 1:1 compatibility with the symbolic interface). To be honest, I'd like to at some point redo the DSL code and, if possible, switch to what has been built for MTK to reduce what we need to maintain here even more.

Anyways, we can discuss more on Wednesday if needed.

TorkelE:
I agree in principle, but think this is a combination of refactoring and new features, and something that can be saved for a later day. Yes, discussing it on Wednesday is probably better to ensure that we are on the same page.

src/reactionsystem.jl
```diff
@@ -756,6 +756,55 @@ function ReactionSystem(iv; kwargs...)
...
+    Each metadata in each `Reaction` of the updated `ReactionSystem` will have the value desiganted in
+    `default_reaction_metadata` (however, `Reaction`s that already have that metadata designated will not
+    have their value updated).
+"""
+function remake_ReactionSystem_internal(rs::ReactionSystem;  default_reaction_metadata = [])
+    # Updates the metadata for all reactions (equation are ignored).
+    updated_equations = [set_default_metadata(rx, default_reaction_metadata) for rx in reactions(rs)]
+    updated_reactions = [set_default_metadata(rx, default_reaction_metadata) for rx in reactions(rs)]
+    @set! rs.eqs = updated_equations
+    @set! rs.rxs = updated_reactions
```
isaacsas, line 779: This is both flattening implicitly by using `reactions(rs)` instead of `get_rxs(rs)`, and also is dropping any non-reaction equations. Remember, `equations(sys)` and `reactions(rs)` include all subsystems' equations and reactions too. To only pull from the top-level system, use `get_eqs(sys)` and `get_rxs(sys)`. Those are the field accessors.

It is also creating multiple copies of the reactions in the new `eqs` and `rxs` fields instead of having the vectors point to the same `Reaction`s in memory. I think you want something more like
```suggestion
    eqtransform(eq) = eq isa Reaction ? set_default_metadata(eq, default_reaction_metadata) : eq
    updated_equations = map(eqtransform, get_eqs(rs))
    @set! rs.eqs = updated_equations
    @set! rs.rxs = Reaction[rx for rx in updated_equations if rx isa Reaction]
```

src/reactionsystem.jl
```diff
@@ -756,6 +756,55 @@ function ReactionSystem(iv; kwargs...)
...
+function set_default_metadata(rx::Reaction, default_metadata)
+    missing_metadata = filter(md -> !in(md[1], entry[1] for entry in rx.metadata), default_metadata)
+    updated_metadata = vcat(rx.metadata, missing_metadata)
+    updated_metadata = convert(Vector{Pair{Symbol, Any}}, updated_metadata)
+    return @set rx.metadata = updated_metadata
+end
+set_default_metadata(eq::Equation, default_metadata) = eq
+
+"""
+    remake_noise_scaling(rs::ReactionSystem, noise_scaling)
```
isaacsas, line 793:
```suggestion
    scale_noise(rs::ReactionSystem, noise_scaling)
```
I think this is a more concise name, and don't really see the need for `remake` here.

You might, however, consider a keyword `preserve_existing_scalings` that allows users to control if existing values are overwritten?

isaacsas (changes requested):

src/reaction_network.jl
```diff
@@ -651,7 +656,7 @@ function push_reactions!(reactions::Vector{ReactionStruct}, sub_line::ExprValues
     for i in 1:maximum(lengs)                       
         # If the `only_use_rate` metadata was not provided, this has to be infered from the arrow used.
         metadata_i = get_tup_arg(metadata, i)
-        if all(arg.args[1] != :only_use_rate for arg in metadata_i.args)
+        if all([arg.args[1] != :only_use_rate for arg in metadata_i.args])
```
isaacsas, line 659:
```suggestion
        if all(arg -> arg.args[1] != :only_use_rate, metadata_i.args)
```
Why allocate a vector here?

src/reactionsystem.jl
```diff
@@ -756,6 +756,71 @@ function ReactionSystem(iv; kwargs...)
...
+    `default_reaction_metadata` (however, `Reaction`s that already have that metadata designated will not
+    have their value updated).
+"""
+function remake_ReactionSystem_internal(rs::ReactionSystem;  default_reaction_metadata = [])
+    rs = set_default_metadata(rs;  default_reaction_metadata)
+    return rs
+end
+
+# For a `ReactionSystem`, updates all `Reaction`'s default metadata.
+function set_default_metadata(rs::ReactionSystem;  default_reaction_metadata = [])
```
isaacsas, line 781: Does the scaling just have to be a parameter or can it be a more general expression? 

Somewhere we probably need a check that the passed in metadata only involves parameters/species/variables that are already in our system, or we also need to add those to the new system.

test/model_simulation/simulate_SDEs.jl
```diff
@@ -119,29 +119,158 @@ let
...
+    noise_scaling_network_2 = @reaction_network begin 
+        @parameters η[1:2]
+        (k1, k2), X1 ↔ X2, ([noise_scaling=η[1]],[noise_scaling=η[2]])  
+    end
+    @unpack k1, k2, η = noise_scaling_network_2
+    sol_2_1 = solve(SDEProblem(noise_scaling_network_2, u0, (0.0, 1000.0), [k1 => 2.0, k2 => 0.66, η[1] => 2.0, η[2] => 2.0]), ImplicitEM(); saveat=1.0)
+    sol_2_2 = solve(SDEProblem(noise_scaling_network_2, u0, (0.0, 1000.0), [k1 => 2.0, k2 => 0.66, η[1] => 2.0, η[2] => 0.2]), ImplicitEM(); saveat=1.0)
+    sol_2_3 = solve(SDEProblem(noise_scaling_network_2, u0, (0.0, 1000.0), [k1 => 2.0, k2 => 0.66, η[1] => 0.2, η[2] => 0.2]), ImplicitEM(); saveat=1.0)
+    @test var(sol_2_1[1,:]) > var(sol_2_2[1,:]) > var(sol_2_3[1,:]) 
+end
```
isaacsas, line 158: I thought I commented on this. Aren't these tests no guaranteed to always pass? The variance could happen for a given random number stream to not satisfy these inequalities right? These need to be deterministic tests that will always succeed (for example, but using the same rng stream each time).

isaacsas: More generally though I don't understand this test. It isn't testing accuracy in any real quantitative way, so what is the point? I'd suggest picking a statistic of a specific system with a specific scaling and checking if you recover that statistic to a given accuracy with enough samples.

test/model_simulation/simulate_SDEs.jl
isaacsas: This has the same issue with the test potentially failing doesn't it?

PR #681, opened 2023-09-18
Documentation of initial conditions that depend on a parameter

TorkelE:
It is possible for an initial condition to be a parameter:
```julia
using Catlyst
rn = @reaction_network begin
  @species X(t)=1.0
  @parameters p=1.0 d=0.1
  p, 0 --> X
  d, X --> ∅
end

using DifferentialEquations, Plots
u0 = []
tspan = (0.0, 10.0)
p = []
oprob = ODEProblem(rn, u0, tspan, p)
sol = solve(oprob)
plot(sol)
```
which is actually quite neat. We don't document (or test) this feature though, so I have added both.

### DIFF
--- a/docs/src/catalyst_functionality/dsl_description.md
+++ b/docs/src/catalyst_functionality/dsl_description.md
@@ -412,6 +412,26 @@ sol = solve(oprob)
 plot(sol)
 ```
 
+## Setting initial conditions that depend on parameters
+It is possible to set the initial condition of one (or several) species so that they depend on some system parameter. This is done in a similar way as default initial conditions, but giving the parameter instead of a value. When doing this, we also need to ensure that the initial condition parameter is a variable of the system:
+```@example tut2
+rn = @reaction_network begin
+  @species X(t)=X0
+  @parameters X0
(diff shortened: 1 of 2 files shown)


docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -412,6 +412,26 @@ sol = solve(oprob)
 plot(sol)
 ```
 
+## Setting initial conditions that depend on parameters
+It is possible to set the initial condition of one (or several) species so that they depend on some system parameter. This is done in a similar way as default initial conditions, but giving the parameter instead of a value. When doing this, we also need to ensure that the initial condition parameter is a variable of the system:
+```@example tut2
+rn = @reaction_network begin
+  @species X(t)=X0
+  @parameters X0
```
isaacsas, line 420:
```suggestion
  @parameters X0
  @species X(t)=X0
```
so it is consistent with how one would write it via the symbolic interface.

isaacsas (commented):
Aside from that one comment LGTM. Thanks!

PR #682, opened 2023-09-19
Test cleanup

TorkelE:
I went through the tests, trying to see if there is anything that can be cut. Most of it is complication time, however, I found some Jump simulations across two files where I could remove some probably superfluous repeats of the simulations. It only removed like 1 minute in the end, had hoped to gain more.

While I went through stuff I removed some unnecessary statements, and only added a few missing StableRNG uses (where `rand` was called without StableRNG input).

### DIFF
--- a/test/dsl/dsl_basics.jl
+++ b/test/dsl/dsl_basics.jl
@@ -2,7 +2,7 @@
 
 ### Fetch Packages and Set Global Variables ###
 
-using Catalyst, ModelingToolkit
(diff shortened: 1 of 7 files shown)


isaacsas (approved):
To be honest at some point we should probably update the jump solution tests to use examples with known solutions / solution statistics as in JumpProcesses, and compare to those. All these ones you are modifying are not such great accuracy tests.

PR #683, opened 2023-09-19
Homotopy Continuation extension and remake

TorkelE:
With the introduction of package extensions (https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) we can add some additional support for packages like HomotopyContinuation without making incompatibilities messier.

This introduces a single function, `hc_steady_states` which finds the steady states of a reaction system using homotopy continuation. If initial conditions are provided, conservation laws are automatically utilised. This feature is stored in a new "HomotopyContinuationExtension" extension, and only available if HomoyopyContinuation is loaded.

This is the full docstring of the new function:
    `hc_steady_states(rs::ReactionSystem, ps; filter_negative=true, neg_thres=-1e-20, u0=typeof(ps)(), kwargs...)`

Uses homotopy continuation to find the steady states of the ODE system corresponding to the provided reaction system.

Arguments:
- `rs::ReactionSystem`: The reaction system for which we want to find the steady states.
- `ps`: The parameter values for which we want to find the steady states.
- `filter_negative=true`: If set to true, solutions with any species concentration <0 is removed from the output.
- `neg_thres=-1e-20`: Determine the minimum values for which a species concentration is to be considred non-negative. Species conentrations ``> neg_thres`` but `< 0.0` are set to `0.0`.
- `u0=typeof(ps)()`: Initial conditions for which we want to find the steady states. For systems with conservation laws this are required to compute conserved quantities.
- `kwargs...`: any additional arguments (like `show_progress= true`) are passed into HomotopyContinuation.jl's `solve` call. 

Examples:
```@repl
rs = @reaction_network begin
    k1, Y --> 2X
    k2, 2X --> X + Y
    k3, X + Y --> Y
    k4, X --> 0
end
ps = [:k3 => 1.0, :k2 => 2.0, :k4 => 1.5, :k1=>8.0]
hc_sol = hc_steady_states(rs, ps)
```
gives
```
[0.5000000000000002, 2.0000000000000004]
[0.0, 0.0]
[4.499999999999999, 5.999999999999999]
```

Notes:
- This is a wrapper around the `solve` function provided by HomotopyContinuation.jl, all credit for this functionality to that package's authors.


ChrisRackauckas:
I do think that the specialization here should be able to be removed when it becomes a standard NonlinearSolve solver. Though I don't know what the time line on that will be.

TorkelE:
When that happens we shoudl be able to just delete the extension, but in the meantime this should be good.

isaacsas:
CI is still failing?

isaacsas:
A test failed.

TorkelE:
You mean this one:
image
it is for Julia 1.6 though, right?

I will try to check what is going on and return back, but if it is failing on 1.6 and not the latest version (if my interpretation is correct), it might be a mess to sort out, especially if the problem is in HC. Either case, will try to figure out what it is.

isaacsas:
SciML is still supporting 1.6 till the next LTS release, so we need to figure out the issue. If it is HC we should open an issue for them to fix it.

TorkelE:
ok

TorkelE:
I think I know it. 1.6 does not support extension, hence this not working. This kind of defeats the entire purpose of this PR. Ideally I would go ahead anyway, but https://pkgdocs.julialang.org/v1/creating-packages/#Transition-from-normal-dependency-to-extension seems like an OK workaround so will go with that (hopefully being able to revert it at some point.

test/runtests.jl
```diff
@@ -46,4 +46,7 @@ using SafeTestsets
     if !Sys.isapple()
         @time @safetestset "Graphs" begin include("visualization/graphs.jl") end
     end
+    
+    ### Tests extensions. ###
+    @time @safetestset "Homotopy Continuation Extension" begin include("extensions/homotopy_continuation.jl") end
```
isaacsas, line 51: The test probably also needs a check that the extension mechanism exists to run conditionally then?

TorkelE: I think this should work, given the work around (but should hold my horses until the 1.6 test actually passes).

However, an alternative would be to, for extensions, simply disable those tests on 1.6. I will let you decide.

Currently, I added 
```julia
if !isdefined(Base, :get_extension) # Workaround for 1.6 compatibility.
    include("../ext/CatalystHomotopyContinuationExtension.jl")
end
```
to Catalyst.jl to make things work on 1.6.

TorkelE: Indeed I spoke to quickly. Is disabling the test for this case as you mentioned a working solution?

isaacsas: Yeah, and going forward we can just do the same for any future extensions.

docs/src/catalyst_applications/homotopy_continuation.md
```diff
@@ -8,136 +8,75 @@ example those which are purely mass action or have only have Hill functions with
 integer Hill exponents). The roots of these can reliably be found through a
 *homotopy continuation* algorithm. This is implemented in Julia through the
 [HomotopyContinuation.jl](https://www.juliahomotopycontinuation.org/) package.
-In this tutorial, we will demonstrate how homotopy continuation can be used to
-find the steady states of mass action chemical reaction networks implemented in
-Catalyst.
+
+Catalyst contains a special homotopy continuation extension that is loaded whenever HomotopyContinuation.jl is. This exports a single function, `hc_steady_states`, that can be used to find the steady states of any `ReactionSystem` structure. If you use this in your research, please [cite the HomotopyContinuation.jl publication](@ref homotopy_continuation_citation).
```
isaacsas, line 12: Perhaps also mention the Catalyst citation?

docs/src/catalyst_applications/homotopy_continuation.md
```diff
@@ -8,136 +8,75 @@ example those which are purely mass action or have only have Hill functions with
...
-system, as well specifying the initial amount of each species,
-HomotopyContinuation can again be used to find steady states. First, we set the
-default values of the system's initial conditions and parameter values. This
-will allow the system to automatically find the conserved amounts.
-```@example hc3
-setdefaults!(two_state_model, [:X1 => 1.0, :X2 => 1.0, :k1 => 2.0, :k2 => 1.0])
-```
-Next, we create a `NonlinearSystem`, while also removing one species via the
-conservation equation.
+To find the steady states of these, an initial condition must also be provided (which is used to compute the system's conserved quantities, in this case `X1+X2`):
```
isaacsas, line 53: I think we should still say what Catalyst is doing under the hood before this. You can say it is done automatically, but we should advertise this functionality is in Catalyst...

isaacsas: We want people to appreciate that this isn't something done by HC for them, but is a Catalyst feature / functionality that many packages are lacking.

docs/src/catalyst_applications/homotopy_continuation.md
```diff
@@ -8,136 +8,75 @@ example those which are purely mass action or have only have Hill functions with
...
 ```@example hc3
-ns = convert(NonlinearSystem, two_state_model; remove_conserved = true)
+ps = [:k1 => 2.0, :k2 => 1.0]
+u0 = [:X1 => 1.0, :X2 => 1.0]
+hc_steady_states(wilhelm_2009_model, ps; u0=u0)
```
isaacsas, line 57:
```suggestion
hc_steady_states(wilhelm_2009_model, ps; u0)

```

docs/src/catalyst_applications/homotopy_continuation.md
```diff
@@ -8,136 +8,75 @@ example those which are purely mass action or have only have Hill functions with
...
-```@example hc3
-# get the X2 symbolic variable
-@unpack X2 = two_state_model
-
-# get its algebraic formula in terms of X1 and parameters
-ceqs = conservedequations(two_state_model)
-X2eqidx = findfirst(eq -> isequal(eq.lhs, X2), ceqs)
-X2eq = ceqs[X2eqidx].rhs
+## Final notes
+- `hc_steady_states` supports any systems where all rates are systems of rational polynomial (such as hill functions with integer hill coefficients).
```
isaacsas, line 61:
```suggestion
- `hc_steady_states` supports any systems where all rates are systems of rational polynomials (such as Hill functions with integer Hill coefficients).

```

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+
+"""
+    hc_steady_states(rs::ReactionSystem, ps; filter_negative=true, neg_thres=-1e-20, u0=typeof(ps)(), kwargs...)
+
+Uses homotopy continuation to find the steady states of the ODE system corresponding to the provided reaction system.
+
+Arguments:
+- `rs::ReactionSystem`: The reaction system for which we want to find the steady states.
+- `ps`: The parameter values for which we want to find the steady states.
+- `filter_negative=true`: If set to true, solutions with any species concentration <0 is removed from the output.
```
isaacsas, line 11: Isn't it then any species concentration < neg_thres?

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+- `neg_thres=-1e-20`: Determine the minimum values for which a species concentration is to be considred non-negative. Species conentrations ``> neg_thres`` but `< 0.0` are set to `0.0`.
+- `u0=typeof(ps)()`: Initial conditions for which we want to find the steady states. For systems with conservation laws this are required to compute conserved quantities.
```
isaacsas, line 13: Why assume `u0=typeof(ps)()`? Just make it `nothing` if nothing is passed in.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+```
+gives
+```
+[0.5000000000000002, 2.0000000000000004]
+[0.0, 0.0]
+[4.499999999999999, 5.999999999999999]
+```
+
+Notes:
+- This is a wrapper around the `solve` function provided by HomotopyContinuation.jl, all credit for this functionality to that package's authors.
```
isaacsas, line 35:
```suggestion
```
This doesn't seem necessary...

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
+### Homotopy Continuation Based Steady State Finding ###
+
+"""
+    hc_steady_states(rs::ReactionSystem, ps; filter_negative=true, neg_thres=-1e-20, u0=typeof(ps)(), kwargs...)
+
+Uses homotopy continuation to find the steady states of the ODE system corresponding to the provided reaction system.
```
isaacsas, line 6:
```suggestion
Uses homotopy continuation via HomotopyContinuation.jl to find the steady states of the ODE system corresponding to the provided reaction system.
```

HISTORY.md
```diff
@@ -1,6 +1,7 @@
 # Breaking updates and feature summaries across releases
 
 ## Catalyst unreleased (master branch)
+- Added a CatalystHomotopyContinuationExtension extension, which exports the `hc_steady_state` function if HomotopyContinuation is exported. `hc_steady_state` finds teh steady states of a reactin system using the homotopy continuation method.
```
isaacsas, line 4: Add an example for using it here.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+
+Notes:
+- This is a wrapper around the `solve` function provided by HomotopyContinuation.jl, all credit for this functionality to that package's authors.
+```
+  """
+function Catalyst.hc_steady_states(rs::ReactionSystem, ps; filter_negative=true, neg_thres=-1e-20, u0=typeof(ps)(), kwargs...)
+    ss_poly = steady_state_polynomial(rs, ps, u0)
+    sols = HC.real_solutions(HC.solve(ss_poly; kwargs...))
+    reorder_sols!(sols, ss_poly, rs)
+    return (filter_negative ? filter_negative_f(sols; neg_thres=neg_thres) : sols)
```
isaacsas, line 42:
```suggestion
    return (filter_negative ? filter_negative_f(sols; neg_thres) : sols)
```

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+end
+
+# For a given reaction system, paraemter values, and initial conditions, find the polynomial that HC solves to find steady states.
+function steady_state_polynomial(rs::ReactionSystem, ps, u0)
+    ns = convert(NonlinearSystem, rs; remove_conserved = true)
+    pre_varmap = map(pair -> pair::Pair{Num,Float64}, [symmap_to_varmap(rs,u0); symmap_to_varmap(rs,ps)])
```
isaacsas, line 48: I don't think you want to promote every parameter to `Float64`, for the same reason it is bad for MTK to do so. Can't you just preserve the user's input type (i.e. if they give a mixed tuple keep using tuples until you substitute in the values, and then substitute in the same type as the user specified)?

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+# For a given reaction system, paraemter values, and initial conditions, find the polynomial that HC solves to find steady states.
+function steady_state_polynomial(rs::ReactionSystem, ps, u0)
+    ns = convert(NonlinearSystem, rs; remove_conserved = true)
+    pre_varmap = map(pair -> pair::Pair{Num,Float64}, [symmap_to_varmap(rs,u0); symmap_to_varmap(rs,ps)]) 
+    conservationlaw_errorcheck(rs, pre_varmap)
+    p_vals = ModelingToolkit.varmap_to_vars(pre_varmap, parameters(ns); defaults = ModelingToolkit.defaults(ns))
+    p_dict  = Dict(parameters(ns) .=> p_vals)
+    eqs_pars_funcs = vcat(equations(ns), conservedequations(rs))
+    eqs_funcs = map(eq -> substitute(eq.rhs - eq.lhs, p_dict), eqs_pars_funcs)
+    eqs = [deregister([mm, mmr, hill, hillr, hillar], eq) for eq in eqs_funcs]
```
isaacsas, line 54: I've kind of wondered for a while if we should just not register all these and let them be substituted in directly by Symbolics? i.e. if one does 
```julia
mm2(X,v,K) = v*X/(X+K)
rn2 = @reaction_network a begin
       mm2(X,v,K), A --> B
       end
equations(rn2)
```
one gets
```julia
julia> equations(rn2)
1-element Vector{Union{Equation, Reaction}}:
 (X*v) / (K + X), A --> B
```
I guess the disadvantage is we can't autodetect them if we want to know about their presence, but we don't do that currently anyways.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+    eqs_intexp = make_int_exps.(eqs)
+    return Catalyst.to_multivariate_poly(remove_denominators.(eqs_intexp))
+end
+
+# If u0s are not given while conservation laws are present, throws an error.
+function conservationlaw_errorcheck(rs, pre_varmap)
+    vars_with_vals = union(first.(pre_varmap), keys(ModelingToolkit.defaults(rs)))
```
isaacsas, line 61:
```suggestion
    vars_with_vals = Set(p[1] for p in pre_varpmap) 
    union!(vars_with_vals, keys(ModelingToolkit.defaults(rs))
```

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+    isempty(intersect(species(rs), vars_with_vals)) || return
```
isaacsas, line 62:
```suggestion
    issubset(species(rs), vars_with_vals) && return
```

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+    isempty(conservedequations(rs)) && return 
+    error("The system have conservation laws but no initial conditions were provided. Please provide initial conditions.")
```
isaacsas, line 64:
```suggestion
    isempty(conservedequations(rs)) || 
        error("The system has conservation laws but initial conditions were not provided for all species.")
```

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+    (expr) ->
+    if istree(expr) && operation(expr) == f
+        args = arguments(expr)
+        invoke_with = map(args) do a
+            t = typeof(a)
+            issym(a) || istree(a) ? wrap(a) => symtype(a) : a => typeof(a)
+        end 
+        invoke(f, Tuple{last.(invoke_with)...}, first.(invoke_with)...)
+    end
+end
```
isaacsas, line 86: Calling invoke seems hacky. 

Perhaps we should just not register the functions to begin with? Was there any good reason we did so originally?

isaacsas: I guess such a change might need to wait for the next breaking release though.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+
+# Parses and expression and return a version where any exponents that are Float64 (but an int, like 2.0) are turned into Int64s.
+make_int_exps(expr) = wrap(Rewriters.Postwalk(Rewriters.PassThrough(___make_int_exps))(unwrap(expr))).val
```
isaacsas, line 89: I have no idea what this is doing, but why do you need to wrap and unwrap? Shouldn't equations already be unwrapped (and stay that way)?

src/Catalyst.jl
```diff
@@ -101,4 +101,13 @@ export @compound
...
 export balance_reaction
 
+### Extensions ###
+
+# HomotopyContinuation
+function hc_steady_states end
+export hc_steady_states
+if !isdefined(Base, :get_extension) # Workaround for 1.6 compatibility.
+    include("../ext/CatalystHomotopyContinuationExtension.jl")
+end
```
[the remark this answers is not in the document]
TorkelE: I was a little bit uncertain how we wanted to do it, but agree with, and prefer to, not to have 1.6 support.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+function ___make_int_exps(expr)
+    !istree(expr) && return expr
+    if (operation(expr) == ^) && isinteger(arguments(expr)[2])
+        return arguments(expr)[1] ^ Int64(arguments(expr)[2])
+    end
```
isaacsas, line 94: Shouldn't this give an error if there is a non-integer exponent, since such a model couldn't be converted to a polynomial form?

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
TorkelE: I and Shashi sat together and fixed this (well Shashi mostly). He was interested in improving this though, so we opened an issue to hopefully have some native support fro this in Symbolics: https://github.com/JuliaSymbolics/Symbolics.jl/issues/975

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
TorkelE: This was also part of Shashi's magic, we had some decent number of iterations to go reach here, so a couple of more natural solutions failed.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
TorkelE: It errors eventually, but I agree that a proper error here is way better, will add.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+    var_names_extended = String.(Symbol.(HC.variables(ss_poly)))
+    var_names = [Symbol(s[1:prevind(s, findlast('_', s))]) for s in var_names_extended]
+    sort_pattern = indexin(MT.getname.(species(rs)), var_names)
+    foreach(sol -> permute!(sol, sort_pattern), sols)
+end
+
+# Filters away solutions with negative species concentrations (and for neg_thres < val < 0.0, sets val=0.0).
+function filter_negative_f(sols; neg_thres=-1e-20)
+    for sol in sols, idx in 1:length(sol)
+        (neg_thres < sol[idx] < 0.0) && (sol[idx] = 0.0)
```
isaacsas, line 121:
```suggestion
        (neg_thres < sol[idx] < 0) && (sol[idx] = 0)
```
Just use `0`, it is more general with regards to types.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+    end
+    return filter(sol -> all(>=(0.0), sol), sols)
```
isaacsas, line 123:
```suggestion
    return filter(sol -> all(>=(0), sol), sols)
```

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
TorkelE: wasn't there some advantage that we have recorded the various derivatives of `mm` etc. , so these are known? Maybe the yare calculated at instant speed anyway though. I have no strong opinions, and maybe we should open an issue to discuss the pros/cons (I will mostly defer to your opinions though)

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,124 @@
...
+
+# If the input is a fraction, removes the denominator.
+function remove_denominators(expr)
+    s_expr = simplify_fractions(expr)
+    !istree(expr) && return expr
+    if operation(s_expr) == /
+        return remove_denominators(arguments(s_expr)[1])
+    end
+    if operation(s_expr) == +
+        return +(remove_denominators.(arguments(s_expr))...)
```
isaacsas, line 105:
```suggestion
        return sum(remove_denominators(arg) for arg in arguments(s_expr))
```
Does this work instead?

isaacsas (changes requested):
I think those are all my comments if you want to make updates.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
isaacsas: I can't remember why we did this change either unfortunately.

isaacsas: The only thought I have is we can programmatically check for hill equations via the function type, which Chris and I discussed adding a JumpProcesses specialization for.

src/Catalyst.jl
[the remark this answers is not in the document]
isaacsas: Let's just make it 1.9+, and note that in the HISTORY file.

TorkelE:
Added your changes, except for a few that feel more like long-term questions regarding symbolic function handling.  Also let `u0` default to `nothing` and didn't use the `subset` check regarding conservation laws, see comments.

src/miscellaneous_functions.jl
```diff
@@ -0,0 +1,38 @@
...
+
+Notes:
+- If a `ReactionSystem` is given, it is coverted to a `ODEFunction` internally. If you plan to compute the stability for a large number of steady states. First covert your ReactionSystem to a ODEFunction using `ODEFunction(convert(ODESystem, rs); jac=true)` and then use this in your input. 
+
+
+"""
+steady_state_stability(s_states::Vector{Float64}, ps, rs::ReactionSystem) = steady_state_stability(s_states, ODEFunction(convert(ODESystem, rs); jac=true))
+# If a ODESystem is given directyl, no conversion is needed.
+function steady_state_stability(s_states::Vector{Float64}, ps, ofun::ODEFunction)
+    maximum(eigen(ofun.jac(ss, last.(ps), Inf)).values) < 0.0
```
isaacsas, line 36: You need to look at the real part of the eigenvalues as they can be complex. Also, a system may not be unstable if the real parts are all non-negative (i.e. some are zero), so we probably shouldn't return a boolean but an enum that indicates, stable, unstable, eigenvalues are non-negative with at least one zero.

Maybe this can be moved to a future PR when we can better work out how to do this (and so we can get this merged now)?

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
isaacsas: Why not use `issubset`? It avoids allocations, unlike using `intersect`...

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -0,0 +1,127 @@
...
+    ss_poly = steady_state_polynomial(rs, ps, u0)
+    sols = HC.real_solutions(HC.solve(ss_poly; kwargs...))
+    reorder_sols!(sols, ss_poly, rs)
+    return (filter_negative ? filter_negative_f(sols; neg_thres) : sols)
+end
+
+# For a given reaction system, paraemter values, and initial conditions, find the polynomial that HC solves to find steady states.
+function steady_state_polynomial(rs::ReactionSystem, ps, u0)
+    ns = convert(NonlinearSystem, rs; remove_conserved = true)
+    pre_varmap = map(pair -> pair, [symmap_to_varmap(rs,u0); symmap_to_varmap(rs,ps)])
```
isaacsas, line 47: Do you have a test that this all works if the user passes a tuple for the parameter map using mixed types like integers and floats (even if the u0 map is a vector)?

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
isaacsas: Also, don't you want to enforce that all species have an initial value?

TorkelE:
- stability computation removed.
- Added Tuple with mixed int/float to tests.
- Now combines p/u0 using:
```julia
[symmap_to_varmap(rs,u0)..., symmap_to_varmap(rs,ps)...]
```
(I tried other stuff, but couldn't come up with something that worked well on this combination which didn't require more lines of code. Given what we are doing next, I presume that the performance damage will be relatively minimal, and I didn't want to reduce readability.
- Now uses this check:
```julia
any(s -> s in vars_with_vals, species(rs))
```
I think part of the problem with your suggestion was that you required initial conditions for *all* species, while I only want to require for some. This way when you have a system with many species and a single conserved quantity you don't have to provide tons of initial conditions. I have also pointed this out in docs properly.

PR #686, opened 2023-09-24
Bump ModelingToolkit

ErikQQY:
The recent 2.0 release of SciMLBase.jl caused many repo errors, let's see if this affect Catalyst.jl


isaacsas:
Looks like StochasticDiffEq needs an update first.

PR #695, opened 2023-09-27
Petab tutorial

TorkelE:
Creates a tutorial on how to use PEtab (https://github.com/sebapersson/PEtab.jl/) to fit the parameters of ODE CRNs to data (adapted from that documentation to provide a shorter summary to get someone started).

What is currently missing:
1) Plots to evaluate optimisation runs are not complete yet (https://github.com/sebapersson/PEtab.jl/pull/83). When that is done, I will complete that tutorial.
2) The `get_best_ps` function is not yet implemented in PEtab (but hopefully will be soon https://github.com/sebapersson/PEtab.jl/pull/85).
3) PEtab does not currently support events, so that section will be empty until it does.
4) It is possible to have different parameters for different experimental/simulation conditions. I started a tutorial for this, but haven't completed it yet (still figuring out the best way to do this).

PR #701, opened 2023-10-07
Inverse problem doc remake and new petab tutorial

TorkelE:
Supersedes:
https://github.com/SciML/Catalyst.jl/pull/674
https://github.com/SciML/Catalyst.jl/pull/695

Since the inverse problem doc remake PR had the same git mess as the petab tutorial one (with the latter probably inheriting from the former). I merged them into one, new one.


TorkelE:
Worth noting is this problem noted in the last PEtab tutorial PR:|

This tutorial is now in a good state where I am happy to receive feedback before merging.

A side note (and a problem) is that I am unable to update the Project.toml file in the docs environment to use the latest PEtab version (2+). I get this error:

(docs) pkg> update PEtab
    Updating registry at `~/.julia/registries/General.toml`
ERROR: Unsatisfiable requirements detected for package SuiteSparse_jll [bea87d4a]:
 SuiteSparse_jll [bea87d4a] log:
 ├─possible versions are: 7.2.0 or uninstalled
 ├─restricted to versions * by an explicit requirement, leaving only versions: 7.2.0
 └─found to have no compatible versions left with Sundials_jll [fb77eaff]
   └─Sundials_jll [fb77eaff] log:
     ├─possible versions are: 3.1.1-6.6.0 or uninstalled
     └─restricted to versions 5.2.1 by an explicit requirement, leaving only versions: 5.2.1

which does not directly seem to involve PEtab. Unsure how to interpret it, but I have had various weird package management errors relating to Sundials of late.

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+3. Be an unknown that we wish to fit to data.
+
+In our case, we wish to fit all three system parameters (*kB*, *kD*, and *kP*). For each, we create a single `PEtabParameter`, and then gather these into a single vector.
+```petab1
+par_kB = PEtabParameter(:kB)
+par_kD = PEtabParameter(:kD)
+par_kP = PEtabParameter(:kP)
+params = [par_kB, par_kD, par_kP]
+nothing # hide
+```
```
sebapersson, line 66: Think it should be nice to mention here that we have default bounds [lb, ub] = [1e-3, 1e]

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+
+### Creating a PEtabModel
+Finally, we combine all inputs in a single `PEtabModel`. To it, we also pass the initial conditions of our simulations (using the `state_map` argument). It is also possible to have [initial conditions with uncertainty](@ref petab_simulation_initial_conditions_uncertainty), [that vary between different simulations](@ref petab_simulation_conditions), or [that we attempt to fit to the data](@ref petab_simulation_initial_conditions_fitted).
+```petab1
+petab_model = PEtabModel(rn, observables, measurements, params; state_map=u0)
+nothing # hide
+```
+
+### Fitting parameters
+We are now able to fit our model to the data. First, we create a `PEtabODEProblem`. Here, we use `petab_model` as the only input, but it is also possible to set various [numeric solver and automatic differentiation option](@ref petab_simulation_conditions) (such as method or tolerance).
```
sebapersson, line 100: Mention briefly that if nothing is chosen PEtab.jl typically chooses good default options based on the problem

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+obs_X = PEtabObservable(X1 + X2, 0.5)
+```
+
+### [Advanced observables noise formulas](@id petab_observables_noise_formula)
+In our basic example we assumed that the normally distributed noise had a standard deviation of `0.5`. However, this value may be a parameter (or indeed any algebraic expression). E.g, we could set
+```petab1
+@parameters σ
+obs_P = PEtabObservable(P, σ)
+```
+and then let the parameter *σ* vary between different [simulation conditions](@ref petab_simulation_conditions). Alternatively we could let the noise scale linearly with the species' concentration:
```
sebapersson, line 143: would also add, or we could estimate *\sigma* (add it as a PEtabParameter)

Estimating noise parameters is quite popular

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+
+### Final notes
+PEtab.jl also supports [multistart optimisation](@ref petab_multistart_optimisation), [automatic equilibration before simulations](https://sebapersson.github.io/PEtab.jl/stable/Brannmark/), and [events](@ref petab_events). Various [plot recipes](@ref petab_plotting) exist for investigating the optimisation process. Please read the [PETab.jl documentation](https://sebapersson.github.io/PEtab.jl/stable/) for a more complete description of the package's features. Below follows additional details of various options and features (generally, PEtab is able to find good default values for most options not specified).
+
+## [Additional features: Observables](@id petab_observables)
+
+### [Defining non-trivial observables](@id petab_observables_observables)
```
sebapersson, line 126: I think it would also be nice to mention scale and offset parameter here? as they are quite common (as is done here https://sebapersson.github.io/PEtab.jl/stable/Julia_obs_noise/)

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+### [Parameter bounds](@id petab_parameters_bounds)
+By default, when fitted, potential parameter values are assumed to be in the interval *[1e-3, 1e3]*. When declaring a `PEtabParameter` it is possible to change these values through the `lb` and `ub` arguments. E.g. we could use
+```petab1
+par_kB = PEtabParameter(:kB; lb=1e-2, ub=1e-2)
+```
+to achieve the more conservative bound *[1e-2, 1e2]* for the parameter *kB*.
+
+### [Parameter scales](@id petab_parameters_scales)
+
+Internally, parameters that are fitted are converted to a logarithmic scale (generally, this is considered advantageous). To prevent this conversion, the `scale=:lin` argument can be used. Here we change the scale of *kB* to linear:
```
sebapersson, line 178: Do we want to add citation here, as it is proven typically better in this paper  https://academic.oup.com/bioinformatics/article/35/17/3073/5280731?login=true

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+
+### [Parameter priors](@id petab_parameters_priors)
+If we has prior knowledge about the distribution of a parameter, it is possible to incorporate this into the model. The prior can be any continuous, univariate, distribution from the [Distributions.jl package](https://github.com/JuliaStats/Distributions.jl). E.g we can use:
+
+```petab1
+using Distributions
+par_kB = PEtabParameter(:kB; prior=Normal(1.0,0.2))
+```
+to set a normally distributed prior (with mean `1.0` and standard deviation `0.2`) on the value of *kB*. By default, the prior is assumed to be on the linear scale of the parameter (before any potential log transform). To specify that the prior is on the logarithmic scale, the `prior_on_linear_scale=false` argument can be provided:
+```petab1
```
sebapersson, line 192: Maybe a bit to nitpicking, but for a prior on parmeter (log) scale Normal is not a good choice as it has negative support, would rather use LogNormal here

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+To record all the parameter sets evaluated (and their objective values) during the optimisation procedure, provide the `save_trace=true` argument to `calibrate_model` (or `calibrate_model_multistart`). This is required for the various [optimisation evaluation plots](@ref petab_plotting) provided by PEtab.jl. If desired, this information can be accessed in the calibration output's `.xtrace` and `.ftrace` fields.
+
+
+## Objective function extraction
+While PEtab.jl provides various tools for analysing the objective function generated by `PEtabODEProblem`, it is also possible to extract this function for customised analysis. Given a `PEtabODEProblem`
+```petab1
+petab_problem = PEtabODEProblem(petab_model)
+nothing # hide
+```
+We can find the:
```
sebapersson, line 374: I think it might be worth to mention here that lower and upper bounds can be accessed from the petab_problem, as well as the order of parameters via `petab_problem.θ_names`

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+1. The condition for triggering the event. This can either indicate a point in time, or a boolean condition.
+2. A rule for updating the event's target
+3. The event's target (either a species or parameter).
+
+Here we create an event which adds `0.5` units of `S` to the system at time `5.0`:
+```petab1
+@unpack S = rn
+event1 = PEtabEvent(5.0, S + 0.5, S)
+```
+Here, the first argument is evaluated to a scalar value, in which case it is interpreted as a time point at which the event happens. If we instead want the event to happen whenever the concentration of `S` falls below `0.5` we set the first argument to a boolean condition indicating this:
```
sebapersson, line 421: Maybe a bit advanced, but might be worth out to point out that for Boolean events like these they are only triggered when transition from `false` to `true`, so `S < 0.5` is only triggered as written when falling below 0.5 (from above), whether `S == 0.5` would be activated when coming from both direction

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+plot(res)
+```
+For a multi-start calibration run, the default output is instead a so-called waterfall plot:
+```petab1
+plot(res_ms)
+```
+![petab waterfall plot](../assets/petab_waterfall.svg)
+(for this, and the next plot, we use a multi-start optimisation result from a different model, which yields less trivial optimisation runs than our catalysis one)
+
+In the waterfall plot, each dot shows the final objective value for a single run in the multi-start process. The runs are ordered by their objective values, and colours designate runs in the same local minimum.
```
sebapersson, line 449: Worth mention that in order to really trust the optmisation results several runs (at least 5) should have converged to the best global minimum?

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+```petab1
+plot(res_ms; plot_type = :best_objective)
+```
+![petab best objective plot](../assets/petab_best_objective.svg)
+
+There exist several types of plots for both types of calibration results. More details of the types of available plots, and how to customise them, can be found [here](@ref https://sebapersson.github.io/PEtab.jl/stable/optimisation_output_plotting/).
+
+
+## [Citations](@id petab_citations)
+If you use this functionality in your research, [in addition to Catalyst](@ref catalyst_citation), please cite the following papers to support the authors of the PEtab.jl package (currently there is no article associated with this package) and the PEtab standard:
```
sebapersson, line 461: If the paper ever escapes the review land we can add it here :)

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+```
+
+We can now simulate our model for the fitted parameter set, and compare the result to the measurements and true solution.
+```petab1
+oprob = ODEProblem(rn, u0, (0.0, 10.0), get_ps(res, petab_problem))
+
+sol = solve(oprob, Tsit5(); saveat=0.1)
+plot!(sol; idxs=4, label="Fitted solution")
+```
+Here we use the `get_ps` function to retrieve a full parameter set using the optimal parameters (alternatively, the `ODEProblem` or fitted simulation can be retrieved directly using the `get_odeproblem` or `get_odesol` [functions](https://sebapersson.github.io/PEtab.jl/dev/API_choosen/#PEtab.get_odeproblem), respectively). The calibration result can also be found in `res.xmin`, however, note that PEtab automatically ([unless a linear scale is selected](petab_parameters_scales)) converts parameters to logarithmic scale, so typically `10 .^res.xmin` are the values of interest. If you investigate the result from this example you might note that the found parameter set (while it fits the data well) does not correspond to the true parameter set. This phenomenon is related to *identifiability*, and is very important for parameter fitting.
```
sebapersson, line 119: Here I also think you should mention the `get_u0` function, because if initial values are estimated then in order to get the correct simulations the correct initial values are also needed (this is automatically taken care of by `get_odeproblem` and `get_odesol`

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+
+### Final notes
+PEtab.jl also supports [multistart optimisation](@ref petab_multistart_optimisation), [automatic equilibration before simulations](https://sebapersson.github.io/PEtab.jl/stable/Brannmark/), and [events](@ref petab_events). Various [plot recipes](@ref petab_plotting) exist for investigating the optimisation process. Please read the [PETab.jl documentation](https://sebapersson.github.io/PEtab.jl/stable/) for a more complete description of the package's features. Below follows additional details of various options and features (generally, PEtab is able to find good default values for most options not specified).
```
sebapersson, line 122:
```suggestion
PEtab.jl also supports [multistart optimisation](@ref petab_multistart_optimisation), [automatic pre-equilibration before simulations](https://sebapersson.github.io/PEtab.jl/stable/Brannmark/), and [events](@ref petab_events). Various [plot recipes](@ref petab_plotting) exist for investigating the optimisation process. Please read the [PETab.jl documentation](https://sebapersson.github.io/PEtab.jl/stable/) for a more complete description of the package's features. Below follows additional details of various options and features (generally, PEtab is able to find good default values for most options not specified).
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,487 @@
...
+m2 = DataFrame(simulation_id="c2", obs_id="obs_P", time=t2, measurement=d2)
+measurements = vcat(m1,m2)
+
+petab_model = PEtabModel(rn, simulation_conditions, observables, measurements, params; state_map=u0)
+nothing # hide
+```
+Note that the `u0` we pass into `PEtabModel` through the `state_map` argument no longer contains the value of *S* (as it is provided by the conditions). 
+
+### [Varying parameters between different simulation conditions](@id petab_simulation_conditions)
+Sometimes, the parameters that are used vary between the different conditions. Consider our catalysis example, if we had performed the experiment twice, using two different enzymes with different catalytic properties, this could have generated such conditions. The two enzymes could e.g. yield different rates (*kP₁* and *kP₂*) for the `SE --> P + E` reaction, but otherwise be identical. Here, the parameters *kP₁* and *kP₂* are unique to their respective conditions. PEtab.jl provides support for cases such as this, and [its documentation](https://sebapersson.github.io/PEtab.jl/dev/Beer/) provided instructions of how to handle them.
```
sebapersson, line 262:
```suggestion
Sometimes, the parameters that are used vary between the different conditions. Consider our catalysis example, if we had performed the experiment twice, using two different enzymes with different catalytic properties, this could have generated such conditions. The two enzymes could e.g. yield different rates (*kP₁* and *kP₂*) for the `SE --> P + E` reaction, but otherwise be identical. Here, the parameters *kP₁* and *kP₂* are unique to their respective conditions. PEtab.jl provides support for cases such as this, and [its documentation](https://sebapersson.github.io/PEtab.jl/stable/Julia_condition_specific/) provided instructions of how to handle them.
```

sebapersson (changes requested):
Looks good! Just some minor comments, primarily on adding small things to the documentation

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: updated

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: added this

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: This part was really only meant if the user wanted to solely use PEtab to generate cost functions (and then optimize these using SciML/Optimization) or use for something else, rather than listing the fields of the problem. Although this should probably be somewhere (but I might leave that out from this tutorial).

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: Definitely :)

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: added

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: might as well, will add.

TorkelE: Added. Generally, I like the idea of having 1-4 useful references on the longer do pages. Not more to make it like a research article, but more to give some useful further readings to the user.

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: Would setting a Normal prior, and the `prior_on_linear_scale=false` be equivalent to a LogNormal prior?

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: added

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: added

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: Just checking, bu this means that if the event continues being true (e.g. when S <0.5 we do something entirely different) the event does not trigger. It only triggers when the transition occur?

docs/src/inverse_problems/petab_ode_param_fitting.md
sebapersson: Yes, but then I agree it should be left at it is

docs/src/inverse_problems/petab_ode_param_fitting.md
sebapersson: Yes, as we have if log(c) ~ N(µ, s) then c ~ LN(µ, s). 

So setting `prior=Normal(3.0, 3.0), prior_on_linear_scale=true` is equivalent to setting `prior=LogNormal(3.0, 3.0), prior_on_linear_scale=false`

docs/src/inverse_problems/petab_ode_param_fitting.md
sebapersson: Exactly, it only triggers when we make transition from `false` to `true` (so when the statement already is true nothing happens trigger wise)

TorkelE: added additional explanation.

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: added extra note of this

TorkelE:
@isaacsas Did you have a look at whether we could merge this one? It passes everything, so should be good to go. There is a couple of stuff which depends on the new doc structure, so it would save me (and Sophie) a decent amount of git effort to have this one merged asap.
(I also indicated at ICSB that this tutorial would be up very soon, so really would like it to be available asap)

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+```
+In this example, setting `prior_on_linear_scale=false` make sense as a (linear) normal distribution is non-zero for negative values (an alternative is to use a log-normal distribution, e.g. `prior=LogNormal(3.0, 3.0)`).
+
+## [Simulation conditions](@id petab_simulation_conditions)
+Sometimes, we have data from several different experimental conditions. Here, when a potential parameter set is evaluated during the fitting process, each experimental condition corresponds to one simulation condition (which produces one simulation). To account for this, PEtab permits the user to define several different simulation conditions, with each condition being defined by specific values for some initial conditions and/or parameters. 
+
+If, for our previous catalysis example, we had measured the system for two different initial values of *S* (*S(0)=1.0* and *S(0)=0.5*), these would correspond to two different simulation conditions. For each condition we define a `Dict` mapping the species to their initial condition (here, *S* is the only species in each `Dict`):
+```@example petab1
+c1 = Dict(:S => 1.0)
+c2 = Dict(:S => 0.5)
```
isaacsas, line 207: Use symbolics?

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+nothing # hide
+```
+Similarly as for observables, we then gather the conditions in another `Dict`, giving each an id tag:
+```@example petab1
+simulation_conditions = Dict("c1" => c1, "c2" => c2)
+nothing # hide
+```
+Again (like for observables), each measurement in the measurements `DataFrame` needs to be associated with a simulation condition id tag (describing which condition that measurements were taken from). Parameters, just like initial conditions, may vary between different conditions. If an initial condition (or parameter) occurs in one condition, it must occur in all of them.
```
isaacsas, line 215:
```suggestion
Again (like for observables), each measurement in the measurements `DataFrame` needs to be associated with a simulation condition id tag (describing which condition those measurements were taken from). Parameters, just like initial conditions, may vary between different conditions. If an initial condition (or parameter) occurs in one condition, it must occur in all of them.
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+
+Here follows a complete version of our basic example, but with measurements both for *S(0)=1.0* and *S(0)=0.5*.
```
isaacsas, line 217:
```suggestion
Here follows a complete version of our basic example, but with measurements both for $S(0)=1$ and $S(0)  = \tfrac{1}{2}$.
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+    ((0.8 .+ 0.4*rand(10)) .* sol[:P][10:10:end], sol.t[10:10:end])
+end
+
+@unpack P = rn
+obs_P = PEtabObservable(P, 0.5)
+observables = Dict("obs_P" => obs_P)
+
+par_kB = PEtabParameter(:kB)
+par_kD = PEtabParameter(:kD)
+par_kP = PEtabParameter(:kP)
```
isaacsas, line 249: symbolics

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+params = [par_kB, par_kD, par_kP]
+
+c1 = Dict(:kB => 1.0)
+c2 = Dict(:kB => 0.5)
```
isaacsas, line 253: symbolics

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+
+using DataFrames
+m1 = DataFrame(simulation_id="c1", obs_id="obs_P", time=t1, measurement=d1)
+m2 = DataFrame(simulation_id="c2", obs_id="obs_P", time=t2, measurement=d2)
+measurements = vcat(m1,m2)
+
+petab_model = PEtabModel(rn, simulation_conditions, observables, measurements, params; state_map=u0)
+nothing # hide
+```
+Note that the `u0` we pass into `PEtabModel` through the `state_map` argument no longer contains the value of *S* (as it is provided by the conditions).
```
isaacsas, line 264:
```suggestion
Note that the `u0` we pass into `PEtabModel` through the `state_map` argument no longer contains the value of $S$ (as it is provided by the conditions). 
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+m2 = DataFrame(simulation_id="c2", obs_id="obs_P", time=t2, measurement=d2)
+measurements = vcat(m1,m2)
+
+petab_model = PEtabModel(rn, simulation_conditions, observables, measurements, params; state_map=u0)
+nothing # hide
+```
+Note that the `u0` we pass into `PEtabModel` through the `state_map` argument no longer contains the value of *S* (as it is provided by the conditions). 
+
+### Varying parameters between different simulation conditions
+Sometimes, the parameters that are used vary between the different conditions. Consider our catalysis example, if we had performed the experiment twice, using two different enzymes with different catalytic properties, this could have generated such conditions. The two enzymes could e.g. yield different rates (*kP₁* and *kP₂*) for the `SE --> P + E` reaction, but otherwise be identical. Here, the parameters *kP₁* and *kP₂* are unique to their respective conditions. PEtab.jl provides support for cases such as this, and [its documentation](https://sebapersson.github.io/PEtab.jl/stable/Julia_condition_specific/) provided instructions of how to handle them.
```
isaacsas, line 267:
```suggestion
Sometimes, the parameters that are used vary between the different conditions. Consider our catalysis example, if we had performed the experiment twice, using two different enzymes with different catalytic properties, this could have generated such conditions. The two enzymes could e.g. yield different rates (`kP₁` and `kP₂`) for the `SE --> P + E` reaction, but otherwise be identical. Here, the parameters `kP₁` and `kP₂` are unique to their respective conditions. PEtab.jl provides support for cases such as this, and [its documentation](https://sebapersson.github.io/PEtab.jl/stable/Julia_condition_specific/) provided instructions of how to handle them.
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+```@example petab4
+u0 = [:S => 1.0, :SE => 0.0, :P => 0.0]
+nothing # hide
+```
+Next, we add `E0` to the parameters we wish to fit:
+```@example petab4
+par_kB = PEtabParameter(:kB)
+par_kD = PEtabParameter(:kD)
+par_kP = PEtabParameter(:kP)
+par_E0 = PEtabParameter(:E0)
```
isaacsas, line 294: symbolics

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+params = [par_kB, par_kD, par_kP, par_E0]
+nothing # hide
+```
+and we can use our updated `rn`, `u0`, and `params` as input to our `PEtabModel`.
+
+### [Uncertain initial conditions](@id petab_simulation_initial_conditions_uncertainty)
+Often, while an initial condition has been reported for an experiment, its exact value is uncertain. This can be modelled by making the initial condition a [parameter that is fitted to the data](@ref petab_simulation_initial_conditions_fitted) and attach a prior to it corresponding to our certainty about its value.
```
isaacsas, line 301:
```suggestion
Often, while an initial condition has been reported for an experiment, its exact value is uncertain. This can be modelled by making the initial condition a [parameter that is fitted to the data](@ref petab_simulation_initial_conditions_fitted) and attaching a prior to it corresponding to our certainty about its value. 
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+par_E0 = PEtabParameter(:E0)
+params = [par_kB, par_kD, par_kP, par_E0]
+nothing # hide
+```
+and we can use our updated `rn`, `u0`, and `params` as input to our `PEtabModel`.
+
+### [Uncertain initial conditions](@id petab_simulation_initial_conditions_uncertainty)
+Often, while an initial condition has been reported for an experiment, its exact value is uncertain. This can be modelled by making the initial condition a [parameter that is fitted to the data](@ref petab_simulation_initial_conditions_fitted) and attach a prior to it corresponding to our certainty about its value. 
+
+Let us consider our initial example, but where we want to add uncertainty to the initial conditions of `S` and `E`. We will add priors on these, assuming normal distributions with mean `1.0` and standard deviation `0.1`. For the synthetic measured data we will use the true values *S(0)=E(0)=1.0*.
```
isaacsas, line 303:
```suggestion
Let us consider our initial example, but where we want to add uncertainty to the initial conditions of `S` and `E`. We will add priors on these, assuming normal distributions with mean `1.0` and standard deviation `0.1`. For the synthetic measured data we will use the true values $S(0)=E(0)=1$.
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+
+@unpack P = rn
+obs_P = PEtabObservable(P, 0.5)
+observables = Dict("obs_P" => obs_P)
+
+par_kB = PEtabParameter(:kB)
+par_kD = PEtabParameter(:kD)
+par_kP = PEtabParameter(:kP)
+par_S0 = PEtabParameter(:S0)
+par_E0 = PEtabParameter(:E0)
```
isaacsas, line 332: symbolics

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+    kP, SE --> P + E
+end
+
+u0 = [:SE => 0.0, :P => 0.0]
+p_true = [:kB => 1.0, :kD => 0.1, :kP => 0.5, :S0=>1.0, :E0 => 1.0]
+
+using OrdinaryDiffEq, Random
+oprob = ODEProblem(rn, [:S=>1.0; :E => 1.0; u0], (0.0, 10.0), p_true)
+sol = solve(oprob, Tsit5(); saveat=0.1)
+data = (0.8 .+ 0.4*rand(10)) .* sol[:P][10:10:end]
```
isaacsas, line 322:
```suggestion
data = (0.8 .+ 0.4*rand(10)) .* sol[rn.P][10:10:end]
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+
+using Catalyst, PEtab
+
+rn = @reaction_network begin
+    @parameters S0 E0
+    @species S(t)=S0 E(t)=E0
+    kB, S + E --> SE
+    kD, SE --> S + E
+    kP, SE --> P + E
+end
```
isaacsas, line 314:
```suggestion
rn = @reaction_network begin
    @parameters S0 E0
    @species S(t)=S0 E(t)=E0
    kB, S + E --> SE
    kD, SE --> S + E
    kP, SE --> P + E
end
rn = complete(rn)
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+
+
+## Objective function extraction
+While PEtab.jl provides various tools for analysing the objective function generated by `PEtabODEProblem`, it is also possible to extract this function for customised analysis. Given a `PEtabODEProblem`
+```@example petab1
+petab_problem = PEtabODEProblem(petab_model)
+nothing # hide
+```
+We can find the:
+1. Objective function in the `petab_problem.compute_cost`. It takes a single argument (`p`) and returns the objective value.
```
isaacsas, line 380:
```suggestion
1. Objective function as `petab_problem.compute_cost`. It takes a single argument (`p`) and returns the objective value.
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+2. Gradient in the `petab_problem.compute_gradient!` field. It takes two arguments (`g` and `p`) with the updated gradient values being written to `g`.
```
isaacsas, line 381:
```suggestion
2. Gradient as `petab_problem.compute_gradient!` field. It takes two arguments (`g` and `p`) with the updated gradient values being written to `g`.
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+3. Hessian in the `petab_problem.compute_hessian` field. It takes two arguments (`H` and `p`) with the updated hessian values being written to `H`.
```
isaacsas, line 382:
```suggestion
3. Hessian as `petab_problem.compute_hessian` field. It takes two arguments (`H` and `p`) with the updated hessian values being written to `H`.
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+```@example petab1
+using Optim
+using QuasiMonteCarlo
+res_ms = calibrate_model_multistart(petab_problem, IPNewton(), 10, "OptimizationRuns"; sampling_method=QuasiMonteCarlo.SobolSample())
+```
+The best result across all runs can still be retrieved using `get_ps(res_ms, petab_problem)`, with the results of the individual runs being stored in the `res_ms.runs` field. 
+
+To load the result in a later session, we can call:
+```@example petab1
+res_ms = PEtabMultistartOptimisationResult("OptimizationRuns")
```
isaacsas, line 407: Does this require saving the data? If so, you should probably show how to do that too. Otherwise how does it find the results on disk?

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+```
+If the OptimizationRuns folder contains the output from several runs, we can designate which to load using the `which_run` argument. Here we loads the second run to be saved in that folder:
```
isaacsas, line 409:
```suggestion
If the OptimizationRuns folder contains the output from several runs, we can designate which to load using the `which_run` argument. Here we load the second run to be saved in that folder:
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+So far, we have assumed that all experiments, after initiation, run without interference. Experiments where conditions change, or where species are added/removed during the time course, can be represented through events (related to [callbacks](@ref advanced_simulations_callbacks)). In PEtab, an event is represented through the `PEtabEvent` structure. It takes three arguments:
+1. The condition for triggering the event. This can either indicate a point in time, or a boolean condition.
+2. A rule for updating the event's target
+3. The event's target (either a species or parameter).
+
+Here we create an event which adds `0.5` units of `S` to the system at time `5.0`:
+```@example petab1
+@unpack S = rn
+event1 = PEtabEvent(5.0, S + 0.5, S)
+```
```
isaacsas, line 425: This should instead be put into the Catalyst model and forwarded appropriately. Does that work?

isaacsas: I suggest putting the events into the Catalyst model and then just showing it forwards ok in this section. You can then refer to the PEtab docs for adding events in their interface.

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+
+```@example petab1
+plot(res)
+```
+For a multi-start calibration run, the default output is instead a so-called waterfall plot:
+```@example petab1
+plot(res_ms)
+```
+![petab waterfall plot](../assets/petab_waterfall.svg)
+(for this, and the next plot, we use a multi-start optimisation result from a different model, which yields less trivial optimisation runs than our catalysis one)
```
isaacsas, line 453:
```suggestion
![petab waterfall plot](../assets/petab_waterfall.svg)

(for this, and the next plot, we use a multi-start optimisation result from a different model, which yields less trivial optimisation runs than our catalysis one)
```

isaacsas (changes requested):

docs/src/catalyst_functionality/example_networks/basic_CRN_examples.md
```diff
@@ -63,7 +63,7 @@ jprob = JumpProblem(rs, dprob, Direct())
 jsol = solve(jprob, SSAStepper())
 
 plot(plot(osol; title = "Reaction Rate Equation ODEs"),
-     plot(jsol; title = "Stochastic Chemical Kinetics Jump Processes");
+     plot(jsol; title = "Gillespie Jump Simulation");
```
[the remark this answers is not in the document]
TorkelE: I thought it was a clear error. In the same doc we use "Gillespie Jump Simulation"
image
and it seemed weird to use different names for the same thing in the same doc...

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+Let us consider a simple catalysis network, where an enzyme (*E*) turns a substrate (*S*) into a product (*P*):
+```petab1
+using Catalyst, PEtab
+
+rn = @reaction_network begin
+    kB, S + E --> SE
+    kD, SE --> S + E
+    kP, SE --> P + E
+end
+```
```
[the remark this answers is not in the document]
TorkelE: I'd rather hold off on this for now, until I have figured out what the scope of these changes actually are.

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: Currently it the optimal way is to use PEtab's interface. I have forgotten if it had to do with the performance or limitations of the SciML events (@sebapersson can probably expand). If I remember right, the idea is to, eventually, add a routine that detects events in Catalyst models and translates them into the PEtab format.
(essentially giving this effect)

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+```@example petab1
+using DataFrames
+measurements = DataFrame(obs_id="obs_P", time=sol.t[10:10:end], measurement=data)
+```
+Since, in our example, all measurements are of the same observable, we can set `obs_id="obs_P"`. If several different observables were measured, `obs_id` would be a vector of the same length as `time` and `measurement`.
+
+### Creating a PEtabModel
+Finally, we combine all inputs in a single `PEtabModel`. To it, we also pass the initial conditions of our simulations (using the `state_map` argument). It is also possible to have [initial conditions with uncertainty](@ref petab_simulation_initial_conditions_uncertainty), [that vary between different simulations](@ref petab_simulation_conditions), or [that we attempt to fit to the data](@ref petab_simulation_initial_conditions_fitted).
+```@example petab1
+petab_model = PEtabModel(rn, observables, measurements, params; state_map=u0)
```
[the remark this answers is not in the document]
TorkelE: This is supported, yes.

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+We are now able to fit our model to the data. First, we create a `PEtabODEProblem`. Here, we use `petab_model` as the only input, but it is also possible to set various [numeric solver and automatic differentiation options](@ref petab_simulation_options) (such as method or tolerance).
+```@example petab1
+petab_problem = PEtabODEProblem(petab_model)
+nothing # hide
+```
+Since no additional input is given, default optiosn are selected by PEtab.jl (and generally, its choices are good).
+
+To fit a parameter set we use the `calibrate_model` function. In addition to our `PEtabODEProblem`, we must also provide an initial guess (which can be generated with the `generate_startguesses` function) and an optimisation algorithm (which needs to be imported specifically). PEtab.jl supports various [optimisation methods and options](@ref petab_optimisation_optimisers). 
+```@example petab1
+using Optim
```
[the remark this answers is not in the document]
TorkelE: It is not, but it is planned for some future update.

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+```@example petab1
+oprob = ODEProblem(rn, u0, (0.0, 10.0), get_ps(res, petab_problem))
+sol = solve(oprob, Tsit5(); saveat=0.1)
+plot!(sol; idxs=4, label="Fitted solution")
+```
+Here we use the `get_ps` function to retrieve a full parameter set using the optimal parameters. Alternatively, the `ODEProblem` or fitted simulation can be retrieved directly using the `get_odeproblem` or `get_odesol` [functions](https://sebapersson.github.io/PEtab.jl/dev/API_choosen/#PEtab.get_odeproblem), respectively (and the initial condition using the `get_u0` function). The calibration result can also be found in `res.xmin`, however, note that PEtab automatically ([unless a linear scale is selected](@ref petab_parameters_scales)) converts parameters to logarithmic scale, so typically `10 .^res.xmin` are the values of interest. If you investigate the result from this example you might note that the found parameter set (while it fits the data well) does not correspond to the true parameter set. This phenomenon is related to *identifiability*, and is very important for parameter fitting.
```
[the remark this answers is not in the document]
TorkelE: There is an upcoming documentation on identifiability. The idea is to (when it is done, which should be very soon) have this link that.

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+We then provide `parameter_map=[:kB => 1.0]` when we assembly our model:
+```@example petab1
+petab_model = PEtabModel(rn, observables, measurements, params; state_map=u0, parameter_map=[:kB => 1.0])
+nothing # hide
+```
+
+### [Parameter bounds](@id petab_parameters_bounds)
+By default, when fitted, potential parameter values are assumed to be in the interval *[1e-3, 1e3]*. When declaring a `PEtabParameter` it is possible to change these values through the `lb` and `ub` arguments. E.g. we could use
+```@example petab1
+par_kB = PEtabParameter(:kB; lb=1e-2, ub=1e-2)
```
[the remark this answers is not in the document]
TorkelE: These are identical?

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: Why use "as" for the Hessian, but not for the Gradient and the objective function?

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+using Optim
+p0 = generate_startguesses(petab_problem, 1)
+res = calibrate_model(petab_problem, p0, IPNewton())
+```
+
+We can now simulate our model for the fitted parameter set, and compare the result to the measurements and true solution.
+```@example petab1
+oprob = ODEProblem(rn, u0, (0.0, 10.0), get_ps(res, petab_problem))
+sol = solve(oprob, Tsit5(); saveat=0.1)
+plot!(sol; idxs=4, label="Fitted solution")
```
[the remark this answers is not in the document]
TorkelE: No, that should be `:P`, silly error. Fixed.

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+4. The simulation condition which generates the measurement ([here](@ref petab_simulation_conditions) is an example where this is used).
+
+([When [pre-equilibration](https://sebapersson.github.io/PEtab.jl/stable/Brannmark/) is used to initiate the system in a steady state, a fifth field is also required])
+
+Measurements are provided in a `DataFrame` where each row corresponds to a measurement. The values are provided in the `obs_id`, `time`, `measurement`, and `simulation_id` columns. In our case we only need to fill in the first three:
+```@example petab1
+using DataFrames
+measurements = DataFrame(obs_id="obs_P", time=sol.t[10:10:end], measurement=data)
+```
+Since, in our example, all measurements are of the same observable, we can set `obs_id="obs_P"`. If several different observables were measured, `obs_id` would be a vector of the same length as `time` and `measurement`.
```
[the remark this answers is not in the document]
TorkelE: No, if we had 10 different measurements of 10 different observables, we would have `obs_id` be a vector of length 10 (to go with the `time` and `measurement` vectors, also both of length 10.

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,494 @@
...
+# Generate synthetic data.
+using OrdinaryDiffEq, Random
+oprob = ODEProblem(rn, u0, (0.0, 10.0), p_true)
+sol = solve(oprob, Tsit5(); saveat=0.1)
+data = (0.8 .+ 0.4*rand(10)) .* sol[:P][10:10:end]
+
+# Plots the true solutions and the (synthetic data) measurements.
+using Plots
+plot(sol; idxs=:P, label="True solution")
+plot!(sol.t[10:10:end], data; label="Measurements", seriestype=:scatter)
```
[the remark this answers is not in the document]
TorkelE: Agreed. Fixed.

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: This was briefly described in the previous section, however, I will add a link to that.

TorkelE: They are automatically saved using the `dir_save` option. It is described in the section, but I can elaborate a bit here.

docs/src/catalyst_functionality/example_networks/basic_CRN_examples.md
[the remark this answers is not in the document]
isaacsas: We should change that figure title then.

docs/src/inverse_problems/petab_ode_param_fitting.md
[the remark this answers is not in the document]
isaacsas: I think some clarification of how multiple measurements are stored in time/measurement here would be helpful.

docs/src/inverse_problems/petab_ode_param_fitting.md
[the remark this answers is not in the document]
isaacsas: OK, it still seems like a poor advertisement for PETab to pick an example it can't actually recover the initial parameters on (even if it is because the example is not identifiable). 

I'd still suggest dropping the non-identifiable example as the opening example where users are getting the flavor for using PETab. It will be more convincing that it is useful if people see it can do a decent job recovering the parameters (and I would show the fit vs. exact values to make this clear).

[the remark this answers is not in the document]
isaacsas: You can then have a non-identifiable example as a second example. (Though I would instead suggest that the opening inverse problem tutorial should instead cover such things since it is more of a generic fitting problem.)

docs/src/inverse_problems/petab_ode_param_fitting.md
isaacsas: Maybe just a reminder since I missed that.

docs/src/inverse_problems/petab_ode_param_fitting.md
isaacsas: What are the limitations / performance issues? 

As I said on Slack, I think there needs to be an explicit bolded/italicized statement then that PETab will ignore events in Catalyst models, so users must then reenter them manually in this way. Otherwise this potentially creates confusion that events in models are being silently dropped.

docs/src/inverse_problems/petab_ode_param_fitting.md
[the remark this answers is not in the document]
isaacsas: What do you mean by the scope?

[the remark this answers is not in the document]
TorkelE: As in the scope of the changes to MTK and what they would mean for us.

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: I added extra stuff to clarify this.

docs/src/inverse_problems/petab_ode_param_fitting.md
TorkelE: I have added a specific note box on this.

docs/src/inverse_problems/petab_ode_param_fitting.md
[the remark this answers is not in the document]
TorkelE: I have created a separate section for this, which I now linked instead of briefly mentioning the concept.

TorkelE:
I have not cut the identifiability thing (which I still prefer this way), and the symbolic usage (the same) I have made changes for everything. Still would need to go through to check that it builds properly and that I didn't miss any of your comments, but otherwise this should be good to go.

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -0,0 +1,539 @@
...
+Whenever we have several events or not, we bundle them together in a single vector which is later passed to the `PEtabODEProblem` using the `events` argument:
+```@example petab1
+events = [event, event2]
+petab_model = PEtabModel(rn, observables, measurements, params; state_map=u0, events=events)
+```
+
+More details on how to use events, including how to create events with multiple targets, can be found in [PEtab.jl's documentation](https://sebapersson.github.io/PEtab.jl/stable/Julia_event/).
+
+!!! note
+    PEtab currently does not support events [created as a part of Catalyst's `ReactionStructure`s](@ref constraint_equations_events) and [implemented through `callbacks`](@ref advanced_simulations_callbacks). Instead, events have to use this interface.
```
isaacsas, line 478:
```suggestion
    PEtab currently ignores events [created as a part of a Catalyst `ReactionSystem` model](@ref constraint_equations_events), and does not support SciML-style events [implemented through `callbacks` to `solve`](@ref advanced_simulations_callbacks). Instead, events have to use this interface.
```

docs/src/inverse_problems/petab_ode_param_fitting.md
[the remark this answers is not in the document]
TorkelE: I have reworded it to say
"If you investigate the result from this example you might note, that even if PEtab.jl has found the global optimum (which fits the data well), this does not actually correspond to the true parameter set. This phenomenon is related to the concept of *identifiability*, which is very important for parameter fitting."
If people are happy with this we will go ahead and merge this.

PR #704, opened 2023-10-25
Create SPR_fitting.md

tstrey:


docs/src/example_networks/SPR_fitting.md
```diff
@@ -0,0 +1,105 @@
...
+    sample_val .*= (1 .+ .1 * rand(Float64, size(sample_val)) .- .01)
+    push!(sample_vals, sample_val)
+end
+
+for val in sample_vals
+    scatter!(p, sample_times, val[2,:]; color = [:blue :red], legend = nothing)
+end
+plot(p)
+```
+Next, we create a function to fit the parameters. Here we are using `NelderMead()`. In order to achieve a better fit, we are incorporating all of our solutions into the loss function. We will fit seperately the association and dissociation signals so for the first estimate, `tend < switch_time`.
```
ChrisRackauckas, line 64: Should probably use a derivative-based method. NelderMead has known convergence issues.

ChrisRackauckas:
This needs to be added to the make.jl of the docs in order to show up.

docs/src/example_networks/SPR_fitting.md
```diff
@@ -0,0 +1,105 @@
+# [Surface Plasmon Resonance Model Example](@id surface_plasmon_resonance_model_Example)
+This tutorial shows how to programmatically construct a ['ReactionSystem'](@ref) corresponding to the chemistry underlying the [Surface Plasmon Resonance Model](https://en.wikipedia.org/wiki/Surface_plasmon_resonance) using [ModelingToolkit](http://docs.sciml.ai/ModelingToolkit/stable/)/[Catalyst](http://docs.sciml.ai/Catalyst/stable/). You can find a simpler constructruction of this model [here](https://docs.sciml.ai/Catalyst/stable/catalyst_applications/parameter_estimation/) in the example of parameter estimation. 
+
+Using the symbolic interface, we will create our states/species, define our reactions, and add a discrete event. This model is simulated and used to generate sample data points with added noise and then used to fit parameters in iterations. This event will correspond to the time at which the antibody stop binding to antigen and switch to dissociating. 
+
+We begin by importing some necessary packages. 
+```julia
+using ModelingToolkit, Catalyst, DifferentialEquations
+using Plots, Optimization
+using OptimizatizationOptimJL, Optim
```
TorkelE, line 10: Is there a reason we have to use both Optim and Optimization+OptimizatizationOptimJL. I though the idea of Optimization was not to need to call these packages? Not my expertise, so I might be wrong (@isaacsas ).

docs/src/example_networks/SPR_fitting.md
```diff
@@ -0,0 +1,105 @@
...
+```
+In this example, the concentration of antigen,`\beta` is varied to determine the constant of proportionality, `\alpha`, `k_{on}`(association rate constant), and `k_{off}` (dissociation rate constant) which characterized the binding interaction between the antigen and the immobilized receptor molecules on the sensor slide. We start by defining our reaction equations, parameters, variables, event, and states. 
+```julia
+@variables t 
+@parameters k_on k_off α
+@species A(t) B(t)
+
+rxs = [(@reaction α*k_on, A --> B), (@reaction k_off, B --> A)]
```
TorkelE, line 18: Is there a reason that we don't use the DSL here? Simply doing
```julia
rs = @reaction_network begin
    (α*k_on, k_off), A <--> B
end
```
seems a lot more intuitive? The main difference is that we cannot add events in the DSL, but we could just add this as a PresettimeCallback: https://docs.sciml.ai/Catalyst/stable/catalyst_applications/advanced_simulations/#advanced_simulations_callbacks

TorkelE:
Will continue going through it tomorrow, but so far it looks good :)

TorkelE:
I have gone through the first half. Unfortunately I get an error after that due to not being sure which combination of optimization packages to use. Once that is figured out, I will go through the last bit :)

docs/src/example_networks/SPR_fitting.md
isaacsas: What do you recommend here? BFGS?

docs/src/example_networks/SPR_fitting.md
isaacsas: The idea is to use a symbolic callback so the callback is encoded in the Catalyst model. Otherwise if one uses the DSL one must use `extend` to merge a separate system with the event in to the Catalyst reaction model.

Once we get discrete event support in the DSL we can rewrite this to fully use the DSL.

docs/src/example_networks/SPR_fitting.md
isaacsas: OptimizationOptimJL is the extension to make Optim optimizers available in Optimization.

tstrey: @TorkelE--yes, Optim is redundant. I'll change that.

isaacsas:
@tstrey this needs to have master merged into it as it is out of date. (The pages.jl needs updating.)

isaacsas:
@tstrey this should probably be moved to be an inverse problem example now instead of a model creation example given the new doc structure. Can you update that?

docs/src/inverse_problems/examples/SPR_fitting.md
```diff
@@ -0,0 +1,110 @@
+# [Surface Plasmon Resonance Model Example](@id surface_plasmon_resonance_model_Example)
+This tutorial shows how to programmatically construct a ['ReactionSystem'](@ref) corresponding to the chemistry underlying the [Surface Plasmon Resonance Model](https://en.wikipedia.org/wiki/Surface_plasmon_resonance) using [ModelingToolkit](http://docs.sciml.ai/ModelingToolkit/stable/)/[Catalyst](http://docs.sciml.ai/Catalyst/stable/). You can find a simpler constructruction of this model [here](https://docs.sciml.ai/Catalyst/stable/catalyst_applications/parameter_estimation/) in the example of parameter estimation.
```
isaacsas, line 2: > https://docs.sciml.ai/Catalyst/stable/catalyst_applications/parameter_estimation/

This seems to be a link from an older version of the docs that no longer works / exists.

docs/src/inverse_problems/examples/SPR_fitting.md
```diff
@@ -0,0 +1,110 @@
+# [Surface Plasmon Resonance Model Example](@id surface_plasmon_resonance_model_Example)
+This tutorial shows how to programmatically construct a ['ReactionSystem'](@ref) corresponding to the chemistry underlying the [Surface Plasmon Resonance Model](https://en.wikipedia.org/wiki/Surface_plasmon_resonance) using [ModelingToolkit](http://docs.sciml.ai/ModelingToolkit/stable/)/[Catalyst](http://docs.sciml.ai/Catalyst/stable/). You can find a simpler constructruction of this model [here](https://docs.sciml.ai/Catalyst/stable/catalyst_applications/parameter_estimation/) in the example of parameter estimation. 
+
+Using the symbolic interface, we will create our states/species, define our reactions, and add a discrete event. This model is simulated and used to generate sample data points with added noise and then used to fit parameters in iterations. This event will correspond to the time at which the antibody stop binding to antigen and switch to dissociating. 
+
+We begin by importing some necessary packages. 
+```julia
```
isaacsas, line 7:
```suggestion
```@example spr_ex
```
All of these should be `@example` blocks so that they run dynamically.

isaacsas: Check out the Documenter.jl documentation for information on how to use such blocks, or look at the other Catalyst tutorials.

docs/src/inverse_problems/examples/SPR_fitting.md
```diff
@@ -0,0 +1,110 @@
...
+using ModelingToolkit, Catalyst
+using Optimization, OptimizatizationOptimJL
+using Plots
+```
+In this example, the concentration of antigen,`\beta` is varied to determine the constant of proportionality, `\alpha`, `k_{on}`(association rate constant), and `k_{off}` (dissociation rate constant) which characterized the binding interaction between the antigen and the immobilized receptor molecules on the sensor slide. We start by defining our reaction equations, parameters, variables, event, and states.
```
isaacsas, line 12: Note that single quotes are for code rendering not math. They won't render \beta as $\beta$. If you build the documentation locally on your computer and check it looks ok you can see issues such as this.

docs/src/inverse_problems/examples/SPR_fitting.md
```diff
@@ -0,0 +1,110 @@
+# [Surface Plasmon Resonance Model Example](@id surface_plasmon_resonance_model_Example)
+This tutorial shows how to programmatically construct a ['ReactionSystem'](@ref) corresponding to the chemistry underlying the [Surface Plasmon Resonance Model](https://en.wikipedia.org/wiki/Surface_plasmon_resonance) using [ModelingToolkit](http://docs.sciml.ai/ModelingToolkit/stable/)/[Catalyst](http://docs.sciml.ai/Catalyst/stable/). You can find a simpler constructruction of this model [here](https://docs.sciml.ai/Catalyst/stable/catalyst_applications/parameter_estimation/) in the example of parameter estimation. 
+
+Using the symbolic interface, we will create our states/species, define our reactions, and add a discrete event. This model is simulated and used to generate sample data points with added noise and then used to fit parameters in iterations. This event will correspond to the time at which the antibody stop binding to antigen and switch to dissociating.
```
isaacsas, line 4:
```suggestion
Using the DSL interface, we will create our states/species, define our reactions, and add a discrete event. This model is simulated and used to generate sample data points with added noise and then used to fit parameters in iterations. This event will correspond to the time at which the antibody stop binding to antigen and switch to dissociating. 
```

docs/src/inverse_problems/examples/SPR_fitting.md
```diff
@@ -0,0 +1,110 @@
...
+    push!(results_list, sol_real(sample_times))
+end
+```
+```@example ceq3
+default(; lw = 3, framestyle = :box, size = (800, 400))
+p = plot()
+plot(p, sample_times, results_list[1][2,:])
+plot!(p, sample_times, results_list[2][2,:])
+plot!(p, sample_times, results_list[3][2,:])
+plot!(p, sample_times, results_list[4][2,:])
```
isaacsas, line 55: Just use a loop.

docs/src/inverse_problems/examples/SPR_fitting.md
```diff
@@ -0,0 +1,110 @@
...
+
+    optf = OptimizationFunction(loss)
+    
+    optprob = OptimizationProblem(optf, pinit)
+    sol = solve(optprob, Optim.NelderMead())
+
+    return sol.u
+end
+
+p_estimate = optimise_p([100.0, 10.0, 1.0], 1.5)
```
isaacsas, line 96:
```suggestion
p_estimate = optimise_p([100.0, 10.0, 1.0], 1.5)
```
Shouldn't `tend` here by `switch_time`? Why are you setting it to be smaller?

docs/src/inverse_problems/examples/SPR_fitting.md
```diff
@@ -0,0 +1,110 @@
...
+```
+Finally, we remake the solution using the estimate, fit the entire time span for some `alpha`, and plot the solution.
```
isaacsas, line 98: But you don't seem to be fitting a second time using the parameters from the first fit?

PR #707, opened 2023-10-28
Bifurcation kit extension

TorkelE:
Enables BifurcationKit `BifurcationProblem`s to be created from `ReactionSystem`s.

Example workflow:
```julia
using Catalyst
wilhelm_2009_model = @reaction_network begin
    k1, Y --> 2X
    k2, 2X --> X + Y
    k3, X + Y --> Y
    k4, X --> 0
    k5, 0 --> X
end

using BifurcationKit
bif_par = :k1
u_guess = [:X => 5.0, :Y => 2.0]
p_start = [:k1 => 4.0, :k2 => 1.0, :k3 => 1.0, :k4 => 1.5, :k5 => 1.25]
plot_var = :X
bprob = BifurcationProblem(wilhelm_2009_model, u_guess, p_start, bif_par; plot_var=plot_var)

p_span = (2.0, 20.0)
opts_br = ContinuationPar(p_min = p_span[1], p_max = p_span[2], max_steps=1000)

bif_dia = bifurcationdiagram(bprob, PALC(), 2, (args...) -> opts_br; bothside=true)

using Plots
plot(bif_dia; xguide="k1", yguide="X")
```


TorkelE:
Functionality is there (just need to add an error check), needs tests and docs. Requires https://github.com/SciML/ModelingToolkit.jl/pull/2337 to merge to handle cases with conservation laws

ext/CatalystBifurcationKitExtension/bifurcation_kit_extension.jl
```diff
@@ -0,0 +1,19 @@
...
+    # Converts symbols to symbolics.
+    (bif_par isa Symbol) && (bif_par = rs.var_to_name[bif_par])
+    (plot_var isa Symbol) && (plot_var = rs.var_to_name[plot_var])
+    (u0_bif isa Vector{Pair{Symbol, Float64}}) && (u0_bif = symmap_to_varmap(rs, u0_bif))
+    (ps isa Vector{Pair{Symbol, Float64}}) && (ps = symmap_to_varmap(rs, ps))
+    (u0 isa Vector{Pair{Symbol, Float64}}) && (u0 = symmap_to_varmap(rs, u0))
+
+    # Creates NonlinearSystem.
+    conservationlaw_errorcheck(rs, vcat(ps, u0))
+    nsys = convert(NonlinearSystem, rs; remove_conserved=true, defaults=Dict(u0))
```
isaacsas, line 15: What happens to existing defaults a user might have specified in building `rs`?

TorkelE: If they are given in `u0`, they are overridden. If not, they should be accounted for.

TorkelE: I have updated the `convert` functions slightly to help with this:
```
function Base.convert(::Type{<:NonlinearSystem}, rs::ReactionSystem; name = nameof(rs),
                      include_zero_odes = true, remove_conserved = false, checks = false,
                      kwargs...)
    spatial_convert_err(rs::ReactionSystem, NonlinearSystem)
    fullrs = Catalyst.flatten(rs)
    remove_conserved && conservationlaws(fullrs)
    ists, ispcs = get_indep_sts(fullrs, remove_conserved)
    eqs = assemble_drift(fullrs, ispcs; combinatoric_ratelaws, remove_conserved,
    error_if_constraint_odes(NonlinearSystem, fullrs)
    eqs, sts, ps, obs, defs = addconstraints!(eqs, fullrs, ists, ispcs; remove_conserved)
    NonlinearSystem(eqs, sts, ps;
                    name,
                    observed = obs,
                    defaults = _merge(defaults,defs),
                    checks,
                    kwargs...)
end
```

TorkelE: This shoudl also be (partially) tested in the test case (in principle I could do various combinations over overriding and not)
```julia
let 
    # Create model
    extended_brusselator = @reaction_network begin
        @species W(t) = 2.0
        @parameters k2 = 0.5
        A, ∅ → X
        1, 2X + Y → 3X
        B, X → Y
        1, X → ∅
        (k1*Y, k2), V <--> W
    end
    @unpack A, B, k1
    u0_guess = [:X => 1.0, :Y => 1.0, :V => 0.0, :W => 0.0]
    p_start = [A => 1.0, B => 4.0, k1 => 0.1]
    
    # Computes bifurcation diagram.
    BifurcationProblem(extended_brusselator, u0_guess, p_start, :B; plot_var=:V, u0 = [:V => 1.0])
    p_span = (0.1, 6.0)
    opt_newton = NewtonPar(tol = 1e-9, max_iterations = 100)
    opts_br = ContinuationPar(dsmin = 0.0001, dsmax = 0.001, ds = 0.0001,
        max_steps = 200000, nev = 2, newton_options = opt_newton,
        p_min = p_span[1], p_max = p_span[2],
        detect_bifurcation = 3, n_inversion = 4, tol_bisection_eigenvalue = 1e-8, dsmin_bisection = 1e-9)
    bif_dia = bifurcationdiagram(bprob, PALC(), 2, (args...) -> opts_br; bothside=true)
    
    # Checks computed V values are correct (Formula: V = k2*(V0+W0)/(k1*Y+k2), where Y=2*B.)
    B_vals = getfield.(bif_dia.γ.branch, :param)
    V_vals = getfield.(bif_dia.γ.branch, :x)
    @test all(V_vals .≈ 0.5*(1.0+2.0) ./ (0.1 .* 2*B_vals .+ 0.5))
    
    # Checks that the bifurcation point is correct.
    @test length(bif_dia.γ.specialpoint) == 3 # Includes start and end point.
    hopf_bif_point = filter(sp -> sp.type == :hopf, bif_dia.γ.specialpoint)[1]
    @test isapprox(hopf_bif_point.param, 1.5, atol=1e-5)

    # Tests that an error is thrown if information of conserved species is not fully provided.
    @test_throws Exception BifurcationProblem(extended_brusselator, u0_guess, p_start, :B; plot_var=:V, u0 = [:X => 1.0])
end
```

rveltz: nev = 2: are you sure you dont have more variables?

TorkelE: I just copied some ContPar code that worked, I will try to remove the necessary fields (and update nev to the number of variables!)

TorkelE:
The required MTK PR have now been merged. However, I think we still would need to wait for a new release to take advantage of it.

src/reactionsystem.jl
```diff
@@ -1456,7 +1461,7 @@ function Base.convert(::Type{<:JumpSystem}, rs::ReactionSystem; name = nameof(rs
     JumpSystem(eqs, get_iv(flatrs), sts, ps;
                observed = MT.observed(flatrs),
                name,
-               defaults = MT.defaults(flatrs),
+               defaults = _merge(defaults,MT.defaults(flatrs)),
```
isaacsas, line 1464: Does merging after flattening give the correct namespacing of parameters and variables? This seems like it is going to cause problems. Merging probably needs to happen before flattening.

TorkelE: Not, that is a good point. I just used the current order of things, but changing it around might be better.

isaacsas: Can you add a test to check this?

TorkelE: You mean providing defaults to a `convert` for a system with sub-systems, which got defaults?

isaacsas: Yes

isaacsas (changes requested):
Can you check this works with coupled ODEs and add a corresponding test? (Do we want to allow coupled algebraic equations here too? If not, we probably need a check that any coupled equations are just ODEs.)

TorkelE:
This should be ready now. Unfortnatley the CI seems bugged out, keep giving stuff like:
```
[2048] signal (2): Interrupt
in expression starting at /home/runner/work/_temp/6799a5b3-ac8d-4e0e-a488-c620d88596be:24
epoll_wait at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
uv__io_poll at /workspace/srcdir/libuv/src/unix/epoll.c:236
uv_run at /workspace/srcdir/libuv/src/unix/core.c:400
ijl_task_get_next at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/partr.c:390
poptask at ./task.jl:974
wait at ./task.jl:983
task_done_hook at ./task.jl:672
jfptr_task_done_hook_27081.clone_1 at /opt/hostedtoolcache/julia/1.9.4/x64/lib/julia/sys.so (unknown line)
_jl_invoke at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/gf.c:2758 [inlined]
ijl_apply_generic at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/gf.c:2940
jl_apply at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/julia.h:1880 [inlined]
jl_finish_task at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/task.c:320
start_task at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-9/src/task.c:1103
unknown function (ip: (nil))
Allocations: 10537356 (Pool: 10521420; Big: 15936); GC: 22
Error: The operation was canceled.
```
in various places, and I have no idea why.

TorkelE:
CI works on my computer, but keeps failing for reasons here. I the PR is done. I think if I rebase on the right stuff it should be good, just tell me when we can merge this one and I will rebase and it should hopefully work.
(I just want to avoid having to rebase this one 8 times over the next week or 2, each time potentially doing some git wrong)

Project.toml
```diff
@@ -22,9 +22,11 @@ Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
...
 
 [weakdeps]
+BifurcationKit = "0f109fa4-8a5d-4b75-95aa-f515264e7665"
 HomotopyContinuation = "f213a82b-91d6-5c5d-acf7-10f1c761b327"
 
 [extensions]
+CatalystBifurcationKitExtension = "BifurcationKit"
 CatalystHomotopyContinuationExtension = "HomotopyContinuation"
 
 [compat]
```
isaacsas, line 32: I think we need compat entries for each weakdep too. See https://github.com/SciML/SciMLBase.jl/blob/master/Project.toml

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
 # [Bifurcation Diagrams](@id bifurcation_diagrams)
-Bifurcation diagrams can be produced for models generated by Catalyst through the
-use of the [BifurcationKit.jl](https://bifurcationkit.github.io/BifurcationKitDocs.jl/stable/)
-package. This tutorial gives a simple example of how to create such a
-bifurcation diagram.
 
-!!! note
-    Catalyst 13.0 and up require at least BifurcationKit v0.2.4.
+Bifurcation diagrams describes how, for a dynamic system, the quantity and quality of its steady states changes with a parameter's value[^1]. These can be computed through the [BifurcationKit.jl](https://github.com/bifurcationkit/BifurcationKit.jl) package. Catalyst provides a simple interface for creating BifurcationKit compatible `BifurcationProblem`s from `ReactionSystem`s. If you use this feature in your research, please [cite the BifurcationKit.jl](@ref bifurcation_kit_citation) and [Catalyst.jl](@ref catalyst_citation) publications.
```
isaacsas, line 3:
```suggestion
Bifurcation diagrams describe how, for a dynamical system, the quantity and type of its steady states change as a parameter is varied[^1]. When using Catalyst-generated models, these can be computed with the [BifurcationKit.jl](https://github.com/bifurcationkit/BifurcationKit.jl) package. Catalyst provides a simple interface for creating BifurcationKit compatible `BifurcationProblem`s from `ReactionSystem`s. If you use this feature in your research, please [cite the BifurcationKit.jl](@ref bifurcation_kit_citation) and [Catalyst.jl](@ref catalyst_citation) publications.
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+
+This tutorial briefly introduces how to use Catalyst in conjecture with BifurcationKit through basic examples, with BifurcationKit.jl providing [a more extensive documentation](https://bifurcationkit.github.io/BifurcationKitDocs.jl/stable/). Especially for more complicated systems, where careful tuning of algorithm options might be required, reading the BifurcationKit documentation is recommended. Finally, BifurcationKit provides many additional features not described here, including [computation of periodic orbits](https://bifurcationkit.github.io/BifurcationKitDocs.jl/stable/periodicOrbit/), [tracking of bifurcation points along secondary parameters](https://bifurcationkit.github.io/BifurcationKitDocs.jl/dev/branchswitching/), and [bifurcation computations for PDEs](https://bifurcationkit.github.io/BifurcationKitDocs.jl/dev/tutorials/tutorials/#PDEs:-bifurcations-of-equilibria).
```
isaacsas, line 5:
```suggestion
This tutorial briefly introduces how to use Catalyst with BifurcationKit through basic examples, with BifurcationKit.jl providing [a more extensive documentation](https://bifurcationkit.github.io/BifurcationKitDocs.jl/stable/). Especially for more complicated systems, where careful tuning of algorithm options might be required, reading the BifurcationKit documentation is recommended. Finally, BifurcationKit provides many additional features not described here, including [computation of periodic orbits](https://bifurcationkit.github.io/BifurcationKitDocs.jl/stable/periodicOrbit/), [tracking of bifurcation points along secondary parameters](https://bifurcationkit.github.io/BifurcationKitDocs.jl/dev/branchswitching/), and [bifurcation computations for PDEs](https://bifurcationkit.github.io/BifurcationKitDocs.jl/dev/tutorials/tutorials/#PDEs:-bifurcations-of-equilibria).
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
 end
 ```
-Next, we specify the system parameters for which we wish to plot the bifurcation
-diagram. We also set the parameter we wish to vary in our bifurcation diagram,
-as well as the interval to vary it over. Finally, we set which variable we wish
-to plot the steady state values of in the bifurcation plot.
+Next we will create a `BifurcationProblem`. In addition to the `ReactionSystem`, we need to provide:
+- The bifurcation parameter (the parameter which is varied in the bifurcation diagram).
+- A full model parameter set. This includes the values of all non-bifurcation parameters, but also a value for the bifurcation parameter (which correspond to the point in parameter space from which the computation of the bifurcation diagram starts).
+- An initial guess of the steady state values of the system at the provided parameter set. Using this point as a start, BifurcationKit uses Newton's method to find an initial steady state from which to compute the bifurcation diagram. Hence, this guess does not need to be very exact (but may be important if the system exhibits multistability for the initial parameter set).
```
isaacsas, line 23:
```suggestion
- An initial guess of the steady state values of the system at the provided parameter set. Using this point as a starting guess for root finding, BifurcationKit calculates an initial steady state from which to compute the bifurcation diagram. Hence, this guess does not need to be very exact (but may be important if the system exhibits multistability for the initial parameter set).
```
It is more sophisticated than plain Newton's Method I think, so probably better to not get into that level of detail.

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+- The species which concentration we wish to plot on the y-axis of the bifurcation diagram (alternatively, a custom value can be provided by using the [`record_from_solution` argument](https://bifurcationkit.github.io/BifurcationKitDocs.jl/stable/periodicOrbit/#.-record*from*solution)).
```
isaacsas, line 24:
```suggestion
- The species or statistic we wish to plot on the y-axis of the bifurcation diagram (a custom value can be provided by using the [`record_from_solution` argument](https://bifurcationkit.github.io/BifurcationKitDocs.jl/stable/periodicOrbit/#.-record*from*solution)).
```
Does this work with symbolic expressions? If we can intercept this to make it work with symbolic expressions that would be cool. (Maybe a good followup PR and issue?)

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+
+We combines all this information to form a `BifurcationProblem`:
```
isaacsas, line 26:
```suggestion
We combine all this information to form a `BifurcationProblem`:
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
-we wish to compute the bifurcation diagram, `p_span[1]`. We thus create a
-modified parameter set where `S = 0.1`. For this parameter set, we also guess
-the steady state of the system. While a good estimate could be provided through
-an ODE simulation, BifurcationKit does not require the guess to be very
-accurate.
+
+BifurcationKit computes bifurcation diagrams using the `bifurcationdiagram` function. From an initial point in the diagram, it tracks the solution (using a continuation algorithm) until the entire diagram is computed (BifurcationKit's continuation can be used for other purposes, however, this tutorial focuses on bifurcation diagram computation). The continuation settings are provided in a `ContinuationPar` structure. In this example, we will only specify three settings, `p_min` and `p_max` (which sets the minimum and maximum values over which the bifurcation parameter is varied) and `max_steps` (the maximum number of continuation steps to take as the bifurcation diagram is tracked). We wish to compute a bifurcation diagram over the interval *(2.0,20.0)*, and will use the following settings:
```
isaacsas, line 37:
```suggestion
BifurcationKit computes bifurcation diagrams using the `bifurcationdiagram` function. From an initial point in the diagram, it tracks the solution (using a continuation algorithm) until the entire diagram is computed (BifurcationKit's continuation can be used for other purposes, however, this tutorial focuses on bifurcation diagram computation). The continuation settings are provided in a `ContinuationPar` structure. In this example, we will only specify three settings, `p_min` and `p_max` (which sets the minimum and maximum values over which the bifurcation parameter is varied) and `max_steps` (the maximum number of continuation steps to take as the bifurcation diagram is tracked). We wish to compute a bifurcation diagram over the interval `(2.0,20.0)`, and will use the following settings:
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+Finally, we compute our bifurcation diagram using:
 ```@example ex1
-oprob = ODEProblem(rn, u0, (0.0, 0.0), p_bstart; jac = true)
-F = (u,p) -> oprob.f(u, p, 0)
-J = (u,p) -> oprob.f.jac(u, p, 0)
-nothing   # hide
+bif_dia = bifurcationdiagram(bprob, PALC(), 2, (args...) -> opts_br; bothside=true)
+nothing # hide
 ```
+Where `PALC()` designates that we wish to use the pseudo arclength continuation method to track our solution. The third argument (`2`) designates the maximum number of recursions when branches of branches are computed (branches appear as continuation encounter certain bifurcation points). For diagrams with highly branches structure (rare for most standard chemical reaction networks) this input is important. Finally, `bothside=true` designates that we wish to perform continuation on both sides of the initial point (which is typically the case).
```
isaacsas, line 49:
```suggestion
Where `PALC()` designates that we wish to use the pseudo arclength continuation method to track our solution. The third argument (`2`) designates the maximum number of recursions when branches of branches are computed (branches appear as continuation encounters certain bifurcation points). For diagrams with highly branched structures (rare for many common chemical reaction networks) this input is important. Finally, `bothside=true` designates that we wish to perform continuation on both sides of the initial point (which is typically the case). 
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
-@unpack S, X = rn
-
-# find their indices in oprob.p and oprob.u0 respectively
-bif_idx  = findfirst(isequal(S), parameters(rn))
-plot_idx = findfirst(isequal(X), species(rn))
-nothing   # hide
+using Plots
+plot(bif_dia; xguide="k1", yguide="X")
 ```
+Here, the steady state concentration of *X* is shown as a function of *k1*'s value. Stable steady states are shown with thick lines, unstable ones with thin lines. The two fold bifurcation points are marked with "bp".
```
isaacsas, line 56: Can we make "fold bifurcation point" a reference to a webpage that describes type of bifurcations?

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
 
-Now, we load the required packages to create and plot the bifurcation diagram.
-We also bundle the information we have compiled so far into a
-`BifurcationProblem`.
-```@example ex1
-using BifurcationKit, Plots, LinearAlgebra, Setfield
+## Additional `ContinuationPar` options
+Most of the options required by the `bifurcationdiagram` function is not provided directly, but instead provided through the `ContinuationPar` structure. For full details, please read the [BifurcationKit documentation](https://bifurcationkit.github.io/BifurcationKitDocs.jl/dev/library/#BifurcationKit.ContinuationPar). However, a few common options, and how they affect the continuation computation, are described here:
```
isaacsas, line 59:
```suggestion
Most of the options required by the `bifurcationdiagram` function are provided through the `ContinuationPar` structure. For full details, please read the [BifurcationKit documentation](https://bifurcationkit.github.io/BifurcationKitDocs.jl/dev/library/#BifurcationKit.ContinuationPar). However, a few common options, and how they affect the continuation computation, are described here:
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+- `p_min` and `p_max`: Sets the interval over which the bifurcation diagram is computed (with the continuation stopping if it reaches these bounds).
```
isaacsas, line 60:
```suggestion
- `p_min` and `p_max`: Set the interval over which the bifurcation diagram is computed (with the continuation stopping if it reaches these bounds).
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
 
-We are now ready to compute the bifurcation diagram:
+## Bifurcation diagrams with disjoint branches
+Let's consider the previous case, but instead compute the bifurcation diagram over the interval *(2.0, 15.0)*:
 ```@example ex1
-bf = bifurcationdiagram(bprob, PALC(), 2, (args...) -> bopts)
-nothing   # hide
+p_span = (2.0, 15.0)
+opts_br = ContinuationPar(p_min = p_span[1], p_max = p_span[2], max_steps = 1000)
+bif_dia = bifurcationdiagram(bprob, PALC(), 2, (args...) -> opts_br; bothside=true)
```
isaacsas, line 82: Maybe we should suggest to Romain a dispatch that if a `ContinuationPar` is passed directly as the fourth argument he just directly build the `(args...) -> opts_br` function? This would make things easier on users.

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+plot(bif_dia; xguide="k1", yguide="X")
 ```
-Finally, we can plot it:
-```@example ex1
-plot(bf; xlabel = string(bif_par), ylabel = string(plot_var))
+Here, in the bistable region, we only see a single branch. The reason is that the continuation algorithm start at our initial guess (here made at *k1 = 4.0* for *(X,Y) = (5.0,2.0)*) and tracks the diagram from there. However, with the upper bound set at *k1=15.0* the bifurcation diagram have a disjoint branch structure, preventing the full diagram from being computed by continuation alone. In this case it could be solved by increasing the bound of *k1=15.0*, however, this is not possible in all cases. In these cases, *deflation* can be used. This is described in the [BifurcationKit documentation](https://bifurcationkit.github.io/BifurcationKitDocs.jl/dev/tutorials/tutorials2/#Snaking-computed-with-deflation).
```
isaacsas, line 85:
```suggestion
Here, in the bistable region, we only see a single branch. The reason is that the continuation algorithm starts at our initial guess (here made at `k1 = 4.0` for `(X,Y) = (5.0,2.0)`) and tracks the diagram from there. However, with the upper bound set at `k1=15.0` the bifurcation diagram has a disjoint branch structure, preventing the full diagram from being computed by continuation alone. In this case it could be solved by increasing the bound from `k1=15.0`, however, this is not possible in all cases. In these cases, *deflation* can be used. This is described in the [BifurcationKit documentation](https://bifurcationkit.github.io/BifurcationKitDocs.jl/dev/tutorials/tutorials2/#Snaking-computed-with-deflation).
```
Please don't use italic blocks (i.e. stars), either use code or math blocks as appropriate. I'm going to stop commenting on this, but please take a pass and update the tutorial to use one of these two.

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+
+
+## Systems with conservation laws
+Some systems are under-determined, and for a given parameter set they have an infinite number of possible steady states, preventing bifurcation diagrams from being computed. Similar to when we [compute single steady states](@ref homotopy_continuation_conservation_laws), we can utilise Catalyst's ability to detect and eliminate conservation laws to resolve this issue. This requires us to provide information of the species concentrations at which we wish to compute the bifurcation diagram. These are provided to the `BifurcationProblem` using the `u0` argument.
```
isaacsas, line 89:
```suggestion
Some systems are under-determined determined at steady-state, so that for a given parameter set they have an infinite number of possible steady state solutions, preventing bifurcation diagrams from being computed. Similar to when we [compute steady states for fixed parameter values](@ref homotopy_continuation_conservation_laws), we can utilise Catalyst's ability to detect and eliminate conservation laws to resolve this issue. This requires us to provide information of the species concentrations at which we wish to compute the bifurcation diagram (to determine the values of conserved quantities). These are provided to the `BifurcationProblem` using the `u0` argument.
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+
+To illustrate this, we will create a simple model of a kinase that is produced and degraded (at rates *p* and *d*). The kinase facilitates the phosphorylation of a protein (*X*), which is dephosphorylated at a constant rate. For this system, we will compute a bifurcation diagram, showing how the concentration of the phosphorylated protein (*Xp*) depends on the degradation rate of the kinase (*d*). We will set the total amount of protein (*X+Xp*) to *1.0*.
```
isaacsas, line 91:
```suggestion
To illustrate this, we will create a simple model of a kinase that is produced and degraded (at rates $p$ and $d$). The kinase facilitates the phosphorylation of a protein ($X$), which is dephosphorylated at a constant rate. For this system, we will compute a bifurcation diagram, showing how the concentration of the phosphorylated protein ($Xp$) depends on the degradation rate of the kinase ($d$). We will set the total amount of protein ($X+Xp$) to $1.0$.
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+p_start = [:p => 1.0, :d => 0.5, :kP => 2.0, :kD => 5.0]
+u0 = [:X => 1.0, :Xp => 0.0]
+bprob = BifurcationProblem(kinase_model, u_guess, p_start, :d; plot_var=:Xp, u0=u0)
+
+p_span = (0.1, 10.0)
+opts_br = ContinuationPar(p_min = p_span[1], p_max = p_span[2], max_steps = 1000)
+bif_dia = bifurcationdiagram(bprob, PALC(), 2, (args...) -> opts_br; bothside=true)
+plot(bif_dia; xguide="d", yguide="Xp")
 ```
+This bifurcation diagram does not contain any interesting features (such as bifurcation points), and only shows how the steady state concentration of *Xp* is reduced as *d* increases. For this example, we will note two facts:
```
isaacsas, line 109:
```suggestion
This bifurcation diagram does not contain any interesting features (such as bifurcation points), and only shows how the steady state concentration of $Xp$ is reduced as $d$ increases. For this example, we will note two facts:
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+- When providing the concentrations for computing the conserved quantities (in `u0`), we only have to designate the concentrations of species that are actually involved in conservation laws. For larger systems, determining which one are may, however, be difficult. In this case, it might still be wise to provide concentrations for all species.
```
isaacsas, line 110: We could add a function to return what species appear in a conservation law. We already store the independent and dependent species and the conservation laws so it would be easy to calculate.

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
+- The steady state guess in `u_guess` does not actually have to fulfil the conserved concentrations provided in `u0`.
```
isaacsas, line 111: Is this really true? Can you add an explanation of the difference between `u_guess` and `u0` here and how they are both used?

docs/src/catalyst_applications/bifurcation_diagrams.md
isaacsas: Why not just make the guess be `u0` and then extract the vector that is needed to pass into BifurfactionKit?

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,118 @@
...
 
-Here, the Hopf bifurcation is marked with a red dot and the fold bifurcations
-with blue dots. The region with a thinner line width corresponds to unstable
-steady states.
+!!! note
+    Computing bifurcation diagrams for [hierarchical models created using compostability](@ref compositional_modeling), that also contain conservations laws, is currently not supported. For these, please [flatten](@ref api_network_extension_and_modification) the system first.
```
isaacsas, line 114:
```suggestion
    Computing bifurcation diagrams for [hierarchical models created by composing multiple systems](@ref compositional_modeling), that also contain conservations laws, is currently not supported. For these, please [flatten](@ref api_network_extension_and_modification) the system first.
```
Why is this not supported? Conversion to ODEs/Nonlinear systems flattens anyways.

src/miscellaneous.jl
```diff
@@ -0,0 +1,13 @@
+# File containing code which I am currently unsure which file in which it belongs. Long-term functions here should be moved elsewhere.
+
+
+# If u0s are not given while conservation laws are present, throws an error.
+# If the system is nested, the function is skipped (conservation laws cannot be computed for nested systems).
+# Used in HomotopyContinuation and BifurcationKit extensions.
+function conservationlaw_errorcheck(rs, pre_varmap)
```
isaacsas, line 7: Maybe just put this in the network api file near the conservation law code? Seems more appropriate to go with that code. Then we can drop having a misc file.

test/runtests.jl
```diff
@@ -52,6 +52,7 @@ using SafeTestsets
     
     ### Tests extensions. ###
     if VERSION >= v"1.9"
```
isaacsas, line 54: This can be dropped now once rebased since we only support 1.9 and up going forward.

isaacsas (changes requested):
Let's get this working with composed systems too. I think it should just require manually flattening in the `BifurcationProblem` before you start making checks and calculating conservation laws.

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: Wouldn't work out of the bat. We could make a routine which takes a symbolic expression and generates a function. Long-term, I think BifurcationKit is the place for a lot of stuff like this. E.g. if it could use `NonlinearSystem`s internally it could just use all the nonlinear solvers from NonlinearSolve.jl (instead of implementing its own). It would also be possible to track the name of the bifurcation parameter and use that in the plots (instead of just calling it `p`).

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: because of the conservation law error check

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: Because people might not make steady state guesses that fulfil their desired conservation law (e.g. just putting all numbers equal to `1`, I usually do that).

TorkelE: yeah, adding extra functionality to conservation laws would be good.

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: `u_guess` is just a guess of the initial steady state.

`u0` is used to compute conservation laws (e.g. what the conserved quantity X1+X2 should be). Yes, it is annoying to have to have both, but feel like if we just use `u_guess` people will do something trivial (like `ones`), and never realise that it is used for conserved quantities computation.

src/miscellaneous.jl
TorkelE: happy to move it there.

(stability stuff is currently in the misc file as well, but yes, I agree that the misc file should be dropped, but figured if I put stuff somewhere it was likely you'd want it moved anyway)

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: Was meant to be a mathematical formula, will witch to $$ (think this was written before you told me to use that instead of **)

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: Still want to use `most common` though. As in I think it is the case for the majority of CRNs, and this just suggest that one should not be surprised not to see a highly branched structure in a CRN (rather than that one should expect it, but correct me if I am wrong)

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: agreed

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: will do
(as mentioned this was before you told me about the $$, and I failed to go back and look at the old tutorials, I agree ** shoudln't be used in these cases as there are better alts)

Project.toml
TorkelE: Added:
```
BifurcationKit = "0.3"

...

HomotopyContinuation = "2.9"
```

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: https://en.wikipedia.org/wiki/Saddle-node_bifurcation

src/miscellaneous.jl
TorkelE: moved

test/runtests.jl
TorkelE: done

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: rewritten this to make everything extra clear

docs/src/catalyst_applications/bifurcation_diagrams.md
TorkelE: Hopefully this will happen soon-ish: https://github.com/bifurcationkit/BifurcationKit.jl/issues/132

TorkelE:
this is ready, think the doc failure has appeared recently though. I will rebase on master and if docs still fail fixing it should be relatively straightforward.

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,120 @@
...
+    k5, 0 --> X
 end
 ```
-Next, we specify the system parameters for which we wish to plot the bifurcation
-diagram. We also set the parameter we wish to vary in our bifurcation diagram,
-as well as the interval to vary it over. Finally, we set which variable we wish
-to plot the steady state values of in the bifurcation plot.
+Next we will create a `BifurcationProblem`. In addition to the `ReactionSystem`, we need to provide:
+- The bifurcation parameter (the parameter which is varied in the bifurcation diagram).
+- A full model parameter set. This includes the values of all non-bifurcation parameters, but also a value for the bifurcation parameter (which correspond to the point in parameter space from which the computation of the bifurcation diagram starts).
```
isaacsas, line 22:
```suggestion
- A full model parameter set. This includes the values of all non-bifurcation parameters, but also a value for the bifurcation parameter (which corresponds to the point in parameter space from which the computation of the bifurcation diagram starts).
```

docs/src/catalyst_applications/bifurcation_diagrams.md
```diff
@@ -1,115 +1,120 @@
...
-with blue dots. The region with a thinner line width corresponds to unstable
-steady states.
+Finally, for additional clarity, we reiterate the purpose of the two `u` arguments used:
+- `u_guess`: A guess of the initial steady states (which BifurcationKit uses to find its starting point). Typically, most trivial guesses work (e.g. setting all species concentrations to `1.0`). `u_guess` *does not* have to fulfil the conserved concentrations provided in `u0`.
+- `u0`: Used to compute the concentrations of any conserved quantities (e.g. in our example $X + Xp = 1.0$). Technically, values are only required for species that are involved in conservation laws (in our case we do not need to provide a value for $K$). However, sometimes determining which species are actually involved in conservation laws can be difficult, and it might be easier to simply provide concentrations for all species.
+
+!!! note
+    Computing bifurcation diagrams for [hierarchical models created by composing multiple systems](@ref compositional_modeling), that also contain conservations laws, is currently not supported. For these, please [flatten](@ref api_network_extension_and_modification) the system first.
```
isaacsas, line 116: I thought we had discussed this. What isn't working here? Conversion to a nonlinear system should already flatten implicitly.

TorkelE: fixed

PR #708, opened 2023-11-01
Tutorial for using Optimization and DiffEqParamEstim for parameter fitting

TorkelE:
Adds a single tutorial page. Currently put under `catalyst_applications`. Once https://github.com/SciML/Catalyst.jl/pull/701 passes I will separately add this to the inverse problem section.

The base tutorial is done (but needs to be reread a bit), but some auxiliary sections need to be added.


isaacsas (commented):
How does this fit in with the current Optimization based tutorial?

Also, why use `CMAEvolutionStrategyOpt`?

docs/src/catalyst_applications/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,168 @@
+# [Parameter Fitting for ODEs using SciML/Optimization.jl and DiffEqParamEstim.jl](@id optimization_parameter_fitting)
+Fitting parameters to data involves solving an optimisation problem (that is, finding the parameter set that optimally fits you model to your data, typically by minimising a cost function). The SciML ecosystem's primary package for solving optimisation problems is [Optimization.jl](https://github.com/SciML/Optimization.jl). While it does not implement any optimisation methods itself, it wraps the large number of optimisation methods that have been implemented in Julia into a single common interface.
```
Vaibhavdixit02, line 2: We have started implementing some methods so just for future it might be safer to just say 

```
It provides access to a large variety of solvers from a single common interface.
```

TorkelE:
> How does this fit in with the current Optimization based tutorial?

That will be moved to an separate example (and rewritten a bit and be a focus on Fitting parameters of oscillatory processes).

> Also, why use CMAEvolutionStrategyOpt?

I just picked one (don't really know much on sensible selection of optimisation methods), happy to change.

Vaibhavdixit02:
I second Sam's point above, CMAES doesn't seem like a good choice here. I would recommend using a first order method (BFGS) instead, you can use it from NLopt so that'll cover multiple solver packages as well which might be useful for users to see.

TorkelE:
Yeah, I just wanted to select a optimiser from another package, I can change to that one.

docs/src/catalyst_applications/optimization_ode_param_fitting.md
TorkelE: Will change

TorkelE: I've updated. I still mentions that it imports and wraps other packages, since I think that is useful for the user to know given what is going on later.

TorkelE:
Updated to use BFGS instead.

docs/src/catalyst_applications/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,168 @@
...
+plot!(fitted_sol; idxs=:P, label="Fitted solution", linestyle=:dash, lw=6, color=:lightblue)
+```
+
+!!! note
+Here, a good exercise is to check the resulting parameter set and note that, while it creates a good fit to the data, it does not actually correspond to the original parameter set. [Identifiability](https://www.sciencedirect.com/science/article/pii/S1364815218307278) is a concept that studies how to deal with this problem.
+
+Say that we instead would like to use the [Broyden–Fletcher–Goldfarb–Shannon](https://en.wikipedia.org/wiki/Broyden%E2%80%93Fletcher%E2%80%93Goldfarb%E2%80%93Shanno_algorithm) algorithm, as implemented by the [NLopt.jl](https://github.com/JuliaOpt/NLopt.jl) package. In this case we would run:
+```@example diffeq_param_estim_1 
+using OptimizationNLopt
+sol = solve(optprob, BFGS())
```
Vaibhavdixit02, line 85:
```suggestion
sol = solve(optprob, NLopt.LD_LBFGS())
```

TorkelE: For this example, do you have a suggestion of a method that I can reach without doing `packagename.optimiser`. Alternatively, have a second one used this way. I want to keep things simple, and it feels weird accessing the two optimisers (currently `NelderMead` and `BFGS`) using two different interfaces.

Vaibhavdixit02: You can change the NelderMEad to Optim.NelderMead that'll still work

TorkelE: Sounds perfect.

Why does some methods need to be accessed like `packagename.optimiser` ad some can be accessed directly?

TorkelE:
This is ready for merging now.

Vaibhavdixit02 (approved):

isaacsas:
I need to go through this carefully still, but I think I’m going to have the same comment about not dynamically generating graphs and such.

TorkelE:
See the reply there. Yes, I agree. I will see what can be done. We could always merge and put in an issue that this should be dealt with at some point of time (although we are starting to get so many issues that it is easy for stuff to get lost there, which makes me slightly reluctant)

TorkelE:
This one is ready for merging

isaacsas:
Ok, I’ll let you know once I’ve read it carefully.

TorkelE:
replaced by https://github.com/SciML/Catalyst.jl/pull/740

PR #709, opened 2023-11-01
Structural Identifiability Extension

TorkelE:
Creates an extension for Catalyst and the StructuralIdentifiability package (https://github.com/SciML/StructuralIdentifiability.jl).

Implements:
- dispatch for calling `assess_local_identifiability` on `ReactionSystem`s
- dispatch for calling `assess_identifiability` on `ReactionSystem`s
- dispatch for calling `find_identifiable_functions` on `ReactionSystem`s
- The `make_si_ode` function, which takes a Catalyst `ReactionSystem` and creates a ODE model of the form used in StructuralIdentifiability.jl, that can be used for some of its more detailed features.

Example usage:
```julia
goodwind_oscillator = @reaction_network begin
    (mmr(P,pₘ,1), dₘ), 0 <--> M
    (pₑ*M,dₑ), 0 <--> E
    (pₚ*E,dₚ), 0 <--> P
end
assess_identifiability(goodwind_oscillator; measured_quantities=[:M])
```
here, we assess global identifiability for the model, given us being able to measure `M`.


TorkelE:
THIS PR is now ready. There are two outstanding issues:
- Handling parameters in exponents (like occurs in e.g. Hill functions). This will be complicated, and require some major work. It will have to be a future update.
- - Handling conservation laws to speed up computation. Hopefully some additional functionality will come of in StructuralIdentfiability to simplify this. What to do about this depend on that.

sumiya11:
Hi Torkel! Thanks again for working on this !

I think the tutorial looks very good.

I wanted to add a couple more simple tests but it looks like I cannot push to this PR. So I will just paste it here, feel free to add these to tests:

```julia
# Test some examples explicitly
let
    rs = @reaction_network begin
        k1, x1 --> x2
    end
    # Measure the source
    id_report = assess_identifiability(rs, measured_quantities = [:x1])
    @test sym_dict(id_report) == Dict(
        :x1 => :globally,
        :x2 => :nonidentifiable,
        :k1 => :globally
    )
    # Measure the target instead
    id_report = assess_identifiability(rs, measured_quantities = [:x2])
    @test sym_dict(id_report) == Dict(
        :x1 => :globally,
        :x2 => :globally,
        :k1 => :globally
    )

    # Example from
    #   Identifiability of chemical reaction networks
    #   DOI: 10.1007/s10910-007-9307-x
    # The rate constants a, b, c are not identifiable even if all of the species
    # are observed.
    rs = @reaction_network begin
        a, A0 --> 2A1
        b, A0 --> 2A2
        c, A0 --> A1 + A2
    end
    id_report = assess_identifiability(rs, measured_quantities = [:A0, :A1, :A2])
    @test sym_dict(id_report) == Dict(
        :A0 => :globally,
        :A1 => :globally,
        :A2 => :globally,
        :a  => :nonidentifiable,
        :b  => :nonidentifiable,
        :c  => :nonidentifiable
    )

    # Test with no parameters
    rs = @reaction_network begin
        1, x1 --> x2
        1, x2 --> x3
    end
    id_report = assess_identifiability(rs, measured_quantities = [:x3])
    @test sym_dict(id_report) == Dict(
        :x1 => :globally,
        :x2 => :globally,
        :x3 => :globally,
    )
    # Will probably be fixed in the 0.5 release of SI.jl
    @test_broken find_identifiable_functions(rs, measured_quantities = [:x3])
end
```

sumiya11:
A couple of papers with examples that can be used for testing:
- [https://www.pnas.org/doi/10.1073/pnas.1215508110](https://www.pnas.org/doi/10.1073/pnas.1215508110). There are also some examples in the supporting information
- https://link.springer.com/article/10.1007/s10910-007-9307-x

isaacsas:
@TorkelE does this work with Symbolics instead of symbols?

I'm concerned that across these PRs you are adding you are making the interface to rely on Symbols all over the place. This is likely to be a major break from ModelingToolkit, and could lead us to have to change all this code down the line.

@ChrisRackauckas can you comment on this? Both the PETab intergration and now the structural identifiability extension are written to use symbols instead of symbolics as the workflow. Is this going to cause issues / inconsistencies with MTK?

isaacsas:
Actually, I think the BifurcationKit extension also uses symbols instead of symbolics right?

test/extensions/structural_identifiability.jl
```diff
@@ -0,0 +1,172 @@
...
+    # Identifiability analysis for StructuralIdentifiability.jl model (declare this overwrites e.g. X2 variable etc.).
+    rs_si = @ODEmodel(
+        X1'(t) = p - k1*X1(t),
+        X2'(t) = k1*X1(t) - k2*X2(t),
+        X3'(t) = k2*X2(t) - 2k3*X3(t),
+        X4'(t) = -d*X4(t) + k3*X3(t),
+        X5'(t) = -d*X5(t) + k3*X3(t),
+        Yi'(t) = kD*Ya(t) - kA*Yi(t)*X3(t),
+        Ya'(t) = -kD*Ya(t) + kA*Yi(t)*X3(t),
+        y1(t) = X1 + Ya,
```
sumiya11, line 134: Should these yi be `X1 + Yi`, `Ya`, `k1`, and `kD` ?

TorkelE:
@sumiya11 Thanks a lot for the test cases and input, I will inccoperate this.

@isaacsas Both this, the BifurcationKit PR, and the rest, can use both symbols and symbolic, and the tests covers both cases.

sumiya11:
> Handling conservation laws to speed up computation. Hopefully some additional functionality will come of in StructuralIdentfiability to simplify this. What to do about this depend on that.

Which functionality in SI would you need?

Note also, it is tricky to use conservation laws together with `find_identifiable_functions` (since if you eliminate x1 by x1 = C - x2, you would need to substitute x1 back after the analysis is done and probably simplify the expressions once again).

So, one feasible thing would be to do this in StructuralIdentfiability instead of Catalyst. There is some ongoing work on this

TorkelE:
Solving this: https://github.com/SciML/StructuralIdentifiability.jl/issues/236 should make it relatively easy to enable conservation laws. There is also some discussion in the Issue on alternative ways to do it, but it will be less straightforward, and I'm still not really sure how.

Yes, handling it all in StructuralIdentfiability would be preferable, since it could cover a wide range of cases. The advantage right now in Catalyst is that we have algorithms implemented that finds these laws on reaction networks, which we then can apply, while SI would need to implement more general methods (but which then would benefit all systems). Long term that would be preferable, but it might also take more time to get everything into place.

TorkelE:
A question for everyone involved:
StructuralIdentifiabiltiy.jl currently return an output on the form of a dictionary taking components to identifiability. E.h.
```julia
:X => :globally
:p => :globally
:d => :globally
```
In Catalyst, the various components have a natural order (initial conditions followed by parameters, each with their own separate order). Currently, when identifability is printed, this can appear quite jumbled, e.g.:
```julia
:d => :globally
:X => :globally
:p => :globally
```
(mixing parameters and initial conditions etc.)

I think it would be nicer to impose order, so that when the user prints the entire identifiability dictionary, they are ordered initial conditions followed by parameters.

This would require the making the output an `OrderedDict`. This while the output of SI functions are simply `Dict`s. It seems undesirable that e.g. `assess_identifiaility` has a different output type when applied to a Catalyst `ReactionSystem` (as opposed to a SI type ODE). How bad do we think this is? Previously I decided not to make the output an `OrderedDict`, but there are advantages. Do anyone have any input?

sumiya11:
> I think it would be nicer to impose order, so that when the user prints the entire identifiability dictionary, they are ordered initial conditions followed by parameters.

I agree, some order can be helpful, especially if there are a lot of quantities. I opened an issue:
https://github.com/SciML/StructuralIdentifiability.jl/issues/244

pogudingleb:
> A question for everyone involved: StructuralIdentifiabiltiy.jl currently return an output on the form of a dictionary taking components to identifiability. E.h.
> 
> ```julia
> :X => :globally
> :p => :globally
> :d => :globally
> ```
> 
> In Catalyst, the various components have a natural order (initial conditions followed by parameters, each with their own separate order). Currently, when identifability is printed, this can appear quite jumbled, e.g.:
> 
> ```julia
> :d => :globally
> :X => :globally
> :p => :globally
> ```
> 
> (mixing parameters and initial conditions etc.)
> 
> I think it would be nicer to impose order, so that when the user prints the entire identifiability dictionary, they are ordered initial conditions followed by parameters.
> 
> This would require the making the output an `OrderedDict`. This while the output of SI functions are simply `Dict`s. It seems undesirable that e.g. `assess_identifiaility` has a different output type when applied to a Catalyst `ReactionSystem` (as opposed to a SI type ODE). How bad do we think this is? Previously I decided not to make the output an `OrderedDict`, but there are advantages. Do anyone have any input?

This is fixed in the current dev version of `StructuralIdentifiability`: if `funcs_to_check` are not provided, states are listed first, otherwise, the results are listed in the same order as in `funcs_to_check`.

TorkelE:
Thanks! This will make things much neater

TorkelE:
now rebased on master

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
```diff
@@ -0,0 +1,144 @@
...
+    osys, conseqs, _ = make_osys(rs; remove_conserved)
+    measured_quantities = make_measured_quantities(rs, measured_quantities, known_p, conseqs; ignore_no_measured_warn)
+    return SI.preprocess_ode(osys, measured_quantities)[1]
+end
+
+### Structural Identifiability Wrappers ###
+
+# Creates dispatch for SI's local identifiability analysis function.
+function SI.assess_local_identifiability(rs::ReactionSystem, args...; measured_quantities = Num[], 
+                                         known_p = Num[], funcs_to_check = Vector(), remove_conserved = true,
```
isaacsas, line 39: Why do you sometimes set the kwargs to be `Num[]` and sometimes `[]` by default across these various functions?

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
```diff
@@ -0,0 +1,144 @@
...
+    # Gives these to SI to create an SI ode model of its preferred form.
+    osys, conseqs, _ = make_osys(rs; remove_conserved)
+    measured_quantities = make_measured_quantities(rs, measured_quantities, known_p, conseqs; ignore_no_measured_warn)
+    return SI.preprocess_ode(osys, measured_quantities)[1]
+end
+
+### Structural Identifiability Wrappers ###
+
+# Creates dispatch for SI's local identifiability analysis function.
+function SI.assess_local_identifiability(rs::ReactionSystem, args...; measured_quantities = Num[],
```
isaacsas, line 38: Shouldn't this and all the other `assess_*` functions get a doc string too, explaining `args` and `kwargs` are forwarded to SI and provide a way to pass additional SI-specific parameters?

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
```diff
@@ -0,0 +1,144 @@
...
+                                  conseqs; ignore_no_measured_warn=false) where {T,S}
+    # Warning if the user didn't give any measured quantities.
+    if ignore_no_measured_warn || isempty(measured_quantities) 
+        @warn "No measured quantity provided to the `measured_quantities` argument, any further identifiability analysis will likely fail. You can disable this warning by setting `ignore_no_measured_warn=true`."
+    end
+
+    # Appends the known parameters to the measured_quantities vector. Converts any Symbols to symbolics.
+    measured_quantities = [measured_quantities; known_p]
+    measured_quantities = [(q isa Symbol) ? Catalyst._symbol_to_var(rs, q) : q for q in measured_quantities]
+    measured_quantities = vector_subs(measured_quantities, conseqs)
```
isaacsas, line 115:
```suggestion
    mqiterator = Iterators.flatten((measured_quantities, known_p))
    mqs = [(q isa Symbol) ? Catalyst._symbol_to_var(rs, q) : q for q in mqiterator]
    mqs = vector_subs(measured_quantities, conseqs)
```
This avoids potentially repeatedly changing the type of `measured_quantities`, and I think is more readable.

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
```diff
@@ -0,0 +1,144 @@
...
+
+    # Creates one internal observation variable for each measured quantity (`___internal_observables`).
+    # Creates a vector of equations, setting each measured quantity equal to one observation variable.
+    @variables t (___internal_observables(t))[1:length(measured_quantities)]
```
isaacsas, line 119: Shouldn't this use the system's `iv` instead of `t`? i.e. `get_iv(rs)`?

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
```diff
@@ -0,0 +1,144 @@
...
+    return vector_subs(funcs_to_check, conseqs)
+end
+
+# Processes the outputs to a better form.
+# Replaces conservation law equations back in the output (so that e.g. Γ are not displayed).
+# Sorts the output according to their input order (defaults to the `[states; parameters]` order).
+function make_output(out, funcs_to_check, conseqs)
+    funcs_to_check = vector_subs(funcs_to_check, conseqs)
+    out = Dict(zip(vector_subs(keys(out), conseqs), values(out)))
+    out = sort(out; by = x -> findfirst(isequal(x, ftc) for ftc in funcs_to_check))
```
isaacsas, line 139:
```suggestion
    sortdict = Dict(ftc => i for (i,ftc) in enumerate(funcs_to_check))
    out = sort(out; by = x -> sortdict[x])
```

test/extensions/structural_identifiability.jl
```diff
@@ -0,0 +1,293 @@
...
+        y1(t) = M(t)
+    )
+    gi_3 = assess_identifiability(goodwind_oscillator_si)
+    li_3 = assess_local_identifiability(goodwind_oscillator_si)
+    ifs_3 = find_identifiable_functions(goodwind_oscillator_si)
+
+    # Check outputs.
+    @test sym_dict(gi_1) == sym_dict(gi_2) == sym_dict(gi_3)
+    @test sym_dict(li_1) == sym_dict(li_2) == sym_dict(li_3)
+    @test length(ifs_1) == length(ifs_2) == length(ifs_3)
```
isaacsas, line 57: Can we hardcode what the answers should be and test against that too?

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
```diff
@@ -0,0 +1,144 @@
...
+# Also compute the, later needed, conservation law equations and list of system symbols (states and parameters).
+function make_osys(rs::ReactionSystem; remove_conserved=true)
+    # Creates the ODESystem corresponding to the ReactionSystem (expanding functions and flattening it).
+    # Creates a list of the systems all symbols (states and parameters).
+    rs = Catalyst.expand_registered_functions(flatten(rs))
+    osys = convert(ODESystem, rs; remove_conserved)
+    vars = [states(rs); parameters(rs)]
+
+    # Computes equations for system conservation laws.
+    # These cannot be computed for hierarchical systems (and hence this is skipped).
```
isaacsas, line 90:
```suggestion
```
Didn't you fix this so it works via the flattening?

isaacsas (changes requested):

Project.toml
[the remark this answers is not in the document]
TorkelE: fixed internally

docs/Project.toml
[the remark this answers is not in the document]
TorkelE: fixed internally

HISTORY.md
```diff
@@ -1,6 +1,18 @@
...
 
 ## Catalyst unreleased (master branch)
+- Added CatalystStructuralIdentifiabilityExtension, which permits StructuralIdentifiability.jl function to be applied directly to Catalyst systems. E.g. use
+```julia
+goodwind_oscillator = @reaction_network begin
+    (mmr(P,pₘ,1), dₘ), 0 <--> M
+    (pₑ*M,dₑ), 0 <--> E
+    (pₚ*E,dₚ), 0 <--> P
+end
+assess_identifiability(goodwind_oscillator; measured_quantities=[:M])
```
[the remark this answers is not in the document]
TorkelE: good point, no it wont'

test/extensions/structural_identifiability.jl
```diff
@@ -0,0 +1,293 @@
...
+    ifs_3 = find_identifiable_functions(rs_si)
+        
+    # Check outputs.
+    @test sym_dict(gi_1) == sym_dict(gi_3)
+    @test sym_dict(li_1) == sym_dict(li_3)
+    @test length(ifs_1)-2 == length(ifs_2)-2 == length(ifs_3) # In the first case, the conservation law parameter is also identifiable.
+
+    # Checks output for the SI converted version of the catalyst model.
+    # For nested systems with conservation laws, conserved quantities like Γ[1], cannot be replaced back.
+    # Hence, here you display identifiability for `Γ[1]` instead of X2.
```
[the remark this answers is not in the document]
TorkelE: This is why I wanted to add metadata of conservation law parameters. Then I can parse the output expression, detect conservation law parameters, and replace them with their expression. But until we add that, I don't see a natural way to do it.

(in the other cases it works because the conservation laws always occur in the context of the replace parameters, e.g. `Γ[1]-X1`. I have a record of this expression that maps it to `X2`. But in some cases, e.g. when identifiable functions are found, you might just get `Γ[1]` as an identifiable function.

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
TorkelE: StructuralIdentifability have internal requirements that some stuff needs to be `Num`s

test/extensions/structural_identifiability.jl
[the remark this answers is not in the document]
isaacsas: Can't use make a list of the conservation law constants, and then use `get_variables` to compare against the list?

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
isaacsas: It will show both I believe, but yours will have 
```julia
    SI.assess_local_identifiability(rs::ReactionSystem, args...; ...)
```
as the function argument, so it is clear it is the method (i.e. version) that takes `ReactionSystem`s.

HISTORY.md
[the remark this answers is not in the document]
TorkelE: fixed

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
isaacsas: Try it on `convert` when you've got Catalyst loaded (`convert` is a base Julia function).

test/extensions/structural_identifiability.jl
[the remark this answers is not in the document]
TorkelE: I think there might be some way to hack it, but I'd rather just go down the metadata route, I think it would be more reliable, and also easier to implement.

test/extensions/structural_identifiability.jl
TorkelE: For all 3, or just the last one? I think the last one could technically be a bit stochastic what expression it produces.

For the others we could.

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
```diff
@@ -0,0 +1,144 @@
...
+- `measured_quantities=observed(rs)`: The quantities of the system we can measure. May either be equations (e.g. `x1 + x2`), or single species (e.g. `:x`). Defaults to the systems observables.
+- `known_p = Num[]`: List of parameters which values are known. 
+- `ignore_no_measured_warn=false`: If set to `true`, no warning is provided when the `measured_quantities` vector is empty. 
+
+Example:
+```julia
+rs = @reaction_network begin
+    (p,d), 0 <--> X
+end
+si_ode(rs; measured_quantities = [:X], known_p = [:p])
```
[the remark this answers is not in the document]
TorkelE: added

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
TorkelE: Got, thanks. Will add docstrings then.

Should these be added to an extensions API section as well?

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
TorkelE: fixed

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
TorkelE: fixed

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
TorkelE: added

test/extensions/structural_identifiability.jl
TorkelE: added

ext/CatalystStructuralIdentifiabilityExtension/structural_identifiability_extension.jl
TorkelE: we do some additional conversions down the kind now, she changed it all to `[]`

docs/Project.toml
```diff
@@ -7,6 +7,7 @@ Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 HomotopyContinuation = "f213a82b-91d6-5c5d-acf7-10f1c761b327"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
+Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
```
isaacsas, line 10: Did you mean to add this depedency? If so, it needs a compat too... Otherwise please remove.

TorkelE: I checked, but I think this is a standard library dependency, and we don't have compats for these (right?).

(it is required because SI uses it to turn off the print out messages, which spam the docs if we have them on)

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+
+Next, we can assess identifiability in the case where we can measure all three species concentrations:
+```example si1
+assess_identifiability(goodwind_oscillator; measured_quantities=[:M, :P, :E], loglevel=Logging.Error)
+```
+in which case all initial conditions and parameters become identifiable.
+
+
+### Indicating known parameters
+In the previous case we assumed that all parameters are unknown, however, this is not necessarily true. If there are parameters which value's are known, we can supply these using the `known_p` argument. Indeed, this might turn other, previously unidentifiable, parameters identifiable. Let us consider the previous example, where we measure the concentration of $M$ only, but also happen to know the production rate of $E$ ($pₑ$):
```
isaacsas, line 49:
```suggestion
In the previous case we assumed that all parameters are unknown, however, this is not necessarily true. If there are parameters with known values, we can supply these using the `known_p` argument. Providing this additional information might also make other, previously unidentifiable, parameters identifiable. Let us consider the previous example, where we measure the concentration of $M$ only, but now assume we also know the production rate of $E$ ($pₑ$):
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+```example si1
+assess_identifiability(gwo; measured_quantities=[:M], known_p=[:pₑ], loglevel=Logging.Error)
+```
+Not only does this turn the previously non-identifiable `pₑ` (globally) identifiable (which is obvious, given that its value is now known), but this additional information improve identifiability for several other network components.
```
isaacsas, line 53:
```suggestion
Not only does this turn the previously non-identifiable `pₑ` (globally) identifiable (which is obvious, given that its value is now known), but this additional information allows identifiability for several other network components.
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+
+To, in a similar manner, indicate that certain initial conditions are known is a work in progress. Hopefully the feature should be an available in the near future.
+
+### Providing non-trivial measured quantities
+Sometimes, we are not actually measuring species species, but rather some combinations of species (or possibly parameters). Here, any algebraic expression can be used in `measured_quantities`. If so, used species and parameters have to first be `@unpack`'ed from the model. Say that we have a model where an enzyme ($E$) is converted between an active and inactive form, which in turns activates the production of a product, $P$:
```
isaacsas, line 58:
```suggestion
Sometimes, we are not actually measuring species, but rather combinations of species (or possibly parameters), which can be captured by `measured_quantities`, which allows any algebraic expression to be provided. Species and parameters have to first be `@unpack`'ed from the model to be used in such expressions. Say that we have a model where an enzyme ($E$) is converted between an active and inactive form, which in turns activates the production of a product, $P$:
```
Why do species and parameters have to be `@unpack`'ed? You can't do `sys.symbol`?

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+```example si2
+using Catalyst, StructuralIdentifiability # hide
+enzyme_activation = @reaction_network begin
+    (kA,kD), Eᵢ <--> Eₐ
+    (Eₐ, d), 0 <-->P
+end
+```
+and we can measure the total amount of $E$ ($=$Eᵢ+Eₐ$), as well as the amount of $P$, we can use the following to assess identifiability:
```
isaacsas, line 66:
```suggestion
and we can measure the total amount of $E$ ($=Eᵢ+Eₐ$), as well as the amount of $P$, we can use the following to assess identifiability:
```

isaacsas (changes requested):
I don't have time to keep going through the tutorial tonight, but it seems to have a lot of typos / wording that could be reworked. Can you give it another careful pass? Have you built it locally and checked it looks ok?

docs/src/inverse_problems/structural_identifiability.md
TorkelE: Since we might go from:

unidentifiable -> locally identifiable
unidentifiable -> globally identifiable
locally identifiable -> globally identifiable

and it is actually somewhat nuanced (and not just identifiable or not), still sure "allows" is correct?
(I'm happy to have it, just checking whether you accounted for this)

docs/src/inverse_problems/structural_identifiability.md
TorkelE: Think I preferred the old start ratehr than "ometimes, we are not actually measuring species, but rather combinations of species (or possibly parameters), which can be captured by `measured_quantities`, which allows any algebraic expression to be provided". changing the other locally and pushing.

isaacsas: Feel free to reword from the old version or my version, but please fix that the old version could be read as two completely independent, disconnected sentences as originally written.

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+
+Structural identifiability can be divided into *local* and *global* identifiability. If a model quantity is locally identifiable, it means that its true value can be determined down to a finite-number of possible options. This also means that there is some limited region around its true value, where the true value is the only possible value. Globally identifiable quantities' values, on the other hand, can be uniquely determined. Again, while identifiability can be confirmed structurally for a model, it does not necessarily mean that they are practically identifiable for some given data.
```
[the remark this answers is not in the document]
TorkelE: New version

"Structural identifiability can be divided into *local* and *global* identifiability. If a model quantity is locally identifiable, it means that its true value can be determined down to a finite-number of possible options. This also means that there is some limited region around the quantity's true value where this true value is the only possible value (and hence, within this region, the quantity is fully identifiable)."

docs/src/inverse_problems/structural_identifiability.md
TorkelE: The code prints the identifiability for all components, so the reader can compare to the previous case (where `pₑ` is not known) and see how things changes. Do you think this is enough to make things make sense?

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+
+Structural identifiability can be divided into *local* and *global* identifiability. If a model quantity is locally identifiable, it means that its true value can be determined down to a finite-number of possible options. This also means that there is some limited region around the quantity's true value where this true value is the only possible value (and hence, within this region, the quantity is fully identifiable). Globally identifiable quantities' values, on the other hand, can be uniquely determined. Again, while identifiability can be confirmed structurally for a model, it does not necessarily mean that they are practically identifiable for some given data.
```
isaacsas, line 12:
```suggestion
Structural identifiability can be divided into *local* and *global* identifiability. If a model quantity is locally identifiable, it means that its true value can be determined down to a finite-number of possible options. This also means that there is some limited region around the quantity's true value where this true value is the only possible value (and hence, within this region, the quantity is fully identifiable). Globally identifiable quantities' values, on the other hand, can be uniquely determined. Again, while identifiability can be confirmed structurally for a quantity, it does not necessarily mean that it is practically identifiable for some given data.
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+- Full variable trajectories (e.g. $x(t)$).
+- Variable initial conditions (e.g. $x(0)$). 
+
+StructuralIdentifiability currently assesses identifiability for the first two only (however, if $x(t)$ is identifiable, then $x(0)$ will be as well).
+
+## Global identifiability analysis
+
+### Basic example
+
+Local identifiability can be assessed using the `assess_identifiability` function. For each model quantity (parameters and variables), it will assess whether they are:
```
isaacsas, line 25: I'm confused here. Do you mean "Global identifiability can be assessed"?

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+
+Structural identifiability (which is what this tutorial considers) can be illustrated by the following differential equation:
+${dx \over dt} = p1*p2*x(t)$
+where, however much data is collected on $x$, it is impossible to determine the distinct values of $p1$ and $p2$. Hence, these parameters are these are non-identifiable (however, their product, $p1*p2$, *is* identifiable).
+
+Catalyst contains a special extension for carrying out structural identifiability analysis using the [StructuralIdentifiability.jl](https://github.com/SciML/StructuralIdentifiability.jl) package. This enables StructuralIdentifiability's `assess_identifiability`, `assess_local_identifiability`, and `find_identifiable_functions` functions to be called directly on Catalyst `ReactionSystem`s. It also implements specialised routines to make these more efficient when applied to reaction network models. How to use these functions is described in the following tutorial, with [StructuralIdentifiability providing a more extensive documentation](https://docs.sciml.ai/StructuralIdentifiability/stable/).
```
isaacsas, line 10:
```suggestion
Catalyst contains a special extension for carrying out structural identifiability analysis of generated reaction rate equation ODE models using the [StructuralIdentifiability.jl](https://github.com/SciML/StructuralIdentifiability.jl) package. This enables StructuralIdentifiability's `assess_identifiability`, `assess_local_identifiability`, and `find_identifiable_functions` functions to be called directly on Catalyst `ReactionSystem`s. This bypasses the need for users to convert `ReactionSystem`s to `ODESystem`s, provides a cleaner interface for calculating identifiability of reaction rate equation models, and implements specialised routines to speed up the identifiability calculations for the generated ODE models. How to use these functions is described in the following tutorial, with [StructuralIdentifiability providing a more extensive documentation](https://docs.sciml.ai/StructuralIdentifiability/stable/). 
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+```example si1
+using Catalyst, Logging, StructuralIdentifiability
+goodwind_oscillator = @reaction_network begin
+    (pₘ/(1+P), dₘ), 0 <--> M
+    (pₑ*M,dₑ), 0 <--> E
+    (pₚ*E,dₚ), 0 <--> P
+end
+assess_identifiability(goodwind_oscillator; measured_quantities=[:M], loglevel=Logging.Error)
+```
+From the output, we find that `E(t)`, `pₑ`, and `pₚ` (the trajectory of $E$, and the production rates of $E$ and $P$, respectively) are non-identifiable. Next, `dₑ` and `dₚ` (the degradation rates of $E$ and $P$, respectively) are locally identifiable. Finally, `P(t)`, `M(t)`, `pₘ`, and `dₘ` (the trajectories of `P` and `M`, and the production and degradation rate of `M`, respectively) are all globally identifiable. We note that we also imported the Logging.jl package, and provided the `loglevel=Logging.Error` input argument. StructuralIdentifiability functions generally provide a large number of output messages. Hence, we will use this argument (which requires the Logging package) throughout this tutorial to decrease the amount of printed text.
```
isaacsas, line 40:
```suggestion
From the output, in the associated reaction rate equation ODE model we find that `E(t)`, `pₑ`, and `pₚ` (the trajectory of $E$, and the production rates of $E$ and $P$, respectively) are non-identifiable. Next, `dₑ` and `dₚ` (the degradation rates of $E$ and $P$, respectively) are locally identifiable. Finally, `P(t)`, `M(t)`, `pₘ`, and `dₘ` (the trajectories of `P` and `M`, and the production and degradation rate of `M`, respectively) are all globally identifiable. We note that we also imported the Logging.jl package, and provided the `loglevel=Logging.Error` input argument. StructuralIdentifiability functions generally provide a large number of output messages. Hence, we will use this argument (which requires the Logging package) throughout this tutorial to decrease the amount of printed text.
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+```example si1
+assess_identifiability(gwo; measured_quantities=[:M], known_p=[:pₑ], loglevel=Logging.Error)
+```
+Not only does this turn the previously non-identifiable `pₑ` (globally) identifiable (which is obvious, given that its value is now known), but this additional information improve identifiability for several other network components.
+
+To, in a similar manner, indicate that certain initial conditions are known is a work in progress. Hopefully the feature should be an available in the near future.
+
+### Providing non-trivial measured quantities
+Sometimes, we are not actually measuring species species, but rather some combinations of species (or possibly parameters). To account for this, `measured_quantities` accepts any algebraic expressions (and not just single species). To form such expressions, species and parameters have to first be `@unpack`'ed from the model. Say that we have a model where an enzyme ($E$) is converted between an active and inactive form, which in turns activates the production of a product, $P$:
```
isaacsas, line 58:
```suggestion
Sometimes, ones may not have measurements of species, but rather some combinations of species (or possibly parameters). To account for this, `measured_quantities` accepts any algebraic expression (and not just single species). To form such expressions, species and parameters have to first be `@unpack`'ed from the model. Say that we have a model where an enzyme ($E$) is converted between an active and inactive form, which in turns activates the production of a product, $P$:
```
Do then need to be unpacked or can one do `sys.A` and such to access them? It might be easier to just suggest the latter if it works.

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+```example si2
+using Catalyst, StructuralIdentifiability # hide
+enzyme_activation = @reaction_network begin
+    (kA,kD), Eᵢ <--> Eₐ
+    (Eₐ, d), 0 <-->P
+end
+```
+and we can measure the total amount of $E$ ($=Eᵢ+Eₐ$), as well as the amount of $P$, we can use the following to assess identifiability:
```
isaacsas, line 66:
```suggestion
If we can measure the total amount of $E$ ($=Eᵢ+Eₐ$), as well as the amount of $P$, we can use the following to assess identifiability:
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+@unpack Eᵢ, Eₐ = enzyme_activation
+assess_identifiability(enzyme_activation; measured_quantities=[Eᵢ+Eₐ, :P], loglevel=Logging.Error)
+nothing # hide
+```
+
+### Assessing identifiability for specified quantities only
+By default, StructuralIdentifiability assesses identifiability for all parameters and variables. It is, however, possible to designate precisely which quantities you want to check using the `funcs_to_check` option. This both includes selecting a smaller subset of parameters and variables to check, or defining customised expressions. Let us consider the Goodwind from previously, and say that we would like to check whether the production parameters ($pₘ$, $pₑ$, and $pₚ$) and the total amount of the three species ($P + M + E$) are identifiable quantities. Here, we would first unpack these (allowing us to form algebraic expressions) and then use the following code:
+```example si1
+@unpack pₘ, pₑ, pₚ, M, E, P = goodwind_oscillator
+assess_identifiability(goodwind_oscillator; measured_quantities=[:M], funcs_to_check=[pₘ, pₑ, pₚ, M + E + P], loglevel=Logging.Error)
```
isaacsas, line 77: Perhaps `quantities_to_identify` or `quants_to_ident` would be clearer for users / more descriptive than `funcs_to_check`?

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+nothing # hide
+```
+
+### Probability of correctness
+The identifiability methods used can, in theory, produce erroneous results. However, it is possible to adjust the lower bound for the probability of correctness using the argument `p` (by default set to `0.99`, that is, at least a $99\%$ chance of correctness). We can e.g. increase the bound through:
+```example si2
+assess_identifiability(goodwind_oscillator; measured_quantities=[:M], p=0.999, loglevel=Logging.Error)
```
isaacsas, line 84:
```suggestion
assess_identifiability(goodwind_oscillator; measured_quantities=[:M], p=0.999, loglevel=Logging.Error)
```
Do we have to use `p`? This seems confusing given it usually means parameters in SciML. Maybe use `identifiability_probability` or `ident_prob`?

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+nothing # hide
+```
+giving a minimum bound of $99.9\%$ chance of correctness. In practise, the bounds used by StructuralIdentifiability are very conservative, which means that while the minimum guaranteed probability of correctness in the default case is $99\%$, in practise it is much higher. While increasing the value of `p` increases the certainty of correctness, it will also increase the time required to assess identifiability.
+
+## Local identifiability analysis
+Local identifiability can be assessed through the `assess_local_identifiability` function. While this is already determined by `assess_identifiability`, assessing local identifiability only have the advantage that it is easier to compute. Hence, there might be models where global identifiability analysis fails (or takes prohibitively long time), where instead `assess_local_identifiability` can be used. This functions takes the same inputs as `assess_identifiability` and returns, for each quantity, `true` if it is locally identifiable (or `false` if it is not). Here, for the Goodwind oscillator, we assesses it for local identifiability only:
```
isaacsas, line 90:
```suggestion
Local identifiability can be assessed through the `assess_local_identifiability` function. While this is already determined by `assess_identifiability`, assessing local identifiability only has the advantage that it is easier to compute. Hence, there might be models where global identifiability analysis fails (or takes a prohibitively long time), where instead `assess_local_identifiability` can be used. This function takes the same inputs as `assess_identifiability` and returns, for each quantity, `true` if it is locally identifiable (or `false` if it is not). Here, for the Goodwind oscillator, we assesses it for local identifiability only:
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+```example si1
+assess_local_identifiability(goodwind_oscillator; measured_quantities=[:M], loglevel=Logging.Error)
+```
+We note that the results are consistent with those produced by `assess_identifiability` (with globally or locally identifiable quantities here all being assessed as at least locally identifiable).
+
+## Finding identifiable functions
+Finally, StructuralIdentifiability provides the `find_identifiable_functions` function. Rather than determining the identifiability of each parameter and initial condition of the model, it finds a minimal set of identifiable functions, such as any other identifiable expression of the model can be generated by these. Let us again consider the Goodwind oscillator, using the `find_identifiable_functions` function we find that identifiability can be reduced to five globally identifiable expressions:
```
isaacsas, line 97: `it finds a minimal set of identifiable functions, such as any other identifiable expression of the model can be generated by these`

I don't understand what is being said here. The last part is confusing in its wording.

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+```example si1
+find_identifiable_functions(goodwind_oscillator; measured_quantities=[:M], loglevel=Logging.Error)
+```
+Again, these results are consistent with those produced by `assess_identifiability`. There, `pₑ` and `pₚ` where found to be globally identifiable. Here, they correspond directly to identifiable expressions. The remaining four parameters (`pₘ`, `dₘ`, `dₑ`, and `dₚ`) occurs as part of more complicated composite expressions.
```
isaacsas, line 101:
```suggestion
Again, these results are consistent with those produced by `assess_identifiability`. There, `pₑ` and `pₚ` where found to be globally identifiable. Here, they correspond directly to identifiable expressions. The remaining four parameters (`pₘ`, `dₘ`, `dₑ`, and `dₚ`) occur as part of more complicated composite expressions.
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+
+The `find_identifiable_functions` functions tries to simplify its output functions to create nice expressions. The degree to which it does this can be adjusted using the `simplify` keywords. Using the `:weak`, `:standard` (default), and `:strong` arguments, increased simplification can be forced (at the expense of longer runtimes).
```
isaacsas, line 103:
```suggestion
`find_identifiable_functions` tries to simplify its output functions to create nice expressions. The degree to which it does this can be adjusted using the `simplify` keywords. Using the `:weak`, `:standard` (default), and `:strong` arguments, increased simplification can be forced (at the expense of longer runtime).
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Creating StructuralIdentifiability compatible ODE models from Catalyst `ReactionSystem`s
+While the functionality described above covers the vast majority of analysis that user might want to perform, the StructuralIdentifiability package supports several additional features. While these does not have inherent Catalyst support, we do provide the `make_si_ode` function to simplify their use. Similarly to the previous functions, it takes a `ReactionSystem` and lists of measured quantities and known parameter values. The output is a [ode of the standard form supported by StructuralIdentifiability](https://docs.sciml.ai/StructuralIdentifiability/stable/tutorials/creating_ode/#Defining-the-model-using-@ODEmodel-macro). It can be created using the following syntax:
```
isaacsas, line 106:
```suggestion
While the functionality described above covers the vast majority of analysis that user might want to perform, the StructuralIdentifiability package supports several additional features. While these does not have inherent Catalyst support, we do provide the `make_si_ode` function to simplify their use. Similar to the previous functions, it takes a `ReactionSystem`, lists of measured quantities, and known parameter values. The output is a [ODE of the standard form supported by StructuralIdentifiability](https://docs.sciml.ai/StructuralIdentifiability/stable/tutorials/creating_ode/#Defining-the-model-using-@ODEmodel-macro). It can be created using the following syntax:
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+```example si1
+si_ode = make_si_ode(goodwind_oscillator; measured_quantities=[:M])
+nothing # hide
+```
+and then used as input to various StructuralIdentifiability functions. In the following example we uses StructuralIdentifiability's `print_for_DAISY` function, printing the model as an expression that can be used by the [DAISY](https://daisy.dei.unipd.it/) software for identifiability analysis[^3].
```
isaacsas, line 111:
```suggestion
and then used as input to various StructuralIdentifiability functions. In the following example we use StructuralIdentifiability's `print_for_DAISY` function, printing the model as an expression that can be used by the [DAISY](https://daisy.dei.unipd.it/) software for identifiability analysis[^3].
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Notes on systems with conservation laws
+Several reaction network models, such as
+```example si2
+using Catalyst, Logging, StructuralIdentifiability # hide
+rs = @reaction_network begin
+  (k1,k2), X1 <--> X2
+end
+```
+contain conservation laws (in this case $Γ = X1 + X2$, where $Γ = X1(0) + X2(0)$ is a constant). Because the presence of such conservation laws makes structural identifiability analysis prohibitively computationally expensive (for all but the simplest of cases), these are automatically eliminated by Catalyst (removing one ODE from the resulting ODE system for each conservation law). For the `assess_identifiability` and `assess_local_identifiability` functions, this will be unnoticed by the user. However, for the `find_identifiable_functions` and `make_si_ode` functions, this may result in one, or several, parameters on the form `Γ[i]` (where `i` is an integer) appearing in the produced expressions. These correspond to the conservation law constants and can be found through
```
isaacsas, line 125:
```suggestion
contain conservation laws (in this case $Γ = X1 + X2$, where $Γ = X1(0) + X2(0)$ is a constant). Because the presence of such conservation laws makes structural identifiability analysis prohibitively computationally expensive (for all but the simplest of cases), these are automatically eliminated by Catalyst (removing one ODE from the resulting ODE system for each conservation law). For the `assess_identifiability` and `assess_local_identifiability` functions, this will be unnoticed by the user. However, for the `find_identifiable_functions` and `make_si_ode` functions, this may result in one, or several, parameters of the form `Γ[i]` (where `i` is an integer) appearing in the produced expressions. These correspond to the conservation law constants and can be found through
```

docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Systems with exponent parameters
+Structural identifiability cannot currently be applied to systems with parameters (or species) in exponents. E.g. this
+```julia
+rn = @reaction_network begin
+    (hill(X,v,K,n),d), 0 <--> X
+end
+assess_identifiability(rn; measured_quantities=[:X])
+```
+is currently not possible. Hopefully this will be a supported feature in the future. For now, these expression will have to be rewritten to not include such exponents. For some cases, e.g. `10^k` this is trivial. However, it is also possible generally (but more involved and often includes introducing additional variables). If you have a system where this is required, it is recommended to contact an expert.
```
isaacsas, line 146:
```suggestion
is currently not possible. Hopefully this will be a supported feature in the future. For now, such expressions will have to be rewritten to not include such exponents. For some cases, e.g. `10^k` this is trivial. However, it is also possible generally (but more involved and often includes introducing additional variables). 
```
I'd just cut that last sentence. I don't think it adds anything helpful as it basically says find a generic other person to help you...

isaacsas (changes requested):
One high level comment; one can also consider SI for other model types (SDEs, jump processes). The current interface associates and conflates SI for a `ReactionSystem` with SI for the rate equation model, which is perhaps not a great idea if we ever want to expand to other system types? But maybe that isn't super relevant here since the SI library only works for ODEs and hence we would need a completely different code base?

For the tests I think it would make more sense to do a set equality test of key/value pairs, so that state/parameter orderings are not encoded in the tests. This seems fragile and a potential place tests could then break down the line. 

Finally, in the tutorial I left some comments on naming of keyword arguments. I'd be in favor of trying to come up with more descriptive names than `funcs_to_check` and `p` for the reasons mentioned there.

If you want to address my comments you are then welcome to merge -- I don't need to see it again.

docs/src/inverse_problems/structural_identifiability.md
isaacsas: It might be worth an issue there. I think they should consider deprecating that and adding a more descriptive keyword. Can't the `p` keyword be used to pass parameters in SciML problems even? If so it probably shouldn't be used in SI if they are moving to be more SciML compatible.

pogudingleb: We do use `p` but if you decide to use a more descriptive name, this sounds good to me. Normally, there is much need in changing this parameter anyway.

pogudingleb:
> One high level comment; one can also consider SI for other model types (SDEs, jump processes). The current interface associates and conflates SI for a `ReactionSystem` with SI for the rate equation model, which is perhaps not a great idea if we ever want to expand to other system types? But maybe that isn't super relevant here since the SI library only works for ODEs and hence we would need a completely different code base?

Just a quick comment here: right now we have ODEs and difference equations (that is, dynamical systems in discrete time). We would be interested in working on other types of models in future if there will be interest in this.

docs/src/inverse_problems/structural_identifiability.md
isaacsas: OK, maybe @TorkelE we can add a more descriptive keyword here and translate it to `p` for them internally, and then once SI gets updated with that keyword we can just switch to passing it through?

isaacsas:
> > One high level comment; one can also consider SI for other model types (SDEs, jump processes). The current interface associates and conflates SI for a `ReactionSystem` with SI for the rate equation model, which is perhaps not a great idea if we ever want to expand to other system types? But maybe that isn't super relevant here since the SI library only works for ODEs and hence we would need a completely different code base?
> 
> Just a quick comment here: right now we have ODEs and difference equations (that is, dynamical systems in discrete time). We would be interested in working on other types of models in future if there will be interest in this.

I'd be very interested in SI functionality for jump processes (i.e. stochastic chemical kinetics) models!

docs/src/inverse_problems/structural_identifiability.md
ChrisRackauckas: Maybe we should name it `pvalue`.

ChrisRackauckas:
Yes, I would be interested to have it easy to recreate https://www.biorxiv.org/content/10.1101/2020.08.10.245233v1.full. @sumiya11 have you looked into this at all?

docs/src/inverse_problems/structural_identifiability.md
ChrisRackauckas: I think in stats everyone calls it a `pvalue` though.

ChrisRackauckas: well I guess it's not quite a pvalue? It's more of a `prob_threshold`?

isaacsas: Yeah, I realized that was what you meant by `pvalue` right after I suggested `probvalue`. I think it is probably ok then.

isaacsas: I personally like `prob_threshold` better.

pogudingleb:
> Yes, I would be interested to have it easy to recreate https://www.biorxiv.org/content/10.1101/2020.08.10.245233v1.full. @sumiya11 have you looked into this at all?

Thanks to pointing out the paper! I think I read it at some point (or another paper of the same authors). They write moment equations for the moments up to some order, and then treat them as ODEs. The question how to determine the right order seem to be open but one can definitely do what they do in the paper (that is, take first or second order) by, for example, [MomentClosure.jl](https://github.com/augustinas1/MomentClosure.jl). 
Is there the moment closure functionality in `Catalyst` ?

ChrisRackauckas:
MomentClosure.jl uses Catalytst.jl, so in theory it could all work.

docs/src/inverse_problems/structural_identifiability.md
TorkelE: I will do prob_threshold . People won't really use it anyway, so a long keyword is not a problem.

docs/src/inverse_problems/structural_identifiability.md
TorkelE: yes, fixed

docs/src/inverse_problems/structural_identifiability.md
TorkelE: agreed, fixed

docs/src/inverse_problems/structural_identifiability.md
TorkelE: Agreed, but again, this is what they sue in StructuralIdentifiability. I can use something else (like for `p`), but figured it would make sense to use what they use.

test/extensions/structural_identifiability.jl
```diff
@@ -0,0 +1,311 @@
...
+    # Check outputs.
+    @test sym_dict(gi_1) == sym_dict(gi_2) == sym_dict(gi_3)
+    @test sym_dict(li_1) == sym_dict(li_2) == sym_dict(li_3)
+    @test length(ifs_1[2:end]) == length(ifs_2) == length(ifs_3) # In the first case, the conservation law parameter is also identifiable.
+
+    # Checks output to manually checked correct answers. 
+    @test isequal(collect(keys(gi_1)),[states(rs_catalyst); parameters(rs_catalyst)])
+    @test isequal(collect(values(gi_1)),[:globally, :locally, :locally, :nonidentifiable, :nonidentifiable, :globally, :globally, :globally, :globally, :locally, :locally, :nonidentifiable, :locally, :globally])
+    @test isequal(collect(keys(li_1)),[states(rs_catalyst); parameters(rs_catalyst)])
+    @test isequal(collect(values(li_1)),[1, 1, 1, 0, 0, 1, 1, 1, 1, 1, 1, 0, 1, 1])
```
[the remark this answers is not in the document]
TorkelE: fixed

TorkelE:
This should be ready now (but must wait for the fix to symbolic currently failing the Catalyst tests)

pogudingleb:
@TorkelE 
Just a quick heads up on the upcoming changes in `StructuralIdentifiability`. The next release (5.2) will:
* use name `prob_threshold` instead of `p` thus addressing [this issue](https://github.com/SciML/StructuralIdentifiability.jl/issues/256)
* have ModelingToolkit as an extension. Since you use it anyway, this shouldn't in principle make a difference
* this extension will have compute >=5.16 with `Symbolics` and [thus](https://github.com/JuliaSymbolics/Symbolics.jl/commit/eacd116c195ffc65e9f54d33ade4d8488f790791) require Julia 1.10

Hope none of this will be a source of problems.

TorkelE:
@pogudingleb Sounds good, thanks for the note

PR #712, opened 2023-11-05
WIP: add discrete_events to DSL

tstrey:
Adds `@discrete_events` option to the DSL such that reaction_system can be can be written in the following way:

```
rn = @reaction_network rxs begin
     @variables t
     @parameters k
     @species A(t) B(t)
     @discrete_events begin
          (t == 2.0) => [B ~ 0.0]
     end
     k, A --> B
end
```


src/reaction_network.jl
```diff
@@ -123,7 +123,11 @@ const forbidden_variables_error = let
 end
 
 # Declares the keys used for various options.
+<<<<<<< Updated upstream
 const option_keys = (:species, :parameters, :variables, :ivs)
+=======
+const option_keys = (:species, :parameters, :variables, :ivs, :discrete_events)
+>>>>>>> Stashed changes
```
isaacsas, line 131: This looks like something has gotten messed up when trying to merge master into your PR. There should be such comments listed.

isaacsas (commented):
If you think the macro is working you should add some tests that use it and make sure the correct system is generated (for example, generate a system with the macro and by hand symbolically and test they give similar results when simulated).

tstrey:
This works for systems like 

```
rn = @reaction_network a begin
              @parameters k
              @variables V(t)
              @species B(t) C(t)
              @discrete_events begin
                      a == 0 => [a ~ a+1, b ~ b-1]
               end
               k, C --> B
        end
```

tstrey:
Needs to be modified for multiple discrete_events.

isaacsas:
@TorkelE ended up adding this as part of Catalyst 14, but thanks for your getting this started!

PR #718, opened 2023-11-10
Chemistry functionality remake

TorkelE:
Changes:
- Add docstrings to chemistry-related functions.
- Adds documentation for chemistry-related functionality.
- Simplifies internal code for chemistry-related functionality.
- Redo `@compound` syntax:
```julia
@variables t
@species C(t) O(t)
@compound CO2(t) ~ C + 2O 
```
- Enables `@compound` to see variables, e.g.
```julia
@variables t
@species C(t) O(t)
o = O
@compound CO2 ~ C + 2o
```
- Add the `compounds` macro, enabling the creation of multiple compounds simultaneously:
```julia
@variables t
@species C(t) O(t)
@compounds begin
    O2 ~ 2O
    CO2 ~ C + 2O
end
```
- Adds the `compounds` option to the DSL, enabling the declaration of compounds within it (the components must be designated as a species, and are not ifered).


TorkelE:
Ok, this shoudl be ready now. I will have a look tomorrow with fresh eyes though to be sure there are no errors.

One feature that it might make sense to add is a `balance_reaction_network` function, which applied `balance_reaction` on each reaction of a `ReactionSystem`, and returns a new `ReactionSystem` with balanced reactions.

Finally, I don't think setting metadata is supported for compounds, which probably should be the case, e.g.
```julia
@compound H2_2(t) [metadata=value] = 2H
@compounds begin
    O2  [metadata=value]= 2O 
end
```

TorkelE:
Ok, this should be ready now.

isaacsas:
Yes, we should support metadata for compounds. 

Should we instead infer the arguments of the compound from the species that make it up?

isaacsas:
I.e. if the species depend on time `t` the compound does, but if they depend on `s` then it does?

TorkelE:
That's a good point. Having a compound with a different independent variable than its compositions would be weird, right? If this is indeed the case I will change it so that the compound takes its components iv (and throw an error if they have different ones.

Supporting metadata should be easy, e.g.
```julia
@compounds begin
    O2  [metadata=value] = 2O 
end
```
But if we want to also give a default value, how would we phrase that?
```julia
@compounds begin
    O2  [metadata=value] = 2O =1.0
end
```
?

isaacsas:
Or the compound could be the union of the `ivs` of the components I guess (for mixed spatial problems).

isaacsas:
Yeah, we should allow default values. Maybe we should use something else to indicate the formula in terms of species? 

```julia
@compounds begin
    O2  [metadata=value] = 1.0 ~ 2O
end
```
But I’m open to other suggestions.

TorkelE:
I will have a look tomorrow at what is actually possible notations using metaprogramming. Leaning towards keeping the current notation, and then add something if default values are required.

isaacsas:
The one reason for `~` is then it looks like an MTK equation.

TorkelE:
With default values being stored in the metadata, would it be possible to define it there? Doesn't seem to work perfectly though, e.g. 
```julia
@variables t X(t) [defaultvalue=1.0]
```
and other alternatives I've tried fail. Stuff like this:
```
@compound CO2 = 0.0 = C + 2O 
```
would be possible.

I think that, for now, I will leave setting default values unsupported (unless we can make it work along the line like `[defaultvalue=1.0]`, in which case there is a workaround. The feature seems incredibly niche, and given that we cannot know whether anyone will actually use compounds a year from now (not to talk about setting default values for them), I don't want to sink too much time into it. If I see an opportunity to do it easily while fixing the metadata stuff I might have a go at it though.

isaacsas:
OK, if it isn't easy to do maybe we should hold off on making an update then. I'd prefer not to potentially break the interface twice (i.e. once with this PR, and then potentially again when we fix it to support metadata and initial values).

TorkelE:
Updated how `@compound` (and similar) have its syntax:
```julia
@compound H2O_1 ~ 2H + O
@compound H2O_2, [output=true] ~ 2H + O
@compound (H2O_3 = 1.5) ~ 2H + O
@compound (H2O_4 = 4, [output=true]) ~ 2H + O
@compound (H2O_5 = p1, [output=true]) ~ 2H + p2*O
```

Currently, I have not been able to figure out how to extract the independent variable from the rhs, so just set it to `t`. Once I've figured that out this should be ready.

TorkelE:
Sorry, I've been juggling 10 branches back and forth and I finally messed up. I accidentally rebased this on https://github.com/SciML/Catalyst.jl/pull/713 (I meant to do that for the structural identifiability PR, which needs to be on that PR to be able to handle mm and hill functions). Talked to Avik and there doesn't seem to be a good solution more than waiting with merging this until https://github.com/SciML/Catalyst.jl/pull/713 is merged.

TorkelE:
Updates:
- Improved internal formatting further.
- Change syntax to `@compound (H2O_2, [output=true]) ~ 2H + O` (parenthesis added).
- Permits adding ivs in the compound.
- If the components have more than one iv combined, ivs must be provided in the compound.

src/chemistry_functionality.jl
```diff
@@ -6,93 +7,195 @@ Symbolics.option_to_metadata_type(::Val{:iscompound}) = CompoundSpecies
...
+"""
+    @compound
+
+Macro that creates a compound species, which is composed of smaller component species.
+
+Example:
+```julia
+@variables t
+@species C(t) O(t)
+@compound CO2(t) = C + 2O
```
isaacsas, line 21:
```suggestion
@compound CO2(t) ~ C + 2O
```

src/chemistry_functionality.jl
```diff
@@ -6,93 +7,195 @@ Symbolics.option_to_metadata_type(::Val{:iscompound}) = CompoundSpecies
...
+- The component species must be defined before using the `@compound` macro.
+"""
+macro compound(expr)
+    make_compound(MacroTools.striplines(expr))
+end
+
+# Declares compound error messages:
+const COMPOUND_CREATION_ERROR_BASE = "Malformed input to @compound. Should use form like e.g. \"@compound CO2 ~ C + 2O\"."
+const COMPOUND_CREATION_ERROR_BAD_SEPARATOR = "Malformed input to @compound. Left-hand side (the compound) and the right-hand side (the components) should be separated by a \"~\" (e.g. \"@compound CO2 ~ C + 2O\"). If the left hand side contains metadata and/or default values, this should be enclosed by \"()\" (e.g. \"@compound (CO2 = 1.0, [output=true]) ~ C + 2O\")."
+const COMPOUND_CREATION_ERROR_DEPENDENT_VAR_REQUIRED = "When the components (collectively) have more than 1 independent variable, independent variables have to be specified for the compound (e.g. `@compound CO2(t,s) ~ C + 2O`). This is required since the @compound macro cannot infer the correct order of the independent variables."
```
isaacsas, line 34: Is this still true or do you infer it as the union? In the tutorial you say you shouldn't give arguments to the compound I think.

src/chemistry_functionality.jl
```diff
@@ -6,93 +7,195 @@ Symbolics.option_to_metadata_type(::Val{:iscompound}) = CompoundSpecies
...
-    escaped_setcomponents_expr = esc(setcomponents_expr)
+Example:
+```julia
+@variables t
+@species C(t) H(t) O(t) 
+@compounds
+    CH4(t) = C + 4H
+    O2(t) = 2O
+    CO2(t) = C + 2O
+    H2O(t) = 2H + O
```
isaacsas, line 110: `~`

src/chemistry_functionality.jl
```diff
@@ -171,6 +275,52 @@ function get_balanced_stoich(reaction::Reaction)
...
+Returns a vector of all possible stoichiometrically balanced `Reaction` objects for the given `Reaction`.
+
+Example:
+```julia
+@variables t
+@species Si(t) Cl(t) H(t) O(t)
+@compound SiCl4(t) = Si + 4Cl
+@compound H2O(t) = 2H + O
+@compound H4SiO4(t) = 4H + Si + 4O
+@compound HCl(t) = H + Cl
```
isaacsas, line 290: `~`

src/chemistry_functionality.jl
```diff
@@ -171,6 +275,52 @@ function get_balanced_stoich(reaction::Reaction)
...
+```
+
+```julia
+@variables t
+@species C(t) H(t) O(t)
+@compound CO(t) = C + O
+@compound CO2(t) = C + 2O
+@compound H2(t) = 2H
+@compound CH4(t) = C + 4H
+@compound H2O(t) = 2H + O
```
isaacsas, line 302: `~`

src/chemistry_functionality.jl
```diff
@@ -171,6 +275,52 @@ function get_balanced_stoich(reaction::Reaction)
...
+```
+
+```julia
+@variables t
+@species Fe(t) S(t) O(t) H(t) N(t)
+@compound FeS2(t) = Fe + 2S
+@compound HNO3(t) = H + N + 3O
+@compound Fe2S3O12(t) = 2Fe + 3S + 12O
+@compound NO(t) = N + O
+@compound H2SO4(t) = 2H + S + 4O
```
isaacsas, line 314: `~`

isaacsas (changes requested):
Does balancing with reactions involving compounds of compounds work / is it tested? 

@TorkelE I'm fine with merging once you've addressed these comments -- I don't need to review again. If compounds of compounds don't work in balancing that should be mentioned in the tutorial and docstring though (or made to work).

TorkelE:
Compound of compounds does not work. Thought it was mentioned, but will add that in (and in docstring as well).

isaacsas:
Can we add an error message in `balance_reaction` if we detect one of the species is a compound of compounds? That should be straightforward I'd hope.

isaacsas:
(Unless I just missed that we already give an error there.)

TorkelE:
Checked, no there is not such error. I will add that.

isaacsas:
OK, thanks! Once you've made your updates feel free to merge if tests pass. (They might not until we get the binomial issue in Symbolics fixed.)

docs/src/catalyst_functionality/chemistry_related_functionality.md
```diff
@@ -0,0 +1,129 @@
...
+    @species C(t) H(t) O(t)
+    @compounds begin
+        C2O ~ C + 2O
+        H2O ~ 2H + O
+        H2CO3 ~ CO2 + H2O
+    end
+    (k1,k2), H2O+ CO2 <--> H2CO3
+end
+```
+When creating compound species using the DSL, it is important to note that *every component must be known to the system as a species, either by being declared using the `@species` option, or by appearing in a reaction*. E.g. the following is not valid
```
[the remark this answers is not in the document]
TorkelE: Yes they can, will rephrase to make this clear.

docs/src/catalyst_functionality/chemistry_related_functionality.md
```diff
@@ -0,0 +1,129 @@
...
+    H2O ~ 2H + O
+end
+unbalanced_reaction = @reaction k, $NH3 + $O2 --> $NO + $H2O
+```
+We can now create a balanced version (where the amount of H, N, and O is the same on both sides):
+```@example chem2
+balanced_reaction = balance_reaction(unbalanced_reaction)[1]
+```
+
+Reactions declared as a part of a `ReactionSystem` (e.g. using the DSL) can be retrieved for balancing using the `reaction` function. Please note that balancing these will not mutate the `ReactionSystem`, but a new reaction system will need to be created using the balanced reactions.
```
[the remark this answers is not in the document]
TorkelE: agreed. I will raise an issue (otherwise I'm bound to forget it).

src/chemistry_functionality.jl
TorkelE: fixed

src/chemistry_functionality.jl
TorkelE: fixed

src/chemistry_functionality.jl
TorkelE: fixed

src/chemistry_functionality.jl
TorkelE: fixed

src/chemistry_functionality.jl
TorkelE: Arguments are required, we couldn't do the union because we wouldn't be able to infer the correct order of the ivs. Will double read through the tutorial to ensure this is written there.

docs/src/catalyst_functionality/chemistry_related_functionality.md
```diff
@@ -0,0 +1,129 @@
+# [Chemistry-related functionality](@id chemistry_functionality)
+
+While Catalyst has primarily been designed around the modelling of biological systems, reaction network models are also common across chemistry. This section describes two types of functionality, that while of general interest, should be especially useful in the modelling of chemical systems.
+- The `@compound` option, which enables the user to designate that a specific species is composed of certain subspecies.
+- The `balance_reaction` function, which enables the user to balance a reaction so the same number of components occur on both sides.
```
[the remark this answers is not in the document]
TorkelE: yes

src/chemistry_functionality.jl
TorkelE: added an extra bit about this.

TorkelE:
This is done now. Will give it a day or two to see what happens with the binomial issue before merging though (in case there is some problem with that).

isaacsas:
We need to wait till we can run tests to completion before merging anything.

TorkelE:
yes

PR #719, opened 2023-11-12
Tutorial for using optimization to optimize for other stuff than parameter fitting

TorkelE:
The advantage with Optimization.jl is that you can optimise all kinds of stuff, and not just do parameter fitting. To point out that this is possible, and describe some of the additional considerations that this entails (which are not relevant in https://github.com/SciML/Catalyst.jl/pull/708), I've created a separate tutorial for this. The main structure is there, but I will need to go through it once or twice more to ensure that the writing is all good.


TorkelE:
This one is read, however https://github.com/SciML/Catalyst.jl/pull/708 shoudl be merged first so that I can ensure that the links from this one to that one works (I didn't want to build this on that branch, to avoid too complicating PR structures when it will be a while until some things might merge, I will just rebase this on master before he merging)

PR #720, opened 2023-11-12
Tutorial on how to find system steady states usings NonlinearSolve.jl

TorkelE:
Adds a new documentation page, demonstrating how to create `NonlinearProblem`s from Catalyst `Reactionnetwork`s, and how to solve these to find steady states.


docs/src/catalyst_applications/nonlinear_solve.md
```diff
@@ -0,0 +1,99 @@
+# [Finding Steady States using NonlinearSolve.jl](@id nonlinear_solve)
+
+We have previously described how `ReactionSystem` steady states can be found through [homotopy continuation](@ref homotopy_continuation). Catalyst also supports the creation of `NonlinearProblems` corresponding to `ODEProblems` with all left-hand side derivatives set to 0. These can be solved using the variety of nonlinear equation-solving algorithms implemented by [NonlinearSolve.jl](https://github.com/SciML/NonlinearSolve.jl), with the solutions corresponding to system steady states. Generally, this approach is preferred when one only wishes to find *a single solution* (with homotopy continuation being preferred when one wishes to find *all solutions*).
```
isaacsas, line 3: Might be worth mentioning Nonlinearsolve is also the only approach when the system does not involve rational(?) functions in the species for rate laws (or is it just mass action and the Catalyst specific rate functions we have like mm that work with homotopy continuation?).

docs/src/catalyst_applications/nonlinear_solve.md
```diff
@@ -0,0 +1,99 @@
...
+\end{aligned}
+```
+To find its steady states we need to solve:
+```math
+\begin{aligned}
+0 &= pₘ - d \cdot mRNA \\
+0 &= pₚ \cdot mRNA - k₁ \cdot P + 2k₂ \cdot P₂ - d \cdot P \\
+0 &= k₁ \cdot P + 2k₂ \cdot P₂ \\
+\end{aligned}
+```
```
isaacsas, line 33: Why not use Latexify to generate these automatically?

docs/src/catalyst_applications/nonlinear_solve.md
```diff
@@ -0,0 +1,99 @@
...
+Finally, we can solve it using the `solve` command, returning the steady state solution:
+```@example nonlinear_solve1
+using NonlinearSolve
+solve(nl_prob)
+```
+
+NonlinearSolve provides [a wide range of potential solvers](https://docs.sciml.ai/NonlinearSolve/stable/solvers/NonlinearSystemSolvers/). If we wish to designate one, it can be supplied as a second argument to `solve`. Here, we use the Newton Trust Region method:
+```@example nonlinear_solve1
+solve(nl_prob, TrustRegion())
```
isaacsas, line 49: Maybe show the solutions the two methods get agree via `isapprox`?

docs/src/catalyst_applications/nonlinear_solve.md
```diff
@@ -0,0 +1,99 @@
...
+The `NonlinearProblem`s generated by Catalyst corresponds to ODEs. A common method of solving these is to simulate the ODE from an initial condition until a steady state is reached. NonlinearSolve supports this through the `DynamicSS` method. To use it, an appropriate ODE solver (and any options you wish it to use) must also be supplied (with [a large number being available](https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/)). Here, we will use the `Tsit5` ODE solver to find the steady states of our dimerisation system.
+```@example nonlinear_solve1
+using OrdinaryDiffEq, SteadyStateDiffEq
+solve(nprob, DynamicSS(Tsit5()))
+```
+Here, we had to import [OrdinaryDiffEq.jl](https://github.com/SciML/OrdinaryDiffEq.jl) (to use `Tsit5`) and [SteadyStateDiffEq.jl](https://github.com/SciML/SteadyStateDiffEq.jl) (to use `DynamicSS`).
+
+Like previously, the found steady state is unaffected by the initial `u` guess. However, when [finding the steady states of systems with conservation laws](@ref nonlinear_solve_conservation_laws) it is important to select a `u` guess corresponding to the initial condition of the ODE simulation which steady state is sought.
```
isaacsas, line 61:
```suggestion
Like previously, the found steady state is unaffected by the initial `u` guess. However, when [finding the steady states of systems with conservation laws](@ref nonlinear_solve_conservation_laws) it is important to select a `u` guess that corresponds to the initial condition for the ODE model for which a steady state is sought.
```

docs/src/catalyst_applications/nonlinear_solve.md
```diff
@@ -0,0 +1,99 @@
...
+
+## [Systems with conservation laws](@id nonlinear_solve_conservation_laws)
+As described in the section on homotopy continuation, when finding the steady states of systems with conservation laws, [additional considerations have to be taken](homotopy_continuation_conservation_laws). E.g. consider the following two-state system:
+```@example nonlinear_solve2
+using Catalyst, NonlinearSolve # hide
+two_state_model = @reaction_network begin
+    (k1,k2), X1 <--> X2
+end
+```
+It has an infinite number of steady states. To make steady state finding possible, information of the system's conserved quantities (here $C=X1+X2$) must be provided. Since these can be computed from system initial conditions (`u0`, i.e. those provided when performing ODE simulations), designating an `u0` is often the best way. This can either be done by performing [ODE simulation-based steady state finding](@ref), using the initial condition as the initial `u` guess. Alternatively, any conservation quantities can be eliminated when the `NonlinearProblem` is created. This feature is supported by Catalyst's [conservation law finding and elimination feature](@ref network_analysis_deficiency).
```
isaacsas, line 72:
```
This can either be done by performing [ODE simulation-based steady state finding](@ref), using the initial condition as the initial `u` guess. Alternatively, any conservation quantities can be eliminated when the `NonlinearProblem` is created.
```
is a bit awkwardly/confusingly worded. I would reword.

docs/src/catalyst_applications/nonlinear_solve.md
```diff
@@ -0,0 +1,99 @@
...
+
+To eliminate conservation laws we simply provide the `remove_conserved = true` argument to `NonlinearProblem`:
+```@example nonlinear_solve2
+p = [:k1 => 2.0, :k2 => 3.0]
+u_guess = [:X1 => 3.0, :X2 => 1.0]
+nl_prob = NonlinearProblem(two_state_model, u_guess, p; remove_conserved = true)
+nothing # hide
+```
+here it is important that the quantities used in `u_guess` correspond to the conserved quantities we wish to use. E.g. here the conserved quantity $X1 + X2= 3.0 + 1.0 = 4$ holds in the input, and will hence do in the output as well. We can now find the steady states using `solve` like before:
```
isaacsas, line 81:
```suggestion
here it is important that the quantities used in `u_guess` correspond to the conserved quantities we wish to use. E.g. here the conserved quantity $X1 + X2= 3.0 + 1.0 = 4$ holds for the initial condition, and will hence also hold in the calculated steady-state as well. We can now find the steady states using `solve` like before:
```

docs/src/catalyst_applications/nonlinear_solve.md
```diff
@@ -0,0 +1,99 @@
...
+```@example nonlinear_solve2
+sol = solve(nl_prob)
+```
+We note that the output only provides a single value. The reason is that the actual system solved only contains a single equation (the other being eliminated with the conserved quantity). To find the values of $X1$ and $X2$ we can [inference the solution object using these as input](@ref simulation_structure_interfacing_solutions):
```
isaacsas, line 85:
```
inference the solution object using these as input
```
is confusing to me. Can you reword?

docs/src/catalyst_applications/nonlinear_solve.md
TorkelE: Very good point.

docs/src/catalyst_applications/nonlinear_solve.md
TorkelE: Latexify has a tendency to jumble up equations a bit, making things come in weird order. Maybe it is better now, will check.

docs/src/catalyst_applications/nonlinear_solve.md
TorkelE: Makes sense, will add.

docs/src/catalyst_applications/nonlinear_solve.md
TorkelE: 👍

docs/src/catalyst_applications/nonlinear_solve.md
TorkelE: yes, will do.

TorkelE:
Ready for merging

PR #721, opened 2023-11-12
Add tests for NonlinearSolve and NonlinearProblem

TorkelE:
Currently, we do not really test the conversions to `NonlinearSystem`s and `NonlinearProblem`s and their solutions. We do some tests on `SteadyStateProblem`s, but these are kind of deprecated anyway.

This PR adds some tests to this. The tests should be good, but need to check whether the tolerances are not good enough that the outputs are close enough to 0. If not, I will have to make stuff more lenient.

### DIFF
--- a/test/miscellaneous_tests/nonlinear_solve.jl
+++ b/test/miscellaneous_tests/nonlinear_solve.jl
@@ -25,7 +25,7 @@ let
     nl_prob = NonlinearProblem(steady_state_network_1, u0, p)
     
     # Solves it using standard algorithm and simulation based algorithm.
-    sol1 = solve(nl_prob, LevenbergMarquardt(); abstol=1e-12, reltol=1e-12).u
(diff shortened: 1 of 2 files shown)


isaacsas:
`1e-18` is smaller than machine epsilon, so seems too stringent for tests to pass. Checking to a relative accuracy of `1e-8` for solutions that are order 1 in size would be fine I think.

TorkelE:
Sounds like a better way, will do that and reduce tolerance to something sensible.

TorkelE:
I had to disable the tests for 1.6 (still unclear why NonlinearSolve fails there, but it is already a problem we are having and I think that package might have stopped supporting 1.6 anyway).

Otherwise, this is ready and good to go.

ChrisRackauckas:
> I had to disable the tests for 1.6 (still unclear why NonlinearSolve fails there, but it is already a problem we are having and I think that package might have stopped supporting 1.6 anyway).

Yes it already had to drop v1.6 because of certain compiler issues.

PR #732, opened 2023-11-21
Update CI strategy

TorkelE:
Add entry:
```
    strategy:
      fail-fast: false
```
to CI file.

After consultation with Avik. Currently, if v1.6 CI fails, the v1 CI is stopped. Especially now when we get random coverall errors on v1.6 (which Avik don't know how to fix more than restarting) this seems good.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,6 +10,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         group:
           - Core


isaacsas:
Let's just move to 1.9 as the base supported version. It appears that MTK and OrdinaryDiffEq did this already anyways.

TorkelE:
skipping 1.6 also sounds good, but having this might make sense anyway?

TorkelE:
Also, steady state solving fails, not sure if it related to https://github.com/SciML/Catalyst.jl/pull/728

isaacsas:
No, I'd prefer we keep this setting consistent with the rest of SciML. If CI is failing we should just stop at that point as we do now rather then continue to use up Github CI resources.

isaacsas:
But updating the 1.6 version to 1.9 will hopefully fix the issues. (And if not we should figure out what is going on and fix them.)

TorkelE:
> No, I'd prefer we keep this setting consistent with the rest of SciML. If CI is failing we should just stop at that point as we do now rather then continue to use up Github CI resources.

Sounds good. Avik argued against it and I cannot really give ay further input.

PR #735, opened 2023-12-01
Enable observables in the DSL

TorkelE:
Enables observables to be created as a part of the DSL. Example:
```julia
using Catalyst 
rn = @reaction_network begin
  @observables begin
    X ~ Xi + Xa
  end
  (p,d), 0 <--> Xi
  (k1,k2), Xi <--> Xa
end
```

if we simulate the system, the observables can now be used for indexing:
```julia
using OrdinaryDiffEq
@unpack Xi, Xa, X, p, d, k1, k2 = rn
u0 = [Xi => 0.0, Xa => 0.0]
ps = [p => 1.0, d => 0.2, k1 => 1.0, k2 => 2.0]
oprob = ODEProblem(rn, u0, (0.0, 10.0), ps)
sol = solve(oprob, Tsit5())
sol[X]
```
```julia
using Plots
plot(sol; idxs=X)
```

Some additional notes:
- If only a single observable is used, the `begin ... end ` block is not required:
```rn = @reaction_network begin
  @observables X ~ Xi + Xa
  (p,d), 0 <--> Xi
  (k1,k2), Xi <--> Xa
end
```
- The right-hand side can have any algebraic expression. However, all stuff there must be declared elsewhere (and it cannot contain other observables).
- Currently, the observables do not have any independent variables. We might want to change this part (although it will require a decent amount of coding).
- Metadata for observables is not supported either. This should probably be added I think, but not sure.

Tests and docs are here, so in principle this is ready
(unless we want to support independent variables in this PR)

Consideration regarding independent variables for observables: If the thing it consists of has several different ones, the iv for the observable should be the union of all those occurring in its formula, right?


TorkelE:
(This PR is built upon https://github.com/SciML/Catalyst.jl/pull/718)

isaacsas:
@TorkelE Can we close this PR since we are using the other one now? It is just causing extraneous CI to be run...

PR #736, opened 2023-12-02
Enable equations in the DSL

TorkelE:
Adds the `@equations` option to the DSL, enabling equations to be added. E.g. to create a model of a cell with a volume (`V`) which grows with a growth factor (`G`), where growth depletes some nutrition (`N`) we can use:
```julia
using Catalyst, OrdinaryDiffEq, Plots

rn = @reaction_network begin
  @parameters k
  @equations begin
    D(V) ~ k*G
    D(N) ~ -k*G
  end
  (p,d), 0 <--> G
end

u0 = [:G => 0.0, :V => 0.1, :N => 10.0]
ps = [:p => 1.0, :d => 0.5, :k => 0.2]
oprob = ODEProblem(rn, u0, (0.0,10.0), ps)
sol = solve(oprob, Tsit5())
plot(sol)
```
image

### Additional details
- If an expression `D(X)` (where `X` is any symbol) occurs in an equation, a differential `D = Differential($(DEFAULT_IV_SYM))` is automatically created and used. If this is used when `D` is used as a species, variable, or parameter name, an error is thrown.
- If the left-hand side is `D(X)` only, `X` is automatically inferred to be a variable (and created as such). Any other symbol occurring in an equation must be declared elsewhere.
- If only a single equation is declared, the `begin ... end` block can be omitted (e.g. `@equations  D(V) ~ k*G`).

### Further considerations
- Currently, DAEs can be created by simply writing equations. However, this causes MTK to throw errors later on. Until we have figured out how to deal with this, this is not supported. An alternative is to force `ODAEProblem`s to be created in these cases, and do some additional stuff there. I have asked about this on Slack.
- I have added an option `@differentials` if the user wants to create custom differentials. However, `ODEProblem`s currently do not support differentials on more than one iv. Hence, the only real use of this option is to create a different name for the default differential (if e.g. one has a species called `D`).
- Documentation can be made more extensive. I will probably wait with that until I rewrite the documentation for the DSL (splitting it into an advanced part), and we have figured out what to do about DAEs.
- I have currently not considered SDEs and Jump systems and how these are affected.


TorkelE:
(this pr depends on https://github.com/SciML/Catalyst.jl/pull/735)

TorkelE:
Currently
- `equations(rn)` returns all equations and reactions.
- `reactions(rn)` returns all reactions.

We are missing something for returning reactions only. Intuitively, that should be `equations` (with something else returning both). However, to keep with MTK standard that should probably (?) remain as it is. So we need a good name for this function. Possibly, we could also add getters for pure differential equations (and algebraic equations) as well.

TorkelE:
(errors are due to being based on a faulty version of https://github.com/SciML/Catalyst.jl/pull/735), rebasing will fix it)

TorkelE:
I think this PR has gotten messed up due to all of the interlocking branch dependencies (depending on two different things in addition to master), not sure how to fix.

TorkelE:
Git history not sustainable anymore, reuploaded as https://github.com/SciML/Catalyst.jl/pull/801

PR #737, opened 2023-12-02
[v14 ready] RDME and graph SSAs

TorkelE:
To avoid rebasing hell this is simply a new version of https://github.com/SciML/Catalyst.jl/pull/663. You might have some input on simpler ways to compute the (non-spatial) mass action Jumps @isaacsas, but otherwise, this is good to go. Currently vertex parameters must be uniform across the grid (and this is enforced). Once @Vilin97 have his PRs merged over at JumpProcesses, this requirement can be lightened. 

I have added support for spatial jumps. Creating and solving a problem is relatively easy:
```julia
using Catalyst, Graphs, JumpProcesses

# Make model.
SIR_system = @reaction_network begin
    α, S + I --> 2I
    β, I --> R
end
trS = @transport_reaction dS S
trI = @transport_reaction dI I
lattice = Graphs.grid([2, 2])
lrs = LatticeReactionSystem(SIR_system, [trS, trI], lattice)

# Parameter values and initial conditions.
u0 = [:S => [999, 1000, 1000, 1000], :I => [1, 0, 0, 0], :R => 0]
pV = [:α => 0.1 / 1000, :β => 0.01]
pE = [:dS => [0.1, 0.1, 0.1, 0.2], :dI => [0.01, 0.01, 0.01, 0.02]]

# Create the problem.
dprob = DiscreteProblem(lrs, u0, (0.0,100.0), (pV,pE))
jprob = JumpProblem(lrs, dprob, NSM())
```

Current limitations:
- Cartesian grids are not supported, only Graphs.
- Of the 5 forms of the hopping rates, we now convert to the most general one (D_{s,i,j}). Once this is merged, I will introduce a routine which detects which is the most compact form that can be provided, and uses that one. To keep this PR short, I am currently holding of on uploading those changes,
- Currently, the constant rate jumps in each compartment must have the same parameter values. Once this is supported in JumpProcesses I will add support for it here as well.

### DIFF
--- a/test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
+++ b/test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
@@ -7,7 +7,8 @@ using Random, Statistics, SparseArrays, Test
 # Fetch test networks.
 include("../spatial_test_networks.jl")
 
-### Correctness Tests ###
+
+### General Tests ###
 
 # Tests that there are no errors during runs for a variety of input forms.
 let
@@ -42,6 +43,7 @@ let
         end
     end
 end
+
 
 ### Input Handling Tests ###
 
@@ -101,6 +103,7 @@ let
         end
     end
 end
+
 
 ### Tests taken from JumpProcesses ###
 
--- a/test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
+++ b/test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
@@ -1,0 +1,157 @@
+### Preparations ###
+
+# Fetch packages.
+using JumpProcesses
+using Random, Statistics, SparseArrays, Test
+
+# Fetch test networks.
+include("../spatial_test_networks.jl")
+
+### Correctness Tests ###
+
+# Tests that there are no errors during runs for a variety of input forms.
+let
+    for grid in [small_2d_grid, short_path, small_directed_cycle]
+        for srs in [Vector{TransportReaction}(), SIR_srs_1, SIR_srs_2]
+            lrs = LatticeReactionSystem(SIR_system, srs, grid)
+            u0_1 = [:S => 999, :I => 1, :R => 0]
+            u0_2 = [:S => round.(Int64, 500.0 .+ 500.0 * rand_v_vals(lrs.lattice)), :I => 1, :R => 0, ]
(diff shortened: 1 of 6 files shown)


TorkelE:
This one shoudl also have better formatting and comments as compared to the previous PR.

isaacsas:
I thought spatially varying mass action jump rates are already supported? That is what a `SpatialMassActionJump` encodes.

TorkelE:
> I thought spatially varying mass action jump rates are already supported? That is what a `SpatialMassActionJump` encodes.

Ahh, didn't know that that was supported already.

test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
```diff
@@ -0,0 +1,160 @@
+### Preparations ###
+
+# Fetch packages.
+using JumpProcesses
+using Random, Statistics, SparseArrays, Test
+
+# Fetch test networks.
+include("../spatial_test_networks.jl")
+
+### Correctness Tests ###
```
Vilin97, line 10: There is no correctness being checked. Only the retcode of the solution.

Vilin97: We should definitely test correctness. You can take the ABC test from JumpProcesses or the any other spatial test.

Vilin97 (dismissed):
1. Coveralls report that all the new lines added are not covered. I don't understand why that is. Is the test accidentally calling some other functions instead of the newly added ones?
2. We should add a correctness test.

test/spatial_reaction_systems/lattice_reaction_systems_jumps.jl
TorkelE: Maybe I should rename that header. We do perform more tests than just checking the output. E.g. the ABC model from JumpProcesses have been converted to Catalyst and is tested later in the file.

TorkelE:
I have renamed the "Correctness Tests" header. Good point of the coverage, the file wasn't being run but added now, thanks!

isaacsas:
Won’t that run CI on the branch for all PRs and merges forever now?

TorkelE:
It will until we remove it, which we will do with the v14 release. And all PRs until then will be for the v14 branch anyway.

This is what we want right, and it is not like there is another way?

PR #730, opened 2023-11-18
Stability computation

TorkelE:
No docs and tests, and internal needs to be improved to handle vector of values (not maps) input prettier, but functionality works. Enables computation of stability o steady states. Conservation laws should be accounted for (not tested yet).

Simple example:
```julia
using Catalyst
import HomotopyContinuation

rs = @reaction_network begin
    (v/10.0 + hill(X, v, K, n), d), 0 <--> X
end
p = [:v => 2.0, :K => 1.0, :n => 3, :d => 1.0]
steady_states = hc_steady_states(rs, p)

steady_state_stability(steady_states[1], rs, p) # true
steady_state_stability(steady_states[2], rs, p) # false
steady_state_stability(steady_states[3], rs, p) # true

steady_state_stability(steady_states, rs, p) # [1, 0, 1]

ss_jac = steady_state_jac(rs) # [1, 0, 1]
steady_state_stability(steady_states, rs, p; ss_jac=ss_jac) # [1, 0, 1]
```

PR #739, opened 2023-12-04
[v14 ready] New stability computation

TorkelE:
New version of: https://github.com/SciML/Catalyst.jl/pull/730 (rebasing was a mess)

Enables computation of stability of steady states. Conservation laws should be accounted for.

Simple example:
```julia
using Catalyst
import HomotopyContinuation

rs = @reaction_network begin
    (v/10.0 + hill(X, v, K, n), d), 0 <--> X
end
p = [:v => 2.0, :K => 1.0, :n => 3, :d => 1.0]
steady_states = hc_steady_states(rs, p)

steady_state_stability(steady_states[1], rs, p) # true
steady_state_stability(steady_states[2], rs, p) # false
steady_state_stability(steady_states[3], rs, p) # true

steady_state_stability(steady_states, rs, p) # [1, 0, 1]

ss_jac = steady_state_jac(rs) # [1, 0, 1]
steady_state_stability(steady_states, rs, p; ss_jac=ss_jac) # [1, 0, 1]
```

Also add a function `is_autnomous` that checks whether a system is autonomous. Add this as an error check when creating `BifurcationProblem`s, `NonlinearSystem`s, and computing steady states using `hc_steady_states`.


src/steady_state_stability.jl
```diff
@@ -0,0 +1,103 @@
...
+        u = [u_dict[var] for var in states(rs)]        
+    end
+    if (p isa Vector{<:Pair}) || (p isa Dict)
+        p_dict = Dict(symmap_to_varmap(rs, p))
+        p = [p_dict[var] for var in parameters(rs)]
+    end
+
+    # Computes stability (by checking that the real part of all eigenvalues are negative).
+    jac = ss_jac(u, p, Inf)
+    return maximum(real.(eigvals(jac))) < 0
```
isaacsas, line 52: This needs research on what algorithms people actually use to calculate stability. I’m doubtful this is a good approach (for example, due to the difficulty in calculating all eigenvalues for large matrices accurately). I wouldn’t be in favor of adding a method that can result in misleading results due to numerical errors.

isaacsas: Have you looked at any literature on methods for calculating stability numerically?

TorkelE: Yeah, you mentioned that last time as well. It is a good point, I have looked around but not found anything (although simple searches typically just give university lecture intro type stuff).

I am asking around in the JuliaLab and will try to find some experts. Realistically, if it is involved, it might take a while until we can implement. In that case I'm still for merging this with appropriate warnings (as it is still a useful feature that e.g. I would have use of). But the point is good, and I haven't actually thought of it.

TorkelE:
Remove autonomy check for `NonlinearSystem`s for now, however, it is probably something we will have to think about at some point (at least the user will get an error quite soon right now anyway).

test/reactionsystem_core/reactionsystem.jl
```diff
@@ -118,13 +118,11 @@ let
     odesys = complete(convert(ODESystem, rs))
     sdesys = complete(convert(SDESystem, rs))
     js = complete(convert(JumpSystem, rs))
-    nlsys = complete(convert(NonlinearSystem, rs))
```
TorkelE, line 121: For this and the next one. Here we test the conversion of a NonlinearSystem that is non-autonomous, which we no longer permit. Hence these tests were removed. I marked the second one as broken since I am not sure if there is some important test here or not.

test/miscellaneous_tests/api.jl
```diff
@@ -446,3 +446,56 @@ let
     neweqs = getfield.(equations(ns), :rhs)
     @test_throws MethodError Catalyst.to_multivariate_poly(neweqs)
 end
+
```
isaacsas, line 449: Add a test with a registered function that takes `t` as a parameter within the rate to make sure that is caught as non-autonomous?

test/reactionsystem_core/reactionsystem.jl
```diff
@@ -158,15 +156,17 @@ let
...
+    # `ReactionSystem`s to be converted to `NonlinearSystem`s.
+    @test_broken false
+    # ns = convert(NonlinearSystem, rs)
+    # nlprob = NonlinearProblem(rs, u, p)
+    # fnl = eval(generate_function(ns)[2])
+    # dunl = similar(du)
+    # @test_broken let # The next line throws an error.
+    #     fnl(dunl, nlprob.u0, nlprob.p)
+    #     @test norm(du - dunl) < 100 * eps()
+    # end
```
isaacsas, line 169: I guess just delete then.

test/runtests.jl
```diff
@@ -33,6 +33,7 @@ using SafeTestsets, Test
     #if GROUP == "All" || GROUP == "Miscellaneous-NetworkAnalysis"
         # Tests various miscellaneous features.
         @time @safetestset "API" begin include("miscellaneous_tests/api.jl") end
+        @time @safetestset "Steady State Stability Computations" begin include("miscellaneous_tests/stability_computation.jl") end
         @time @safetestset "Compound Species" begin include("miscellaneous_tests/compound_macro.jl") end
         @time @safetestset "Reaction Balancing" begin include("miscellaneous_tests/reaction_balancing.jl") end
         @time @safetestset "Units" begin include("miscellaneous_tests/units.jl") end
```
isaacsas, line 39: Just in terms of core functionality I would put `Units` after `API`, and probably put this stability calculation at the end or second to last. This and balancing seem the least "core".

src/reactionsystem.jl
```diff
@@ -1203,6 +1203,39 @@ function dependants(rx, network)
...
+    dep_var_param_eqs = [ModelingToolkit.get_variables(eq) for eq in filter(eq -> !(eq isa Reaction), equations(rs))]
+    if isempty(dep_var_param_rxs) && isempty(dep_var_param_eqs)
+        dep_var_param = []
+    else
+        dep_var_param = reduce(vcat,[dep_var_param_rxs; dep_var_param_eqs])
+    end
+
+    # Checks for iv and spatial ivs
+    any(isequal(get_iv(rs), var) for var in dep_var_param) && (return false)
+    any(isequal(siv, var) for siv in get_sivs(rs) for var in dep_var_param) && (return false)
```
isaacsas, line 1235:
```suggestion
function is_autonomous(rs::ReactionSystem)

    # Get all variables occurring in reactions and equations.
    vars = Set()
    for eq in equations(rs)
        (eq isa Reaction) ? get_variables!(vars, eq.rate) : get_variables!(vars, eq)
    end

    # Checks for iv and spatial ivs
    (get_iv(rs) in vars) && return false
    any(in(vars), get_sivs(rs)) && return false
```

isaacsas: Does that work?

docs/src/steady_state_functionality/steady_state_stability_computation.md
isaacsas, line 1: I haven't really read this carefully, but please at least show one example on changing the tolerance.

docs/src/steady_state_functionality/steady_state_stability_computation.md
```diff
@@ -0,0 +1,57 @@
+# Steady state stability computation
+After system steady states have been found using [HomotopyContinuation.jl](@ref homotopy_continuation), [NonlinearSolve.jl](@ref nonlinear_solve), or other means, their stability can be computed using Catalyst's `steady_state_stability` function. Systems with conservation laws will automatically have these removed, permitting stability computation on systems with singular Jacobian.
+
+!!! warn 
+    Catalyst currently computes steady state stabilities using the naive approach of checking whether a system's largest eigenvalue real part is negative. While more advanced stability computation methods exist (and would be a welcome addition to Catalyst), there is no direct plans to implement these. Furthermore, Catalyst uses a arbitrary tolerance $tol ≈ 1.5*10^-7$ to determine whether a computed eigenvalue is far away enough from 0 to be reliably used. This selected threshold, however, have not been subject to further analysis (and can be changed through the `tol` argument).
```
isaacsas, line 5:
```suggestion
    Catalyst currently computes steady state stabilities using the naive approach of checking whether a system's largest eigenvalue real part is negative. While more advanced stability computation methods exist (and would be a welcome addition to Catalyst), there is no direct plans to implement these. Furthermore, Catalyst uses a tolerance `tol = 10*sqrt(eps())` to determine whether a computed eigenvalue is far away enough from 0 to be reliably used. This threshold can be changed through the `tol` keyword argument.
```
I don't think it is arbitrary, it is motivated by the defaults for `isapprox` but a bit more conservative.

src/steady_state_stability.jl
```diff
@@ -0,0 +1,135 @@
...
+
+Notes:
+- The input `u` can (currently) be both a vector of values (like `[1.0, 2.0]`) and a vector map 
+(like `[X => 1.0, Y => 2.0]`). The reason is that most methods for computing stability only
+produces vectors of values. However, if possible, it is recommended to work with values on the
+form of maps.
+- Catalyst currently computes steady state stabilities using the naive approach of checking whether
+a system's largest eigenvalue real part is negative. While more advanced stability computation 
+methods exist (and would be a welcome addition to Catalyst), there is no direct plans to implement 
+these. Furthermore, Catalyst uses a arbitrary tolerance tol ~ 1.5*10^-7 to determine whether a
```
isaacsas, line 43:
```suggestion
these. Furthermore, Catalyst uses a tolerance `tol = 10*sqrt(eps())` to determine whether a 
```

src/steady_state_stability.jl
```diff
@@ -0,0 +1,135 @@
...
+- The input `u` can (currently) be both a vector of values (like `[1.0, 2.0]`) and a vector map 
+(like `[X => 1.0, Y => 2.0]`). The reason is that most methods for computing stability only
+produces vectors of values. However, if possible, it is recommended to work with values on the
+form of maps.
+- Catalyst currently computes steady state stabilities using the naive approach of checking whether
+a system's largest eigenvalue real part is negative. While more advanced stability computation 
+methods exist (and would be a welcome addition to Catalyst), there is no direct plans to implement 
+these. Furthermore, Catalyst uses a arbitrary tolerance tol ~ 1.5*10^-7 to determine whether a 
+computed eigenvalue is far away enough from 0 to be reliably used. This selected threshold, 
+however, have not been subject to further analysis (and can be changed through the `tol` argument).
```
isaacsas, line 45:
```suggestion
computed eigenvalue is far away enough from 0 to be reliably used. This selected threshold can be changed through the `tol` argument.
```

src/steady_state_stability.jl
```diff
@@ -0,0 +1,135 @@
...
+
+    # Generates the Jacobian at the steady state (technically, `ss_jac` is an `ODEProblem` with dummy values for `u0` and `p`).
+    J = zeros(length(u), length(u))
+    ss_jac = remake(ss_jac; u0 = u, p = ps)
+    ss_jac.f.jac(J, ss_jac.u0, ss_jac.p, Inf)
+    
+    # Computes stability (by checking that the real part of all eigenvalues is negative).
+    max_eig = maximum(real(ev) for ev in eigvals(J))
+    if abs(max_eig) < tol
+        error("The system Jacobian's maximum eigenvalue at the steady state is within the tolerance range (abs($max_eig) < $tol). Hence, stability could not be reliably determined. If you still wish to compute the stability, reduce the `tol` argument.")
```
isaacsas, line 70:
```suggestion
        error("The system Jacobian's maximum eigenvalue at the steady state is within the tolerance range (abs($max_eig) < $tol). Hence, stability could not be reliably determined. If you still wish to compute the stability, reduce the `tol` argument, but note that floating point error in the eigenvalue calculation could lead to incorrect results.")
```

isaacsas (changes requested):

src/Catalyst.jl
```diff
@@ -42,6 +42,7 @@ import Base: (==), hash, size, getindex, setindex, isless, Sort.defalg, length,
 import MacroTools, Graphs
 import Graphs: DiGraph, SimpleGraph, SimpleDiGraph, vertices, edges, add_vertices!, nv, ne
 import DataStructures: OrderedDict, OrderedSet
+import LinearAlgebra.eigvals
```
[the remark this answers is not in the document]
TorkelE: changed. Especially for a few of these niche ones, it might make sense to have some kind of comment on why they are needed/which feature uses them. I feel we are starting to accumulate quite a few dependencies, and it would probably make sense to keep track of which we have and why.

test/reactionsystem_core/reactionsystem.jl
TorkelE: happy to do, have deleted now

docs/src/steady_state_functionality/steady_state_stability_computation.md
TorkelE: sounds good

docs/src/api.md
```diff
@@ -166,6 +166,7 @@ speciesmap
 paramsmap
 reactionparamsmap
 isspecies
+is_autonomous
```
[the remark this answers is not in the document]
TorkelE: yes, will update to use this

test/miscellaneous_tests/api.jl
TorkelE: added

docs/src/steady_state_functionality/steady_state_stability_computation.md
TorkelE: added

PR #740, opened 2023-12-04
New SciML/Optimization for fitting parameters of ODEs tutorial tutorial

TorkelE:
A new version of this one: https://github.com/SciML/Catalyst.jl/pull/708

Adds a tutorial of how to use DiffEqParamEstim and Optimization for parameter fitting. Those packages are still not properly adapted to symbolic indexing, so normal vectors have to be used. However, I still think we go with that (I can add an additional note box at the beginning if we want to highlight this).

I still think this is better than self-coding this stuff (easy-to-miss stuff like what happens if a solver fails) so think this is useful (PEtab is still probably the best alternative). Also, hopefully, we can use this as a template workflow that we think should work without indexing, and something DiffEqParamEstim + Optimization should work towards fixing.


docs/pages.jl
```diff
@@ -16,8 +16,10 @@ pages = Any["Home" => "index.md",
                                            "catalyst_applications/homotopy_continuation.md",
                                            "catalyst_applications/nonlinear_solve.md",
                                            "catalyst_applications/bifurcation_diagrams.md"],
-            "Inverse Problems" => Any["inverse_problems/parameter_estimation.md",
-                                      "inverse_problems/petab_ode_param_fitting.md",
-                                      "inverse_problems/structural_identifiability.md"],
-            "FAQs" => "faqs.md",
```
isaacsas, line 22: You removed the FAQs?

docs/pages.jl
```diff
@@ -16,8 +16,10 @@ pages = Any["Home" => "index.md",
...
+                                        "inverse_problems/petab_ode_param_fitting.md",
```
isaacsas, line 19: Why is this here and below?

docs/src/inverse_problems/examples/ode_fitting_oscillation.md
```diff
@@ -1,24 +1,16 @@
-# [Parameter Estimation](@id parameter_estimation)
-The parameters of a model, generated by Catalyst, can be estimated using various
-packages available in the Julia ecosystem. Refer
-[here](https://docs.sciml.ai/Overview/stable/highlevels/inverse_problems/) for
-more extensive information. Below follows a quick tutorial of how
-[Optimization.jl](https://docs.sciml.ai/Optimization/stable/) can be used to fit
-a parameter set to data.
+# [Fitting Parameters for an Oscillatory System](@id parameter_estimation)
+In this examples we will use [Optimization.jl](https://github.com/SciML/Optimization.jl) to fit the parameters of an oscillatory system (the Brusselator) to data. Here, special consideration is taken to avoid reaching a local minimum. Instead of fitting the entire time series directly, we will start with fitting parameter values for the first period, and then use those as an initial guess for fitting the next (and then these to find the next one, and so on). Using this procedure is advantageous for oscillatory systems, and enables us to reach the global optimum.
```
isaacsas, line 2:
```suggestion
In this example we will use [Optimization.jl](https://github.com/SciML/Optimization.jl) to fit the parameters of an oscillatory system (the Brusselator) to data. Here, special consideration is taken to avoid reaching a local minimum. Instead of fitting the entire time series directly, we will start with fitting parameter values for the first period, and then use those as an initial guess for fitting the next (and then these to find the next one, and so on). Using this procedure is advantageous for oscillatory systems, and enables us to reach the global optimum.
```

docs/src/inverse_problems/examples/ode_fitting_oscillation.md
```diff
@@ -1,24 +1,16 @@
...
 using Catalyst
-using DifferentialEquations
-using SciMLSensitivity
+using OrdinaryDiffEq
 using Optimization
-
-# for the ADAM optimizer
-using OptimizationOptimisers
+using OptimizationOptimisers # Required for the ADAM optimizer.
+using SciMLSensitivity # Required for `Optimization.AutoZygote()` automatic differentiation option.
```
isaacsas, line 10:
```suggestion
using OptimizationOptimisers # for the ADAM optimizer
using SciMLSensitivity # for Optimization.AutoZygote() AD backend
```

docs/src/inverse_problems/examples/ode_fitting_oscillation.md
```diff
@@ -124,14 +117,14 @@ The final parameter estimate is then
...
 
-## Why we fit the parameters in iterations.
-The reason we chose to fit the model on a smaller interval to begin with, and
+## Why we fit the parameters in iterations
+As previously mentioned, the reason we chose to fit the model on a smaller interval to begin with, and
 then extend the interval, is to avoid getting stuck in a local minimum. Here
 specifically, we chose our initial interval to be smaller than a full cycle of
 the oscillation. If we had chosen to fit a parameter set on the full interval
-immediately we would have received an inferior solution.
+immediately we would have received the wrong solution.
```
isaacsas, line 127:
```suggestion
immediately we would have obtained a poorer fit and less accurate estimate for the parameters.
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
+# [Parameter Fitting for ODEs using SciML/Optimization.jl and DiffEqParamEstim.jl](@id optimization_parameter_fitting)
+Fitting parameters to data involves solving an optimisation problem (that is, finding the parameter set that optimally fits you model to your data, typically by minimising a cost function). The SciML ecosystem's primary package for solving optimisation problems is [Optimization.jl](https://github.com/SciML/Optimization.jl). It provides access to a variety of solvers from a single common interface, wrapping a large number of optimisation methods that have been implemented in Julia into this interface.
```
isaacsas, line 2:
```suggestion
Fitting parameters to data involves solving an optimisation problem (that is, finding the parameter set that optimally fits your model to your data, typically by minimising a cost function). The SciML ecosystem's primary package for solving optimisation problems is [Optimization.jl](https://github.com/SciML/Optimization.jl). It provides access to a variety of solvers via a single common interface by wrapping a large number of optimisation libraries that have been implemented in Julia.
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+
+This tutorial both demonstrate how to create parameter fitting cost functions using the [DiffEqParamEstim.jl](https://github.com/SciML/DiffEqParamEstim.jl) package, and how to use Optimization.jl to minimise these. Optimization.jl can also be used in other contexts, such as finding parameter sets that maximises the magnitude of some system behaviour. More details on how to use these packages can be found in their [respective](https://docs.sciml.ai/Optimization/stable/) [documentations](https://docs.sciml.ai/DiffEqParamEstim/stable/).
```
isaacsas, line 4:
```suggestion
This tutorial demonstrates both how to create parameter fitting cost functions using the [DiffEqParamEstim.jl](https://github.com/SciML/DiffEqParamEstim.jl) package, and how to use Optimization.jl to minimise these. Optimization.jl can also be used in other contexts, such as finding parameter sets that maximise the magnitude of some system behavior. More details on how to use these packages can be found in their [respective](https://docs.sciml.ai/Optimization/stable/) [documentations](https://docs.sciml.ai/DiffEqParamEstim/stable/).
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+data_ts = data_sol.t[2:end]
+data_vals = (0.8 .+ 0.4*rand(10)) .* data_sol[:P][2:end]
+
+# Plots the true solutions and the (synthetic data) measurements.
+using Plots
+plot(true_sol; idxs=:P, label="True solution", lw=8)
+plot!(data_ts, data_vals; label="Measurements", seriestype=:scatter, ms=6, color=:blue)
+```
+
+Next, we will use DiffEqParamEstim.jl to build a loss function that describing how well our model's simulations fit the data.
```
isaacsas, line 37:
```suggestion
Next, we will use DiffEqParamEstim.jl to build a loss function to measure how well our model's solutions fit the data.
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+```@example diffeq_param_estim_1 
+using DiffEqParamEstim, Optimization
+p_dummy = [:kB => 0.0, :kD => 0.0, :kP => 0.0]
+oprob = ODEProblem(rn, u0, (0.0, 10.0), p_dummy)
+loss_function = build_loss_objective(oprob, Tsit5(), L2Loss(data_ts, data_vals), Optimization.AutoForwardDiff(); maxiters=10000, verbose=false, save_idxs=4)
```
isaacsas, line 42:
```suggestion
loss_function = build_loss_objective(oprob, Tsit5(), L2Loss(data_ts, data_vals), Optimization.AutoForwardDiff(); 
    maxiters=10000, verbose=false, save_idxs=4)
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+nothing # hide
+```
+To `build_loss_objective` we provide the following arguments:
+- `oprob`: The `ODEProblem` with which we simulate our model (using some dummy parameter values, since we do not know these).
+- `Tsit5()`: The numeric integrator we wish to simulate our model with.
+- `L2Loss(data_ts, data_vals)`: Defines the loss function. While [other alternatives](https://docs.sciml.ai/DiffEqParamEstim/stable/getting_started/#Alternative-Cost-Functions-for-Increased-Robustness) are available, `L2Loss` is the simplest one (measuring the sum of squared distances between model simulations and data measurements). Its first argument is the time points at which the data is collected, and the second is the data's values.
+- `Optimization.AutoForwardDiff()`: Our choice of [automatic differentiation](https://en.wikipedia.org/wiki/Automatic_differentiation) framework.
+
+Furthermore, we can pass any number of additional optional arguments, these are then passed to the internal `solve()` function (that are used to simulate our ODE). Here we provide the following additional arguments:
```
isaacsas, line 51:
```suggestion
Furthermore, we can pass any number of additional optional arguments, these are then passed to the internal `solve()` function (which is used to solve our ODE). Here we provide the following additional arguments:
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+- `maxiters=10000`: If the ODE integrator takes a very large number of steps, that is typically the sign of a very poor fit. Reducing the `maxiters` threshold reduces the time we waste on evaluating such models.
```
isaacsas, line 52:
```suggestion
- `maxiters=10000`: If the ODE integrator takes a very large number of steps, that can be a sign of a very poor fit (or stiffness in the ODEs, but that is not a concern for our current example). Reducing the `maxiters` threshold reduces the time we waste on evaluating such models. 
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+- `Optimization.AutoForwardDiff()`: Our choice of [automatic differentiation](https://en.wikipedia.org/wiki/Automatic_differentiation) framework.
+
+Furthermore, we can pass any number of additional optional arguments, these are then passed to the internal `solve()` function (that are used to simulate our ODE). Here we provide the following additional arguments:
+- `maxiters=10000`: If the ODE integrator takes a very large number of steps, that is typically the sign of a very poor fit. Reducing the `maxiters` threshold reduces the time we waste on evaluating such models. 
+- `verbose=false`: The simulation of models with highly unsuitable parameter sets typically generate various warnings (such as simulation terminations due to reaching the `maxiters` value). To avoid an overflow of such (here unnecessary) warnings as we evaluate a large number of parameter sets, we turn warnings off.
```
isaacsas, line 53:
```suggestion
- `verbose=false`: The simulation of models with highly unsuitable parameter sets typically generate various warnings (such as premature simulation termination due to reaching `maxiters` timesteps). To avoid an overflow of such (here unnecessary) warnings, as we evaluate a large number of parameter sets, we turn warnings off.
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+- `save_idxs=4`: The measured species (*P*) is the 4th species in our species vector (`species(rn)`). To ensure that the concentration of the right species is evaluated against the data, we set the numeric integrator to only save the value of this species.
```
isaacsas, line 54:
```suggestion
- `save_idxs=4`: The measured species ($P$) is the 4th species in our species vector (`species(rn)`). Since we only assume data is available for $P(t)$ there is no reason to save any other species.
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+
+Now we can create an `OptimizationProblem` using our `loss_function` and some initial guess of parameter values from which the optimiser will start:
+```@example diffeq_param_estim_1 
+optprob = OptimizationProblem(loss_function, [1.0, 1.0, 1.0])
+nothing # hide
+```
+
+!!! note
+    `OptimizationProblem` cannot currently accept parameter values in the form of a map (e.g. `[:kB => 1.0, :kD => 1.0, :kP => 1.0]`). These must be provided as individual values (using the same order as the parameters occur in in the `parameters(rs)` vector). Similarly, `build_loss_objective`'s `save_idxs` uses the species index, rather than the species directly. These inconsistencies should be remedied in future package releases.
```
isaacsas, line 63:
```suggestion
    `OptimizationProblem` cannot currently accept parameter values in the form of a map (e.g. `[:kB => 1.0, :kD => 1.0, :kP => 1.0]`). These must be provided as individual values (using the same order as the parameters occur in in the `parameters(rs)` vector). Similarly, `build_loss_objective`'s `save_idxs` uses the species index, rather than the species directly. These inconsistencies should be remedied in future DiffEqParamEstim releases.
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+
+Finally, we can optimise `optprob` to find the parameter set that best fits our data. Optimization.jl only provide a few optimisation methods natively. However, for each supported optimisation package, it provides a corresponding wrapper-package to import that optimisation package for use with Optimization.jl. E.g., if we wish to use [Optim.jl](https://github.com/JuliaNLSolvers/Optim.jl)'s [Nelder-Mead](https://en.wikipedia.org/wiki/Nelder%E2%80%93Mead_method) method, we must install and import the OptimizationOptimJL package. A summary of all, by Optimization.jl, supported optimisation packages can be found [here](https://docs.sciml.ai/Optimization/stable/#Overview-of-the-Optimizers). Here, we import the Optim.jl package and uses it to minimise our cost function (thus finding a parameter set that fits the data):
```
isaacsas, line 65:
```suggestion
Finally, we can optimise `optprob` to find the parameter set that best fits our data. Optimization.jl only provides a few optimisation methods natively. However, for each supported optimisation package, it provides a corresponding wrapper-package to import that optimisation package for use with Optimization.jl. E.g., if we wish to use [Optim.jl](https://github.com/JuliaNLSolvers/Optim.jl)'s [Nelder-Mead](https://en.wikipedia.org/wiki/Nelder%E2%80%93Mead_method) method, we must install and import the OptimizationOptimJL package. A summary of all, by Optimization.jl, supported optimisation packages can be found [here](https://docs.sciml.ai/Optimization/stable/#Overview-of-the-Optimizers). Here, we import the Optim.jl package and uses it to minimise our cost function (thus finding a parameter set that fits the data):
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+```@example diffeq_param_estim_1 
+using OptimizationOptimJL
+optsol = solve(optprob, Optim.NelderMead())
```
isaacsas, line 68:
```suggestion
optsol = solve(optprob, Optim.NelderMead())
```
Using `NelderMead` is likely to get you criticized (though I've personally had problems where it seems to work more robustly than all the fancy first-derivative derivative based optimizers, or the methods often stated as better derivative-free alternatives to it).

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+
+We can now simulate our model for the corresponding parameter set, checking that it fits our data.
+```@example diffeq_param_estim_1 
+oprob_fitted = remake(oprob; p=optsol.u)
+fitted_sol = solve(oprob_fitted, Tsit5())
+plot!(fitted_sol; idxs=:P, label="Fitted solution", linestyle=:dash, lw=6, color=:lightblue)
+```
+
+!!! note
+    Here, a good exercise is to check the resulting parameter set and note that, while it creates a good fit to the data, it does not actually correspond to the original parameter set. [Identifiability](https://www.sciencedirect.com/science/article/pii/S1364815218307278) is a concept that studies how to deal with this problem.
```
isaacsas, line 80:
```suggestion
    Here, a good exercise is to check the resulting parameter set and note that, while it creates a good fit to the data, it does not actually correspond to the original parameter set. [Identifiability](https://www.sciencedirect.com/science/article/pii/S1364815218307278) is a concept that studies how to deal with this problem.
```
I will simply repeat my protest that introducing parameter estimation to beginners by fitting an example that is not identifiable is not a good approach, and identifiability should be introduced and explored separately. The first example in all cases should be one where fitting works, and followup examples can then show that it doesn't always work because of identifiability (among other reasons).

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+
+Say that we instead would like to use the [Broyden–Fletcher–Goldfarb–Shannon](https://en.wikipedia.org/wiki/Broyden%E2%80%93Fletcher%E2%80%93Goldfarb%E2%80%93Shanno_algorithm) algorithm, as implemented by the [NLopt.jl](https://github.com/JuliaOpt/NLopt.jl) package. In this case we would run:
+```@example diffeq_param_estim_1 
+using OptimizationNLopt
+sol = solve(optprob, NLopt.LD_LBFGS())
+nothing # hide
+```
+
+## Optimisation problems with data for multiple species
+Imagine that, in our previous example, we had measurements the concentration of both *S* and *P*:
```
isaacsas, line 90:
```suggestion
Imagine that, in our previous example, we had measurements of the concentration of both *S* and *P*:
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+```@example diffeq_param_estim_1 
+optprob_S_P = OptimizationProblem(loss_function_S_P, [1.0,1.0, 1.0])
+optsol_S_P = solve(optprob_S_P, Optim.NelderMead())
+oprob_fitted_S_P = remake(oprob; p=optsol_S_P.u)
+fitted_sol_S_P = solve(oprob_fitted_S_P, Tsit5())
+plot!(fitted_sol_S_P; idxs=[:S, :P], label="Fitted solution", linestyle=:dash, lw=6, color=[:lightblue :pink])
+```
+
+## Setting parameter constraints and boundaries
+Sometimes, it is desired to set boundaries on parameter values. Indeed, this can speed up the optimisation process (by preventing searching through unfeasible parts of parameter space) or be a prerequisite of some optimisation methods. This can be done by passing the `lb` (lower bounds) and `up` (upper bounds) arguments to `OptimizationProblem`. These are vectors (of the same length as the number of parameters), with each argument corresponding to the boundary value of the parameter with the same index (as used in the `parameters(rn)` vector). If we wish to constrain each parameter to the interval *(0.1, 10.0)* this can be done through:
```
isaacsas, line 117:
```suggestion
Sometimes, it is desired to set boundaries on parameter values. Indeed, this can speed up the optimisation process (by preventing searching through unfeasible parts of parameter space), and can also be a requirement for some optimisation methods. This can be done by passing the `lb` (lower bounds) and `up` (upper bounds) arguments to `OptimizationProblem`. These are vectors (of the same length as the number of parameters), with each argument corresponding to the boundary value of the parameter with the same index (as used in the `parameters(rn)` vector). If we wish to constrain each parameter to the interval $(0.1, 10.0)$ this can be done through:
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+```@example diffeq_param_estim_1 
+optprob = OptimizationProblem(loss_function, [1.0, 1.0, 1.0]; lb = [0.1, 0.1, 0.1], ub = [10.0, 10.0, 10.0])
+nothing # hide
+```
+
+In addition to boundaries, Optimization.jl also supports setting [linear and non-linear constraints](https://docs.sciml.ai/Optimization/stable/tutorials/constraints/#constraints) on its output solution.
```
isaacsas, line 123:
```suggestion
In addition to boundaries, Optimization.jl also supports setting [linear and non-linear constraints](https://docs.sciml.ai/Optimization/stable/tutorials/constraints/#constraints) on its output solution for some optimizers.
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Parameter fitting with known parameters
+If we from previous knowledge know that *kD = 0.1*, and only would like to fit the values of *kD* and *kP*, this can be achieved through `build_loss_objective`'s `prob_generator` argument. First, we create a function (`fixed_p_prob_generator`) that modifies our `ODEProblem` to incorporate this knowledge:
```
isaacsas, line 126:
```suggestion
If from previous knowledge we know that $kD = 0.1$, and only want to fit the values of $kD$ and $kP$, this can be achieved through `build_loss_objective`'s `prob_generator` argument. First, we create a function (`fixed_p_prob_generator`) that modifies our `ODEProblem` to incorporate this knowledge:
```
Why are we fitting `kD` if we know it?

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+```@example diffeq_param_estim_1 
+fixed_p_prob_generator(prob, p) = remake(prob; p = vcat(p[1], 0.1, p[2]))
+nothing # hide
+```
+Here, it takes the `ODEProblem` (`prob`) we simulates, and the parameter set used, during the optimisation process (`p`), and creates a modified `ODEProblem` (by setting a customised parameter vector [using `remake`](@ref simulation_structure_interfacing_remake)). Now we create our modified loss function:
```
isaacsas, line 131:
```suggestion
Here, it takes the `ODEProblem` (`prob`) we simulate, and the parameter set used, during the optimisation process (`p`), and creates a modified `ODEProblem` (by setting a customised parameter vector [using `remake`](@ref simulation_structure_interfacing_remake)). Now we create our modified loss function:
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+```@example diffeq_param_estim_1 
+loss_function_fixed_kD = build_loss_objective(oprob, Tsit5(), L2Loss(data_ts, data_vals), Optimization.AutoForwardDiff(); prob_generator = fixed_p_prob_generator, maxiters=10000, verbose=false, save_idxs=4)
+nothing # hide
+```
+
+We can create an optimisation problem from this one like previously, but keeping in mind that it (and its output results) only contains two parameter values (*kB* and *kP):
```
isaacsas, line 137:
```suggestion
We can create an optimisation problem from this one like previously, but keep in mind that it (and its output results) only contains two parameter values (*kB* and *kP):
```

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -0,0 +1,173 @@
...
+```@example diffeq_param_estim_1 
+optprob_fixed_kD = OptimizationProblem(loss_function_fixed_kD, [1.0, 1.0])
+optsol_fixed_kD = solve(optprob_fixed_kD, Optim.NelderMead())
+nothing # hide
+```
+
+## Fitting parameters on the logarithmic scale
+Often it can be advantageous to fit parameters on a [logarithmic, rather than linear, scale](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1008646). The best way to proceed is to simply replace each parameters in the model definition by its logarithmic version:
```
isaacsas, line 145:
```suggestion
Often it can be advantageous to fit parameters on a [logarithmic, rather than linear, scale](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1008646). The best way to proceed is to simply replace each parameter in the model definition by its logarithmic version:
```
This doesn't seem like the right approach. One should just build a loss function wrapper that does the conversion for you instead of changing the model definition of parameters. This should be a trivial transformation to implement.

isaacsas (changes requested):
What about fitting multiple data sets that should have the same parameters simultaneously? (i.e. if the experiment has been run multiple times) Do we show this anywhere?

docs/src/inverse_problems/examples/ode_fitting_oscillation.md
TorkelE: I prefer the old version. Especially spelling out the AD part (As I imagine most people are unfamiliar with this, and especially just "AD backend" doesn't say much.

docs/src/inverse_problems/optimization_ode_param_fitting.md
TorkelE: OK. It probably is trivial, but I have to admit I don't understand how to do it from your example. It sounds like the best solution is to make an example problem where does that. put that under example problems, and then reference that.

Generally, PEtab handling this kind of stuff automatically (if you want) is one of the advantages of just using that one.

docs/src/inverse_problems/optimization_ode_param_fitting.md
TorkelE: will fix that

docs/src/inverse_problems/optimization_ode_param_fitting.md
TorkelE: I'm basically self-learned on optimisation, and don't really know the pros/cons of different algorithms. If you have any advice that would be useful

docs/pages.jl
TorkelE: Sorry, seemed to have gotten messed up in the latest merge. Fixed now.

docs/pages.jl
TorkelE: Sorry, seemed to have gotten messed up in the latest merge. Fixed now.

docs/src/inverse_problems/examples/ode_fitting_oscillation.md
TorkelE: Also, I generally try to make my comments full sentences, not sure if there is some standard regarding this though.

TorkelE:
> What about fitting multiple data sets that should have the same parameters simultaneously? (i.e. if the experiment has been run multiple times) Do we show this anywhere?

I think I looked at demonstrating this as well, but it was just really messy so I decided not to do it (was a while ago, so don't really remember). There is a section mentioning that this is possible, and linking to where DiffEqParamEstim describes it though:
```
## Parameter fitting to multiple experiments
Say that we had measured our model for several different initial conditions, and would like to fit our model to all these measurements simultaneously. This can be done by first creating a [corresponding `EnsembleProblem`](@ref advanced_simulations_ensemble_problems). How to then create loss functions for these are described in more detail [here](https://docs.sciml.ai/DiffEqParamEstim/stable/tutorials/ensemble/).
```
We also show how to do this for PEtab.

TorkelE:
Things updated and rebased on the latest v14 branch, thanks for the input!

isaacsas:
Feel free to merge when you feel this is done and sufficiently revised.

TorkelE:
Thanks for the input :)

 I will have a second read-through tonight just to be sure, and will merge afterwards.

PR #750, opened 2023-12-21
Performance advice for Jump simulation tutorial

TorkelE:
Adds a single tutorial, providing some advice on Jump simulation performance:

- Managing of solution saving using `save_positions` and `saveat`.
- Types of jumps and why e.g. `k, E + S --> E + P` is better than `k*E, S --> P`.
- How to select a good solver and why e.g. `RSSA` is probably better than `Direct` (and `RSSACR` should be preferred for big models).
- Mention of the existence and principle of hybrid simulation, however, notes that these are currently only a work in progress.


TorkelE:
Also, @isaacsas, If you know a good reference book/article on Jump simulation it would be useful to know (so I could put a reference to it here, for further reading).

isaacsas:
I think for bio readers this is a nice one that is focused on algorithms: Luca Marchetti, Corrado Priami, Vo Hong Thanh,
"Simulation Algorithms for Computational Systems Biology".

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+# `ConstantRateJump`s
+@reaction k*log(X), Y --> X
+@reaction mm(X,v,K), 0 --> X
+
+# `VariableRateJump`s
+@reaction k*(1+sin(t)), 0 --> X
+@reaction X/t, X + Y --> XY
+nothing # hide
+```
+Throughout a simulation, `VariableRateJump`s' rates require updating more frequently than `ConstantRateJump`s' rates, which in turn requires updating more frequently than `MassActionJump`s' rates. Since the performance of jump simulations is heavily affected by how frequently jump rates are computed, keeping the number of `ConstantRateJump`s and `VariableRateJump`s as small as possible is advantageous. Primarily, there exist two common cases where models are written in a way so that sub-optimal jump types are generated, both of which we describe below.
```
isaacsas, line 76: ```Throughout a simulation, `VariableRateJump`s' rates require updating more frequently than `ConstantRateJump`s' rates, which in turn requires updating more frequently than `MassActionJump`s' rates. ```
This isn't really the issue (or even true). I would more say that evaluating propensities and state changes due to `MassActionJump`s are more computationally efficient due to their constrained form than updating them for `ConstantRateJump`s (which are more general in the propensity functions and state changes they support but lose computational efficiency due to this flexibility). `VariableRateJump`s require more computationally expensive simulation algorithms than standard SSAs, which require root finding to calculate next event times (except when using rejection-based approaches via rate bounds, but Catalyst doesn't support that yet so it probably isn't worth mentioning).

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### Using piecewise constant, time-dependant, function instead of callbacks.
+Let us consider a reaction with some input that depends on time. We assume that the input is initially $0$, but after some critical time $t=5.0$ it is increased to $1$. This can be implemented through a step function:
+```@example jump_simulation_performance_1
+input(t) = t > 5.0
+r = @reaction input(t), 0 --> X
+nothing # hide
+```
+Here, the production of species $X$ is switched on at the time $t=5.0$. This reaction (which rate depends on `t`) will generate a `VariableRateJump`, which is bad for jump simulation performance. However, since the rate is piecewise constant, it can instead be implemented by setting it to a constant parameter $i$, and then use a [*callback*](@ref advanced_simulations_callbacks) to update it at the critical times. This will again generate an equivalent model, but with the reaction encoded as a `MassActionJump` (rather than a `VariableRateJump`).
```
isaacsas, line 94:
```suggestion
Here, the production of species $X$ is switched on at the time $t=5.0$. This reaction (for which the rate depends on `t`) will generate a `VariableRateJump`, which is the least performant jump type in simulations, and can not be used with the standard SSAs of JumpProcesses. However, since the rate is piecewise constant, it can alternatively be implemented by setting it to a constant parameter $i$, and then using a [*discrete callback*](@ref advanced_simulations_callbacks) to update it at the switching times. This will again generate an equivalent model, but with the reaction encoded as a `MassActionJump` (rather than a `VariableRateJump`). Generally such explicit time-discontinuities should be encoded via discrete callbacks instead of as `VariableRateJump`s if possible as simulation methods for the latter typically assume the system's propensities evolve continuously in-between jumps.
```
Do you want to show how to do this?

Should we show how to actually simulate a `VariableRateJump` model? (The interface isn't perfect, but it can be done...)

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+end
+u0 = [:X => 10]
+tspan = (0.0, 1000.0)
+ps = [:p => 1000.0, :d => 100.0]
+
+dprob = DiscreteProblem(rn, u0, tspan, ps)
+jprob = JumpProblem(rn, dprob, Direct())
+nothing # hide
+```
+Here (as throughout most of Catalyst's documentation) we have used the `Direct()` solver (which corresponds to Gillespie's original direct method [^2][^3], also called the *stochastic simulation algorithm*). This method was originally published in 1976, and since then, many additional methods for performing jump simulations of CRN models have been developed.
```
isaacsas, line 112:
```suggestion
Here (as throughout most of Catalyst's documentation) we have used the `Direct()` SSA solver (which corresponds to Gillespie's original direct method [^2][^3]). This method was originally published in 1976, and since then, many additional methods for simulating stochastic chemical kinetics models have been developed. 
```

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+tspan = (0.0, 1000.0)
+ps = [:p => 1000.0, :d => 100.0]
+
+dprob = DiscreteProblem(rn, u0, tspan, ps)
+jprob = JumpProblem(rn, dprob, Direct())
+nothing # hide
+```
+Here (as throughout most of Catalyst's documentation) we have used the `Direct()` solver (which corresponds to Gillespie's original direct method [^2][^3], also called the *stochastic simulation algorithm*). This method was originally published in 1976, and since then, many additional methods for performing jump simulations of CRN models have been developed. 
+
+Gillespie's direct method will, after a jump has been performed, recompute the rates of *all* possible jumps in the system. This is typically not required. E.g. consider the following system:
```
isaacsas, line 114:
```suggestion
Gillespie's direct method will, after a jump has been performed, recompute the propensities of *all* possible jumps in the system (i.e. of all reactions). This is typically not required. E.g. consider the following system:
```

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### Brief (and optional) introduction to jump simulations
+Jump simulations are continuous time, discrete space, simulations (where the system's state, throughout the simulation, are the discrete copy numbers of each species). The system's state changes at discrete time points (in so-called jumps). For CRNs, these jumps correspond to the occurrence of individual reactions. Here, *stochastic chemical kinetics* describes how the reactions are distributed across time. Typically, the frequency of each reaction depends on its *propensity* (which in turn depends on its *rate* and *substrates*). These depend on the state of the system, and must thus be re-computed whenever the system's state changes. Hence, jump simulations' runtimes are heavily dependent on how frequently these propensities must be recomputed.
+
+Typically, propensities are recomputed only whenever a jump occurs. This means that jump simulations' runtimes, very roughly, scale with the number of jumps. Runtimes typically become prohibitively expensive for:
```
isaacsas, line 7:
```suggestion
Typically, propensities are recomputed only whenever a jump occurs. This means that jump simulations' run-times, generally scale with the number of jumps that occur. Run-times typically become prohibitively expensive for:
```

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+
+Gillespie's direct method will, after a jump has been performed, recompute the rates of *all* possible jumps in the system. This is typically not required. E.g. consider the following system:
+```@example jump_simulation_performance_2
+@reaction_network begin
+    k1, X1 --> X2
+    k2, X2 --> X3
+    k3, X3 --> 0
+end
+```
+Here, the rate of the `k1, X1 --> X2` and `k2, X2 --> X3` reactions does not depend on the amount of $X3$ in the system. Hence, their rates are unaffected by the occurrence of the `k3, X3 --> 0` reaction. Performant jump simulation methods have clever ways to determine which rates require recomputing after the occurrence of each reaction, which improves their performance. Many of these depend on so-called dependency graphs (which track which reactions' rates are affected by the occurrence of which reactions). Catalyst automatically builds such dependency graphs, which means that most jump simulators can be used without any additional input.
```
isaacsas, line 122:
```suggestion
Here, the propensities of the `k1, X1 --> X2` and `k2, X2 --> X3` reactions, `k1*X1*X2` and `k2*X2` respectively, do not depend on the amount of $X3$ in the system. Hence, the value of their propensities are unchanged by the occurrence of the `k3, X3 --> 0` reaction. Performant jump simulation methods have several ways to determine which rates require recomputing after the occurrence of a reaction, which improves their computational performance. Many of these depend on dependency graphs, which track which other reactions' propensities must be recomputed after the occurrence of a given reaction. Catalyst automatically builds such dependency graphs, which means that all JumpProcesses SSAs can be used without any additional inputs.
```

Note, one needs to be careful about using rate in place of propensity. In Catalyst rate usually refers to the `rate` field of a `Reaction`, which is not the propensity function (the latter is more the stochastic rate law).

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+
+A full list of jump simulation method implemented by JumpProcesses can be found [here](https://docs.sciml.ai/JumpProcesses/stable/jump_types/#Jump-Aggregators-for-Exact-Simulation). Generally, `RSSA()` (the rejection SSA method [^4][^5]) is recommended for small models, with `RSSACR()` (the rejection SSA with composition-rejection method [^6]) typically being more performant for larger models. For models that are simulated a large number of times, it can be worthwhile to try a few different jump simulation methods to determine which one is most performant in each given case.
```
isaacsas, line 124: I think we need a little bit of discussion / explanation that SSAs are called aggregators and are responsible for calculating next jump times and type (i.e. which reaction occurs next and at what time). `SSAStepper` is then the way to advance a simulation using a selected aggregator (i.e. SSAs) from time to time in JumpProcesses for systems with constant and mass action rates. (Then we can point out the difference for systems with VariableRates, which require using ODE/SDE timesteppers instead of SSAStepper.)

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+jprob = JumpProblem(rn, dprob, RSSA())
+sol = solve(jprob, SSAStepper())
+
+plt_1 = plot(sol; idxs=2)
+plt_2 = plot(sol; idxs=3)
+plot(plt_1, plt_2, size=(1200,400))
+```
+We note that the copy numbers $Pₐ$ are highly stochastic. Meanwhile, $M$ exists in so large copy numbers that its trajectory can be considered approximately deterministic. For models like this one, jump simulations are required to capture the stochastic behaviour of the promoter. However, the large number of jumps generated by $M$ makes such simulations excessively expensive. Here, *hybrid simulations* can be used. Hybrid simulations employ different simulation strategies for their different components. In our example, the state of the promoter could be simulated using a jump simulation, while $M$ could be simulated using an ODE.
+
+Hybrid simulations for Catalyst models are currently not supported. However, it is a work in progress. If you require such simulations, please [raise an issue](https://github.com/SciML/Catalyst.jl/issues) and we can notify you of the current state of our implementation. Alternatively, other CRN modelling tools that support hybrid simulations exist.
```
isaacsas, line 153:
```suggestion
Hybrid simulations for Catalyst models are currently not supported, though they are supported by the lower-level solvers in JumpProcesses and DifferentialEquations. It is work in progress to interface Catalyst to such lower-level solvers, and to provide better optimized hybrid methods. If you require such simulations, please [raise an issue](https://github.com/SciML/Catalyst.jl/issues) and we can notify you of the current state of our implementation, or give suggestions on how one can directly interface with the lower-level hybrid solver interface.
```

isaacsas (changes requested):

docs/src/catalyst_applications/jump_simulation_performance.md
TorkelE: Yes. I was trying to keep things simple by referring to only rates, rather than propensities and rates, The idea was to make it easier for someone who hasn't studied the theory and wasn't familiar with these terms. We can rewrite it to use both, but in that case I might want a more throughout introduction of propensities at the earlier stage.

TorkelE:
I tried writing the tutorial assuming that someone wasn't familiar with rates, propensities, and rate laws, and hence basically just used rate instead of propensities. I don't min making it more rigorous, but then I want to add a proper introduction of what these things are. I think if we just start using propensity people not familiar with it will get confused really quickly.

(long term I want a better introduction to this somewhere in the docs as well, just so that it can be cross-referenced in situations like this)

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+
+Typically, propensities are recomputed only whenever a jump occurs. This means that jump simulations' runtimes, very roughly, scale with the number of jumps. Runtimes typically become prohibitively expensive for:
+- Simulations of large models (either with many different species or where some species occur in large copy numbers).
+- Simulations over long time spans.
+- Simulations that are performed a large number of times.
+
+More throughout description of these types of simulations can be found in the literature [^1].
+
+## Managing of solution saving
+By default, `solve` saves the value of the solution at the time of every jump. For simulations where a large number of jumps occur, this can cause memory to quickly fill up. Typically, for simulations with a large number of jumps, we want to [disable this feature](https://docs.sciml.ai/JumpProcesses/dev/tutorials/discrete_stochastic_example/#save_positions_docs) and instead set the save frequency manually. To exemplify this, let us consider a simple production/degradation model:
```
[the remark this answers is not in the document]
TorkelE: I agree. ish. Basically, at some point I think it would make sense to split out model simulation in the same manner as we did inverse problems. When we do, I think it would be useful to have a page introduction simulations (more properly than in the introduction to Catalyst). It would be one tutorial, with one section each for ODE, SDE, and Jumps. I think this should go there, and then this tutorial can make references to that. Basically, this is not supposed to become the master tutorial for Jump simulations, but rather a continuation on the general introduction.

I agree this order of things might be not the best though. Do you think it would make sense to do it this way? In that case, I can make a PR to:
- Create a simulation section
- Create an introduction to simulations (could also cover some additional useful stuff, like difference between DIfferentialEquartions and OrdinaryDiffEq/StochasticDiffEq/JumpProcesses).

I think it would also make sense to move the part where we show the equations (e.g. explaining how higher order terms are handled) from the Introduction to Catalyst tutorial to this tutorial.

Next, this and the ODE performance tutorial could be rebased on that, and we might consider what to put in this tutorial (given what already is in the base one).

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+nothing # hide
+```
+Let us simulate this model using the default options, and plot the results. Furthermore, we use [BenchmarkTools.jl's](https://github.com/JuliaCI/BenchmarkTools.jl) `@btime` macro to measure the time it takes to plot the output solution:
+```@example jump_simulation_performance_1
+using BenchmarkTools
+jprob = JumpProblem(rn, dprob, Direct())
+sol = solve(jprob, SSAStepper())
+@btime plot(sol)
+```
+This simulation generates a very large number of jumps, with the solution saved and plotted at each time such a jump occurs. If the number of jumps is high enough, the memory limits may be exceeded (causing Julia to crash). Here, we do not reach that limit, however, the performance of the `plot(sol)` is affected by the number of points it must render (rendering too much in a single plot is another potential cause of crashed due to memory strain).
```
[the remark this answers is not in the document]
TorkelE: In this case I think it is less than a second, so should be no worry (the biggest worry is probably that the effect, in this case, is not that significant).

docs/src/catalyst_applications/jump_simulation_performance.md
```diff
@@ -0,0 +1,163 @@
...
+However, if we were to simulate our model now, the solution would only actually be saved at its initial and final times. To remedy this, we provide the `saveat = 1.0` argument to the `solve` command, causing the solution to be saved at every `1.0`th time unit (at `t` = `0.0`, `1.0`, `2.0`, ... `999.0`, `1000.0`).
+```@example jump_simulation_performance_1
+sol = solve(jprob, SSAStepper(),saveat=1.0)
+nothing # hide
+```
+we can now plot the new solution:
+```@example jump_simulation_performance_1
+@btime plot(sol)
+```
+Here, we note that the time to plot the simulation was reduced (for this example, admittedly from an already low level). Furthermore, the plots look different since the trajectory is sampled at much sparser time points.
```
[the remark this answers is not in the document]
TorkelE: Didn't know it was actually possible to do `sol(t)` to get value at certain time points. Likely most users won't know that either. We probably should explain that somewhere, or reference a description, before doing into detail about that here.

Also, wouldn't think vaguely be the case for ODEs/SDEs as well, just that the interpolation there is better?

ChrisRackauckas:
Is it worth mentioning small case static array stuff?

isaacsas:
We should probably benchmark that case and see if there is a speed up. I don’t think we’ve ever done so.

ChrisRackauckas:
I have, you can see it here:

https://github.com/sdwfrost/Gillespie.jl?tab=readme-ov-file#benchmarks

It's a bit old but it works well for small problems. In fact, for the smallest of small problems we should probably make a different codegen approach, but that's a whole different story.

PR #751, opened 2023-12-25
[v14 ready] Tutorial for ensuring performance for ODE simulations

TorkelE:
Creates a new doc page, providing some performance advice for ODE simulations. Stuff covered:

- Stiff vs non-stiff ODEs.
- Selecting a solver for ODE simualtions.
- Various options for Jacobian computations (symbolic jacobian, sparse jacobian, lienar solver selection, and preconditioners).
- Parallelisation on CPUs.
- Parallelisation on GPUs.

The GPU parallelization part is not ready, I have only started. However, I need to go through some stuff with Utkarsh before finishing that part.

Also, if someone have a link to a good reference book introducing this kind of stuff, that would be useful


docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,315 @@
...
+
+We can now simulate our model using a GPU-based ensemble algorithm. Currently, two such algorithms are available, `EnsembleGPUArray` and `EnsembleGPUKernel`. Their differences are that
+* Only `EnsembleGPUKernel` requires arrays to be static arrays (although it is still advantageous for `EnsembleGPUArray`).
+* While `EnsembleGPUArray` can use standard ODE solvers, `EnsembleGPUKernel` requires specialised versions (such as `GPUTsit5`). A list of available such solvers can be found [here](https://docs.sciml.ai/DiffEqGPU/dev/manual/ensemblegpukernel/#specialsolvers).
+
+Generally, it is recommended to use `EnsembleGPUArray` for large models (that have at least $100$ variables), and `EnsembleGPUKernel` for smaller ones. Here we simulate our model using both approaches (noting that `EnsembleGPUKernel` requires `GPUTsit5`):
+```@example ode_simulation_performance_5
+esol1 = solve(eprob, Tsit5(), EnsembleGPUArray(CUDA.CUDABackend()); trajectories=100)
+esol2 = solve(eprob, GPUTsit5(), EnsembleGPUKernel(CUDA.CUDABackend()); trajectories=100)
+nothing # hide
```
utkarsh530, line 308: Confirmed that this code works for me.

```
Status `~/DiffEqGPU_Catalyst/Project.toml`
  [052768ef] CUDA v5.1.1
  [479239e8] Catalyst v13.5.1
  [071ae1c0] DiffEqGPU v3.3.0
⌅ [1dea7af3] OrdinaryDiffEq v6.66.0
  [90137ffa] StaticArrays v1.9.0
Info Packages marked with ⌅ have new versions available but compatibility constraints restrict them from upgrading. To see why use `status --outdated`
```

utkarsh530 (changes requested):

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: thanks for checking :)

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+
+Generally, this small checklist provides a quick guide for dealing with ODE performance:
+1. If performance is not critical, use [the default solver choice](@ref ode_simulation_performance_solvers) and do not worry further about the issue.
+2. Determine whether your problem is [non-stiff nor stiff](@ref ode_simulation_performance_stiffness), and use the `Tsit5` solver for the former case and `Rodas5P` for the latter.
```
ChrisRackauckas, line 8: Definitely no. Absolutely not. This is not a good idea, and the docs do not recommend this.

ChrisRackauckas: This is almost certainly worse than the default solver choice. This "step" should just be removed. Either someone should read the solver selection recommendations or they should use the default method. Most likely, they should use the default method.

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+3. If more performance would be useful, read about solver selection (both in [this tutorial](@ref ode_simulation_performance_solvers) and [OrdinaryDiffEq's documentation](https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/)) and then try a few different solvers to find one with good performance.
+4. If you have a large ODE, try the [various options for efficient Jacobian computation](@ref ode_simulation_performance_jacobian) (noting that some are non-trivial to use, and should only be investigated if required).
```
ChrisRackauckas, line 10:
```suggestion
4. If you have a large ODE (approximately size 100 or more), try the [various options for efficient Jacobian computation](@ref ode_simulation_performance_jacobian) (noting that some are non-trivial to use, and should only be investigated if required).
```

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+```@example ode_simulation_performance_1
+sol2 = solve(oprob, Rodas5P())
+plot(sol2)
+``` 
+This time the simulation was successfully completed, which can be confirmed by checking the return code:
+```@example ode_simulation_performance_1
+sol2.retcode
+``` 
+
+Generally, ODE solvers can be divided into [*explicit* and *implicit* solvers](https://en.wikipedia.org/wiki/Explicit_and_implicit_methods). Roughly, explicit solvers (which include `Tsit5`) are better for non-stiff problems, with implicit solvers (like `Rodas5P`) being required for stiff problems. While we could use implicit solvers for all problems (to guarantee successful simulations irrespective of stiffness), these are typically less performant on non-stiff problems (as compared to the explicit solvers). An important property of implicit solvers is that they require the *computation of a Jacobian* as part of their routine. This means that the various options for efficient Jacobian computation [described later in this tutorial](@ref ode_simulation_performance_jacobian) are only relevant to implicit solvers.
```
ChrisRackauckas, line 49: Roughly is doing a lot of work here, since Rodas5P isn't actually implicit.

ChrisRackauckas: > these are typically less performant on non-stiff problems (as compared to the explicit solvers)

these are typically less performant on non-stiff problems (as compared to the explicit solvers) on equations that are non-stiff.

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+
+Finally, we should note that this is not an exact science, and sometimes explicit solvers can successfully solve a stiff problem. E.g. if we change the parameters values of our previous Brusselator model to `[:A => 1.0, :B => 4.0]`, the `Tsit5` will successfully be able to simulate it.
```
ChrisRackauckas, line 51: This is not correct. Changing the parameters can change a problem from stiff to non-stiff. It is not only about equation form.

ChrisRackauckas: If there is a fundamental principle about stiffness, it's based on the pseudospectra of the Jacobian. A reasonable simplification is the condition number of the Jacobian, i.e. the ratio between the maximum and minimum eigenvalue, or roughly the scale between the fastest and slowest operations. So it's not the case that parameters don't effect that. And in direct contrast, it's is almost certainly a function of parameters.

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+using DifferentialEquations
+solve(oprob)
+nothing # hide
+``` 
+While the default choice is typically enough for most single simulations, if performance is important, it can be worthwhile exploring the available solvers to find one that is especially suited for the given problem. A complete list of possible ODE solvers, with advice on optimal selection, can be found [here](https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/). This section will give some general advice.
+
+The most important part of solver selection is to select one appropriate for [the problem's stiffness](@ref ode_simulation_performance_stiffness). Generally, the `Tsit5` solver is recommended for non-stiff problems, and `Rodas5P` for stiff problems. For large stiff problems (with many species), `QNDF` can be a good choice. We can illustrate the impact of these choices by simulating our production/degradation model using the `Tsit5`, `BS5` (an explicit solver yielding [low error in the solution](@ref ode_simulation_performance_error)), `Rodas5P`, and `QNDF` solvers (benchmarking their respective performance using [BenchmarkTools.jl](https://github.com/JuliaCI/BenchmarkTools.jl)):
```
ChrisRackauckas, line 74: `FBDF` instead of `QNDF`. Say a cutoff of around 50.

Instead of BS5, show Vern7.

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+
+### [Building the Jacobian symbolically](@id ode_simulation_performance_symbolic_jacobian)
+By default, OrdinaryDiffEq computes the Jacobian using [*automatic differentiation*](https://en.wikipedia.org/wiki/Automatic_differentiation) (however, using [*finite differences*](https://en.wikipedia.org/wiki/Finite_difference) is [also possible](https://docs.sciml.ai/DiffEqDocs/stable/features/performance_overloads/)). Since Catalyst builds its ODEs symbolically, it is able to *compute an analytic Jacobian symbolically*. Typically, this is advantageous when some rows of the Jacobian are highly dense (that is, some system species interact with most other species of the system).
```
ChrisRackauckas, line 99: It's really only advantageous when sparse

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+### [Linear solver selection](@id ode_simulation_performance_symbolic_jacobian_linear_solver)
+When implicit solvers use e.g. the Newton-Raphson method to, in each simulation time step, solve a (typically non-linear) equation, they actually solve a linearised version of this equation. For this they use a linear solver, the choice of which can impact performance. To specify one, we use the `linsolve` option (given to the solver function, *not* the `solve` command). E.g. to use the `KLUFactorization` linear solver we run
+```@example ode_simulation_performance_3
+using DifferentialEquations
+solve(oprob, Rodas5P(linsolve = KLUFactorization()))
+nothing # hide
+```
+Please note that this requires the full DifferentialEquations package (rather than just the more lightweight OrdinaryDiffEq). A full list of potential linear solvers can be found [here](https://docs.sciml.ai/LinearSolve/dev/solvers/solvers/#Full-List-of-Methods), however, the default choice typically performs well.
```
ChrisRackauckas, line 133: It does not. It only requires LinearSolve.

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+```@example ode_simulation_performance_3
+oprob = ODEProblem(brusselator, u0, tspan, p; sparse=true)
+nothing # hide
+```
+
+### [Linear solver selection](@id ode_simulation_performance_symbolic_jacobian_linear_solver)
+When implicit solvers use e.g. the Newton-Raphson method to, in each simulation time step, solve a (typically non-linear) equation, they actually solve a linearised version of this equation. For this they use a linear solver, the choice of which can impact performance. To specify one, we use the `linsolve` option (given to the solver function, *not* the `solve` command). E.g. to use the `KLUFactorization` linear solver we run
+```@example ode_simulation_performance_3
+using DifferentialEquations
+solve(oprob, Rodas5P(linsolve = KLUFactorization()))
```
ChrisRackauckas, line 130: This is the wrong solver for anything of this size and we should demonstrate this more correctly.

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+    else
+        Pl = Plprev
+    end
+    Pl, nothing
+end
+nothing # hide
+```
+Next, `incompletelu` can be supplied to our solver using the `precs` argument:
+```@example ode_simulation_performance_3
+solve(oprob, Rodas5P(precs = incompletelu))
```
ChrisRackauckas, line 158: this won't use the preconditioner

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+nothing # hide
+```
+Finally, we note that since matrix-free linear solvers (like `KrylovJL_GMRES`) by default do not build a Jacobian. Hence, if we want to use them with a preconditioner we must tell them to build it. This can be done using the `concrete_jacobian=true` option:
+```@example ode_simulation_performance_3
+solve(oprob, Rodas5P(linsolve = KrylovJL_GMRES(), precs = incompletelu, concrete_jac=true))
```
ChrisRackauckas, line 163: you need to skip straight to here because the line above is just incorrect

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+- `prob`: The problem that it modifies at the start of each individual run (which will be the same as `EnsembleProblem`'s first argument).
+- `i`: The index of the specific simulation.
+- `repeat`: The repeat of the specific simulations. We will ignore this one for now, but more details (like how it is different from `i`) is provided [here](https://docs.sciml.ai/DiffEqDocs/stable/features/ensemble/#Building-a-Problem).
+
+and output the `ODEProblem` simulated in the ith simulation.
+
+Let us assume that we wish to simulate our model 100 times, for $kP = 0.01, 0.02, ..., 0.99, 1.0$. We can now define our `prob_func` as:
+```@example ode_simulation_performance_4
+function prob_func(prob, i, repeat) 
+    prob[:kP] = 0.01*i
```
ChrisRackauckas, line 210: this isn't proper prob indexing.

isaacsas:
Why don’t we just use some of the examples from the Catalyst paper that show different performance characteristics to illustrate which solvers and configurations work best in different cases? Then we’ve already done the benchmarking and know what is fastest.

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+```@example ode_simulation_performance_4
+plot(esol.u[47])
+```
+To extract the amount of $P$ produced in each simulation, and plot this against the corresponding $kP$ value, we can use:
+```@example ode_simulation_performance_4
+plot(0.01:0.01:1.0, map(sol -> sol[:P][end], esol.u), xguide="kP", yguide="P produced", label="")
+```
+
+Above, we have simply used `EnsembleProblem` as a convenient interface to run a large number of similar simulations. However, these problems have the advantage that they allow the passing of an *ensemble algorithm* to the `solve` command, which describes a strategy for parallelising the simulations. By default, `EnsembleThreads` is used. This parallelises the simulations using [multithreading](https://en.wikipedia.org/wiki/Multithreading_(computer_architecture)) (parallelisation within a single process), which is typically advantageous for small problems. An alternative is `EnsembleDistributed` which instead parallelises the simulations using [multiprocessing](https://en.wikipedia.org/wiki/Multiprocessing) (parallelisation across multiple processes). To do this, we simply supply this additional solver to the solve command:
```
ChrisRackauckas, line 234: Typically advantageous for shared memory devices.

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+We can now simulate our model using a GPU-based ensemble algorithm. Currently, two such algorithms are available, `EnsembleGPUArray` and `EnsembleGPUKernel`. Their differences are that
+* Only `EnsembleGPUKernel` requires arrays to be static arrays (although it is still advantageous for `EnsembleGPUArray`).
+* While `EnsembleGPUArray` can use standard ODE solvers, `EnsembleGPUKernel` requires specialised versions (such as `GPUTsit5`). A list of available such solvers can be found [here](https://docs.sciml.ai/DiffEqGPU/dev/manual/ensemblegpukernel/#specialsolvers).
+
+Generally, it is recommended to use `EnsembleGPUArray` for large models (that have at least $100$ variables), and `EnsembleGPUKernel` for smaller ones. Here we simulate our model using both approaches (noting that `EnsembleGPUKernel` requires `GPUTsit5`):
+```@example ode_simulation_performance_5
+esol1 = solve(eprob, Tsit5(), EnsembleGPUArray(CUDA.CUDABackend()); trajectories=100)
+esol2 = solve(eprob, GPUTsit5(), EnsembleGPUKernel(CUDA.CUDABackend()); trajectories=100)
```
ChrisRackauckas, line 307: Needs more trajectories to make sense. 10,000 trajectories at least. Modern GPUs have over 10,000 CUDA cores so there's effectively no performance disadvantage to doing less.

docs/src/catalyst_applications/ode_simulation_performance.md
```diff
@@ -0,0 +1,318 @@
...
+nothing # hide
+```
+Note that we have to provide the `CUDA.CUDABackend()` argument to our ensemble algorithms (to designate our GPU backend, in this case, CUDA).
+
+Just like OrdinaryDiffEq is able to utilise parallel CPU processes to speed up the linear solve part of ODE simulations, GPUs can also be used. More details on this can be found [here](https://docs.sciml.ai/DiffEqGPU/stable/tutorials/within_method_gpu/). This is only recommended when ODEs are very large (at least 1,000 species), and typically not applicable to CRNs.
```
ChrisRackauckas, line 312: mostly because they tend to have unstructured sparsity.

docs/src/catalyst_applications/ode_simulation_performance.md
ChrisRackauckas: We should recommend `using MKL` and trying UMFPACK

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: updated

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: Will update to use the new SII

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: updated.

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: updated

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: fixed

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: step 2 removed

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: fixed

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: You mean that Rodas5P is wrong for large problems (where we would do something like this) and should be changed, or that since preconditioners and stuff like this is not good for a 2-state brusselators we should demonstrate on a larger model?

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: Does preconditioners make sense for linear solvers except for `KrylovJL_GMRES` then? Or should `concrete_jac=true` be used for the previous case?

docs/src/catalyst_applications/ode_simulation_performance.md
TorkelE: Didn't find any examples for MKL, so was unsure on how to implement it here.

TorkelE:
Did you have any further input on the preconditioners @ChrisRackauckas? 

Also, what is the progress on moving default solvers to OrdinaryDiffEq? If that is coming soon then useful, but it would be useful to have a quick and dirt solver selection advice that did not depend on reading through all the doc, nor loading the full DifferenitalEquations suite.

PR #755, opened 2023-12-30
warning fixes

isaacsas:
Fixes some incorrect usage of fields in MTK systems instead of getters.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -55,7 +55,7 @@ plot(bif_dia; xguide="k1", yguide="X")
 - Updated Bifurcation documentation with respect to this new feature.
 
 ## Catalyst 13.5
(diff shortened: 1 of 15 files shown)


isaacsas:
needs https://github.com/JuliaSymbolics/Symbolics.jl/pull/1029 for tests to pass

TorkelE:
Looks good.

Out of curriosity, why should we use `ModelingToolkit.getname()` instead of `getname` directly?

isaacsas:
> Looks good.
> 
> Out of curriosity, why should we use `ModelingToolkit.getname()` instead of `getname` directly?

No good reason -- I was making changes to get rid of the warnings and had put something else there at one point, and the ModelingToolkit just got kept. As long as that function is exported by MTK the qualifier wouldn't be needed.

PR #757, opened 2023-12-31
WIP: update docs for SII and other changes

isaacsas:
A very large number of examples throughout the docs now error due to using `Symbol`s in ways that are no longer supported / work so we are going to have to update all over to get things working again. We can work on it in this PR.


ChrisRackauckas:
One thing to make sure is with the getp/setp and getu/setu we should show properly caching the function for a lean cost function.

isaacsas:
If you look at the tutorial I updated I repeatedly reuse the function on different object (problems, integrators, and solutions). But probably in the callback section we should emphasize that this should be precomputed and stored in the callback.

isaacsas:
@TorkelE can you handle the remaining errors? They are from

1. SDE related indexing (just needs the SII/SciMLBase fixes, then we need to bump the doc/Project.toml to the right SciMLBase release).
2. Issues with the BifurcationTutorial
3. Broken links in a number of the tutorials.
4. Unreferenced footnotes in many tutorials. (I've made this only be warnings, but it would be good to fix them.)

You should be able to push directly to this PR, but let me know if you can't.

isaacsas:
I also suggest doing a full local doc build once everything is fixed, and carefully looking through all the tutorials to check the outputs / figures are working and look correct (give that we can no longer preview it via CI).

TorkelE:
Will have a look. I had a look through earlier and stuff should be fine, but maybe something appeared. Unreferenced footnote are intentional. They are basically further readings for some cases, but where I don't directly reference them in the tutorial.

TorkelE:
I try to build local, but that fails on my computer these days (or most Julia code), which limits its utility.

TorkelE:
There are a couple of indexing PRs up already that should fix some of the issues, so will go through once that is passed.

isaacsas:
Sounds good! If it is failing on your computer have you deleted the manifest in the docs folder and re-resolve/updated using doc/Project.toml?

TorkelE:
will try that, thanks :)

docs/src/catalyst_applications/advanced_simulations.md
```diff
@@ -181,20 +182,28 @@ sol = solve(oprob; callback = ps_cb)
...
 
+# the following code creates a function, `setk` that can
+# change parameters within an integrator
+using ModelingToolkit: setp
+setk = setp(rn, :k)
+
+# now we create our callback and solve
 condition = [5.0]
-affect!(integrator) = integrator[:k] = 5.0
+affect!(integrator) = (setk(integrator, 5.0); nothing)
```
TorkelE, line 206: Apparently, the bugs introduced by the SII update will be fixed together with the MTK v9 release, which should have been last week, so hopefully (fingers very crossed) it should be soon.

Either way, I think we should use the 
```julia
integrator.ps[:k] = 5.0
```
notation throughout tutorials, rather than
```julia
setk = setp(rn, :k)
setk(integrator, 5.0)
```

docs/src/catalyst_applications/advanced_simulations.md
```diff
@@ -31,13 +31,14 @@ rn = @reaction_network begin
     v0 + hill(X,v,K,n), ∅ --> X
     deg, X --> ∅
 end
+rn = complete(rn)
```
TorkelE, line 34: I don't think `complete()` here (and in other places are needed). We should also introduce a section somewhere (I am writing right now) explain what `complete` is.

docs/src/catalyst_applications/advanced_simulations.md
```diff
@@ -217,69 +226,67 @@ because the callback modifies our `ODEProblem` during the simulation, and this
 modification remains during the second simulation. An improved workflow to avoid
 this issue is:
 ```@example ex2
```
TorkelE, line 228: Still think this will use the old (modified) p, so at least that must be added back in.

docs/src/catalyst_applications/nonlinear_solve.md
```diff
@@ -86,7 +86,7 @@ here it is important that the quantities used in `u_guess` correspond to the con
 ```@example nonlinear_solve2
 sol = solve(nl_prob)
 ```
-We note that the output only provides a single value. The reason is that the actual system solved only contains a single equation (the other being eliminated with the conserved quantity). To find the values of $X1$ and $X2$ we can [directly query the solution object for these species' values, using the species themselves as inputs](@ref simulation_structure_interfacing_solutions):
+We note that the output only provides a single value. The reason is that the actual system solved only contains a single equation (the other being eliminated with the conserved quantity). To find the values of $X1$ and $X2$ we can [directly query the solution object for these species' values, using the species themselves as inputs](@ref simulation_structure_interfacing):
```
TorkelE, line 89: Why change this? Doesn't it make sense to directly reference the section where this is explained?

docs/src/catalyst_applications/simulation_structure_interfacing.md
```diff
@@ -11,35 +11,91 @@ using Catalyst
 rn = @reaction_network begin
     (k1,k2), X1 <--> X2
 end
+```
+
+To later enable us to index solutions and problems using symbolic variables we
+also need to indicate that we are finished constructing our system using `complete`:
+```@example ex1
+rn = complete(rn)
+nothing  # hide
```
TorkelE, line 20: I don't think this belongs here, we should put a proper (but not long-winded) explanation of `complete` and similar somewhere. Then, we can reference that whenever we want to introduce complete systems.

docs/src/catalyst_applications/simulation_structure_interfacing.md
```diff
@@ -52,82 +108,77 @@ oprob2.tspan
...
 ```@example ex1
-integrator[:X1] = 10.0
+oprob2.p
 ```
-Please read [this](@ref advanced_simulations_ssa_callbacks) with regards to updating integrators of `JumpProblem`s.
 
+Note that using `Symbol`s like `:k1` for parameters is not currently supported with `remake`.
 
-## [Interfacing solution objects](@id simulation_structure_interfacing_solutions)
+## Interfacing integrator objects
```
TorkelE, line 138: I think git got stuff confused here., should be fine though

docs/src/catalyst_applications/simulation_structure_interfacing.md
```diff
@@ -52,82 +108,77 @@ oprob2.tspan
...
+oprob3 = remake(oprob; u0 = [X1 => 10.0])
 nothing    # hide
 ```
-will only update the initial conditions.
-
+will only update `X1`'s value in the initial conditions.
 
-## Interfacing integrator objects
-
-During a simulation, the solution is stored in an integrator object, we will here describe how to interface with these. The primary circumstance under which a user may wish to do so is when using [callbacks](@ref advanced_simulations_callbacks). We can create an integrator by calling `init` on our problem ([while circumstances where the user might want to use `init` function exist](https://docs.sciml.ai/DiffEqDocs/stable/basics/integrator/#Initialization-and-Stepping), since integrators are automatically created during simulations, these are rare):
```
TorkelE, line 67: Git got messed up here. This is marked as deleted, which it is not. Can we salvage this, should we ignore it, or do we start from the beginning on a new PR once the other stuff is merged?

isaacsas:
I assume this has been handled / will be handled in other docs so closing.

PR #761, opened 2024-01-10
Add MTK structure indexing tests

TorkelE:
Adds a range of tests that should cover all cases of indexing for all relevant problem types. Will need to wait for merging until various fixes in SymbolicIndexingInterface, as well as updated support for indexing parameters via `.ps[]` notation.


AayushSabharwal (commented):
Tests don't run because `Direct` isn't defined. They're very comprehensive, but I think `ReactionSystem` might need to implement SII for this to work.

TorkelE:
Good catch. Practically this will need the latest MTK update to run (and the accompanying PRs), as that (should) fix a lot of the issues, and also enable the `struct.ps[p]` notation for get/set index which we test here.

Ideally also https://github.com/SciML/Catalyst.jl/pull/735 to make it easier to test the observables.

TorkelE:
Also, are you sure about `ReactionSystem` needing to implement SII? Whenever a problem is generated, it always has to be converted to the corresponding system type. Except for the stuff which is broken with MTK models as well, things works with `ReactionSystem`s, so I think it should be fine (unless you know something I don't)

isaacsas:
`ReactionSystem`s are `AbstractSystem`s that conform to the MTK interface, shouldn't SII work with a general `AbstractSystem`?

TorkelE:
Closed due to git history mess

PR #763, opened 2024-01-14
Add continuous events to DSL

TorkelE:
Adds the ability to create continuous events in the DSL. Uses a general event PR that discrete events can also use. tests currently cover both continuous and discrete events (for when the latter gets added, it was easier to do them both at once). Documentation not added. We can do that later, or post-merging (DSL docs need revamping anyway).

Expected syntax is used. Events are encapsulated by a `@continuous_events` option block. Within it, they are written as normal equations are declared. Example:
```julia
    # Creates model via DSL.
    rn_dsl = @reaction_network rn begin
        @parameters thres=1.0
        @variables Z(t)
        @continuous_events begin
            [t - 2.5] => [p ~ p + 0.2]
            [X - thres, Y - X] => [X ~ X - 0.5, Z ~ Z + 0.1]
        end

        (p, dX), 0 <--> X
        (p, dY), 0 <--> Y
    end
```


TorkelE:
This PR is based on https://github.com/SciML/Catalyst.jl/pull/736, and should be merged after that one.

TorkelE:
Git history not sustainable anymore, reuploaded as https://github.com/SciML/Catalyst.jl/pull/801

PR #765, opened 2024-01-16
[v14 ready] Change basic model examples to library of basic models

TorkelE:
Basically, we have a couple of standard models that we use recurringly throughout tutorials. I expanded the basic model examples doc page (also renaming it to "Library of Basic Chemical Reaction Network Models", instead of "Examples ...") to include the various common examples (8 in total). 

I also expanded each section a little bit (e.g. giving 2-3 sentences of background, and changing up a bit what kind of simulations we perform).

In the future, when we use a Brusselator model, I figured we could now instead coss-reference to that section, rather that e.g. link to Wikipedia (the brusselator specific section do actually contain a Wikipedia link though).


TorkelE:
This one should be ready for merging now as well

PR #771, opened 2024-01-25
Enable dsl observed

TorkelE:
Merges https://github.com/SciML/Catalyst.jl/pull/735 into v14. 

Original PR test:

Enables observables to be created as a part of the DSL. Example:
```julia
using Catalyst 
rn = @reaction_network begin
  @observables begin
    X ~ Xi + Xa
  end
  (p,d), 0 <--> Xi
  (k1,k2), Xi <--> Xa
end
```

if we simulate the system, the observables can now be used for indexing:
```julia
using OrdinaryDiffEq
@unpack Xi, Xa, X, p, d, k1, k2 = rn
u0 = [Xi => 0.0, Xa => 0.0]
ps = [p => 1.0, d => 0.2, k1 => 1.0, k2 => 2.0]
oprob = ODEProblem(rn, u0, (0.0, 10.0), ps)
sol = solve(oprob, Tsit5())
sol[X]
```
```julia
using Plots
plot(sol; idxs=X)
```

Some additional notes:
- If only a single observable is used, the `begin ... end ` block is not required:
```rn = @reaction_network begin
  @observables X ~ Xi + Xa
  (p,d), 0 <--> Xi
  (k1,k2), Xi <--> Xa
end
```
- The right-hand side can have any algebraic expression. However, all stuff there must be declared elsewhere (and it cannot contain other observables).
- Currently, the observables do not have any independent variables. We might want to change this part (although it will require a decent amount of coding).
- Metadata for observables is not supported either. This should probably be added I think, but not sure.

Tests and docs are here, so in principle this is ready
(unless we want to support independent variables in this PR)

Consideration regarding independent variables for observables: If the thing it consists of has several different ones, the iv for the observable should be the union of all those occurring in its formula, right?


docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -567,3 +567,60 @@ species(rn)
...
+  @observables Xtot ~ X + XY
+  (pX,dX), 0 <--> X
+  (pY,dY), 0 <--> Y
+  (kB,kD), X + Y <--> XY
+end
+```
+
+Finally, some general rules for creating observables:
+- Observables can depend on any species, parameters, or variables, but not on other observables.
+- All observables components must be declared somewhere (i.e., they cannot only appear as a part of the observables formula).
```
isaacsas, line 621:
```suggestion
- All observables components appearing on the right side of the `~` must be declared somewhere (i.e., they cannot only appear as a part of the observables formula).
```
Does interpolation of external symbolics work too? We should probably support that.

docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -567,3 +567,60 @@ species(rn)
...
+- Only a single `@observables` option block can be used in each `@reaction_network` call.
+- The left-hand side of the observables expression must be a single symbol, indicating the observable's name.
```
isaacsas, line 623:
```suggestion
- The left-hand side of the observables expression must be a single symbol, indicating the observable's name.
```

docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -567,3 +567,60 @@ species(rn)
...
+- Metadata can, however, be provided, e.g through `@observables (Xtot, [description="Total amount of X"]) ~ X + XY`.
+- The right-hand side of the observables expression can be any valid algebraic expression.
+- Observables are (by default, but this can be changed) considered `variables` (and not `species`). This can be changed by e.g. pre-declaring them using the `@species` option.
```
isaacsas, line 626: Give an example then where you can predefine an observable as a species to make concrete?

src/reaction_network.jl
```diff
@@ -682,6 +687,71 @@ function recursive_find_reactants!(ex::ExprValues, mult::ExprValues,
...
+        observed_vars = Expr(:block, :(@variables))
+
+        for (idx, obs_eq) in enumerate(observed_eqs.args)
+            # Extract the observable, checks errors, and continues the loop if the observable has been declared.        
+            obs_name, ivs, defaults, metadata = find_varinfo_in_declaration(obs_eq.args[2])
+            isempty(ivs) || error("An observable ($obs_name) was given independent variable(s). These should not be given, as they are inferred automatically.")
+            isnothing(defaults) || error("An observable ($obs_name) was given a default value. This is forbidden.")
+            in(obs_name, forbidden_symbols_error) && error("A forbidden symbol ($(obs_eq.args[2])) was used as an observable name.")
+            if (obs_name in species_declared) 
+                isnothing(metadata) || error("Metadata was provided to observable $obs_name in the `@observables` macro. However, the obervable was also declared separately (using either @species or @variables). When this is done, metadata cannot be provided within the @observables declaration.")
```
isaacsas, line 708:
```suggestion
                isnothing(metadata) || error("Metadata was provided to observable $obs_name in the `@observables` macro. However, the obervable was also declared separately (using either @species or @variables). When this is done, metadata should instead be provided within the original @species or @variable declaration.")
```

src/reactionsystem.jl
```diff
@@ -516,6 +516,10 @@ struct ReactionSystem{V <: NetworkProperties} <:
     function ReactionSystem(eqs, rxs, iv, sivs, states, spcs, ps, var_to_name, observed,
                             name, systems, defaults, connection_type, nps, cls, cevs, devs,
                             complete::Bool = false; checks::Bool = true)
+        # Filters away any potential obervables from `states` and `spcs`.
+        obs_vars = [obs_eq.lhs for obs_eq in observed]
+        states = filter(state -> !any(isequal(state, obs_var) for obs_var in obs_vars), states)
+        spcs = filter(spc -> !any(isequal(spc, obs_var) for obs_var in obs_vars), spcs)
```
isaacsas, line 522: I don't think this is the place to do this. It would be reasonable to have a check and error message here, but this shouldn't rearrange a user's state and species vectors.

Can you move this into the DSL logic, and make this just a consistency check that `states` and `spcs` don't contain an observable?

test/dsl/dsl_options.jl
```diff
@@ -339,3 +339,236 @@ let
...
+    end
+    rn4 = @reaction_network rn_observed begin
+        @variables X(t) [description="description"]
+        @observables X ~ X1 + X2
+        k, 0 --> X1 + X2
+    end
+    @test isequal(observed(rn3)[1].rhs, observed(rn4)[1].rhs)
+    @test isequal(observed(rn3)[1].lhs.metadata, observed(rn4)[1].lhs.metadata)
+    @test isequal(states(rn3), states(rn4))
+end
```
isaacsas, line 511: Do you have a version explicitly declaring an observable as a species (with metadata) too?

test/dsl/dsl_options.jl
```diff
@@ -339,3 +339,236 @@ let
...
+
+# Tests various erroneous declarations throw errors.
+let 
+    # Independent variable in @compounds.
```
isaacsas, line 515:
```suggestion
    # Independent variable in @observables.
```

test/dsl/dsl_options.jl
```diff
@@ -339,3 +339,236 @@ let
...
+    # System with observable in observable formula.
+    @test_throws Exception @eval @reaction_network begin
+        @observables begin
+            X ~ X1 + X2
+            X2 ~ 2X
+        end
+        (p,d), 0 <--> X1 + X2
+    end
+
+    # Multiple @compounds options
```
isaacsas, line 530:
```suggestion
    # Multiple @observables options
```

isaacsas (dismissed):
- Can you add a test that when predeclaring a species or variable observable it doesn't end up in `states(sys)` or `species(sys)`?
- I wasn't quite clear -- should this be correctly inferring ivs as the union of the rhs terms?
- Can you add a check that interpolation of an external variable/species/parameter works on the lhs of an observable?

docs/src/catalyst_functionality/dsl_description.md
TorkelE: It should work, but I will add a test that it does.

docs/src/catalyst_functionality/dsl_description.md
TorkelE: This suggestion is identical to the original text.

docs/src/catalyst_functionality/dsl_description.md
TorkelE: done

TorkelE:
I will add additional tests. Yes, the ivs of the observables is the same as the ivs of the right hand stuff (the union, using the order with which they appear in the actual reaction network.

docs/src/catalyst_functionality/dsl_description.md
isaacsas: Yeah, I had a comment originally but removed it, but forgot to remove the "comment" on Github.

test/dsl/dsl_options.jl
TorkelE: added now

src/reactionsystem.jl
TorkelE: This has been moved now

PR #772, opened 2024-01-26
mention constant species

ranocha:
I found a way to set constant species in the tests - and also in an online tutorial. Thus, I think it's reasonable/useful to include it in the docs so that it becomes public and stable API.

### DIFF
--- a/docs/src/catalyst_functionality/dsl_description.md
+++ b/docs/src/catalyst_functionality/dsl_description.md
@@ -285,7 +285,7 @@ where here `t` always denotes Catalyst's time variable. Please note that many
 user-defined functions can be called directly, but others will require
 registration with Symbolics.jl ([see the faq](@ref user_functions)).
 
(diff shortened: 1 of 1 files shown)


docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -412,6 +412,74 @@ sol = solve(oprob)
 plot(sol)
 ```
 
+## Constant/fixed species
+It is possible to fix the concentration of a species in a reaction. Without fixing
```
isaacsas, line 416:
```suggestion
It is possible to fix the amount of a species in a reaction. Without fixing
```
Since we support either concentration or "number of" units.

isaacsas:
@TorkelE this looks good to me. Did I set the correct branch you are using for v14? If so, we can merge this there.

TorkelE:
Looks good to me, yes, this is the correct branch.

PR #784, opened 2024-02-13
Handle test errors from MTKv9 update (including `ReactionSystem` completeness)

TorkelE:
For MTK v9 all inputs to various systems have to be complete. Which means we must either add `complete(::ReactionSystem)` wherever these are required, or make them complete by default. Since this can be sorted out on the v14 branch even before  MTK v9 I figured we could do it in a separate branch.

Currently, I make `ReactionSystem`s complete by default. which can be changed by a DSL option/optional argument:
```julia
@reaction_network begin
    @incompelte
    (p,d), 0 <--> X
end
ReactionSystem(rxs, t; complete = false)
```
However, we can flip the default really easily (while also renaming `@incompelte` to `@compelte`

Deciding which to go for is still a problem. I prefer this, since it only introduces the concept of completeness when it is actually required. I think it is useful to follow how MTK does stuff, but for something which has a major impact on Catalyst (and is not just notation), I think we should do what works best, independently of MTK (whatever the case, I hope to add a "differences to MTK" section somewhere in the docs).


docs/pages.jl
```diff
@@ -17,7 +17,7 @@ pages = Any["Home" => "index.md",
                                            "catalyst_applications/nonlinear_solve.md",
                                            "catalyst_applications/bifurcation_diagrams.md"],
             "Inverse Problems" => Any["inverse_problems/optimization_ode_param_fitting.md",
-                                        "inverse_problems/petab_ode_param_fitting.md",
+                                        #"inverse_problems/petab_ode_param_fitting.md",
```
TorkelE, line 20: PEtab is not MTK v9 updated yet, so commented out here (permitting docs to build).

docs/src/catalyst_applications/advanced_simulations.md
```diff
@@ -292,8 +292,9 @@ plot(sol)
 
 ## Scaling the noise magnitude in the chemical Langevin equations
 When using the CLE to generate SDEs from a CRN, it can sometimes be desirable to
-scale the magnitude of the noise terms. This can be done by introducing a *noise
-scaling parameter*. First, we simulate a simple two-state CRN model using the
+scale the magnitude of the noise terms. Here, each reaction of the system generates a separate noise term in the CLE. If you require identical scaling for all reactions, the `@default_noise_scaling` option can be used. Else, you can supply a `noise_scaling` metadata for each individual reaction, describing how to scale the noise for that reaction.
+
+We begin with considering the first approach. First, we simulate a simple two-state CRN model using the
```
TorkelE, line 297: Additions in this file is because this PR builds upon the noise scaling one, so showing its changes as changes here. Will disappear once that gets merged.

docs/src/catalyst_functionality/compositional_modeling.md
```diff
@@ -5,16 +5,47 @@ can construct the earlier repressilator model by composing together three
...
+We can test whether a system is complete using the `ModelingToolkit.iscomplete` function:
+```@example ex0
+ModelingToolkit.iscomplete(degradation_component)
+```
+To make a incomplete system complete, we can use the `complete` function:
+```@example ex0
+degradation_component_complete = complete(degradation_component)
+ModelingToolkit.iscomplete(degradation_component_complete)
+```
+
```
TorkelE, line 36: If we change completeness around, this will have to change. I also have an in-depth introduction in a local branch that goes through stuff like this. Basically, once all of this is over things like this should/will get rewritten anyway.

docs/src/catalyst_functionality/compositional_modeling.md
```diff
@@ -5,16 +5,47 @@ can construct the earlier repressilator model by composing together three
...
 ## Compositional modeling tooling
 Catalyst supports two ModelingToolkit interfaces for composing multiple
 [`ReactionSystem`](@ref)s together into a full model. The first mechanism for
 extending a system is the `extend` command
 ```@example ex1
 using Catalyst
 basern = @reaction_network rn1 begin
+  @incomplete
```
TorkelE, line 44: Currently, the `@incompelte` option marks that a reaction system (created via the DSL) is _not_ complete when created. Can change how to manage this later.

src/Catalyst.jl
```diff
@@ -48,8 +48,8 @@ end
 function default_t()
     return ModelingToolkit.t_nounits
 end
-const DEFAULT_t = default_t()
-const DEFAULT_IV_SYM = Symbol(DEFAULT_t)
+const DEFAULT_IV = default_t()
+const DEFAULT_IV_SYM = Symbol(DEFAULT_IV)
```
TorkelE, line 52: Wrong name was used before.

src/networkapi.jl
```diff
@@ -1624,7 +1624,7 @@ function validate(rs::ReactionSystem, info::String = "")
             rxunits *= get_unit(sub)^rx.substoich[i]
         end
 
-        if rxunits != rateunits
+        if !isequal(rxunits, rateunits)
```
TorkelE, line 1627: Caused error with MTK v9, this seemed to fix it. Guess that either of these could suddenly be `Num`s or stuff like that.

src/reaction_network.jl
```diff
@@ -123,7 +123,7 @@ const forbidden_variables_error = let
 end
 
 # Declares the keys used for various options.
-const option_keys = (:species, :parameters, :variables, :ivs, :compounds, :observables)
+const option_keys = (:species, :parameters, :variables, :ivs, :compounds, :observables, :default_noise_scaling, :incomplete)
```
TorkelE, line 126: Stuff that follows are either the noise scaling (because this builds on that PR), or the compelteness.

test/dsl/dsl_model_construction.jl
```diff
@@ -10,8 +10,9 @@ t = default_t()
 using StableRNGs
 rng = StableRNG(12345)
 
-# Fetch test networks.
+# Fetch test networks and functions.
 include("../test_networks.jl")
+include("../test_functions.jl")
```
TorkelE, line 15: In the end I could not pu these "test helper functions" in the runtests.jl file. I created a separate file, which is fetched whenever it is needed (like here).

test/dsl/dsl_model_construction.jl
```diff
@@ -122,17 +123,16 @@ let
...
 
     for networks in identical_networks_1
-        f1 = ODEFunction(convert(ODESystem, networks[1]), jac = true)
-        f2 = ODEFunction(convert(ODESystem, networks[2]), jac = true)
-        g1 = SDEFunction(convert(SDESystem, networks[1]))
-        g2 = SDEFunction(convert(SDESystem, networks[2]))
         for factor in [1e-2, 1e-1, 1e0, 1e1, 1e2, 1e3]
-            u0 = factor * rand(rng, length(get_unknowns(networks[1])))
-            p = factor * rand(rng, length(get_ps(networks[1])))
+            u0_1 = rnd_u0(networks[1], rng; factor)
```
TorkelE, line 127: Uses a helper function that creates a random u0 (another one does ps) vector (and as a symbolic map).

test/dsl/dsl_model_construction.jl
```diff
@@ -122,17 +123,16 @@ let
...
+            p_1 = rnd_ps(networks[1], rng; factor)
+            u0_2 = Pair.(unknowns(networks[2]), last.(u0_1))
+            p_2 = Pair.(parameters(networks[2]), last.(p_1))
             t = rand(rng)
-            @test all(abs.(f1(u0, p, t) .≈ f2(u0, p, t)))
-            @test all(abs.(f1.jac(u0, p, t) .≈ f2.jac(u0, p, t)))
-            @test all(abs.(g1(u0, p, t) .≈ g2(u0, p, t)))
+            
+            @test f_eval(networks[1], u0_1, p_1, t) ≈ f_eval(networks[2], u0_2, p_2, t)
```
TorkelE, line 133: Uses the new `f_eval` function to evaluate the ODESystem `f` function.

TorkelE: Alsmot always, like here, `f_eval` is only called once for each network, do no use using a in-place version. Because most networks have different `du` vector sizes, reusing it across different systems is hard.

test/dsl/dsl_options.jl
```diff
@@ -629,4 +629,24 @@ let
...
+        @incomplete
+        (p,d), 0 <--> X
+    end
+
+    # Check that the correct compeltness is achived.
+    @test ModelingToolkit.iscomplete(rn1)
+    @test !ModelingToolkit.iscomplete(rn2)
+    rn2 = complete(rn2)
+    @test ModelingToolkit.iscomplete(rn2)
+end
```
TorkelE, line 650: Test for the new `@incomplete` functionality.

test/extensions/bifurcation_kit.jl
```diff
@@ -9,7 +9,7 @@ rng = StableRNG(12345)
 
 # Brusselator extended with conserved species.
 # Runs full computation, checks values corresponds to known values.
-# Checks that teh correct bifurcation point is found at the correct position.
+# Checks that the correct bifurcation point is found at the correct position.
```
TorkelE, line 12: I did a search through the docs for any occurrences of "teh"...

test/extensions/homotopy_continuation.jl
```diff
@@ -2,14 +2,17 @@
...
+
 ### Run Tests ###
 
 # Tests for network without conservation laws.
 # Tests for Symbol parameter input.
 # Tests for Symbolics initial condition input.
 # Tests for different types (Symbol/Symbolics) for parameters and initial conditions.
 # Tests that attempts to find steady states of system with conservation laws, while u0 is not provided, gives an error.
-let 
+@test_broken let # Requires https://github.com/SciML/ModelingToolkit.jl/issues/2566 to be fixed
```
TorkelE, line 15: One of the tests that are currently broken. Basically, we cannot do homotopy continuation on systems with conservation laws until that issue gets sorted out.

test/extensions/homotopy_continuation.jl
```diff
@@ -50,7 +53,7 @@ end
 # Tests correctness in presence of default values.
 # Tests where some default values are overwritten with other values.
 # Tests where input ps/u0 are tuples with mixed types.
-let    
+@test_broken let # Requires https://github.com/SciML/ModelingToolkit.jl/issues/2566 to be fixed
```
TorkelE, line 56: Same problem as the one above.

test/miscellaneous_tests/nonlinear_solve.jl
```diff
@@ -20,67 +23,68 @@ let
...
+    sol2 = solve(nlprob, DynamicSS(Rosenbrock23()); abstol=1e-12, reltol=1e-12)
     
     # Tests solutions are correct.
-    @test isapprox(sol1[1], p[1] / p[2]; atol=1e-10)
-    @test isapprox(sol1[2]^3 / factorial(3), p[3] / p[4]; atol=1e-10)
-    @test isapprox(sol1[3], (p[5] * p[2]) / (p[6] * p[1]); atol=1e-10)
-    @test isapprox(sol2[1], p[1] / p[2]; atol=1e-10)
-    @test isapprox(sol2[2]^3 / factorial(3), p[3] / p[4]; atol=1e-10)
-    @test isapprox(sol2[3], (p[5] * p[2]) / (p[6] * p[1]); atol=1e-10)
+    @test sol1[:X1] ≈ nlprob.ps[:k1] / nlprob.ps[:k2] atol=1e-10
```
TorkelE, line 35: Uses symbolic (instead of integer) indexing.

Loads of test where also written before I was familiar with nice ways to use `≈`, `@test`, and `atol`, so I rewrote quite a few tests using this while I was at it.

test/miscellaneous_tests/nonlinear_solve.jl
```diff
@@ -20,67 +23,68 @@ let
...
+    @test nlprob_eval_2.f(nlprob_eval_2.u0, nlprob_eval_2.p)[1] ≈ 0.0 atol=1e-10
+    @test nf_manual(sol1, last.(ps)) ≈ 0.0 atol=1e-10
+    @test nf_manual(sol2, last.(ps)) ≈ 0.0 atol=1e-10
 end
 
 # Checks for system with conservation laws.
 # Checks using interfacing with output solution.
 let 
+    # Conservation laws currently broken (you get stuck in an infinite loop in MTK or something).
+    return (@test_broken false)
```
TorkelE, line 81: Another conservation law issue, this one only appears when a `XProblem` is created (because these creates a `MTKParameters` object, which is the real culprit, not sure what the problem in these are though).

test/miscellaneous_tests/symbolic_stoichiometry.jl
```diff
@@ -3,7 +3,8 @@ t = default_t()
 
 ### Base Tests ###
 
-let
+# Probably requires us to permit Int64 parameters for this to work. 
+@test_broken let
```
TorkelE, line 7: Both symbolic stoichiometry tests are broken. These are the ones that maybe you should have a look at, since there is quite a lot going on. I am happy to write new ones, but afraid that if I start changing in these I will inadvertently change something that was intended to be there and I did not realise.

test/miscellaneous_tests/symbolic_stoichiometry.jl
```diff
@@ -188,7 +189,10 @@ end
 
 ### Simple Solving Tests on SIR Model ###
 
-let
+# Gives an error. I can fix so there is not, but unsure what part of the code is what we actually are
+# testing for, and what we are not. Sam, if you have any opinions on what to preserve in this test or not,
+# tell me before I try to rewrite. 
+@test_broken let
```
TorkelE, line 195: Again, this is also broken,

test/miscellaneous_tests/units.jl
```diff
@@ -6,6 +6,10 @@ const MT = ModelingToolkit
 ### Run Tests ###
 
 let
+    # Units not working right now. Not familiar with Unitful, someone else would have to have a look.
+    # Probably we should just figure DynamicQuantities out directly and deprecate all of this.
+    return (@test_broken false)
+
```
TorkelE, line 12: Unitful is broken, I presume we are just going to abandon this and write new tests depending on DynamicQuantitites?

test/model_simulation/make_jacobian.jl
```diff
@@ -26,18 +29,20 @@ let
...
     for factor in [1e-2, 1e-1, 1e0, 1e1, 1e2], repeat in 1:10
-        u = factor * rand(rng, 3)
-        p = factor * rand(rng, 3)
-        local osys = convert(ODESystem, jacobian_network_2)
-        local test_jac = ODEFunction(osys, jac = true).jac(u, p, 0.0)
-        local real_jac = [-1-2 * p[3] * u[2] * u[1] -p[3]*u[1]*u[1] 1.0;
-                          -2*p[3]*u[2]*u[1] -1-p[3] * u[1] * u[1] 1;
-                          2*p[3]*u[2]*u[1] p[3]*u[1]*u[1] -1.0]
-        @test all(abs.(test_jac .- real_jac) .< 1e-9)
+        u = Dict(rnd_u0(jacobian_network_2, rng; factor))
```
TorkelE, line 35: Uses Dict to simplify evaluating `real_jac` in 3 lines from here.

test/model_simulation/simulate_ODEs.jl
```diff
@@ -103,26 +104,32 @@ let
...
-            prob1 = ODEProblem(networks[1], u0, (0.0, 10000.0), p)
-            sol1 = solve(prob1, Rosenbrock23(), saveat = 1.0)
-            prob2 = ODEProblem(networks[2], u0, (0.0, 10000.0), p)
-            sol2 = solve(prob2, Rosenbrock23(), saveat = 1.0)
-            @test all(abs.(hcat((sol1.u .- sol2.u)...)) .< 1e-7)
+            u0_1 = rnd_u0(networks[1], rng; factor)
+            p_1 = rnd_ps(networks[1], rng; factor)
+            (i == 3) && (p_1 =rnd_ps_Int64(networks[1], rng))
+            u0_2 = last.(u0_1)
+            p_2 = last.(p_1)
```
TorkelE, line 111: This depends on `ReactionSystem` not doing some weird reordering (which we do not want to promise in the future). However, here we loop through quite a few networks and just generate parameters. If we'd want to make order not matter here, we'd have to write out the vectors for each case manually, and handle them separately. I can see how to do this, but non-trivial amount of work. I'd rather do that at some later overhaul of the tests than right now (when we have tons of stuff to fix).

test/reactionsystem_structure/higher_order_reactions.jl
```diff
@@ -51,7 +51,10 @@ let
 end
 
 # Tests that the discrete jump systems are equal.
-let
+# Currently fails because binomial only takes Int input (and X is Float64).
+# I see several solutions, but depends on whether we allow species to be specified as Int64.
+# I have marked this one as broken for now.
+@test_broken let
```
TorkelE, line 57: This test will remain broken until we decide how to handle non-Float64 definitions of species. Which will be messy, because if we do we have to rewrite the `@compound` stuff to permit that as well.

Once sorted out, we can fix these tests using the best solution (which depends on whether that is allowed or not).

test/reactionsystem_structure/reactionsystem.jl
```diff
@@ -143,66 +146,29 @@ end
...
-let
-    p = rand(rng, length(k))
-    u = rand(rng, length(k))
-    t = 0.0
-    G = [p p p p]' .* sdenoise(u, p, t)
-    sdesys_noise_scaling = convert(SDESystem, rs; noise_scaling = k)
-    sf = SDEFunction{false}(sdesys_noise_scaling, unknowns(rs),
-                            parameters(sdesys_noise_scaling))
-    G2 = sf.g(u, p, t)
-    @test norm(G - G2) < 100 * eps()
```
TorkelE, line 205: These were moved in the noise scaling PR to the SDe simulation file (so not removed).

test/reactionsystem_structure/reactionsystem.jl
```diff
@@ -282,13 +248,13 @@ let
     maj = MassActionJump(symmaj.param_mapper(pars), symmaj.reactant_stoch, symmaj.net_stoch,
                          symmaj.param_mapper, scale_rates = false)
     for i in midxs
-        @test abs(jumps[i].scaled_rates - maj.scaled_rates[i]) < 100 * eps()
+        @test_broken abs(jumps[i].scaled_rates - maj.scaled_rates[i]) < 100 * eps()
```
TorkelE, line 251: some of these new Jump-related problems, I think you have some idea what they are about?

test/test_functions.jl
```diff
@@ -0,0 +1,44 @@
+# Various functions that are useful for running the tests, and used across several test sets.
+
+
+# Initial Condition/Parameter Generators ###
```
TorkelE, line 4: These are hopefully self-explanatory? The int ones are needed (or useful) in certain cases (e.g. I think in some Hill function cases they are good to have).

test/miscellaneous_tests/api.jl
```diff
@@ -513,7 +521,9 @@ let
 end
 
 # Test conservation law elimination.
+# Conservation laws currently broken (you get stuck in an infinite loop in MTK or something).
 let
+    return (@test_broken false)
```
TorkelE, line 526: No conservation law stuff now (or at least none that creates `XProblem`s, until one more thing gets fixed).

test/reactionsystem_structure/reactionsystem.jl
```diff
@@ -143,66 +146,29 @@ end
...
-    p = rand(rng, length(k))
-    u = rand(rng, length(k))
-    t = 0.0
-    G = [p p p p]' .* sdenoise(u, p, t)
-    sdesys_noise_scaling = convert(SDESystem, rs; noise_scaling = k)
-    sf = SDEFunction{false}(sdesys_noise_scaling, unknowns(rs),
-                            parameters(sdesys_noise_scaling))
-    G2 = sf.g(u, p, t)
-    @test norm(G - G2) < 100 * eps()
+    @test_broken let # The next line throws an error.
```
TorkelE, line 168: I don't really understand this set of tests (or not exactly why some stuff is done like it is). If you tell me what you are OK/Not ok with changing here I can fix, but until then I'd rather not start poking in it.

HISTORY.md
```diff
@@ -22,8 +22,9 @@ rn = @reaction_network begin
     @parameters η
     k, 2X --> X2, [noise_scaling=η]
 end
-getnoisescaling(rn)
+get_noise_scaling(rn)
```
TorkelE, line 25: This and other related changes is due to this one building on the noise scaling PR

test/dsl/dsl_basics.jl
```diff
@@ -63,63 +67,58 @@ end
...
-    AAA = A^2 + B
-    rn = @reaction_network rn begin
-        @parameters k
-        @species A(t) B(t) C(t) D(t)
-        k*$AAA, C --> D
-    end
-    rn2 = ReactionSystem([Reaction(k*AAA, [C], [D])], t; name=:rn)
-    @test rn == rn2
+    # Declares parameters and species used across the test.
+    @parameters α k k1 k2
```
TorkelE, line 73: Same as stuff on the lhs, Git seems to mark some stuff as new when it shouldn't.

test/dsl/dsl_basics.jl
```diff
@@ -272,7 +282,7 @@ let
     @test rn == mixedsys
 end
 
-# Test variables that appear only in rates and aren't ps
+# Test that variables that appear only in rates and aren't ps
```
TorkelE, line 285: Because I was going through the files from start to finish, I straightened out some of the writing while I was at it.

test/dsl/custom_functions.jl
```diff
@@ -7,6 +6,9 @@ t = default_t()
 using StableRNGs
 rng = StableRNG(12345)
 
+# Fetch test functions.
+include("../test_functions.jl")
```
TorkelE, line 10: There was no way to declare the helper function in the runtests.jl file, so I put them in a separate script.

test/dsl/dsl_basics.jl
```diff
@@ -2,11 +2,15 @@
 
 ### Fetch Packages and Set Global Variables ###
 
+# Fetch packages.
 using Catalyst, ModelingToolkit
+
+# Set creates the `t` independent variable.
```
TorkelE, line 8: We already merged the `t = default_t()` change. We can always revert later, this is relatively easy to do.

test/dsl/dsl_options.jl
TorkelE: (need changing after final decision on completeness)

test/miscellaneous_tests/api.jl
```diff
@@ -100,6 +106,17 @@ let
     @test pmat == prodstoichmat(rnmat) == Matrix(prodstoichmat(rnmat, sparse = true))
 end
 
+# Tests `reactionparamsmap`, `reactionrates`, and `symmap_to_varmap` getters.
```
TorkelE, line 109: Moved up from a different place. I haven't removed any tests nor written any actual new test blocks. However, for some files the order was rather jumbled. As part of the make I fixed this. Hopefully we won't have to deal with this again now.

test/miscellaneous_tests/api.jl
```diff
@@ -606,20 +625,6 @@ let
     test_stoich(Int, rn2)
 end
 
-### Miscelenesous Tests ###
```
TorkelE, line 609: As mentioned, moved up to a better place in the file.

test/miscellaneous_tests/compound_macro.jl
```diff
@@ -60,36 +67,7 @@ let
     @test iscompound(rn.NH3_5)
 end
 
-### Test Various Independent Variables ###
```
TorkelE, line 63: Again, moved down. You should see the pattern now. Realise this is a bit annoying to read, was intending to do this at the beginning of the summer (when everything else was sorted out). But as we wanted to go through and remake all the tests, I figured it was best to do this now to avoid two major reamkes of the test files.

test/model_simulation/simulate_ODEs.jl
```diff
@@ -37,25 +40,27 @@ let
...
-
-identical_networks_1 = Vector{Pair}()
-
+# Compares simulations generated through Catalyst with those generated by manually created functions.
 let
+    # Manually declares ODEs to compare Catalyst-generated simulations to.
+    catalyst_networks = []
+    manual_networks = []
+    u0_syms = []
+    ps_syms = []
```
TorkelE, line 62: Many cases have had similar reameks now. One  vector to store one set of networks, one for another. Next I provide the list of the order the parameters and variables appears in in the manually created networks. These can be used to generate the correct u0/ps vectors from the randomly generated maps.

test/model_simulation/simulate_ODEs.jl
```diff
@@ -99,64 +116,88 @@ let
...
     end
-    push!(identical_networks_1, reaction_networks_weird[2] => real_functions_5)
+    push!(catalyst_networks, reaction_networks_weird[2])
+    push!(manual_networks, real_functions_5)
+    push!(u0_syms, [:X, :Y, :Z])
+    push!(ps_syms, [:k1, :k2, :k3, :k4])
 
-    for (i, networks) in enumerate(identical_networks_1)
+    for (rn_catalyst, rn_manual, u0_sym, ps_sym) in zip(catalyst_networks, manual_networks, u0_syms, ps_syms)
         for factor in [1e-2, 1e-1, 1e0, 1e1]
```
TorkelE, line 125: I considered changing some of these when we re-run tests (and I did some stochastic ones). But practically, we are not talking about any really time, so I left them as they were.

test/model_simulation/simulate_ODEs.jl
```diff
@@ -99,64 +116,88 @@ let
...
+            dt = zeros(length(u0_1))
+            rn_manual(dt, u0_2, ps_2, 0.0)
+            @test dt ≈ f_eval(rn_catalyst, u0_1, ps_1, 0.0)
+
+            # Compares that simulations are identical. 
+            oprob_1 = ODEProblem(rn_catalyst, u0_1, (0.0, 100.0), ps_1)
+            oprob_2 = ODEProblem(rn_manual, u0_2, (0.0, 100.0), ps_2)
+            sol1 = solve(oprob_1, Rosenbrock23())
+            sol2 = solve(oprob_2, Rosenbrock23())
+            @test sol1[u0_sym] ≈ sol2.u
```
TorkelE, line 143: Indexes the solution simualted with Catalyst with the symbol vector showing the order the species appears in in the manual functions.

test/model_simulation/simulate_SDEs.jl
```diff
@@ -93,81 +103,197 @@ let
...
+            @test duW ≈ g_eval(rn_catalyst, u0_1, ps_1, 0.0)
+
+            # Compares simulation with identical seed. 
+            # Cannot test the third network (requires very fiddly parameters for CLE to be defined).
+            if nameof(rn_catalyst) != :rnc9
+                sprob_1 = SDEProblem(rn_catalyst, u0_1, (0.0, 1.0), ps_1)
+                sprob_2 = SDEProblem(rn_manual.f, rn_manual.g, u0_2, (0.0, 1.0), ps_2; noise_rate_prototype = rn_manual.nrp)
+                sol1 = solve(sprob_1, ImplicitEM(); seed = 1234)
+                sol2 = solve(sprob_2, ImplicitEM(); seed = 1234)
+                @test sol1[u0_sym] ≈ sol2.u
```
TorkelE, line 136: Testing bith the functions and the simulatiosn doesn't really take more times. Figured it was good to be rigorous in at least one case and check both.

test/model_simulation/simulate_SDEs.jl
```diff
@@ -93,81 +103,197 @@ let
...
+        ps = rnd_ps(rn, rng)
+        sprob = SDEProblem(rn, u0, (0.0, 1.0), ps)
+        init(sprob, ImplicitEM())
+    end
+end
+
 ### Noise Scaling ###
 
-# Tests with a single noise scaling parameter.
+# Compares noise scaling with manually declared noise function.
```
TorkelE, line 155: This one noise scaling test was added here and not in the noise scaling PR. Because the required helper functions were created in this PR, this test did not come until this one.

test/model_simulation/simulate_jumps.jl
```diff
@@ -91,56 +103,60 @@ let
...
-            sol1 = solve(prob1, SSAStepper())
-            prob2 = JumpProblem(DiscreteProblem(u0, (0.0, 1000.0), p), Direct(),
-                                networks[2]...)
-            sol2 = solve(prob2, SSAStepper())
-            for i in 1:length(u0)
-                vals1 = getindex.(sol1.u, i)
-                vals2 = getindex.(sol1.u, i)
-                (mean(vals2) > 0.001) && @test 0.8 < mean(vals1) / mean(vals2) < 1.25
-                (std(vals2) > 0.001) && @test 0.8 < std(vals1) / std(vals2) < 1.25
+            if nameof(rn_catalyst) == :rnh7
```
TorkelE, line 126: Still no idea what happens with this one case. I have probably spent 3 hours trying to figure it out. Really got no idea at this stage.

test/reactionsystem_structure/reactionsystem.jl
```diff
@@ -1,15 +1,24 @@
 ### Fetch Packages and Set Global Variables ###
 
-# Fetch pakcages.
-using Catalyst, LinearAlgebra, JumpProcesses, Test, OrdinaryDiffEq, StochasticDiffEq
-t = default_t()
+# Fetch packages.
```
TorkelE, line 3: There is a couple of really long tests here that I didn't want to touch. They pass, but at some point it would be good if we could look at them together to decide how they can be remade.

test/spatial_reaction_systems/lattice_reaction_systems_ODEs.jl
```diff
@@ -12,6 +11,9 @@ include("../spatial_test_networks.jl")
 using StableRNGs
 rng = StableRNG(12345)
 
+# Sets defaults
```
TorkelE, line 14: The spatial tests are already good. Further changes here will wait until that  part is done (the internal branches I have already do some modifications)

test/test_functions.jl
```diff
@@ -0,0 +1,54 @@
...
+    issetequal(keys(syms_dict), syms) || error("Map symbols ($(keys(syms_dict))) and symbol vector symbols ($(syms)) do not match.")
+    return [syms_dict[sym] for sym in syms]
+end
+
+### System function evaluation ###
+
+# Evaluates the the drift function of the ODE corresponding to a reaction network.
+function f_eval(rs::ReactionSystem, u, p, τ)
+    prob = ODEProblem(rs, u, (0.0, 0.0), p)
+    return prob.f(prob.u0, prob.p, τ)
```
TorkelE, line 41: Technically, the order can affect stuff here. But until Chris decides that this is something he actually want to support there is not much to do. Generally, for variables, this is really not much of a problem, probably won't be in the future, but hopefully some improvement is made soon-ish.

test/visualization/latexify.jl
```diff
@@ -1,11 +1,15 @@
 #! format: off
 
-### Fetch Packages and Reaction Networks ###
-using Catalyst, Latexify
+### Preparations ###
+
+# Fetch packages.
+using Catalyst, Latexify, Test
```
TorkelE, line 6: I don't think `using Test` is required when `@safetestset` is used. But I figured we should be consistent whether to use it or not. I decided to just put it in everywhere.
(simplifies when copying tests to another file to run them, for a starter)

test/visualization/latexify.jl
```diff
@@ -131,50 +138,6 @@ let
     ", "\r\n"=>"\n")
 end
 
-# Test empty system.
```
TorkelE, line 134: Moving things around again

test/extensions/homotopy_continuation.jl
```diff
@@ -18,16 +24,18 @@ let
     ps = [k1 => 1.0, k2 => 2.0, k3 => 2.0, k4 => 2.0]
     u0 = [:X1 => 2.0, :X2 => 2.0, :X3 => 2.0, :X2_2X3 => 2.0]
 
-    hc_ss = hc_steady_states(rs, ps; u0=u0, show_progress=false)[1]
-    f = ODEFunction(convert(ODESystem, rs))
-    @test f(hc_ss, last.(ps), 0.0)[1] == 0.0
+    # Computes the single steady state, checks that when given to the ODE rhs, all are evaluated to 0.
+    hc_ss = hc_steady_states(rs, ps; u0=u0, show_progress=false)
+    hc_ss = Pair.(unknowns(rs), hc_ss[1])
```
TorkelE, line 29: hc_steadt_states guarantees that the order is the same as in species(rs). At some point this should have SII support (but would ideally require additional support in SII/MTK/SciMLBase)

docs/src/catalyst_functionality/compositional_modeling.md
```diff
@@ -5,16 +5,44 @@ can construct the earlier repressilator model by composing together three
...
+We can test whether a system is complete using the `ModelingToolkit.iscomplete` function:
+```@example ex0
+ModelingToolkit.iscomplete(degradation_component)
+```
+To make a incomplete system complete, we can use the `complete` function:
+```@example ex0
+degradation_component_complete = complete(degradation_component)
+ModelingToolkit.iscomplete(degradation_component_complete)
+```
+
```
TorkelE, line 35: Post v14 I would like to revisit this an write an improved description. I already have some write-up in local branches. For now, I figured it would be best to simple describe what is required. Next, we can look at the best way to actually document everything.

docs/src/catalyst_functionality/compositional_modeling.md
```diff
@@ -5,16 +5,44 @@ can construct the earlier repressilator model by composing together three
...
 ## Compositional modeling tooling
 Catalyst supports two ModelingToolkit interfaces for composing multiple
 [`ReactionSystem`](@ref)s together into a full model. The first mechanism for
 extending a system is the `extend` command
 ```@example ex1
 using Catalyst
-basern = @reaction_network rn1 begin
+basern = @network_component rn1 begin
```
TorkelE, line 42: For now, I went with `@network_component` as the non-complete version of the macro.

docs/src/catalyst_functionality/programmatic_CRN_construction.md
```diff
@@ -61,6 +61,9 @@ system to be the same as the name of the variable storing the system.
 Alternatively, one can use the `name = :repressilator` keyword argument to the
 `ReactionSystem` constructor.
 
+!!! warn
+       All `ReactionSystem`s created programmatically (i.e. by calling `ReactionSystem` with some input, rather than using `@reaction_network`) are created as *incomplete*. To simulate them, they must first be made *complete*. This can be done using the `complete` function, i.e. by calling `repressilator = complete(repressilator)`. An expanded description on *completeness* can be found [here](@ref completeness_note).
+
```
TorkelE, line 66: As mentioned earlier, this PR does not make the most extensive rewrite of the docs. But I do add minimally what would be required for someone to be able to run things given the new changes.

src/expression_utils.jl
```diff
@@ -94,18 +94,6 @@ function is_escaped_expr(expr)
...
-# (e.g. :([a=1.0, b=2.0])) to a vector where the assignment instead uses symbols and pairs 
-# (e.g. :([a=>1.0, b=>2.0])). Used to e.g. convert default reaction metadata to a form that can be 
-# evaluated as actual code.
-function expr_equal_vector_to_pairs(expr_vec)
-    pair_vector = :([])
-    foreach(arg -> push!(pair_vector.args, arg.args[1] => arg.args[2]), expr_vec.args) 
-    return pair_vector
-end
-
 ### Old Stuff ###
```
TorkelE, line 97: Turns out this one was not really needed.

src/reaction_network.jl
```diff
@@ -1,45 +1,4 @@
...
-macro reaction_network(name::Symbol, ex::Expr, parameters...)
-    error(deprication_message)
-end
-macro reaction_network(name::Expr, ex::Expr, parameters...)
-    error(deprication_message)
-end
-macro reaction_network(ex::Expr, parameters...)
-    error(deprication_message)
-end
-
```
TorkelE, line 42: So, we could keep these around. The reason I had them removed is that now when we have both `@reaction_network` and `@network_component`, it seems best that the former calls the latter (but with a `complete` call), and the latter does the actual stuff. When I tried to implement, I had a problem that some of these were more specific, and where called instead. Hence I removed them.

They have been around since the change, so maybe it makes sense to remove them anyway (although might be useful if someone learns from an old tutorial).

When we figured out the new macro implementation, I don''t mind adding these back in, if possible.

src/reaction_network.jl
```diff
@@ -186,36 +145,74 @@ emptyrn = @reaction_network empty
 # an empty network with random generated name
 emptyrn = @reaction_network
 ```
+
+ReactionSystems generated through `@reaction_network` are complete.
 """
+# macro reaction_network(args...)
+#     return :(complete(@network_component $(args... )))
+# end
```
TorkelE, line 153: Ideally, we'd only want to implement something like this macro, where `@reaction_network` calls `@reaction_component` (which does the actual work).

Unfortunately, this implementation does not work with interpolation. E.g.
```julia
name = :NAME
rn = @reaction_network $NAM begin
    d, X --> 0
end

k = 1
rn = @reaction_network begin
    d + 1 + $k, X --> 0
end
``` 
gives `name` and `k` `not defined`, respectively. Figured I'd let you have a look, since you know interpolation well and might have an easy solution.

src/reaction_network.jl
```diff
@@ -186,36 +145,74 @@ emptyrn = @reaction_network empty
...
 macro reaction_network(name::Symbol, ex::Expr)
-    make_reaction_system(MacroTools.striplines(ex); name = :($(QuoteNode(name))))
+    :(complete($(make_reaction_system(MacroTools.striplines(ex); name = :($(QuoteNode(name)))))))
```
TorkelE, line 155: Here I just reimplements the two macros identically, until we can figure out how to use one set of implementations.

src/reaction_network.jl
```diff
@@ -480,7 +477,7 @@ end
 
 # When compound species are declared using the "@compound begin ... end" option, get a list of the compound species, and also the expression that crates them.
 function read_compound_options(opts)
-    # If the compound option is used retrive a list of compound species (need to be added to teh reaction system's species), and the option that creates them (used to declare them as compounds at the end).
+    # If the compound option is used retrive a list of compound species (need to be added to the reaction system's species), and the option that creates them (used to declare them as compounds at the end).
```
TorkelE, line 480: I did a control+F through the docs to find all instances of "teh" so that I could fix them

src/reaction_network.jl
```diff
@@ -810,7 +807,7 @@ function check_default_noise_scaling!(default_reaction_metadata, options)
         if (length(options[:default_noise_scaling].args) != 3) # Becasue of how expressions are, 3 is used.
             error("@default_noise_scaling should only have a single input, this appears not to be the case: \"$(options[:default_noise_scaling])\"")
         end
-        push!(default_reaction_metadata.args, :(noise_scaling=$(options[:default_noise_scaling].args[3])))
+        push!(default_reaction_metadata.args, :(:noise_scaling => $(options[:default_noise_scaling].args[3])))
```
TorkelE, line 810: This change here saves us calling that function in exprs_utils.jl

src/reactionsystem.jl
```diff
@@ -205,7 +205,7 @@ end
 # Checks if a metadata input has an entry :only_use_rate => true
 function metadata_only_use_rate_check(metadata)
     only_use_rate_idx = findfirst(:only_use_rate == entry[1] for entry in metadata)
-    isnothing(only_use_rate_idx) && return true
+    isnothing(only_use_rate_idx) && return false
```
TorkelE, line 208: fixing an error

src/reactionsystem.jl
```diff
@@ -544,7 +544,7 @@ struct ReactionSystem{V <: NetworkProperties} <:
     # inner constructor is considered private and may change between non-breaking releases.
     function ReactionSystem(eqs, rxs, iv, sivs, unknowns, spcs, ps, var_to_name, observed,
                             name, systems, defaults, connection_type, nps, cls, cevs, devs,
-                            metadata = nothing, complete::Bool = false; checks::Bool = true)
+                            metadata, complete; checks::Bool = true)
```
TorkelE, line 547: Not fully sure how you want this function, but this is basically the base constructor from the args, so did like I understood it and our conversation from Wednesday.

src/reactionsystem.jl
```diff
@@ -1469,6 +1471,8 @@ function spatial_convert_err(rs::ReactionSystem, systype)
     isspatial(rs) && error("Conversion to $systype is not supported for spatial networks.")
 end
 
+COMPLETENESS_ERROR = "A ReactionSystem must be complete before it can be converted to other system types. A ReactionSystem can be marked as complete using the `compelte` function."
+
```
TorkelE, line 1475: Long text that is used repeatedly. When we refactor this file we might want to put it in a better place.

src/reactionsystem.jl
```diff
@@ -1490,6 +1494,7 @@ function Base.convert(::Type{<:ODESystem}, rs::ReactionSystem; name = nameof(rs)
                       include_zero_odes = true, remove_conserved = false, checks = false,
                       default_u0 = Dict(), default_p = Dict(), defaults = _merge(Dict(default_u0), Dict(default_p)),
                       kwargs...)
+    iscomplete(rs) || error(COMPLETENESS_ERROR)
```
TorkelE, line 1497: Currently throws error whenever non-compelte rs is used as input to any convert call.

src/reactionsystem.jl
```diff
@@ -1660,6 +1672,7 @@ function DiffEqBase.ODEProblem(rs::ReactionSystem, u0, tspan,
     pmap = symmap_to_varmap(rs, p)
     osys = convert(ODESystem, rs; name, combinatoric_ratelaws, include_zero_odes, checks,
                    remove_conserved)
+    osys = complete(osys)
```
TorkelE, line 1675: Error for non-compelte `ReactionSystem` input is already thrown in the convert call. This converts the osys (or similar for the other cases) to convert so that it can be simulated.

test/dsl/custom_functions.jl
```diff
@@ -33,32 +35,13 @@ let
...
-    g1 = SDEFunction(convert(SDESystem, custom_function_network_1))
-    g2 = SDEFunction(convert(SDESystem, custom_function_network_2))
     for factor in [1e-2, 1e-1, 1e0, 1e1, 1e2]
-        u0 = factor * rand(rng, length(get_unknowns(custom_function_network_1)))
-        p = factor * rand(rng, length(get_ps(custom_function_network_2)))
-
-        # needed as this code assumes an ordering of the parameters and species...
-        p2 = permute_ps(p, custom_function_network_1, custom_function_network_2)
-
+        u0 = rnd_u0(custom_function_network_1, rng; factor)
```
TorkelE, line 39: Using the new functionality and helper functions, quite a few tests can be written more concisely, which is nice.

test/dsl/custom_functions.jl
```diff
@@ -33,32 +35,13 @@ let
...
+        ps = rnd_ps(custom_function_network_1, rng; factor)
         t = rand(rng)
-        @test all(abs.(f1(u0, p, t) .- f2(u0, p2, t)) .< 10e-10)
-        @test all(abs.(f1.jac(u0, p, t) .- f2.jac(u0, p2, t)) .< 10e-10)
-        @test all(abs.(g1(u0, p, t) .- g2(u0, p2, t)) .< 10e-10)
+        @test f_eval(custom_function_network_1, u0, ps, t) ≈ f_eval(custom_function_network_2, u0, ps, t)
+        @test jac_eval(custom_function_network_1, u0, ps, t) ≈ jac_eval(custom_function_network_2, u0, ps, t)
+        @test g_eval(custom_function_network_1, u0, ps, t) ≈ g_eval(custom_function_network_2, u0, ps, t)
```
TorkelE, line 44: Many tests was written by me 6 years ago. Given that I know a bit more of how various functions work, I can rewrite them better now. While I was going through the tests I took the time to do this as well.

test/dsl/dsl_basics.jl
```diff
@@ -110,16 +117,7 @@ let
...
-@testset "make_reaction_system can be called from another module" begin
-    ex = quote
-        (Ka, Depot --> Central)
-        (CL / Vc, Central --> 0)
-    end
-    # Line number nodes aren't ignored so have to be manually removed
-    Base.remove_linenums!(ex)
-    @test eval(Catalyst.make_reaction_system(ex)) isa ReactionSystem
-end
-
```
TorkelE, line 122: This one unfortunately is marked as deleted/added. I put into a `let end` block like our other tests. If there is some special reason to have it in this form, we can move it back.

TorkelE: Seems it was also moved one step down.

test/dsl/dsl_model_construction.jl
```diff
@@ -376,16 +368,7 @@ let
...
 
-# Test names work.
-let
-    rn = @reaction_network SIR1 begin
-        k1, S + I --> 2I
-        k2, I --> R
-    end
-    @test nameof(rn) == :SIR1
-end
-
```
TorkelE, line 387: This one is in between two related tests. I moved it to the end.

test/miscellaneous_tests/api.jl
```diff
@@ -487,12 +497,17 @@ let
...
     # Test symmap_to_varmap.
-    sir = @reaction_network sir begin
+    sir = @network_component sir begin
         β, S + I --> 2I
         ν, I --> R
     end
-    subsys = @reaction_network subsys begin k, A --> B end
+    subsys = @network_component subsys begin 
+        k, A --> B 
+    end
```
TorkelE, line 506: unfold system that got messed up by the formater

test/miscellaneous_tests/api.jl
```diff
@@ -606,20 +621,6 @@ let
...
-    rn = @reaction_network begin
-        (p,d), 0 <--> X
-        (kB,kD), 2X <--> X
-    end
-    @unpack p, d, kB, kD = rn
-    isequal(reactionparamsmap(rn), Dict([p => 1, d => 2, kB => 3, kD => 4]))
-    issetequal(reactionrates(rn), [p, d, kB, kD])
-    isequal(symmap_to_varmap(rn, [:p => 1.0, :kB => 3.0]), [p => 1.0, kB => 3.0])
-end
-
```
TorkelE, line 622: This is the test that got moved to the top

test/miscellaneous_tests/compound_macro.jl
```diff
@@ -60,36 +67,7 @@ let
...
-    @compound SO2(t,x,y) ~ S + 2O
-    
-    # Checks they have the correct independent variables.
-    @test issetequal(arguments(ModelingToolkit.unwrap(CO2)), [t])
-    @test issetequal(arguments(ModelingToolkit.unwrap(NH4)), [x])
-    @test issetequal(arguments(ModelingToolkit.unwrap(H2O)), [t, x])
-    @test issetequal(arguments(ModelingToolkit.unwrap(PH4)), [t, x])
-    @test issetequal(arguments(ModelingToolkit.unwrap(SO2)), [t, x, y])
-end
-
```
TorkelE, line 89: Got moved to better place in the same file. Content not actually changed. Sorry that this makes it harder to look, but since we wanted to refactor the tests,it was easier to do this stuff now as well, rather than having to go back and do it all again.

TorkelE: This test now start start around line 105

test/miscellaneous_tests/compound_macro.jl
```diff
@@ -120,37 +99,38 @@ let
...
-    @test !iscompound(components(C3H5OH3)[1])
-    @test !iscompound(components(C3H5OH3)[2])
-    @test iscompound(components(C3H5OH3)[3])
+    @variables x y z
+    @species C(t) H(x) N(x) O(t) P(t,x) S(x,y)
 
-    @test isequal([C, H, OH], components(C3H5OH3))
-    @test isequal([O, H], components(components(C3H5OH3)[3]))
-    @test isequal([3, 5, 3], coefficients(C3H5OH3))
```
TorkelE, line 150: This test (unchanged) now starts at around line 270

test/model_simulation/simulate_ODEs.jl
```diff
@@ -1,30 +1,33 @@
...
-                           range(0.0, 100 / factor, length = 101))
-        @test all(abs.(first.(sol.u) .- analytic_sol) .< 0.1)
+        u0 = rnd_u0(exponential_decay, rng; factor)    
+        t_stops = range(0.0, 100 / factor, length = 101)
+        p = rnd_ps(exponential_decay, rng; factor)
+        prob = ODEProblem(exponential_decay, u0, (0.0, t_stops[end]), p)
+
+        sol = solve(prob, Rosenbrock23(), saveat = t_stops, abstol = 1e-14, reltol = 1e-14)
+        analytic_sol = [u0[1][2] * exp(-p[1][2] * t) for t in t_stops]
+        @test sol[:X] ≈ analytic_sol
```
TorkelE, line 30: Lots of simulation tests are basically makred as entirely new. The process is the same, but typically most lines had some change to accommodate new approach to maps etc.

test/model_simulation/simulate_SDEs.jl
```diff
@@ -93,47 +103,88 @@ let
...
+    end
+    
+    u0_1 = rnd_u0(noise_scaling_network, rng)
+    ps_1 = rnd_ps(noise_scaling_network, rng)
+    u0_2 = map_to_vec(u0_1, [:X1, :X2])
+    ps_2 = map_to_vec(ps_1, [:η1, :η2, :p, :k1, :k2, :d])
+    
+    @test g_eval(noise_scaling_network, u0_1, ps_1, 0.0) == real_g_3(zeros(2, 4), u0_2, ps_2, 0.0)
+end
+
```
TorkelE, line 187: New noise scaling test that uses `g_eval`

test/programmatic_model_creation/component_based_model_creation.jl
```diff
@@ -21,6 +27,7 @@ let
     specs = [m, P, R]
     pars = [α₀, α, K, n, δ, β, μ]
     @named rs = ReactionSystem(rxs, t, specs, pars)
+    rs = complete(rs)
```
TorkelE, line 30: This is one of those test files that have really long test blocks where it is hard to see what exactly is being tested. Long-term it would be nice to split these up. But lets fix v14 first.

test/reactionsystem_structure/reactionsystem.jl
```diff
@@ -451,18 +475,19 @@ let
...
     du2 = zeros(4)
     sprob = SDEProblem(ssys, u0map, tspan, pmap; check_length = false)
-    sprob.f(du1, u0, p, 1.0)
+    sprob.f(du1, sprob.u0, sprob.p, 1.0)
     fs!(du2, u0, p, 1.0)
     @test isapprox(du1, du2)
     dg1 = zeros(4, 4)
     dg2 = zeros(4, 4)
-    sprob.g(dg1, u0, p, 1.0)
+    sprob.g(dg1, sprob.u0, sprob.p, 1.0)
```
TorkelE, line 490: Could maybe have used test functions like `g_eval` in these cases. However, I did not want to make more changes than absolutely necessary here, as I was not fully sure what was part of the tests and what is just necessary to get there.

test/visualization/latexify.jl
```diff
@@ -131,50 +138,6 @@ let
...
-let
```
TorkelE, line 135: Now on line 215

test/visualization/latexify.jl
```diff
@@ -131,50 +138,6 @@ let
...
-        k*Y, Y --> ∅
-    end
-
-    # Latexify.@generate_test latexify(rn)
-    @test_broken latexify(rn) == replace(
-    raw"\begin{align*}
-    \varnothing &\xrightarrow{p} (m + n)\mathrm{X}  
-    \end{align*}
-    ", "\r\n"=>"\n")
-end
```
TorkelE, line 155: now line 225

test/visualization/latexify.jl
```diff
@@ -131,50 +138,6 @@ let
...
-let
-    rn = @reaction_network begin
-        (p,d), 0 <--> X
-    end
-    chem_latex = latexify(rn; env = :arrows)
-    @test chem_latex == latexify(rn; env = :chem)
-    @test chem_latex == latexify(rn; env = :chemical)
-    @test chem_latex == latexify(rn; env = :arrow)
-    @test_throws Exception latexify(rn; env = :wrong_env)
-end
```
TorkelE, line 168: This test no starts on line 195

test/visualization/latexify.jl
```diff
@@ -131,50 +138,6 @@ let
...
-
-# Tests that the `mathrm` option affects the output.
-let
-    rn = @reaction_network begin
-        (k1,k2), 2X <--> X2
-    end
-    @test latexify(rn; mathrm = true) != latexify(rn; mathrm = false)
-end
-
```
TorkelE, line 177: Now line 207

test/dsl/dsl_model_construction.jl
TorkelE: That is, line 4429

test/dsl/dsl_basics.jl
TorkelE: ok, manged to sort this ouit now, diff should look better.

test/dsl/dsl_basics.jl
```diff
@@ -110,7 +117,8 @@ let
     @test rn == rn2
 end
 
-@testset "make_reaction_system can be called from another module" begin
```
TorkelE, line 113: Chanegd the `@testset` block to ` let .. end` block like for all of our other tests. Not sure if there is a reason for this here, since it is all already in a testset?

test/model_simulation/u0_n_parameter_inputs.jl
```diff
@@ -19,15 +19,15 @@ include("../test_networks.jl")
 # Tests various ways to input u0 and p for various functions.
 let
     test_network = reaction_networks_standard[7]
-    test_osys = convert(ODESystem, test_network)
+    test_osys = complete(convert(ODESystem, test_network))
     @parameters p1 p2 p3 k1 k2 k3 v1 K1 d1 d2 d3 d4 d5
     @species X1(t) X2(t) X3(t) X4(t) X5(t) X6(t) X(t)
 
     for factor = [1e-2, 1e-1, 1e0, 1e1, 1e2, 1e3]
-        u0_1 = factor*rand(rng,length(get_unknowns(test_network)))
```
TorkelE, line 27: This file have not been properly updated. However, it is really simalr to the indexing test file that I have created to check SII. When I push that PR, that will contain a revamp (and big expansion) of these tests as well. So we will deal with them them

src/reaction_network.jl
TorkelE: This issue seems realted to this one: https://github.com/JuliaLang/julia/issues/37691

test/miscellaneous_tests/symbolic_stoichiometry.jl
isaacsas: Let's leave them broken in this PR and followup in another. I added it to the V14 todo.

test/miscellaneous_tests/units.jl
isaacsas: We'll try to support both in your other PR.

test/model_simulation/simulate_jumps.jl
isaacsas: Can you give me a standalone MWE on this example that breaks to play with? I'm not even sure what network this is. (i.e. that I can just cut and paste and doesn't require pulling a network out of another file.)

test/model_simulation/simulate_jumps.jl
```diff
@@ -91,56 +103,61 @@ let
...
-        @test mean(vals1) > mean(vals2)
+    no_param_network = @reaction_network begin 
+        (1.2, 5), X1 ↔ X2 
     end
+    u0 = rnd_u0_Int64(no_param_network, rng)
+    dprob = DiscreteProblem(no_param_network, u0, (0.0, 1000.0))
+    jprob = JumpProblem(no_param_network, dprob, Direct(); rng)
+    sol = solve(jprob, SSAStepper())
+    @test mean(sol[:X1]) > mean(sol[:X2])
 end
```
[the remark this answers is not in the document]
isaacsas: I don't like these relative size tests as they are not going to necessarily surface a truely incorrect solution.

test/model_simulation/u0_n_parameter_inputs.jl
isaacsas: ok

test/programmatic_model_creation/programmatic_model_expansion.jl
```diff
@@ -58,6 +62,7 @@ let
 end
 
 # Tests creating a network and adding reactions.
+# This test seems weird?
```
isaacsas, line 65:
```suggestion
```
Not really something to comment on in the test. Add a PR comment on Github if you have questions about a test. This is just checking that we end up with the right number of parameters and states. (But will be deleted anyways so doesn't matter...)

test/reactionsystem_structure/higher_order_reactions.jl
```diff
@@ -36,49 +40,75 @@ let
...
+    # higher_order_network_alt2 = @reaction_network begin
+    #     p, ∅ ⟼ X1
+    #     r1 * binomial(X1, 2), 2X1 ⟾ 3X2
+    #     mm(X1, r2, K) * binomial(X2, 3), 3X2 ⟾ X3 + 2X4
+    #     r3 * binomial(X3, 1) * binomial(X4, 2), X3 + 2X4 ⟾ 3X5 + 3X6
+    #     r4 * X2 * binomial(X5, 3) * binomial(X6, 3), 3X5 + 3X6 ⟾ 3X5 + 2X7 + 4X8
+    #     r5 * binomial(X5, 3) * binomial(X7, 2) * binomial(X8, 4), 3X5 + 2X7 + 4X8 ⟾ 10X9
+    #     r6 * binomial(X9, 10), 10X9 ⟾ X10
+    #     d * binomial(X10, 2), 2X10 ⟾ ∅
+    # end
```
isaacsas, line 72: This passed before. We need to keep it working now. It is a valuable test that is pretty comprehensive...

isaacsas: Please open issues on MTK/Symbolics if something like `binomial` is now broken that used to work.

test/reactionsystem_structure/higher_order_reactions.jl
```diff
@@ -36,49 +40,75 @@ let
...
+    affect2!(int) = (int.u[1] -= 2; int.u[2] += 3;)
+    affect3!(int) = (int.u[2] -= 3; int.u[3] += 1; int.u[4] += 2;)
+    affect4!(int) = (int.u[3] -= 1; int.u[4] -= 2; int.u[5] += 3; int.u[6] += 3;)
+    affect5!(int) = (int.u[5] -= 3; int.u[6] -= 3; int.u[5] += 3; int.u[7] += 2; int.u[8] += 4;)
+    affect6!(int) = (int.u[5] -= 3; int.u[7] -= 2; int.u[8] -= 4; int.u[9] += 10;)
+    affect7!(int) = (int.u[9] -= 10; int.u[10] += 1;)
+    affect8!(int) = (int.u[10] -= 2;)
+
+    higher_order_network_alt2 = ConstantRateJump.([rate1, rate2, rate3, rate4, rate5, rate6, rate7, rate8], 
+                                [affect1!, affect2!, affect3!, affect4!, affect5!, affect6!, affect7!, affect8!])
```
isaacsas, line 93: This is a nice test too, so we should get the old Catalyst version fixed and keep your new JumpProcesses version too and compare against both going forward.

test/reactionsystem_structure/reactions.jl
```diff
@@ -85,6 +85,6 @@ let
     r1 = Reaction(k, [X], [X2], [2], [1])
     r2 = Reaction(k, [X], [X2], [2], [1]; metadata=metadata)
 
-    @test isequal(Catalyst.get_noise_scaling(r1), 1.0)
+    @test_throws Exception Catalyst.get_noise_scaling(r1)
     @test isequal(Catalyst.get_noise_scaling(r2), η)
```
isaacsas, line 89: Do you have a `has_noise_scaling` function to test too (or is it tested somewhere else)?

test/reactionsystem_structure/reactionsystem.jl
isaacsas: Just trying to test the ODE rhs function gives the same values as the NonlinearProblem. We can switch to just use the function within the NonlinearProblem (I think maybe it wasn't implemented/working when we originally wrote these tests).

test/reactionsystem_structure/reactionsystem.jl
isaacsas: OK, when you are done with this PR I'll take a pass on these then.

isaacsas (changes requested):
There may be some older comments I missed deleting, so ask if any seem out of date or already settled. Sorry about that!

src/networkapi.jl
TorkelE: yes, I have a follow up PR for actually dealing with units, this is just the minimal amount of change for tests to pass.

test/miscellaneous_tests/symbolic_stoichiometry.jl
TorkelE: This is fixed in https://github.com/SciML/Catalyst.jl/pull/797

test/miscellaneous_tests/units.jl
TorkelE: DynamicQuantities working right now, still trying to figure out Unitful.

src/reactionsystem.jl
TorkelE: Good point, didn't directly think of the extensions. Both BifurcationKit/HomotopyContionuation should throw an error if applied to non-complete systems. I can add a tests for this though (or add an error at a higher level, so it is thrown already in `BifurcationProblem`/`hc_steady_states`, rather than in the created `NonlinearProblem`.

I could make StructuralIdentifiability require completeness. But think it does not do that for MTK models. I will double check what SI does when you given it a normal MTK system, and we use the same for Catalyst's `ReactionSystem`s, that would make sense right?

src/reactionsystem.jl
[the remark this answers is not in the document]
TorkelE: This would have been caught by (and actually was) regular CI. The problem was that when we merged the noise scaling one (and the other early MTK v9 PRs) all tests was erroring, so we merged without CI. 

In this case I checked the new features manually, mainly using the newly written tests, and not the ones unrelated to the PR, hence it slipped through. This is just a very good example while we are right to not merge any code if not all of CI is passing.

ext/CatalystBifurcationKitExtension/bifurcation_kit_extension.jl
```diff
@@ -13,7 +13,7 @@ function BK.BifurcationProblem(rs::ReactionSystem, u0_bif, ps, bif_par, args...;
 
     # Creates NonlinearSystem.
     Catalyst.conservationlaw_errorcheck(rs, vcat(ps, u0))
-    nsys = convert(NonlinearSystem, rs; remove_conserved=true, defaults=Dict(u0))
+    nsys = complete(convert(NonlinearSystem, rs; remove_conserved=true, defaults=Dict(u0)))
```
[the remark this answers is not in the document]
TorkelE: Both `XProblem` and `convert(XSystem, rs)` require the output to be complete. In the extensions, if the input `rs` is not complete, the `convert` call should throw an error. Next, because `convert(NonlinearSystem, rs; ...)` always yield an incomplete system, this has to be completed to be passed into `NonlinearProblem` or later on.

I think this bit is analogue e.g. `XProblem` first passing `rs` to `xsys = convert(XSystem, rs)` (which throw an error if `rs` is not complete), and then internally marks the output `x = complete(xsys).

I will add some tests to the extensions to ensure they throw an error if an input system is not complete, I agree this should not be allowed.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -45,7 +45,7 @@ end
 # For a given reaction system, parameter values, and initial conditions, find the polynomial that HC solves to find steady states.
 function steady_state_polynomial(rs::ReactionSystem, ps, u0)
```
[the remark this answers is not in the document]
TorkelE: Yes, `convert(NonlinearSystem, rs; remove_conserved = true)` throws an error if `rs` is not complete.

src/reactionsystem.jl
```diff
@@ -1469,6 +1471,8 @@ function spatial_convert_err(rs::ReactionSystem, systype)
     isspatial(rs) && error("Conversion to $systype is not supported for spatial networks.")
 end
 
+COMPLETENESS_ERROR = "A ReactionSystem must be complete before it can be converted to other system types. A ReactionSystem can be marked as complete using the `compelte` function."
```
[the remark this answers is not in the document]
TorkelE: Will do a ctrl+f for this. Should have done that already, know from experience that it happens a lot (it is that and "teh")

test/dsl/custom_functions.jl
```diff
@@ -250,7 +139,7 @@ end
 
 # Tests `Equation`s.
 let
-    @variables T X(T) Y(T)
+    @variables X(t) Y(t)
```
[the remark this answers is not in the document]
TorkelE: I will add some tests for that.

test/reactionsystem_structure/higher_order_reactions.jl
TorkelE: I did, I even wrote a PR to DynamicQuantities to get it solved as well, which got something, but not all the way. I will continue pushing for this though. Not sure when it will be fixed though. The `@test_broken false` can be removed.

test/reactionsystem_structure/higher_order_reactions.jl
TorkelE: agreed

docs/src/catalyst_functionality/compositional_modeling.md
```diff
@@ -5,16 +5,44 @@ can construct the earlier repressilator model by composing together three
 identically repressed genes, and how to use compositional modeling to create
 compartments.
 
+## [A note on *completeness*](@id completeness_note)
```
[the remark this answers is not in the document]
TorkelE: That make sense.  Do you want to change this around now, or when we get to the docs parts of the v14 fixes?

test/model_simulation/simulate_jumps.jl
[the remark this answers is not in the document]
TorkelE: I fully agree. Most of these are more or less similar to the first bunch of tests I wrote for Catalyst. Phasing these outs for more strict tests should be done.

test/reactionsystem_structure/higher_order_reactions.jl
isaacsas: If this issue is DynamicQuantites I would prefer we drop units over dropping this test and the use of binomials. We can then add units back if/when they update it.

docs/src/catalyst_functionality/compositional_modeling.md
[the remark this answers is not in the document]
isaacsas: We can hold off to a separate doc update, but can you add a TODO item on this so we don't forget?

test/model_simulation/simulate_ODEs.jl
```diff
@@ -37,25 +40,27 @@ let
...
-        @test abs.(sol.u[end][3] * sol.u[end][4] / sol.u[end][5] - p[4] / p[3]) <
-              10000 * eps()
-        @test abs.((sol.u[end][6]^2 / factorial(2)) / (sol.u[end][7]^3 / factorial(3)) -
-                   p[6] / p[5]) < 1e-8
-        @test abs.(sol.u[end][8] - p[7] / p[8]) < 10000 * eps()
+    for factor in [1e-1, 1e0, 1e1]
+        u0 = rnd_u0(known_equilibrium, rng; factor)    
+        p = rnd_ps(known_equilibrium, rng; factor, min = 0.1)
+        prob = ODEProblem(known_equilibrium, u0, (0.0, 100000.0), p)
+        sol = solve(prob, Vern7(); abstol = 1e-12, reltol = 1e-12)
```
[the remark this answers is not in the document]
isaacsas: But they passed previously with a larger tolerance right? So what has changed that they need to be so small. Something seems suspicious about this.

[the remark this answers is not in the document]
TorkelE: No, not really. E.g. for the one where I sat tolerances to `1e-14` we previously had a check like
```julia
 @test all(abs.(first.(sol.u) .- analytic_sol) .< 0.1)
```
essentially using an absolute tolerance of `1e-1`, this one we rand for a longer time (I think in some cases the asymptotic approach towards the equilibrium is really slow).

test/reactionsystem_structure/reactionsystem.jl
TorkelE: Sounds good, yeah, I think most of the tests were written when we didn't have access to lots of stuff we have now, and we also have gotten a lot further on how to write Catalyst code

test/model_simulation/simulate_jumps.jl
TorkelE: Yes, I will do this. I will just try to do it as minimal as possible before I send it to you (which in this case turned out surprisingly hard)

src/reaction_network.jl
TorkelE: removed

test/dsl/custom_functions.jl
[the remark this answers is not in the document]
TorkelE: I also agree that there is only gain from changing around the iv here. I have changed this back to `T`

test/miscellaneous_tests/api.jl
```diff
@@ -444,6 +450,10 @@ let
 end
 
 # Test defaults.
+# Uses mutating stuff (`setdefaults!`) and order dependent input (`species(rn) .=> u0`).
+# If you want to test this here @Sam I can write a new one that simualtes using defaults.
+# If so, tell me if you have anything specific you want to check though, or I will just implement 
+# it as I would.
```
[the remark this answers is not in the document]
TorkelE: sounds good

PR #787, opened 2024-02-14
Run JuliaFormatter.format()

LilithHafner:
Using JuliaFormatter v1.0.47


isaacsas:
This should probably get a target of the v14 branch. With so many things broken in the lead up to MTK 9 we are doing all dev work there currently, and I suspect reformatting master and then updating that branch with those changes would be a headache.

LilithHafner:
I'll try again once v14 is released.

isaacsas:
Sounds good!

PR #793, opened 2024-02-28
Fix `merge!` to include the `var_to_name` map

lmshk:
Previously, parameters defined in the second `ReactionSystem` would not be accessible as properties on the merged system. This also affected `@add_reactions`.

### DIFF
--- a/src/networkapi.jl
+++ b/src/networkapi.jl
@@ -1559,6 +1559,7 @@ function Base.merge!(network1::ReactionSystem, network2::ReactionSystem)
     union!(get_observed(network1), get_observed(network2))
     union!(get_systems(network1), get_systems(network2))
     merge!(get_defaults(network1), get_defaults(network2))
+    merge!(getfield(network1, :var_to_name), getfield(network2, :var_to_name))
     union!(MT.get_continuous_events(network1), MT.get_continuous_events(network2))
     union!(MT.get_discrete_events(network1), MT.get_discrete_events(network2))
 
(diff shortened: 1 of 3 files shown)


isaacsas:
Hi, thanks for the PR! We can't merge this at this time as currently any updates to master will break doc builds unfortunately (due to bugs in ModelingToolkit's V8 releases prior to the release of V9). We'll therefore have to wait till Catalyst V14 is merged into master to start making changes.

lmshk:
In the meantime, the following workaround can be used: Instead of `@add_reactions system begin ... end`, use:

    let more =
        @reaction_network begin
            ...
        end
        merge!(system, more)
        merge!(system.var_to_name, more.var_to_name}
        system
    end

The second `merge!` is idempotent and will therefore become a no-op once this PR is merged.

isaacsas:
Happy to hear you found a workaround! I would strongly recommend you change your workflow to creating a second system with any additional reactions and then using `ModelingToolkit.extend` to merge it into the first. This is the preferred way to merge systems together, and will likely be the only way in Catalyst 14 (since it is the official ModelingToolkit way to merge systems).

isaacsas:
See this tutorial: https://docs.sciml.ai/Catalyst/stable/catalyst_functionality/compositional_modeling/#Compositional-modeling-tooling

lmshk:
Sounds like this PR will become obsolete then, feel free to close.

isaacsas:
We've removed these mutating functions for Catalyst V14 so I'm going to close this.

PR #800, opened 2024-04-01
Noise scaling update

TorkelE:
Updates the noise scaling feature with:
- Ensures that species/variables/parameters introduced by the noise scaling metadata is always properly added to the new system.
- Add additional tests.

### DIFF
--- a/src/reactionsystem.jl
+++ b/src/reactionsystem.jl
@@ -1227,27 +1227,19 @@ function set_default_metadata(rs::ReactionSystem;  default_reaction_metadata = [
     # Currently, `noise_scaling` is the only relevant metadata supported this way.
     drm_dict = Dict(default_reaction_metadata)
     if haskey(drm_dict, :noise_scaling)
-        # Finds parameters, species, and variables in the noise scaling term.
-        ns_expr = drm_dict[:noise_scaling]
+        # Finds parameters, species, and variables in the noise scaling term.       
+        ns_expr = drm_dict[:noise_scaling] 
         ns_syms = [Symbolics.unwrap(sym) for sym in get_variables(ns_expr)]
-        ns_ps = filter(ModelingToolkit.isparameter, ns_syms)
-        ns_sps = filter(Catalyst.isspecies, ns_syms)
-        ns_vs = filter(sym -> !Catalyst.isspecies(sym) && !ModelingToolkit.isparameter(sym), ns_syms)
-
+        ns_ps = Iterators.filter(ModelingToolkit.isparameter, ns_syms)
+        ns_sps = Iterators.filter(Catalyst.isspecies, ns_syms)
+        ns_vs = Iterators.filter(sym -> !Catalyst.isspecies(sym) &&
+                                        !ModelingToolkit.isparameter(sym), ns_syms)
         # Adds parameters, species, and variables to the `ReactionSystem`.
-        if any(!any(isequal(p1, p2) for p2 in get_ps(rs)) for p1 in ns_ps)
-            @set! rs.ps = unique([get_ps(rs); ns_ps])
-        end
-        if any(!any(isequal(sp1, sp2) for sp2 in get_species(rs)) for sp1 in ns_sps)
-            sps_new = unique([get_species(rs); ns_sps])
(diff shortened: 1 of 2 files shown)


src/reactionsystem.jl
```diff
@@ -1221,7 +1222,34 @@ function set_default_metadata(rs::ReactionSystem;  default_reaction_metadata = [
...
     @set! rs.rxs = Reaction[rx for rx in updated_equations if rx isa Reaction]
-    
+
+    # Special routine to handle `Reaction` metadata that can contain new symbolic variables.
+    # Currently, `noise_scaling` is the only relevant metadata supported this way.
+    drm_dict = Dict(default_reaction_metadata)
+    if haskey(drm_dict, :noise_scaling)
+        # Finds parameters, species, and variables in the noise scaling term.
+        ns_expr = drm_dict[:noise_scaling]
+        ns_syms = [Symbolics.unwrap(sym) for sym in get_variables(ns_expr)]
```
isaacsas, line 1232: `unwrapping` should be handled in the `Reaction` constructor. We shouldn't be storing things that aren't unwrapped as we don't want to have to think about this internally later on (at least every internal Catalyst function I've written assumes its input is unwrapped as this is the MTK convention).

src/reactionsystem.jl
```diff
@@ -1221,7 +1222,34 @@ function set_default_metadata(rs::ReactionSystem;  default_reaction_metadata = [
...
+            sps_new = unique([get_species(rs); ns_sps])
+            vs_old = get_unknowns(rs)[length(get_species(rs))+1 : end]
+            @set! rs.species = sps_new
+            @set! rs.unknowns = [sps_new; vs_old]
+        end
+        if any(!any(isequal(v1, v2) for v2 in get_unknowns(rs)) for v1 in ns_vs)
+            us_new = unique([get_unknowns(rs); ns_vs])
+            @set! rs.unknowns = us_new
+        end
+    end
```
isaacsas, line 1251: Does `union!` not work?

isaacsas (changes requested):

src/reactionsystem.jl
TorkelE: Right now we just use the same convention for `Reaction` metadata as MTK/Symbolics does for symbolic variables (i.e. metadata is stored as they are provided by the user, if the user sets a metadata as `Num` it remain as a `Num`).

If you want we can add a specific routine in the `Reaction` constructor that checks if there is any `noise_scaling` metadata, and if it is `Num` unwraps it at that stage.

src/reactionsystem.jl
TorkelE: I never really knew about `union!` until I discovered that you use it in a few cases, so not that familiar with it. That said I should start using it, will implement it here.

TorkelE: Yeah, I can probably use generators, I think at this stage it was late and I was getting increasingly annoyed at all the effort doing this extra step. Will update to use generators though.

TorkelE: actually, I am a little bit uncertain exactly how you suggest we use `union!` and generators in these cases

isaacsas, line 1251:
```suggestion
        ns_syms = [Symbolics.unwrap(sym) for sym in get_variables(ns_expr)]
        ns_ps = Iterators.filter(ModelingToolkit.isparameter, ns_syms)
        ns_sps = Iterators.filter(Catalyst.isspecies, ns_syms)
        ns_vs = Iterators.filter(sym -> !Catalyst.isspecies(sym) && !ModelingToolkit.isparameter(sym), ns_syms)

        # Adds parameters, species, and variables to the `ReactionSystem`.
        @set! rs.ps = union(get_ps(rs), ns_ps])
         sps_new = union(get_species(rs), ns_sps)
         @set! rs.species = sps_new
         vs_old = @view get_unknowns(rs)[length(get_species(rs))+1 : end]            
         @set! rs.unknowns = union(sps_new, vs_old]
         @set! rs.unknowns = union(get_unknowns(rs), ns_vs)
        end       
    end
```
If this is acceptable and works / tests pass I'm fine with merging after switching to this.

PR #801, opened 2024-04-01
Equations and events reupload

TorkelE:
I have tried to rebase https://github.com/SciML/Catalyst.jl/pull/736 on master. Unfortunately, there have been too many changes, and it is not feasible anymore. This PR builds that one.

## Creating events in the DSL

MTK does not currently have a wrapper around the `@continuous_events` or `@discrete_events` options. However, we here use the same syntax as for programmatic description, but in a `begin ... end` block, with one event on each line. If there is only a single event, it can just be given after `@X_events`.
```julia
# When t == 2.5, increase p by 0.2.
rs = @reaction_network begin
    @continuous_events begin
        [t ~ 2.5] => [p ~ p + 0.2]
    end
    (p,d), 0 <--> X
end

# Every 2.0 time unit, increase d by 0.2.
rs = @reaction_network begin
    @discrete_events begin
        2.0 => [d ~ d + 0.2]
    end
    (p,d), 0 <--> X
end

# At time unit 2.0, increase d by 0.2.
rs = @reaction_network begin
    @discrete_events begin
        [2.0] => [d ~ d + 0.2]
    end
    (p,d), 0 <--> X
end

# After every timestep, if X > 0.2, increase d by 0.2.
rs = @reaction_network begin
    @discrete_events begin
        X > 0.1 => [d ~ d + 0.2]
    end
    (p,d), 0 <--> X
end
```

## Creating equations in the DSL
Equations use the `@equations` option. This utilises some MTK stuff, but the notation is similar as when creating stuff programmatically. Both differential and algebraic equations are supported.
```julia
rs_dsl = @reaction_network hybrid_rs begin
    @variables X2
    @parameters v
    @equations begin
        D(V) ~ X - v*V
        X2  ~ 2*X
    end
(p,d), 0 <--> X
end
```
Parameters and variables used within equations must be declared as such The exception is if we have a differential equation on the form
```julia
D(sym) ~ ...
```
where `sym` is a single symbol. Here, `sym` is inferred to be a variable (it can still be declared in `@variables`, if you e.g. want to attach metadata). In these cases, `D` is inferred to be a differential with respect to the independent variable.

It is also possible to create new differentials using the `@differentials` option:
```julia
rs_dsl = @reaction_network hybrid_rs begin
    @differentials begin
        DD = Differential(t)
    end
    @variables V(t)
    @equations begin
        DD(V) ~ -V
    end
    (p,d), 0 <--> X
end
```
these can then be used in equations. Note that if a non-default differential is used, the `V` variable have to be decalred.

## Working with hybrid equations
These work like before, however, with the following additions:
- Algebraic equations can now be used. These require using `structural_simplify` on the final `XSystem`. All relevant problem calls now have a `structural_simpligy = false` kwarg. Setting this to `true` ensures that `structural_simplify` (not `compelte`) is called on the internally created `ODESystem`.
- Hybrid systems can now be used to create Nonlinear Systems. When this happens, andy term `D(...)` (where `D` is some differential) are automatically set to `0`).
- Hybrid systems can be converted into SDEs. Currently, we do not support adding noise equations to variables (which would have to be done at a later point).

There are a couple of things which are broken, awaiting fixes in MTK, but these are rather tangential, and the main functionality works well.


Project.toml
```diff
@@ -6,6 +6,7 @@ version = "13.5.1"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 DiffEqBase = "2b5f629d-d688-5b77-993f-72d75c75574e"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+DynamicPolynomials = "7c1d4256-1411-5781-91ec-d7bc3513ac07"
```
TorkelE, line 9: Required to handle HomotopyCotninuation for systems featuring equations.

Project.toml
```diff
@@ -54,12 +55,13 @@ RuntimeGeneratedFunctions = "0.5.12"
 Setfield = "1"
 StructuralIdentifiability = "0.5.1"
 SymbolicUtils = "1.0.3"
-Symbolics = "5.14"
+Symbolics = "5.27"
```
TorkelE, line 58: We need some of the latest Symbolics updates

Project.toml
```diff
@@ -54,12 +55,13 @@ RuntimeGeneratedFunctions = "0.5.12"
...
 Unitful = "1.12.4"
 julia = "1.9"
 
 [extras]
 BifurcationKit = "0f109fa4-8a5d-4b75-95aa-f515264e7665"
+DiffEqCallbacks = "459566f4-90b8-5000-8ac3-15dfb0a30def"
```
TorkelE, line 64: For tests comparing events to callbacks (so we don't have to load full DifferentialEquations.jl)

Project.toml
isaacsas: Shouldn’t this be an extra?

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -53,7 +53,8 @@ function steady_state_polynomial(rs::ReactionSystem, ps, u0)
     eqs_pars_funcs = vcat(equations(ns), conservedequations(rs))
     eqs = map(eq -> substitute(eq.rhs - eq.lhs, p_dict), eqs_pars_funcs)
     eqs_intexp = make_int_exps.(eqs)
-    return Catalyst.to_multivariate_poly(remove_denominators.(eqs_intexp))
+    ss_poly = Catalyst.to_multivariate_poly(remove_denominators.(eqs_intexp))
+    return poly_type_convert(ss_poly)
```
TorkelE, line 57: Basically, if you have an algebraic equation in your system, you get a polynomial of a type HC cannot interpret, so I need to convert it to the correct one. This does that.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -104,4 +105,13 @@ function filter_negative_f(sols; neg_thres=-1e-20)
         (neg_thres < sol[idx] < 0) && (sol[idx] = 0)
     end
     return filter(sol -> all(>=(0), sol), sols)
+end
+
+# Sometimes (when polynomials are created from hybrid CRN/DAEs), the steady state polynomial have the wrong type.
+# This converts it to the correct type, which homotopy continuation can handle.
+const WRONG_POLY_TYPE = Vector{DynamicPolynomials.Polynomial{DynamicPolynomials.Commutative{DynamicPolynomials.CreationOrder}, DynamicPolynomials.Graded{DynamicPolynomials.LexOrder}}}
```
TorkelE, line 112: Sometimes you get a polynomial system of `WRONG_POLY_TYPE`, which can be converted to `CORRECT_POLY_TYPE`. Long-term this could probably be sorted out in HC, but until then I put this bit in here.

src/Catalyst.jl
```diff
@@ -37,11 +37,15 @@ import ModelingToolkit: check_variables,
...
 
+# Event-related MTK stuff.
+import ModelingToolkit: PresetTimeCallback
+
 import Base: (==), hash, size, getindex, setindex, isless, Sort.defalg, length, show
 import MacroTools, Graphs
 import Graphs: DiGraph, SimpleGraph, SimpleDiGraph, vertices, edges, add_vertices!, nv, ne
 import DataStructures: OrderedDict, OrderedSet
 import Parameters: @with_kw_noshow
+import Symbolics: occursin, wrap
```
TorkelE, line 48: At some point I want to go through this file, as I am really unsure where I should put various import statements, so they end up a bit all over the place, but that is for after v14.

src/reaction_network.jl
```diff
@@ -82,7 +82,8 @@ const forbidden_variables_error = let
 end
 
 # Declares the keys used for various options.
-const option_keys = (:species, :parameters, :variables, :ivs, :compounds, :observables, :default_noise_scaling)
+const option_keys = (:species, :parameters, :variables, :ivs, :compounds, :observables, :default_noise_scaling,
+                     :differentials, :equations, :continuous_events, :discrete_events)
```
TorkelE, line 86: adds te 4 new supported DSL options:
```
@differentials
@equations
@continuous_events
@continuous_events
```

src/reaction_network.jl
```diff
@@ -353,20 +354,28 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
     option_lines = Expr[x for x in ex.args if x.head == :macrocall]
 
     # Get macro options.
-    length(unique(arg.args[1] for arg in option_lines)) < length(option_lines) && error("Some options where given multiple times.")
+    if length(unique(arg.args[1] for arg in option_lines)) < length(option_lines)
+         error("Some options where given multiple times.")
+    end
```
TorkelE, line 359: move long single line to two lines.

src/reaction_network.jl
```diff
@@ -353,20 +354,28 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
...
     options = Dict(map(arg -> Symbol(String(arg.args[1])[2:end]) => arg,
                        option_lines))
 
     # Reads options.
     default_reaction_metadata = :([])
     check_default_noise_scaling!(default_reaction_metadata, options)
     compound_expr, compound_species = read_compound_options(options)
+    continuous_events_expr = read_events_option(options, :continuous_events)
+    discrete_events_expr = read_events_option(options, :discrete_events)
```
TorkelE, line 368: fetches continuous and discret event information.

src/reaction_network.jl
```diff
@@ -353,20 +354,28 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
...
 
     # Parses reactions, species, and parameters.
     reactions = get_reactions(reaction_lines)
     species_declared = [extract_syms(options, :species); compound_species]
     parameters_declared = extract_syms(options, :parameters)
-    variables = extract_syms(options, :variables)
+    variables_declared = extract_syms(options, :variables)
```
TorkelE, line 374: Previously variables and declared variables (using `@variables`) where the same. Now some variables can be extracted from differential equations (only those on the form `D(v) ~ ...`).

src/reaction_network.jl
```diff
@@ -353,20 +354,28 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
...
+
+    # Reads more options.
+    vars_extracted, add_default_diff, equations = read_equations_options(options, variables_declared)
+    variables = vcat(variables_declared, vars_extracted)
```
TorkelE, line 378: Reads the equations. Extracts variables from diff equations on the form
```julia
D(V) ~ ...
```
Also, if a diff equation on the form
```julia
D(V) ~ ...
```
is used, `D` is assumed to be the default differential with respect to the IV, so we make a note that this differential should be declared as part of the DSL output.

src/reaction_network.jl
```diff
@@ -391,6 +400,9 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
     species = vcat(species_declared, species_extracted)
     parameters = vcat(parameters_declared, parameters_extracted)
 
+    # Create differential expression.
+    diffexpr = create_differential_expr(options, add_default_diff, [species; parameters; variables])
+
```
TorkelE, line 405: Creates an expression with all the differentials that should be declared as part of the DSL, e.g.
```julia
begin
    D = Differential(t)
end
```

src/reaction_network.jl
```diff
@@ -402,7 +414,7 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
 
     # Creates expressions corresponding to actual code from the internal DSL representation.
     sexprs = get_sexpr(species_extracted, options; iv_symbols = ivs)
-    vexprs = haskey(options, :variables) ? options[:variables] : :()
+    vexprs = get_sexpr(vars_extracted, options, :variables)
```
TorkelE, line 417: Before we could just fetch the input to `@variables`, now we have to do like for species (where we have both an expression like that, and a few extracted ones)

src/reaction_network.jl
```diff
@@ -412,6 +424,9 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
     for reaction in reactions
         push!(rxexprs.args, get_rxexprs(reaction))
     end
+    for equation in equations
+        push!(rxexprs.args, equation)
+    end
```
TorkelE, line 429: Add all extracted equations to those given to the ReactionSystem

src/reaction_network.jl
```diff
@@ -799,6 +816,100 @@ function make_observed_eqs(observables_expr)
...
+# `vars_extracted`: A vector with extracted variables (lhs in pure differential equations only).
+# `dtexpr`: If a differentialequation is defined, the default derrivative (D ~ Differential(t)) must be defined.
+# `equations`: a vector with the equations provided.
+function read_equations_options(options, variables_declared)
+    # Prepares the equations. First, extracts equations from provided option (converting to block form if requried).
+    # Next, uses MTK's `parse_equations!` function to split input into a vector with the equations.
+    eqs_input = haskey(options, :equations) ? options[:equations].args[3] : :(begin end)
+    eqs_input = option_block_form(eqs_input)
+    equations = Expr[]
+    ModelingToolkit.parse_equations!(Expr(:block), equations, Dict{Symbol, Any}(), eqs_input)
```
TorkelE, line 829: This first bit creates the equation expression that we actually use

src/reaction_network.jl
```diff
@@ -799,6 +816,100 @@ function make_observed_eqs(observables_expr)
...
+        lhs = eq.args[2]
+        # if lhs: is an expression. Is a function call. The function's name is D. Calls a single symbol.
+        if (lhs isa Expr) && (lhs.head == :call) && (lhs.args[1] == :D) && (lhs.args[2] isa Symbol)
+            diff_var = lhs.args[2]
+            if in(diff_var, forbidden_symbols_error)
+                error("A forbidden symbol ($(diff_var)) was used as an variable in this differential equation: $eq")
+            end
+            add_default_diff = true
+            in(diff_var, variables_declared) || push!(vars_extracted, diff_var)
+        end
```
TorkelE, line 853: This bits looks for Differential equations on the form
```
D(V) ~ ...
```
When these are found, it adds `V` as something that should be declared a variable, and note that we have used the default differential `D`.

Project.toml
TorkelE: Because it is only relevant to an extension? If I understand the question correct (unfortunately) no, I asked about it on the Julai slack earlier:
image

isaacsas: Actually, it should already be being pulled installed as a Symbolics dependency anyways it seems.

src/reactionsystem.jl
```diff
@@ -551,14 +557,19 @@ struct ReactionSystem{V <: NetworkProperties} <:
...
+
         # unit checks are for ODEs and Reactions only currently
         nonrx_eqs = Equation[eq for eq in eqs if eq isa Equation]
         if checks && isempty(sivs)
             check_variables(unknowns, iv)
             check_parameters(ps, iv)
             nonrx_eqs = Equation[eq for eq in eqs if eq isa Equation]
             !isempty(nonrx_eqs) && check_equations(nonrx_eqs, iv)
-            check_equations(equations(cevs), iv)
+            !isempty(cevs) && check_equations(equations(cevs), iv)
```
TorkelE, line 572: only check this if there actually are any

src/reactionsystem.jl
```diff
@@ -617,70 +628,77 @@ function ReactionSystem(eqs, iv, unknowns, ps;
                         continuous_events = nothing,
                         discrete_events = nothing,
                         metadata = nothing)
-                        
-    name === nothing &&
+
```
TorkelE, line 631: When I did this tehre was some functions in reactionssyetm.jl that I had to go through line by line to figure out what was going on. While I did this I took the time to add comments what was going on, and to fix some weird formating. I only touched functions that I needed to go through for this PR, the remaining ones I left for another time.

src/reactionsystem.jl
```diff
@@ -617,70 +628,77 @@ function ReactionSystem(eqs, iv, unknowns, ps;
                         continuous_events = nothing,
                         discrete_events = nothing,
                         metadata = nothing)
-                        
-    name === nothing &&
+               
+    # Error checks
+    if name === nothing &&
         throw(ArgumentError("The `name` keyword must be provided. Please consider using the `@named` macro"))
+    end
```
TorkelE, line 635: I find these 
```julia
condition &&
    do something
```
that the formater have created confusing to parse. I tried to change them to either be a `if` statement, or be on a single line.

src/reactionsystem.jl
```diff
@@ -617,70 +628,77 @@ function ReactionSystem(eqs, iv, unknowns, ps;
...
+    # Extracts independent variables (iv and sivs), dependent variables (species and variables)
+    # and parameters. Sorts so that species comes before variables in unknowns vector.
     iv′ = value(iv)
     sivs′ = if spatial_ivs === nothing
         Vector{typeof(iv′)}()
     else
         value.(MT.scalarize(spatial_ivs))
     end
-    unknowns′ = sort!(value.(MT.scalarize(unknowns)), by = !isspecies) # species come first
+    unknowns′ = sort!(value.(MT.scalarize(unknowns)), by = !isspecies)
```
TorkelE, line 654: remove commeent that are on same line as code. This information is now part of the comment in the beginning of the block

src/reactionsystem.jl
```diff
@@ -617,70 +628,77 @@ function ReactionSystem(eqs, iv, unknowns, ps;
...
+    # Adds their (potential) default values to the defaults vector.
     var_to_name = Dict()
     MT.process_variables!(var_to_name, defaults, unknowns′)
     MT.process_variables!(var_to_name, defaults, ps′)
     MT.collect_var_to_name!(var_to_name, eq.lhs for eq in observed)
 
-    nps = if networkproperties === nothing
-        NetworkProperties{Int, get_speciestype(iv′, unknowns′, systems)}()
+    # Computes network properties.
+    if networkproperties === nothing
```
TorkelE, line 695: felt like writing it this way is much easier to read (but correct me if I am wrong).

src/reactionsystem.jl
```diff
@@ -694,77 +712,125 @@ function ReactionSystem(rxs::Vector, iv = Catalyst.DEFAULT_IV; kwargs...)
 end
 
 # search the symbolic expression for parameters or unknowns
-# and save in ps and sts respectively. vars is used to cache results
-function findvars!(ps, sts, exprtosearch, ivs, vars)
+# and save in ps and us respectively. vars is used to cache results
+function findvars!(ps, us, exprtosearch, ivs, vars)
```
TorkelE, line 716: Quite a few places we used e.g. `sts` assuming that something is a state. When these was renamed to unknowns I also renamed these variables to make that clear.

src/reactionsystem.jl
```diff
@@ -694,77 +712,125 @@ function ReactionSystem(rxs::Vector, iv = Catalyst.DEFAULT_IV; kwargs...)
...
     if (spatial_ivs !== nothing)
         for siv in (MT.scalarize(spatial_ivs))
             push!(ivs, value(siv))
         end
     end
-    sts = OrderedSet{eltype(sts_in)}(sts_in)
+
+    # Initialises the new unknowns and parameter vectors.
+    # Preallocates the `vars` set, which is used by `findvars!`
+    us = OrderedSet{eltype(us_in)}(us_in)
```
TorkelE, line 750: again, unknowns not states

isaacsas:
Please keep this PR, which is getting humongous focused on the changes needed for the DSL functionality. Style changes because you find something personally confusing, which others find standard and easy to read, should not be made as part of this PR.

src/reactionsystem.jl
```diff
@@ -694,77 +712,125 @@ function ReactionSystem(rxs::Vector, iv = Catalyst.DEFAULT_IV; kwargs...)
...
-            MT.isparameter(spec) ? push!(ps, spec) : push!(sts, spec)
-        end
-    end
-
+    # Loops through all reactions, adding encountered quantities to the unknown and parameter vectors.
     for rx in rxs
-        findvars!(ps, sts, rx.rate, ivs, vars)
-        for s in rx.substoich
-            (s isa Symbolic) && findvars!(ps, sts, s, ivs, vars)
+        # Loops through all reaction substrates and products, extracting these.
```
TorkelE, line 761: Loops through all reactions in one loop. Next, goes through each part of the reaction in order.

src/reactionsystem.jl
```diff
@@ -694,77 +712,125 @@ function ReactionSystem(rxs::Vector, iv = Catalyst.DEFAULT_IV; kwargs...)
...
 function ReactionSystem(iv; kwargs...)
     ReactionSystem(Reaction[], iv, [], []; kwargs...)
 end
 
-
+# Loops through all events in an supplied event vector, adding all unknowns and parameters found in
+# its condition and affect functions to their respective vectors (`ps` and `us`).
+function find_event_vars!(ps, us, events::Vector, ivs, vars)
+    foreach(event -> find_event_vars!(ps, us, event, ivs, vars), events)
+end
```
TorkelE, line 808: Special functiosn for extracting variables/parameters that are part of events. Can be used for both discrete/continuous events.

src/reactionsystem.jl
```diff
@@ -927,7 +993,7 @@ function get_indep_sts(rs::ReactionSystem, remove_conserved = false)
     sts = get_unknowns(rs)
     nps = get_networkproperties(rs)
     indepsts = if remove_conserved
-        filter(s -> (s ∈ nps.indepspecs) && (!isbc(s)), sts)
+        filter(s -> ((s ∈ nps.indepspecs) || (!isspecies(s))) && (!isbc(s)), sts)
```
TorkelE, line 996: bugfix

src/reactionsystem.jl
```diff
@@ -1548,17 +1614,26 @@ function Base.convert(::Type{<:NonlinearSystem}, rs::ReactionSystem; name = name
...
+    NonlinearSystem(eqs, us, ps;
                     name,
                     observed = obs,
                     defaults = _merge(defaults,defs),
                     checks,
                     kwargs...)
 end
 
+# Finds and differentials in an expression, and sets these to 0.
+function remove_diffs(expr)
```
TorkelE, line 1629: Sets differentials to 0 when nonlinear systems are created.

src/reactionsystem.jl
```diff
@@ -1587,20 +1662,20 @@ function Base.convert(::Type{<:SDESystem}, rs::ReactionSystem;
     spatial_convert_err(rs::ReactionSystem, SDESystem)
 
     flatrs = Catalyst.flatten(rs)
-    error_if_constraints(SDESystem, flatrs)
+    #error_if_constraints(SDESystem, flatrs)
```
TorkelE, line 1665: we now allow equations in SDEs

TorkelE:
Needs a new MTK release to pass tests, I will finish writing the description of how the equations work after lunch.

src/reactionsystem.jl
```diff
@@ -1548,17 +1614,31 @@ function Base.convert(::Type{<:NonlinearSystem}, rs::ReactionSystem; name = name
     ists, ispcs = get_indep_sts(fullrs, remove_conserved)
     eqs = assemble_drift(fullrs, ispcs; combinatoric_ratelaws, remove_conserved,
                          as_odes = false, include_zero_odes)
-    error_if_constraint_odes(NonlinearSystem, fullrs)
```
TorkelE, line 1551: No longer needed as nonlinear system can be created from hybrid models

src/reactionsystem.jl
```diff
@@ -1669,12 +1748,21 @@ function DiffEqBase.ODEProblem(rs::ReactionSystem, u0, tspan,
...
-    osys = complete(osys)
+
+    # Handles potential differential algebraic equations (which requires `structural_simplify`).
+    if structural_simplify 
+        (osys = MT.structural_simplify(osys))
+    elseif has_alg_equations(rs)
+        error("The input ReactionSystem has algebraic equations. This requires setting `structural_simplify=true` within `ODEProblem` call.")
+    else
+        osys = complete(osys)
+    end
```
TorkelE, line 1764: If `structural_simplify = true` we use that, if not we use `complete`. If there are algebraic equations but `structural_simplify = true` was not used, throw an error (these requires `structural_simplify`).

test/programmatic_model_creation/component_based_model_creation.jl
```diff
@@ -362,7 +363,7 @@ let
     rn2 = extend(osys, rn)
     rn2 = complete(rn2)
     rnodes = complete(convert(ODESystem, rn2))
-    @test_throws ErrorException convert(NonlinearSystem, rn2)
+    # @test_throws ErrorException convert(NonlinearSystem, rn2)
```
TorkelE, line 366: This is now possible.

test/runtests.jl
```diff
@@ -8,6 +8,8 @@ using SafeTestsets
     @time @safetestset "Reactions" begin include("reactionsystem_structure/reactions.jl") end
     @time @safetestset "ReactionSystem" begin include("reactionsystem_structure/reactionsystem.jl") end
     @time @safetestset "Higher Order Reactions" begin include("reactionsystem_structure/higher_order_reactions.jl") end
+    @time @safetestset "Parameter Type Designation" begin include("reactionsystem_structure/designating_parameter_types.jl") end
```
TorkelE, line 11: Didn't actually get properly added in the last PR

TorkelE:
Figured I should have another go through and see if I missed anything, and was immediately punished. On the plus side I discovered another MTK bug: https://github.com/SciML/SciMLBase.jl/issues...

TorkelE:
Ok, I've updated to the new MTK version where this was fixed, all good now.

src/reaction_network.jl
```diff
@@ -799,6 +816,101 @@ function make_observed_eqs(observables_expr)
...
+    # Next, uses MTK's `parse_equations!` function to split input into a vector with the equations.
+    eqs_input = haskey(options, :equations) ? options[:equations].args[3] : :(begin end)
+    eqs_input = option_block_form(eqs_input)
+    equations = Expr[]
+    ModelingToolkit.parse_equations!(Expr(:block), equations, Dict{Symbol, Any}(), eqs_input)
+
+    # Loops through all equations, checks for lhs of the form `D(X) ~ ...`.
+    # When this is the case, the variable X and differential D are extracted (for automatic declaration).
+    # Also performs simple error checks.
+    vars_extracted = []
```
isaacsas, line 834: Can this be typed as a `Symbol` array?

src/reactionsystem.jl
```diff
@@ -551,14 +557,19 @@ struct ReactionSystem{V <: NetworkProperties} <:
             (p isa Symbolics.BasicSymbolic) || error("Parameter $p is not a `BasicSymbolic`. This is required.")
         end
 
+        # Filters away any potential obervables from `states` and `spcs`.
+        obs_vars = [obs_eq.lhs for obs_eq in observed]
+        unknowns = filter(state -> !any(isequal(state, obs_var) for obs_var in obs_vars), unknowns)
+        spcs = filter(spc -> !any(isequal(spc, obs_var) for obs_var in obs_vars), spcs)
```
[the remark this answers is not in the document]
isaacsas: (Part of the reason for this is that a user might be incorrectly understanding how observables and variables/species work, so it is better to error and let them correct their model I think.)

src/reactionsystem.jl
```diff
@@ -617,70 +628,77 @@ function ReactionSystem(eqs, iv, unknowns, ps;
...
-    name === nothing &&
+               
+    # Error checks
+    if name === nothing
         throw(ArgumentError("The `name` keyword must be provided. Please consider using the `@named` macro"))
+    end
     sysnames = nameof.(systems)
-    (length(unique(sysnames)) == length(sysnames)) ||
-        throw(ArgumentError("System names must be unique."))
+    (length(unique(sysnames)) == length(sysnames)) || throw(ArgumentError("System names must be unique."))
```
isaacsas, line 637: This kind of change is just going to get undone when we next run the formatter I think. More generally, it is noise in the PR that makes understanding what you've changed harder. Please keep such reformatting type changes to separate PRs to help keep your PRs as short as possible.

src/reactionsystem.jl
```diff
@@ -694,76 +712,115 @@ function ReactionSystem(rxs::Vector, iv = Catalyst.DEFAULT_IV; kwargs...)
...
-        for p in rx.prodstoich
-            (p isa Symbolic) && findvars!(ps, sts, p, ivs, vars)
+
+        # Adds all quantitites encountered in the reaction's rate.
+        findvars!(ps, us, rx.rate, ivs, vars)
+
+        # Extracts all quantitites encountered within stoichiometries.
+        for stoichiometry in (rx.substoich, rx.prodstoich), sym in stoichiometry
+            (sym isa Symbolic) && findvars!(ps, us, sym, ivs, vars)
         end
```
isaacsas, line 781: Aren't these changes going to alter the order of parameters and species potentially? That may break tests for downstream packages (which really shouldn't be relying on order, but I think some may).

src/reactionsystem.jl
```diff
@@ -694,76 +712,115 @@ function ReactionSystem(rxs::Vector, iv = Catalyst.DEFAULT_IV; kwargs...)
...
+        union!(us, unknowns(osys))
+        union!(ps, parameters(osys))
     else
         fulleqs = rxs
-    end
+    end 
+
+    # Loops through all events, adding encountered quantities to the unknwon and parameter vectors.
+    find_event_vars!(ps, us, continuous_events, ivs, vars)      
+    find_event_vars!(ps, us, discrete_events, ivs, vars)
```
isaacsas, line 799: Should these be in MTK then and added to the systems there?

src/reactionsystem.jl
isaacsas: I thought the point is to return the independent unknowns which include non-species as the first return argument. The independent species are returned as the second argument. There is no reason to return both anymore in this function?

src/reactionsystem.jl
```diff
@@ -1548,17 +1605,59 @@ function Base.convert(::Type{<:NonlinearSystem}, rs::ReactionSystem; name = name
     ists, ispcs = get_indep_sts(fullrs, remove_conserved)
```
isaacsas, line 1605: Here `ists` are no longer the independent unknowns now with your change, won't they be identical to the `ispcs`?

src/reactionsystem.jl
```diff
@@ -1548,17 +1605,59 @@ function Base.convert(::Type{<:NonlinearSystem}, rs::ReactionSystem; name = name
...
     eqs = assemble_drift(fullrs, ispcs; combinatoric_ratelaws, remove_conserved,
                          as_odes = false, include_zero_odes)
-    error_if_constraint_odes(NonlinearSystem, fullrs)
-    eqs, sts, ps, obs, defs = addconstraints!(eqs, fullrs, ists, ispcs; remove_conserved)
+    eqs, us, ps, obs, defs = addconstraints!(eqs, fullrs, ists, ispcs; remove_conserved)
+
+    # Throws a warning if there are differential equations in non-standard format.
+    # Next, sets all differential terms to `0`.
+    nonlinear_convert_differentials_check(rs)
+    eqs = [remove_diffs(eq.lhs) ~ remove_diffs(eq.rhs) for eq in eqs]
```
isaacsas, line 1613: I think we should error if there are higher than first-order derivatives present. The generated nonlinear system doesn't really make sense to me in that case.

src/reactionsystem.jl
```diff
@@ -1548,17 +1605,59 @@ function Base.convert(::Type{<:NonlinearSystem}, rs::ReactionSystem; name = name
...
+        (eq isa Reaction) && continue
+        if Symbolics._occursin(Symbolics.is_derivative, eq.rhs) ||
+                !Symbolics.istree(eq.lhs) ||
+                !isequal(Symbolics.operation(eq.lhs), Differential(get_iv(rs))) || 
+                (length(arguments(eq.lhs)) != 1) ||
+                !any(isequal(arguments(eq.lhs)[1]), nonspecies(rs))
+            @warn "You are attempting to convert a `ReactionSystem` coupled with differential equations. However, some of these differentials are not of the form `D(x) ~ ...` where: 
+                    (1) The left-hand side is a differential of a single variable with respect to the time independent variable, and
+                    (2) The right-hand side does not contain any differentials.
+                  This is permitted (and all differential will be set to `0`), however, it is recommended to proceed with caution to ensure that the produced nonlinear equation is the desired one."
```
isaacsas, line 1655: I think this should be an error with a keyword to override it if a user really wants to do this. I can't really see, especially for higher order derivatives, that this will generally be correct so we should really make sure users want to do this via forcing them to pass the override keyword.

src/reactionsystem.jl
```diff
@@ -1669,12 +1767,21 @@ function DiffEqBase.ODEProblem(rs::ReactionSystem, u0, tspan,
...
-    osys = complete(osys)
+
+    # Handles potential differential algebraic equations (which requires `structural_simplify`).
+    if structural_simplify 
+        (osys = MT.structural_simplify(osys))
+    elseif has_alg_equations(rs)
+        error("The input ReactionSystem has algebraic equations. This requires setting `structural_simplify=true` within `ODEProblem` call.")
+    else
+        osys = complete(osys)
+    end
```
isaacsas, line 1783: This feels like adding a keyword for something we don't own -- if a user wants to simplify why not require them to explicitly convert and call `structural_simplify`?

src/reactionsystem.jl
```diff
@@ -1698,13 +1805,22 @@ function DiffEqBase.SDEProblem(rs::ReactionSystem, u0, tspan,
                                p = DiffEqBase.NullParameters(), args...;
                                name = nameof(rs), combinatoric_ratelaws = get_combinatoric_ratelaws(rs),
                                include_zero_odes = true, checks = false, check_length = false,
-                               remove_conserved = false, kwargs...)
+                               remove_conserved = false, structural_simplify = false, kwargs...)
```
isaacsas, line 1808: Same comment.

isaacsas (changes requested):
We can leave the reformats I didn't address as is. But please stop rewriting / refactoring perfectly fine existing code as part of PRs that are adding new functionality. It just makes the PR much longer, increasing the review time (it is much easier to find time here and there to review 10 200-line PRs than one 2000 line PR). Modifying existing code in ways that does not change what it is doing or fix a bug should be left for separate PRs.

src/reactionsystem.jl
isaacsas: Nevermind, I misunderstood your change.

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -94,7 +95,7 @@ end
 function reorder_sols!(sols, ss_poly, rs::ReactionSystem)
     var_names_extended = String.(Symbol.(HC.variables(ss_poly)))
     var_names = [Symbol(s[1:prevind(s, findlast('_', s))]) for s in var_names_extended]
-    sort_pattern = indexin(MT.getname.(species(rs)), var_names)
+    sort_pattern = indexin(MT.getname.(unknowns(rs)), var_names)
```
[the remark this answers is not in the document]
TorkelE: Yes (we should end up using the same for this as if the user provides rates containing non-polynomial stuff)

src/reactionsystem.jl
TorkelE: That should probably (definitely) be the case. I will raise an issue at MTK that they should start doing this (and when they do, we can remove this line).

docs/src/catalyst_functionality/dsl_description.md
```diff
@@ -701,3 +701,66 @@ rn = @reaction_network begin
...
+
+## Incorporating (differential) equations into reaction network models
+Some models cannot be purely described as reaction networks. E.g. consider the growth of a cell, where the rate of change in cell's volume depends on some growth factor. Here, the cell's volume would be described by a normal equation. Such equations can be incorporated into a model using the `@equations` option. Here, we create a model where a growth factor ($G$) is produced and degraded at a linear rates, and the rate of change in cell volume ($V$) os linear to the amount of growth factor:
+```@example eqs1
+using Catalyst #hide
+rn = @reaction_network begin
+  @equations begin
+    D(V) ~ G
+  end
+  (p,d), 0 <--> G
```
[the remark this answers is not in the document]
TorkelE: Yeah, I think that makes sense. All of the volume equations are a bit awkward since I am unsure whether to go full out on modelling it properly or not.

Either way, these docs are entirely rebuilt in a follow-up PR, so let's handle that there. I mostly wrote these back in the day since I thought there was a chance this might be up for a while before I actually got to revamping the equation dosc properly.

src/reactionsystem.jl
TorkelE: Agree.

In my defense there was a large number of formatting changes that I wanted to do that I left out for this reason. I only changed these ones because I (apparently erroneously) thought it was an error introduced by the formater.

src/reactionsystem.jl
TorkelE: I am more than happy to error in this case as well, will update.

src/reactionsystem.jl
TorkelE: Yes it will.

Long term this felt like a better way to phrase the code, especially as we add more steps (like checking various metadata fields).

As you say, stuff should not rely on order, but I agree it might break _something_. That said, if it improves the code I don't think we should adapt for someone doing something they shouldn't do. In either case we def should note this in the History file.

I am still OK with reverting this though.

src/reactionsystem.jl
TorkelE: added such a keyword, and made a note of it in the error.

src/reaction_network.jl
TorkelE: updated

src/Catalyst.jl
```diff
@@ -30,18 +30,23 @@ using ModelingToolkit: Symbolic, value, istree, get_unknowns, get_ps, get_iv, ge
...
+                        namespace_parameters, rename, renamespace, getname, flatten,
+                        is_alg_equation, is_diff_equation
 
 # internal but needed ModelingToolkit functions
 import ModelingToolkit: check_variables,
                         check_parameters, _iszero, _merge, check_units,
                         get_unit, check_equations, iscomplete
 
+# Event-related MTK stuff.
+import ModelingToolkit: PresetTimeCallback
```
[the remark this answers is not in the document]
TorkelE: I remember adding this, but have forgotten why actually. Will try removing it and if tests passes that is good.

src/reactionsystem.jl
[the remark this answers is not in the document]
isaacsas: Then shouldn't this be part of the DSL? I still don't think that `ReactionSystem`s should be "fixing" bad input but should instead just error.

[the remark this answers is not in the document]
isaacsas: I'd be willing to have this in the two-argument constructor, but I don't think it should be in the inner constructor. The inner constructor implicitly assumes the collection of unknowns and parameters are correct and doesn't do discovery.

src/reactionsystem.jl
isaacsas: As long are you are ok with making updates to downstream unmaintained packages we'd like to keep working that break from this I am ok with the change. (i.e. stuff like SBMLToolkit, FiniteStateProjection, etc.) I just don't want to have to go and submit PRs to these various packages if they rely on order and we change it now.

src/reactionsystem.jl
isaacsas: But isn't it weird to add it for Problem(::ReactionSystem) but not have it for Problem(::ODESystem) (for example)? Seems like something that could confuse users? 

If you really feel strongly about it we can keep it, but I just think it introduces a strange inconsistency for the user workflow by having such a keyword just for `ReactionSystem`s.

src/reactionsystem.jl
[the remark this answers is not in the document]
TorkelE: I will look at revamping where this is done. Generally (in agreement with our previous discussion) I think that stuff like this should be handled outside of the DSL.

src/reactionsystem.jl
TorkelE: Generally, most users who are not developing stuff will not be dealing with non`ReactionSystem` systems anyway, I think this should be fine.

PR #805, opened 2024-04-08
New indexing and simulation input tests

TorkelE:
Reupload from https://github.com/SciML/Catalyst.jl/pull/761 (where git could not keep track with all the various branches).

Adds various tests for indexing and creating problems.


test/visualization/plotting.jl
```diff
@@ -1,35 +0,0 @@
-### Not currently run by runtests
-
```
TorkelE, line 2: These tests where not run, and are superseded by new indexing tests, which does this but are also more extensive

test/meta/mtk_problem_inputs.jl
```diff
@@ -0,0 +1,243 @@
...
+
+    # Loops through all potential parameter sets, checking their inputs yield errors.
+    for ps in [ps_valid; ps_invalid], u0 in [u0_valid; u0s_invalid]
+        # Handles problems with/without tspan separately. Special check ensuring that valid inputs passes.
+        for XProblem in [ODEProblem, SDEProblem, DiscreteProblem]
+            if (ps == ps_valid) && (u0 == u0_valid)
+                XProblem(rn, u0, (0.0, 1.0), ps); @test true;
+            else
+                # Several of these cases do not throw errors (https://github.com/SciML/ModelingToolkit.jl/issues/2624).
+                @test_broken false
```
TorkelE, line 228: Basically a hack so that we get test broken results in the log (since some tests in the loop passes and some fail, cannot just use `test_broken`

test/meta/mtk_structure_indexing.jl
```diff
@@ -0,0 +1,389 @@
...
+    @test remake(jprob; p = [rn.p1 => 5.0]).massaction_jump.scaled_rates[1] == 10.0
+    @test remake(jprob; p = [:p1 => 6.0]).massaction_jump.scaled_rates[1] == 12.0
+    @test remake(jprob; p = [p1 => 4.0, p2 => 3.0]).massaction_jump.scaled_rates[1] == 12.0
+    @test remake(jprob; p = [rn.p1 => 5.0, rn.p2 => 4.0]).massaction_jump.scaled_rates[1] == 20.0
+    @test remake(jprob; p = [:p1 => 6.0, :p2 => 5.0]).massaction_jump.scaled_rates[1] == 30.0
+
+    # Checks that updating an integrators parameter values does not affect mass action rate until after
+    # `reset_aggregated_jumps!` have been applied as well (wt which point the correct rate is achieved).
+    jint.ps[p1] = 4.0
+    @test jint.cb.condition.ma_jumps.scaled_rates[1] != 8.0
```
isaacsas, line 361: It might be a bit better to check the previous value is unchanged, so we know the exact expected behavior is happening.

test/runtests.jl
```diff
@@ -57,4 +56,8 @@ using SafeTestsets
     @time @safetestset "HomotopyContinuation Extension" begin include("extensions/homotopy_continuation.jl") end
     @time @safetestset "Structural Identifiability Extension" begin include("extensions/structural_identifiability.jl") end
     
+    ### Upstream SciML and DiffEq tests. ###
+    @time @safetestset "MTK Structure Indexing" begin include("meta/mtk_structure_indexing.jl") end   
+    @time @safetestset "MTK Problem Inputs" begin include("meta/mtk_problem_inputs.jl") end
```
isaacsas, line 61: I'd put these higher up in the listing; after we test solution accuracy maybe?

TorkelE: Yeah, I will move them up there

test/meta/mtk_structure_indexing.jl
TorkelE: Yes, that makes sense. I will update it to do that.

PR #815, opened 2024-05-09
Coupled eqs doc expansion

TorkelE:
Create new doc for coupled equations

### DIFF
--- a/docs/src/model_creation/coupled_crn_equations.md
+++ b/docs/src/model_creation/coupled_crn_equations.md
@@ -35,7 +35,7 @@ nothing # hide
 We can check the ODE this model would generate [using Latexify](@ref ref):
 ```@example coupled_eqs_basic_example
 using Latexify
-latexify(coupled_crn; form=:ode)
+latexify(coupled_crn; form = :ode)
 ```
 Finally, the model can be simulated using standard syntax (while providing initial conditions for both $V$ and $G$):
 ```@example coupled_eqs_1


docs/pages.jl
```diff
@@ -9,7 +9,7 @@ pages = Any[
         "model_creation/dsl_description.md",
         # DSL - Advanced
         "model_creation/programmatic_CRN_construction.md",
-        "model_creation/constraint_equations.md",
+        "model_creation/coupled_crn_equations.md",
```
TorkelE, line 12: The old file is still around. Do you think it would make sense to turn it into an example of how to combine equations + events + CRN models? It is a bit more straightforward with the examples, while this new one simply goes through all options one by one.

docs/src/model_creation/coupled_crn_equations.md
```diff
@@ -0,0 +1,438 @@
...
+u0 = [X => 0.1, A => 1.0, D(A) => 1.0]
+tspan = (0.0, 10.0)
+p = [:p => 1.0, :d => 0.5, :ω => 2.0]
+oprob = ODEProblem(higher_order_model, u0, tspan, p; structural_simplify = true)
+sol = solve(oprob)
+plot(sol)
+```
+We note that while modelling of higher-order coupled CRN/differential equation models using the DSL is a bit convoluted (due to the requirement of re-declaring the differential outside of the DSL), this is not the case for [programmatic modelling](@ref coupled_equations_programmatic). Hence, for these models, this is instead the recommended approach.
+
+## [Coupled CRN/equation models for programmatic modelling](@id coupled_equations_programmatic)
```
TorkelE, line 376: Right now, I start with going through how coupling work with the DSL. Then, at the end I have a separate section describing the details one needs to change when using a programmatic workflow (here and most other cases it is straightforward, except for the base syntax the two approaches are really similar).

TorkelE: How to compose `ODESystem`s with `ReactionSystem`s will be in the compositional modelling part (or at least I think that was our plan when we discussed this a while ago, still happy to change things around)

PR #816, opened 2024-05-09
Programmatic modelling doc update

TorkelE:
update docs for programmatic modelling


docs/src/model_creation/programmatic_CRN_construction.md
```diff
@@ -1,251 +0,0 @@
-# [Programmatic Construction of Symbolic Reaction Systems](@id programmatic_CRN_construction)
```
TorkelE, line 1: The difference between this and the new file too large for displaying them side by side is not helpful. I have no opinions on filename though, just used a new one to differentiate the new file.

docs/src/model_creation/programmatic_modelling.md
```diff
@@ -0,0 +1,530 @@
...
+    @reaction $d_rate, $X3 --> 0
+]
+nothing # hide 
+```
+
+!!! note
+    When using interpolation, expressions like `2$spec` won't work; the multiplication symbol must be explicitly included like `2*$spec`.
+
+
+## [Additional options for programmatic model creation](@id programmatic_modelling_additional_options)
```
TorkelE, line 508: Anything missing here? The other parts that I can remember (events, equations) are described in the specific sections for these.

PR #818, opened 2024-05-09
Catalyst introduction doc remake

TorkelE:
updates the catalyst intro doc


docs/src/introduction_to_catalyst/introduction_to_catalyst.md
```diff
@@ -1,301 +1,166 @@
...
+mm_model = @reaction_network begin
+ kB, S + E --> SE
+ kD, SE --> S + E
+ kP, SE --> P + E
+end
 ```
-Assuming [Graphviz](https://graphviz.org/) is installed and command line
-accessible, within a Jupyter notebook we can also graph the reaction network by
-```julia
-g = Graph(repressilator)
```
TorkelE, line 70: Last time we discussed I think we said that it made sense to move the graph visualisation out of the intro tutorial, and instead have a specific section.

docs/src/introduction_to_catalyst/introduction_to_catalyst.md
```diff
@@ -1,301 +1,166 @@
...
+```@example intro_1
+u0 = [:S => 99, :I => 1, :R => 0]
+tspan = (0.0, 20.0)
+ps = [:β => 0.025, :γ => 0.2]
+oprob = ODEProblem(sir_model, u0, tspan, ps)
 ```
-Alternatively, we can use ModelingToolkit-based symbolic species variables to
-specify these mappings like
-```@example tut1
-t = default_t()
```
TorkelE, line 120: All MTK and internal stuff is moved to the thrid intro page (Advanced introduction to Catalyst, or something like that, not sure what we think the best name is)

docs/src/introduction_to_catalyst/introduction_to_catalyst.md
```diff
@@ -327,32 +185,20 @@ k X^2 Y^3
 and the ODE model
 ```math
 \begin{align*}
-\frac{dX}{dt} &=  -2 k X^2 Y^3, &
-\frac{dY}{dt} &=  -3 k X^2 Y^3, &
+\frac{dX}{dt} &= -2 k X^2 Y^3, &
+\frac{dY}{dt} &= -3 k X^2 Y^3, &
```
TorkelE, line 189: Not sure if we maybe shoudl add CLE equations after here as well.

TorkelE:
The plan is to basically split this one. This one then because a pure intro to Catalyst that do not really assume that the user have read anything else (but are familiar with basic Julai syntax). Next, we have a "Advanced introduction to Catalyst" or "Introduction to Catalyst and its internals", or something (not really sure of its name) where we go through stuff like how system conversions work, symbolic representations, and give examples of the various types of modelling approaches (DSL, Programmatic, Compositional).

docs/src/introduction_to_catalyst/introduction_to_catalyst.md
```diff
@@ -1,301 +1,168 @@
...
+Before starting this tutorial, we must (unless this has already been done) [install and activate](@ref catalyst_for_new_julia_users_packages) the Catalyst package:
 ```julia
-latexify(repressilator)
+using Pkg
+Pkg.add("Catalyst")
+using Catalyst
+```
+
+## [Creating basic Catalyst models](@id introduction_to_catalyst_model_creation)
+Catalyst exports the [`@reaction_network`](@ref) [macro]([macro](https://docs.julialang.org/en/v1/manual/metaprogramming/#man-macros)), which provides the main way of creating CRN models (an alternative, and in many ways more potent, approach is described [later](@ref programmatic_CRN_construction)). This is followed by a `begin ... end` block which encapsulates the model's reactions. E.g here we create a simple [birth-death model](@ref ref) (where a single species $X$ is both produced and degraded):
```
TorkelE, line 12: There is quite a few `(@ref ref)` here that will be broken links if someone click on them (I can add them in once the other docs are merged as well as this one).

docs/src/introduction_to_catalyst/introduction_to_catalyst.md
```diff
@@ -1,301 +1,168 @@
...
-symbolic ODEs. We could instead pass `odesys` directly like
-```@example tut1
-odesys = complete(odesys)
-oprob2 = ODEProblem(odesys, u₀symmap, tspan, psymmap)
-nothing   # hide
+Next, we import the [StochasticDiffEq.jl](https://github.com/SciML/StochasticDiffEq.jl) package (which is used for all SDE simulations in Julia) and use it to simulate our SDE. 
+```@example intro_1
+using StochasticDiffEq
+sol = solve(sprob, STrapezoid())
+sol = solve(sprob, STrapezoid(); seed = 1234) # hide
```
TorkelE, line 85: Generally, I pu hidden seed simulations throughout the docs so that they look the same each time they are built. However, I agree that we should not depend on this. For all cases I have done this, I go through and ensure that the stochastic behaviour is more or less the same (e.g. I would not repeat what we had previously and have a workflow that depends on a seed in the early stage, but which might give a different result if this is modified).

docs/src/introduction_to_catalyst/introduction_to_catalyst.md
TorkelE: (the dedicated model visualisation doc page is written and should be merged as soon as the tests can be run again)

docs/src/introduction_to_catalyst/introduction_to_catalyst.md
```diff
@@ -306,53 +173,80 @@ k \frac{X^2}{2!} \frac{Y^3}{3!} \\
...
+\frac{dY}{dt} &= -3 k X^2 Y^3, &
 \frac{dZ}{dt} &= k X^2 Y^3.
 \end{align*}
 ```
 
----
-## Notes
-1. For each of the preceding models we converted the `ReactionSystem` to, i.e.,
-   ODEs, jumps, or SDEs, we had two paths for conversion:
+## [Catalyst represents its models symbolically, and these can be generated programmatically](@id introduction_to_catalyst_sym_prog)
```
TorkelE, line 196: This example is probably a bit long, but I thought it was really nice to be able to show of both automatic generation using programmatic modelling, _and_ how to utilise symbolic representations.

docs/src/introduction_to_catalyst/introduction_to_catalyst.md
```diff
@@ -306,53 +173,80 @@ k \frac{X^2}{2!} \frac{Y^3}{3!} \\
...
+tspan = (0.0, 10.)
+ps = [K => 0.1, n => 3, d => 0.5]
+oprob = ODEProblem(repressilator, u0, tspan, ps)
+sol = solve(oprob, Tsit5())
+plot(sol)
+```
+
+If we consider the code above, we use the species and parameters (e.g. `K` and `X`) to designate values and create our reactions. This is possible because the `@species` and `@parameters` macros create these as [*symbolic variables*](@ref ref). This enables us to form algebraic expressions using them. E.g. here we create an expression describing $X$'s production rate:
+```@example intro_prog_sym
+X_prod = K/(K + Z^3)
```
TorkelE, line 232: I think this part is quite nice in showing of how the symbolic representation is useful. Originally I used normal hill functions to model the repressilator, but then the print-out of this was just
```
hillr(X, v, K, n)
```
which was kind of underwhelming. This gave a better result (and is also a fair way to model a repressilator)

docs/src/introduction_to_catalyst/introduction_to_catalyst.md
```diff
@@ -306,53 +173,80 @@ k \frac{X^2}{2!} \frac{Y^3}{3!} \\
...
-   SDEs, including options to build functions for evaluating Jacobians and/or
-   multithreaded versions of derivative evaluation functions. See the options
-   for
-   [`ODEProblem`s](https://mtk.sciml.ai/dev/systems/ODESystem/#DiffEqBase.ODEProblem)
-   and
-   [`SDEProblem`s](https://mtk.sciml.ai/dev/systems/SDESystem/#DiffEqBase.SDEProblem).
 
 ---
-## References
+## [References](@id introduction_to_catalyst_references)
```
TorkelE, line 251: Generally, I just put reference tags on all sections now, even if not used. It is not strictly required (especially this one would surprise me). However, from the previous year of working across multiple branches, it sometimes lead to annoying merge things where you modify other files when you write a tutorial. This way, if you write a new tutorial and have to reference some other section, you do not have to worry about modifying other files.

PR #822, opened 2024-05-12
Complex balancing

vyudu:
## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
- [x] Any new documentation only uses public API
  
## Additional context

A first pass at the complex-balance implementation. Does not seem like Graphs.jl has a matrix-tree implementation, so I have the matrix tree implementation here for now, but I'll work on adding it as a PR to Graphs.jl (and in particular implementing a method of generating the spanning trees that doesn't simply generate every combination of n-1 edges).


isaacsas (changes requested):
Let me start by saying this is a great first PR!

Here are a first set of comments -- these are more general Julia practices / Catalyst internals comments than method specifc or performance specific. 

I need to finish going through the paper and then I will give a second review more focused on the algorithmic details, but you can at least address these for now.

TorkelE:
I second that GitKraken is a really nice tool, if you haven't checked it out yet.

src/networkapi.jl
```diff
@@ -1652,3 +1652,138 @@ function validate(rs::ReactionSystem, info::String = "")
...
+    for v in 1:n
+        rootedTrees = [reverse(Graphs.bfs_tree(t, v, dir=:in)) for t in trees]
+        π[v] = sum([treeweight(t) for t in rootedTrees])
+    end
+    
+    # sum the contributions
+    return π 
+end
+
+function massactionrate(rs::ReactionSystem, rxn_idx::Int, dir::Int, conc::Vector)
```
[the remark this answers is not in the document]
vyudu: Ah, this was an artifact of a previous implementation that I forgot to delete. Good to know about those two functions, though, will implement the latter check

src/networkapi.jl
```diff
@@ -1652,3 +1652,138 @@ function validate(rs::ReactionSystem, info::String = "")
...
+"""
+    complexbalanced(rs::ReactionSystem, rates::Vector)
+
+Constructively compute whether a network will have complex-balanced equilibrium
+solutions, following the method in [this paper](https://link.springer.com/article/10.1007/s10910-015-0498-2#Sec3). 
+"""
+
+using MetaGraphs, Combinatorics, LinearAlgebra
+
+function complexbalanced(rs::ReactionSystem, rates::Vector)
```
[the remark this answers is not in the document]
vyudu: I think I made this change correctly, but let me know. Fortunately it seemed that I don't actually need to use the fixed parameter vector directly, just the mapping.

test/network_analysis/network_properties.jl
```diff
@@ -37,6 +37,9 @@ let
     @test isweaklyreversible(MAPK, subnetworks(MAPK)) == false
     cls = conservationlaws(MAPK)
     @test Catalyst.get_networkproperties(MAPK).rank == 15
+
+    rates = rand(numparams(MAPK))
```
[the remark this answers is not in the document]
vyudu: Ditto on this - let me know if how I'm constructing the parameter maps in the test file in the new commit is correct.

Project.toml
```diff
@@ -14,6 +15,7 @@ LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
+MetaGraphs = "626554b9-1ddb-594c-aa3c-2596fe9399a5"
```
isaacsas, line 18:
```suggestion
```
Can you remove the MetaGraphs dependency here?

src/networkapi.jl
```diff
@@ -1481,5 +1481,158 @@ function validate(rs::ReactionSystem, info::String = "")
...
+        error("The number of reaction rates must be equal to the number of parameters")
+    end
+
+    sm = speciesmap(rs)
+    cm = reactioncomplexmap(rs)
+    complexes, D = reactioncomplexes(rs)
+    rxns = reactions(rs)
+    nc = length(complexes); nr = numreactions(rs); nm = numspecies(rs)
+
+    if !all(r->ismassaction(r, rs), rxns) error("Not mass action") end
```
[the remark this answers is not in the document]
isaacsas: Also, currently `ismassaction` would still declare that something like `k1*k2, A --> B` is mass action, but this would break your code here. After checking all reactions are mass action, what you need to do is something like
```julia
ratevalues = [substitute(rate, rates) for rate in reactionrates(rs)]
```
This should numerically evaluate `k1*k2` using the substitution vector/dict the user provided. It isn't the most efficient way to do this, but is fine for the first version of this code.

src/networkapi.jl
```diff
@@ -1481,5 +1481,158 @@ function validate(rs::ReactionSystem, info::String = "")
...
+Constructively compute whether a network will have complex-balanced equilibrium
+solutions, following the method in [this paper](https://link.springer.com/article/10.1007/s10910-015-0498-2#Sec3). Accepts a map of rates [k1 => 1.0, k2 => 2.0,...]k2
+"""
+
+using Combinatorics, LinearAlgebra
+
+function iscomplexbalanced(rs::ReactionSystem, rates::Dict{Any, Float64}) 
+    if length(rates) != numparams(rs) 
+        error("The number of reaction rates must be equal to the number of parameters")
+    end
```
isaacsas, line 1496:
```suggestion
```
See below -- this isn't how we should do the check I think since even for mass action reactions rates may not be just a single symbolic.

src/networkapi.jl
```diff
@@ -1481,5 +1481,158 @@ function validate(rs::ReactionSystem, info::String = "")
...
+"""
+    iscomplexbalanced(rs::ReactionSystem, rates::Vector)
+
+Constructively compute whether a network will have complex-balanced equilibrium
+solutions, following the method in [this paper](https://link.springer.com/article/10.1007/s10910-015-0498-2#Sec3). Accepts a map of rates [k1 => 1.0, k2 => 2.0,...]k2
+"""
+
+using Combinatorics, LinearAlgebra
+
+function iscomplexbalanced(rs::ReactionSystem, rates::Dict{Any, Float64})
```
isaacsas, line 1493: This should work for Vector/tuple mappings too as we commonly suggest them for users, along with specifying `Symbols` for the first component. You can use `symmap_to_varmap` to convert `Symbol` mappings to symbolic mappings, see https://docs.sciml.ai/Catalyst/stable/api/catalyst_api/#Catalyst.symmap_to_varmap.

src/networkapi.jl
```diff
@@ -1481,5 +1481,158 @@ function validate(rs::ReactionSystem, info::String = "")
...
+#     end
+# 
+#     for (i, e) in collect(enumerate(edges(rwg)))
+#         rxn = rxns[i]
+#         set_prop!(rwg, Graphs.src(e), Graphs.dst(e), :reaction, rxn)
+#         set_prop!(rwg, Graphs.src(e), Graphs.dst(e), :rate, rates[rxn.rate])
+#     end
+# 
+#     rwg
+# end
```
isaacsas, line 1563:
```suggestion
```

src/networkapi.jl
```diff
@@ -1481,5 +1481,158 @@ function validate(rs::ReactionSystem, info::String = "")
...
+
+function ratematrix(rs::ReactionSystem, rates::Dict{Any, Float64}) 
+    if length(rates) != numparams(rs) 
+        error("The number of reaction rates must be equal to the number of parameters")
+    end
```
isaacsas, line 1568:
```suggestion
```
See comments above about how to numerically calculate the rates (I'd suggest validating and calculating them above, and then just passing the final numeric rate vector into this function. It may help speedup this function by passing in a more concrete type for `rates`, i.e. one that doesn't depend on `Any`).

src/networkapi.jl
```diff
@@ -1481,5 +1481,158 @@ function validate(rs::ReactionSystem, info::String = "")
...
+
+    complexes, D = reactioncomplexes(rs)
+    n = length(complexes)
+    rxns = reactions(rs)
+    ratematrix = zeros(n, n)
+
+    for r in 1:length(rxns)
+        rxn = rxns[r]
+        s = findfirst(x->x==-1, D[:,r])
+        p = findfirst(x->x==1, D[:,r])
```
isaacsas, line 1578:
```suggestion
        s = findfirst(==(-1), @view D[:,r])
        p = findfirst(==(1), @view D[:,r])
```
This avoids allocating a copy vector for each column.

src/networkapi.jl
isaacsas: I would just not type `rates` for simplicity, but in the docstring mention the input options (vectors, tuples, dict mappings as accepted by ModelingToolkit for specifying parameter values).

src/networkapi.jl
```diff
@@ -1481,5 +1481,158 @@ function validate(rs::ReactionSystem, info::String = "")
     validated
 end
 
+"""
+    iscomplexbalanced(rs::ReactionSystem, rates::Vector)
```
isaacsas, line 1485:
```suggestion
    iscomplexbalanced(rs::ReactionSystem, rates)
```

src/networkapi.jl
```diff
@@ -1481,5 +1481,158 @@ function validate(rs::ReactionSystem, info::String = "")
...
+
+# TODO: implement Winter's algorithm for generating spanning trees 
+function spanningtrees(g::SimpleGraph) 
+        
+end
+
 # Checks if a unit consist of exponents with base 1 (and is this unitless).
-unitless_exp(u) = istree(u) && (operation(u) == ^) && (arguments(u)[1] == 1)
\ No newline at end of file
+unitless_exp(u) = istree(u) && (operation(u) == ^) && (arguments(u)[1] == 1)
```
isaacsas, line 1637: This should go before your added code since it goes with the unit validation above.

isaacsas (changes requested):
Can you add one more test that uses a network that is complex balanced with multiple nonlinearities and non-trivial rate expressions (i.e. reactions like `k1*k2, 3A + 2B --> 3*C`)? Just to stress the code a bit more.

In terms of performance optimization I think that can be a followup. So I'm happy to merge once you address this last set of comments.

src/networkapi.jl
isaacsas: Also, probably it is more descriptive to call the second argument `parametermap` or similar as it should be a mapping giving numeric values for all system parameters.

src/networkapi.jl
isaacsas: Inside the function you are welcome to convert the inputted mapping to a dictionary for ease of use if you want (and needed). It might make sense to just have three dispatches
```julia
function iscomplexbalanced(rs::ReactionSystem, parametermap::Vector)
     pdict = Dict(parametermap)
     iscomplexbalanced(rs, pdict)
end

function iscomplexbalanced(rs::ReactionSystem, parametermap::Tuple)
     pdict = Dict(parametermap)
     iscomplexbalanced(rs, pdict)
end

function iscomplexbalanced(rs::ReactionSystem, parametermap::Dict)
    your existing code
end

iscomplexbalanced(rs::ReactionSystem, parametermap) = error("parametermap is not in an allowable format.)
```
or similar.

src/networkapi.jl
```diff
@@ -1481,5 +1481,158 @@ function validate(rs::ReactionSystem, info::String = "")
...
+    end
+
+    # generate all spanning trees
+    ug = SimpleGraph(SimpleDiGraph(g))
+    trees = collect(Combinatorics.combinations(collect(edges(ug)), n-1))
+    trees = SimpleGraph.(trees)
+    trees = filter!(t->isempty(Graphs.cycle_basis(t)), trees)
+    # trees = spanningtrees(g)
+
+    # constructed rooted trees for every edge, compute sum
```
isaacsas, line 1612:
```suggestion
    # constructed rooted trees for every vertex, compute sum
```

src/networkapi.jl
```diff
@@ -1481,5 +1481,158 @@ function validate(rs::ReactionSystem, info::String = "")
...
+    prod = 1
+    for e in edges(t)
+        s = Graphs.src(e); t = Graphs.dst(e)
+        prod *= distmx[s, t] 
+    end
+    prod 
+end
+
+# TODO: implement Winter's algorithm for generating spanning trees 
+function spanningtrees(g::SimpleGraph)
```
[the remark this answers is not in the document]
vyudu: Folded this into a future PR

src/networkapi.jl
vyudu: I've changed it to the following - does this solve this concern? 

```julia
if length(parametermap) != numparams(rs) error... end
```

isaacsas: Sure, that is fine to check.

isaacsas:
CI is still failing?

isaacsas:
This looks like it might have been rebased on an older version of master?

isaacsas:
(Generally I find it easier to merge instead of rebase to update a PR.)

TorkelE:
@vyudu I tried to merge this onto master for you, but couldn't get it to work unfortunately. If you check the "Files changed" tab at the top of the PR you see that there are a very large number of changes:
image

What probably have happened is that this was a PR of some branch, and then it got rebased on an updated master, and somewhere some of the changes of the branches it originally was branched off of was carried through when this one was rebased on the master again.

Since you are developing on your own master branch, I would try (as Sam pointed out here, and to me earlier) to simply merge your master branch with Catalyst's master branch, and then push the PR again. If that does not work out, and the changes introduced here are very straightforward (i.e. just adds functions at the bottom of the network_analysis.jl file, and some tests) it might be easiest to simply copy these into a new, clean, branch and make a new PR from that.

src/Catalyst.jl
```diff
@@ -6,9 +6,10 @@ module Catalyst
 using DocStringExtensions
 using SparseArrays, DiffEqBase, Reexport, Setfield
 using LaTeXStrings, Latexify, Requires
-using JumpProcesses: JumpProcesses, JumpProblem, 
-                     MassActionJump, ConstantRateJump, VariableRateJump,
-                     SpatialMassActionJump
+using LinearAlgebra, Combinatorics
+using JumpProcesses: JumpProcesses,
+                     JumpProblem, MassActionJump, ConstantRateJump,
+                     VariableRateJump
```
isaacsas, line 12:
```suggestion
using LinearAlgebra, Combinatorics
using JumpProcesses: JumpProcesses, JumpProblem, 
                     MassActionJump, ConstantRateJump, VariableRateJump,
                     SpatialMassActionJump
```

isaacsas:
Thanks @vyudu great first PR.

TorkelE:
> Thanks @vyudu great first PR.

Second that! And sorry for giving you poor advice on the rebase/merge thing. I tried @isaacsas's method properly now as well and it works great

PR #830, opened 2024-05-13
merge Catalyst version 14 into master

isaacsas:


isaacsas:
@TorkelE this is failing because of `==` equality tests between solution objects, which as I said a while back we really shouldn't be doing...

isaacsas:
https://github.com/SciML/Catalyst.jl/blob/893714347a29e7d74882e300650536bbaf6382e1/test/miscellaneous_tests/events.jl#L247 doesn't use `StableRng`s for the SDEProblem?

ChrisRackauckas:
But it's relative, so both solves should change if it changes, so I think that one is okay.

TorkelE:
> @TorkelE this is failing because of == equality tests between solution objects, which as I said a while back we really shouldn't be doing...

Do we know what? I can check tomorrow. Testing solutions for equality have worked for 6 years, so while moving away from it might be good, inn this case it probably suggest that something else have gone wrong as well, which we would want to know about.

> doesn't use StableRngs for the SDEProblem?

SDEProblems cannot take these. I either use fixed seeds, or have these generated through StableRNGs though.

TorkelE:
Here is a previous issue in StableRNG seeds for SDEs: https://github.com/SciML/StochasticDiffEq.jl/issues/569. Might be a while until it gets added though, not sure if anyone is actively working on SDEs.

PR #835, opened 2024-05-15
Test restructure and parallelisation

TorkelE:
Do not change any actual code. However:
- Moves around some test files to different categories (e.g. some stuff is now tested with the main reaction system structure tests).
- Moves a couple of tests to new files (so tests from old files that fit better in new ones).
- Parallelise tests.
- Slight updates to comments for clarity.


test/dsl/dsl_advanced_model_construction.jl
```diff
@@ -323,58 +330,6 @@ let
...
-    rx = Reaction(k*k2*D, [A, B], [C, C2], [E, 1], [F, 1])
-    @named ivstest = ReactionSystem([rx], s; spatial_ivs = [x])
-
-    @test ivstest == rn
-    @test issetequal(unknowns(rn), [D, E, F, A, B, C, C2])
-    @test issetequal(species(rn), [A, B, C, C2])
-    @test isequal(ModelingToolkit.get_iv(rn), s)
-    @test issetequal(Catalyst.get_sivs(rn), [x])
-end
-
```
TorkelE, line 377: These are DSL options, so I moved them to the file for DSL option tests. We probably should add some additional `@variables` tests at some point, and maybe for `@ivs` tests as well.

test/miscellaneous_tests/api.jl
```diff
@@ -300,119 +300,6 @@ let
...
-        (k2, k1), A + E <--> AE
-        (k4, k3), AE <--> B + E
-        k5, B --> 0
-        k6, 0 --> A
-    end
-    rev = false
-    weak_rev = false
-    testreversibility(rn, reactioncomplexes(rn)[2], rev, weak_rev)
-end
-
```
TorkelE, line 415: Moved to network analysis tests. There were quite a few things in the API test file that I think maybe should be moved, but was honestly unsure. Only moved this and the conservation law stuff because those where the only ones I felt very certain about.

test/model_simulation/u0_n_parameter_inputs.jl
```diff
@@ -1,74 +0,0 @@
-#! format: off
```
TorkelE, line 1: This is tested, and in waaay more depth, in the new upstream test files.

test/programmatic_model_creation/programmatic_model_expansion.jl
```diff
@@ -1,34 +0,0 @@
-#! format: off
```
TorkelE, line 1: The empty test network test was moved to the reaction system structure test. This file is pretty much deprecated when we removed mutating functions, and hence removed. 

It feels silly and bad that we end up with a folder with only a single file (the composing model test). However, I'd like to test more for hierarchical/composed systems, so we will likely add more test files here? In which case it would make sense to keep this one.

test/reactionsystem_core/reactionsystem_structure.jl
```diff
@@ -107,14 +107,6 @@ let
...
 end
 
-# Test show.
-let
-    io = IOBuffer()
-    show(io, rs)
-    str = String(take!(io))
-    @test count(isequal('\n'), str) < 30
-end
-
```
TorkelE, line 117: I have grouped `show` tests together.

test/reactionsystem_core/reactionsystem_structure.jl
```diff
@@ -651,18 +668,6 @@ let
...
-let
-    @parameters a
-    @species A(t) B(t) C(t)[1:2]
-    rx1 = Reaction(a, [A, C[1]], [C[2], B], [1, 2], [2, 3])
-    io = IOBuffer()
-    show(io, rx1)
-    str = String(take!(io))
-    @test str == "a, A + 2*(C(t))[1] --> 2*(C(t))[2] + 3*B"
-end
-
```
TorkelE, line 665: I have grouped `show` tests together.

test/reactionsystem_core/reactionsystem_structure.jl
```diff
@@ -621,14 +646,6 @@ let
...
 end
 
-# Test balanced_bc_check.
-let
-    @species A(t) [isbcspecies = true]
-    rx = @reaction k, 2 * $A + B --> C + $A
-    @test_throws ErrorException ReactionSystem([rx], t; name = :rs)
-    @named rs = ReactionSystem([rx], t; balanced_bc_check = false)
-end
-
```
TorkelE, line 631: I have grouped boundary condition tests together (still in the same file).

isaacsas (changes requested):
My only concern is the sheer number of groups and whether that has implications on test time potentially taking longer, especially as some groups have very small numbers of tests within them...

Do you have any quantitative information on how this impacts total time for all tests to run?

TorkelE:
I kind of agree. When I did it I didn't realise there would be a separate run for each group. Maybe we create separate groups (independent from the previous grouping), and just split it into 3-4 sets depending on runtimes?

isaacsas:
I think that makes more sense. Something like all the symbolic DSL/ReactionSystem stuff, all the solving stuff, and the extensions seems like a natural split. We can further split solving stuff if needed to model types or such.

TorkelE:
Agreed. For now, maybe just call stuff "TestGroup1", "TestGroup2", etc.? That way we don't have to be super committed to what goes into which group.

isaacsas:
Sure. Changing around the group names later doesn’t seem like a big deal anyways — it isn’t user facing in any way.

test/reactionsystem_core/reactionsystem_structure.jl
```diff
@@ -747,14 +747,9 @@ end
...
-    empty_network_1 = @reaction_network
-    eqs, iv, ps, name, systems = unpacksys(empty_network_1)
-    @test length(eqs) == 0
-    @test nameof(iv) == :t
-    @test length(get_unknowns(empty_network_1)) == 0
-    @test length(ps) == 0
+    empty_network = @reaction_network
+    @test length(equations(empty_network)) == 0
+    @test nameof(independent_variable(empty_network)) == :t
```
isaacsas, line 752: This is a deprecated function. More generally why have you dropped using the getters here? We need to test them too.

isaacsas: https://github.com/SciML/ModelingToolkit.jl/blob/e16c18091ac71ad45b2e87d4028e015deb70f713/src/systems/abstractsystem.jl#L319

TorkelE: Didin't realise that the test was specific to using the `get_eqs` getter instead of the `equations` getters. Figured it was just about checking the content of the empty network. If this is and important part of the test, shouldn't we just use the full set? 

Good point about `independent_variable`, I will revert that and the others.

PR #838, opened 2024-05-16
Source folder restructure

TorkelE:
Moves around functions in the src folders:
- Renamed "reaction_network.jl" to "dsl.jl".
- Split the "reactionsystem.jl" and "networkapi.jl" files into 4 files (one for the `Reaction` struct, one for the `ReactionSystem` struct, one for conversion, and one for network analysis).
- Moved some constant declarations from the "dsl.jl" file to the "Catalyst.jl" file.
- Moved some functions from the "dsl.jl" file to the "expression_utils.jl" file.
- In many files, reorder function order to group them more naturally.
- Added some comments and section comments (generally very minor).


src/network_analysis.jl
```diff
@@ -0,0 +1,702 @@
...
+            !isconstant(product) && push!(specset, product)
+        end
+    end
+    specs = collect(specset)
+    newps = Vector{eltype(p)}()
+    for rx in rxs
+        Symbolics.get_variables!(newps, rx.rate, p)
+    end
+    rxs, specs, newps   # reactions and species involved in reactions of subnetwork
+end
```
TorkelE, line 416: This was the kind of situation I mean when I said that I thought having subfunctions first and main functions second was confusing. In a file like this where almost everything has docstrings, when I look through it (on a first glance) looks like this function is grouped with the previous one (`deficiency`), and not with the subsequent one (`subnetworks`). In other situations I don't mind as much.

src/expression_utils.jl
isaacsas, line 1: I'm not sure I see the point of keeping this file around. This had been down to one function -- why not move it all into the dsl file, which is the only place these functions are used? The decision on what is a utility vs. non-utility seems somewhat arbitrary to me.

src/networkapi.jl
isaacsas, line 1: Is this file deleted? It seems just empty from my Github view but I am not sure.

src/reaction_structure.jl
isaacsas, line 1: I would just calls this `reactions.jl` since it encompasses more than just the `Reaction` type (i.e. functionality associated with reactions too). Also, "structure" is often confusing to use in talking about Julia code as one has to be clear about if one means the structure in the sense of a Julia `struct` vs. the structure of something -- in a file name it is extra confusing. (I make this later comment as I notice in many docs you now say reaction structure or reaction system structure, and generally I think this is ambiguous language we should avoid as it can be confusing which of the two options one means.)

isaacsas, line 1: Seems like the ordering in this file has still been changed from what it previously was in `ReactionSystem.jl`.

src/reaction_structure.jl
```diff
@@ -0,0 +1,545 @@
...
+    print(io, rx.rate, ", ")
+    print_rxside(io, rx.substrates, rx.substoich)
+    arrow = rx.only_use_rate ? "⇒" : "-->"
+    print(io, " ", arrow, " ")
+    print_rxside(io, rx.products, rx.prodstoich)
+end
+
+function print_rxside(io::IO, specs, stoich)
+    # reactants/substrates
+    if isempty(specs)
```
isaacsas, line 258: Order

src/reaction_structure.jl
```diff
@@ -0,0 +1,545 @@
...
+                print(io, stoich[i], "*", prspec)
+            end
+
+            (i < length(specs)) && print(io, " + ")
+        end
+    end
+    nothing
+end
+
+"""
```
isaacsas, line 278: If you want a comment "title" for the following functions (and I'd move `show` here too) you could call this 

```julia
# dispatches for Base functions.
```

src/reaction_structure.jl
```diff
@@ -0,0 +1,545 @@
...
+    end
+    Reaction(rate, subs, prods, substoich, prodstoich, netstoich, rx.only_use_rate, rx.metadata)
+end
+
+function apply_if_nonempty(f, v)
+    isempty(v) && return v
+    s = similar(v)
+    map!(f, s, v)
+    s
+end
```
isaacsas, line 338: order

src/reactionsystem_structure.jl
isaacsas, line 1: Similar comment to above -- I would just keep this named "reactionsystem" since it encompasses more than the structure and avoids confusion about what "structure" means.

src/reactionsystem_conversions.jl
```diff
@@ -0,0 +1,900 @@
...
+              " when there are constraint equations.")
+    nothing
+end
+
+
+### Utility ###
+
+# Function returning `true` for species which shouldn't change from the reactions, 
+# including non-species variables.
+drop_dynamics(s) = isconstant(s) || isbc(s) || (!isspecies(s))
```
isaacsas, line 437: Please place before `assemble_oderhs` where it is first used.

src/reactionsystem_conversions.jl
```diff
@@ -0,0 +1,900 @@
...
+            for str in view(strs, 2:length(strs))
+                var = getproperty(var, Symbol(str), namespace = true)
+            end
+        else
+            throw(ArgumentError("System $(nameof(sys)): variable $sym does not exist"))
+        end
+    end
+    var
+end
+
```
isaacsas, line 873: order

src/reactionsystem_structure.jl
```diff
@@ -0,0 +1,1438 @@
...
+function MT.unknowns(sys::ReactionSystem)
+    sts = get_unknowns(sys)
+    systems = get_systems(sys)
+    if !isempty(systems)
+        sts = unique!([sts; reduce(vcat, namespace_variables.(systems))])
+        sort!(sts; by = !isspecies)
+        return sts
+    end
+    return sts
+end
```
isaacsas, line 606: These should be after the `get_*` functions as they are higher level and build from using `get_*` functions.

isaacsas: That is also the order I had before...

src/reactionsystem_structure.jl
```diff
@@ -0,0 +1,1438 @@
...
+any subsystems that are of type `ReactionSystem`. To get the species and non-species
+variables in the system and all subsystems, including non-`ReactionSystem` subsystems, uses
+`unknowns(network)`.
+
+Notes:
+- If `ModelingToolkit.get_systems(network)` is non-empty will allocate.
+"""
+function species(network)
+    sts = get_species(network)
+    systems = filter_nonrxsys(network)
```
isaacsas, line 679: where is `filter_nonrxsys`? it should precede this function.

src/reactionsystem_structure.jl
```diff
@@ -0,0 +1,1438 @@
...
+    isempty(systems) && return sts
+    unique([sts; reduce(vcat, map(sys -> species(sys, species(sys)), systems))])
+end
+
+# Special species but which take a set of states and names spaces them according to another
+# `ReactionSystem`.
+function species(network::ReactionSystem, sts)
+    [MT.renamespace(network, st) for st in sts]
+end
```
isaacsas, line 688: order...

isaacsas (changes requested):
Seems like there are still some places that functions are got moved till after they were called (sometimes into a different of the new sections).

src/reaction_structure.jl
isaacsas: Generally when I want to refer to the struct I would just say `ReactionSystem` and/or `Reaction` with an appropriate link to the struct's definition in docs. That seems more clear about what we are referring to.

src/Catalyst.jl
```diff
@@ -69,6 +69,26 @@ function __init__()
...
+
+# Declares symbols which may neither be used as parameters nor unknowns.
+const forbidden_symbols_skip = Set([:ℯ, :pi, :π, :t, :∅])
+const forbidden_symbols_error = union(Set([:im, :nothing, CONSERVED_CONSTANT_SYMBOL]),
+                                      forbidden_symbols_skip)
+const forbidden_variables_error = let
+    fvars = copy(forbidden_symbols_error)
+    delete!(fvars, :t)
+    fvars
+end
```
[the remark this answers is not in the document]
TorkelE: `forbidden_symbols_error` is used in the DSL, expression utils (which we might move to dsl), the main reaction system files, and in the spatial modelling files. `forbidden_variables_error` seemed relevant enough that it probably should be with the other two.

src/latexify_recipes.jl
```diff
@@ -174,27 +134,74 @@ function chemical_arrows(rn::ReactionSystem; expand = true,
...
+        any_nonrx_subsys(subsys) && (return true)
     end
-    error("Unrecognised form argument given: $form. This should be either reactions (default), :ode, or :sde.")
+    false
 end
 
-function Latexify.infer_output(env, rs::ReactionSystem, args...)
-    env in [:arrows, :chem, :chemical, :arrow] && return chemical_arrows
+# Recursively traverses an expression and removes things like X^1, 1*X. Will not actually have any effect on the expression when used as a function, but will make it much easier to look at it for debugging, as well as if it is transformed to LaTeX code.
+function recursive_clean!(expr)
```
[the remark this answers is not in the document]
TorkelE: `recursive_clean!` is used by some Latexify functions.

src/expression_utils.jl
TorkelE: Thing here are used by the DSL, chemistry, and spatial modelling files (pretty much everywhere where we implement a macro).

However, I do agree that this is (and definitely was) pretty much an appendix to the DSL file. I tried to sort it so that stuff here is generally relevant to expression handling, while the dsl file stuff is things that are directly related to the DSL. There is functions here though that are only used by the DSL file.

src/networkapi.jl
TorkelE: When I check it out now at homeit is not in the repo, but I will double check again before merging.

src/reaction_structure.jl
TorkelE: that makes sense, I agree.

TorkelE: Do you suggest we call the file "ReactionSystem.jl" or "reactionsystem.jl"?

src/reaction_structure.jl
TorkelE: will move back

src/reaction_structure.jl
TorkelE: That is a good one, will use that

src/reaction_structure.jl
TorkelE: will move

src/reactionsystem_conversions.jl
TorkelE: can do

src/reactionsystem_conversions.jl
TorkelE: will move

src/reactionsystem_structure.jl
TorkelE: Can do. However, in this case I thought it would make more sense to have MTK-specific getters before Catalyst-specific ones.

src/reactionsystem_structure.jl
TorkelE: My impression was that this was meant as a general API function as well (although we technically do not export it I guess). I figured that it made sense to roughly sore the network API functions from standard to obscure, and those felt like it was at the far end. But if you want it internal I can move it up.

src/reactionsystem_structure.jl
TorkelE: Figured they were both species functions, so I had this order, but can move if you want.

src/reaction_structure.jl
isaacsas: Lower case since upper case usually means the module name file in Julia.

src/reactionsystem_structure.jl
isaacsas: The `get_*` functions are more low-level compared to something like `unknowns` which is doing more than a direct field access (it aggregates across systems). Functions like `unknowns` are higher level and implemented in terms of getters, so it makes sense to define getters first (by my normal convention I was using before).

src/reaction_structure.jl
TorkelE: Right, yeah, I agree 100% with that.

src/reactionsystem_structure.jl
isaacsas: I’d still prefer to stick with the convention of functions appearing before their use as I had previously.

src/reactionsystem_structure.jl
isaacsas: Order of use. This version is called by the other version so should precede it.

src/expression_utils.jl
isaacsas: Ok, if these are primarily used in multiple places factoring them out seems reasonable.

PR #851, opened 2024-05-20
Full system balancing

TorkelE:
Creates a function `balance_system`, which takes a `ReactionSystem` and returns a version with all its reactions balanced.

```julia
rs = @reaction_network begin
    @species C(t) O(t) H(t)
    @compounds begin
        H2(t) = 2H
        CH4(t) = C + 4H
        O2(t) = 2O
        CO2(t) = C + 2O
        H2O(t) = 2H + O
    end
    1.0, C + H2 --> CH4
    2.0, CH4 + O2 --> CO2 + H2O
end
rs_balanced = balance_system(rs)
```

### DIFF
--- a/src/chemistry_functionality.jl
+++ b/src/chemistry_functionality.jl
@@ -346,3 +346,16 @@ function create_matrix(reaction::Catalyst.Reaction)
 
     return A
 end
+
+"""
+    balance_system(rs::ReactionSystem)
+
+From a system, creates a new system where each reaction is a balanced version of the corresponding
+reaction of the original system. For more information, consider the `balance_reaction` function
+(which is internally applied to each system reaction).
+"""
+function balance_system(rs::ReactionSystem)
+    @set! rs.eqs = [(eq isa Reaction) ? balance_reaction(eq) : eq for eq in get_eqs(rs)]
+    @set! rs.rxs = [balance_reaction(rx) for rx in get_rxs(rs)]
+    return rs
+end
--- a/src/chemistry_functionality.jl
+++ b/src/chemistry_functionality.jl
@@ -259,7 +259,7 @@ function balance_reaction(reaction::Reaction)
     end
 
     isempty(balancedrxs) && (@warn "Unable to balance reaction.")
-    (length(balancedrxs) > 1) && (@warn "Infinite balanced reactions from ($reaction) are possible, returning a basis for them. Note that we do not check if they preserve the set of substrates and products from the original reaction.")
+    (length(balancedrxs) > 1) && (@warn "The space of possible balanced versions of the reaction ($reaction) is greater than 1. This prevents the selection of a single appropriate balanced reaction. Instead, a basis for balanced reactions is returned. Note that we do not check if they preserve the set of substrates and products from the original reaction.")
     return balancedrxs
 end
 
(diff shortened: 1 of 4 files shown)


src/chemistry_functionality.jl
```diff
@@ -346,3 +346,16 @@ function create_matrix(reaction::Catalyst.Reaction)
...
 end
+
+"""
+    balance_system(rs::ReactionSystem)
+
+From a system, creates a new system where each reaction is a balanced version of the corresponding
+reaction of the original system. For more information, consider the `balance_reaction` function
+(which is internally applied to each system reaction).
+"""
+function balance_system(rs::ReactionSystem)
```
isaacsas, line 357: What does this do if there are no solutions / infinitely many solutions?

isaacsas: It seems like it ignores those cases?

TorkelE: I think we should throw an error, right?

isaacsas: If you get back multiple basis functions you should at least give a warning and pick one. If you get one you should give an error that it can’t balanced.

src/chemistry_functionality.jl
```diff
@@ -346,3 +346,45 @@ function create_matrix(reaction::Catalyst.Reaction)
...
+# or several, balanced alternatives.
+function get_balanced_reaction(rx::Reaction)
+    brxs = balance_reaction(rx)
+
+    # In case there are no, or multiple, solutions to the balancing problem. 
+    if isempty(brxs)
+        error("Could not balance reaction `$rx`, unable to create a balanced `ReactionSystem`.")
+    end
+    if length(brxs) > 1
+        error("Infinite number of balanced reactions possible for reaction ($rx) are possible. No method for automatically generating a valid reaction is currently implemented in `balance_system`.")
```
TorkelE, line 384: I was unsure how to actually construct a proper reaction from the base. Had a look at some outputs but didn't really understand. At some point we should support it, but not really a priority.

TorkelE: Although we probably could raise a issue that it should be fixed

isaacsas: We should probably update these error messages as they are imprecise (the old one too). The issue isn't an infinite number of solutions (there are always an infinite number), it is whether there are multiple independent solution vectors (i.e. the dimension of the solution space is greater than one). So the error messages should probably be updated to reflect this.

isaacsas: Modulo that comment this is fine with me.

TorkelE:
New messages:

waring:
"The space of possible balanced versions of the reaction ($reaction) is greater than 1. This prevents the selection of a single appropriate balanced reaction. Instead, a basis for balanced reactions is returned. Note that we do not check if they preserve the set of substrates and products from the original reaction."

error (for system-level balancing)
"The space of possible balanced versions of the reaction ($reaction) is greater than 1. This prevents the selection of a single appropriate balanced reaction. No method to, in this case, automatically generate a valid reaction is currently implemented in `balance_system`.""

src/chemistry_functionality.jl
```diff
@@ -259,7 +259,7 @@ function balance_reaction(reaction::Reaction)
     end
 
     isempty(balancedrxs) && (@warn "Unable to balance reaction.")
-    (length(balancedrxs) > 1) && (@warn "Infinite balanced reactions from ($reaction) are possible, returning a basis for them. Note that we do not check if they preserve the set of substrates and products from the original reaction.")
+    (length(balancedrxs) > 1) && (@warn "The space of possible balanced versions of the reaction ($reaction) is greater than 1. This prevents the selection of a single appropriate balanced reaction. Instead, a basis for balanced reactions is returned. Note that we do not check if they preserve the set of substrates and products from the original reaction.")
```
isaacsas, line 262:
```suggestion
    (length(balancedrxs) > 1) && (@warn "The space of possible balanced versions of the reaction ($reaction) is greater than one-dimension. This prevents the selection of a single appropriate balanced reaction. Instead, a basis for balanced reactions is returned. Note that we do not check if they preserve the set of substrates and products from the original reaction.")
```

isaacsas (commented):
If tests pass happy for you to merge this.

PR #852, opened 2024-05-20
Add two new reaction metadata

TorkelE:
I have added two more `Reaction` metadata that we support with getters:
- `description`
- `misc`

There are two main uses of this PR:
- These two metadata should be usable by the user to support any metadata they would like to use, without us having to support the addition of user-designated metadata.
- They are good metadata to introduce the concept of reaction metadata in the docs (rather than noise scaling, which is non-obvious what it is). This way I do not have to introduce `Reaction` metadata in the SDE tutorials.

### DIFF
--- a/src/reaction.jl
+++ b/src/reaction.jl
@@ -470,16 +470,16 @@ end
 """
 has_noise_scaling(reaction::Reaction)
 
-Checks whether a specific reaction has the metadata field `noise_scaing`. If so, returns `true`, else
-returns `false`.
+Returns `true` if the input reaction has the `noise_scaing` metadata field assigned, else `false`.
 
 Arguments:
-- `reaction`: The reaction for which we wish to check.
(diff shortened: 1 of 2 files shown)


isaacsas (commented):
Adding a 'misc' metadata seems redundant? Isn't metadata itself is the `misc` of a `Reaction`?

TorkelE:
Yes, ish.

It is redundant if we advice the users that they can add whatever metadata they want themselves. However, since we right now don't do that, I figured this was an easy workaround (without actually looking like it). I.e. whatever metadata a user wants to add they can just stick in the `misc` field.

In practise, I looked through the metadata permitted for symbolic variables and picked `misc` and `description` from that list as they seemed to make sense as easy implementation for `Reaction` metadata.

isaacsas (approved):

PR #853, opened 2024-05-20
Add `conservedquantity` metadata

TorkelE:
Add a `conservedquantity = true` metadata to all conserved quantity parameters. These permits e.g. detecting these in e.g. StrucuralIdentifiability outputs, and eliminating them from here.

### DIFF
--- a/src/network_analysis.jl
+++ b/src/network_analysis.jl
@@ -543,7 +543,7 @@ end
 ### Conservation Laws ###
 
 # Implements the `conservationquantity` parameter metadata.
-struct ConservationQuantity end
+struct ConservedParameter end
 Symbolics.option_to_metadata_type(::Val{:conservationquantity}) = ConservationQuantity
 
 """
--- a/src/network_analysis.jl
+++ b/src/network_analysis.jl
@@ -544,7 +544,7 @@ end
 
 # Implements the `conservationquantity` parameter metadata.
 struct ConservedParameter end
-Symbolics.option_to_metadata_type(::Val{:conservationquantity}) = ConservationQuantity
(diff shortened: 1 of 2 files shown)


src/network_analysis.jl
```diff
@@ -542,6 +542,23 @@ end
 
 ### Conservation Laws ###
 
+# Implements the `conservationquantity` parameter metadata.
+struct ConservationQuantity end
+Symbolics.option_to_metadata_type(::Val{:conservationquantity}) = ConservationQuantity
+
+"""
+    isconservationquantity(p)
```
isaacsas, line 550: Why not just `isconserved`? Seems shorter and clear.

src/network_analysis.jl
```diff
@@ -542,6 +542,23 @@ end
 
 ### Conservation Laws ###
 
+# Implements the `conservationquantity` parameter metadata.
+struct ConservationQuantity end
```
isaacsas, line 546:
```suggestion
struct ConservedParameter end
```

src/network_analysis.jl
```diff
@@ -542,6 +542,23 @@ end
...
+Symbolics.option_to_metadata_type(::Val{:conservationquantity}) = ConservationQuantity
```
isaacsas, line 547:
```suggestion
Symbolics.option_to_metadata_type(::Val{:conserved}) = ConservationQuantity
```
Just doing `@parameter A [conserved = true]` seems cleaner and clear?

isaacsas (commented):
Just a few comments on naming, happy for you to merge after addressing them.

PR #855, opened 2024-05-20
Add `get_variables` dispatch for `Reaction`

TorkelE:
The `get_variables` function is currently defined for `Equation`s but not `Reaction`s. I've added it. It goes through all fields, checking and adds all symbolic variables to an output vector.

### DIFF
--- a/src/reaction.jl
+++ b/src/reaction.jl
@@ -343,6 +343,30 @@ end
 MT.is_diff_equation(rx::Reaction) = false
 MT.is_alg_equation(rx::Reaction) = false
 
+"""
+    get_variables(rx::Reaction)
+
+Returns all symbolic variables that are part of a reaction. This includes all variables 
+encountered in:
+    - Rates.
+    - Among substrates and products.
+    - Among stoichiometries.
+    - Among potential noise scaling metadata.
+"""
+function ModelingToolkit.get_variables(rx::Reaction)
+    sym_vars = get_variables(rx.rate)
+    sym_vars = unique!([sym_vars; rx.substrates; rx.products])
+    for stoich in rx.substoich
+        sym_vars = unique!([sym_vars; get_variables(stoich)])
+    end
+    for stoich in rx.prodstoich
+        sym_vars = unique!([sym_vars; get_variables(stoich)])
+    end
+    if has_noise_scaling(rx)
+        sym_vars = unique!([sym_vars; get_variables(get_noise_scaling(rx))])
+    end
+    return sym_vars
+end
 
 ### Dependency-related Functions ###
 
--- a/src/reaction.jl
+++ b/src/reaction.jl
@@ -344,7 +344,7 @@ MT.is_diff_equation(rx::Reaction) = false
 MT.is_alg_equation(rx::Reaction) = false
 
 """
-    get_variables(rx::Reaction)
+    get_symbolics(set, rx::Reaction)
 
 Returns all symbolic variables that are part of a reaction. This includes all variables 
 encountered in:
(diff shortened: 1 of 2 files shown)


isaacsas (changes requested):
If we are going to add it we should probably add the in-place version and then just call that for the allocating version. It might also be worth using a `Set` and converting to a vector at the end instead. See https://github.com/JuliaSymbolics/Symbolics.jl/blob/1da13fdc375fccd6cd0a779a53b06be4ec7a54ec/src/utils.jl#L47-L79

isaacsas:
Actually, we already have an inplace version that takes `Set`s, so why not just wrap that and then convert to a vector at the end?

TorkelE:
In practise, I just used what Symbolics define for `Equation`s:
```julia
get_variables(eq::Equation) = unique(vcat(get_variables(eq.lhs), get_variables(eq.rhs)))
```
and I figured the above definition was most faithful to that.

Little bit uncertain exactly what rewrite you want, but if you have a different way to phrase the function content I am happy to update.

src/reaction.jl
```diff
@@ -343,6 +343,42 @@ end
...
+    - Among substrates and products.
+    - Among stoichiometries.
+    - Among potential noise scaling metadata.
+"""
+function ModelingToolkit.get_variables!(set, rx::Reaction)
+    get_variables!(set, rx.rate)
+    append!(set, rx.substrates)
+    append!(set, rx.products)
+    for stoichs in (rx.substoich, rx.prodstoich), stoich in stoichs
+        get_variables!(set, stoich)
```
isaacsas, line 375: Shouldn't this check if `stoich` is a Symbolic? If not there is no reason to call `get_variables!`?

src/reaction.jl
```diff
@@ -343,6 +343,42 @@ end
...
+encountered in:
+    - Rates.
+    - Among substrates and products.
+    - Among stoichiometries.
+    - Among potential noise scaling metadata.
+"""
+function ModelingToolkit.get_variables!(set, rx::Reaction)
+    get_variables!(set, rx.rate)
+    append!(set, rx.substrates)
+    append!(set, rx.products)
```
isaacsas, line 373:
```suggestion
    foreach(sub -> push!(set, sub), rx.substrates)
    foreach(prod -> push!(set, prod), rx.products)
```
`append!` would restrict this function to only work on vectors (i.e. `Sets` wouldn't work).

src/reaction.jl
```diff
@@ -343,6 +343,42 @@ end
...
+    for stoichs in (rx.substoich, rx.prodstoich), stoich in stoichs
+        get_variables!(set, stoich)
+    end
+    if has_noise_scaling(rx)
+        get_variables!(set, get_noise_scaling(rx))
+    end
+    return unique!(set)
```
isaacsas, line 380:
```suggestion
    return (set isa AbstractVector) ? unique!(set) : set
```
If passing a `Set` don't need `unique!` (not sure if even defined for them).

isaacsas (changes requested):
A few minor changes. Once addressed and tests pass feel free to merge.

src/reaction.jl
TorkelE: That make sense, good catch.

src/reaction.jl
TorkelE: Yes, I didn't really think of the difference between sets and vectors, good point

src/reaction.jl
TorkelE: Yeah, that is definitely an improvement.

PR #856, opened 2024-05-21
Update `Reaction` constructor

TorkelE:
- Permits `[]` and `nothing` to be used interchangeably (previously there were some cases which unintentionally would not work).
- Ensures that the `Reaction` structure is concretely typed.

### DIFF
--- a/src/reaction.jl
+++ b/src/reaction.jl
@@ -194,14 +194,11 @@ function Reaction(rate, subs::Vector, prods::Vector, substoich::Vector{S}, prods
     end
     
     # Computes the net stoichiometries.
-    netstoich = if isnothing(netstoich)
-        get_netstoich(subs, prods, substoich, prodstoich)
-    else
-        if typeof(netstoich) != Vector{Pair{BasicSymbolic{Real}, stoich_type}}
-            netstoich = Pair{BasicSymbolic{Real}, stoich_type}[
-                            value(ns[1]) => convert(stoich_type, ns[2]) for ns in netstoich]
-        end
-        netstoich
+    if isempty(netstoich)
+        netstoich = get_netstoich(subs, prods, substoich, prodstoich)
+    elseif typeof(netstoich) != Vector{Pair{BasicSymbolic{Real}, stoich_type}}
+        netstoich = Pair{BasicSymbolic{Real}, stoich_type}[
+                        value(ns[1]) => convert(stoich_type, ns[2]) for ns in netstoich]
     end
 
     # Handles metadata (check that all entries are unique, remove potential `only_use_rate`
--- a/src/reaction.jl
+++ b/src/reaction.jl
@@ -72,6 +72,13 @@ function metadata_only_use_rate_check(metadata)
     return Bool(metadata[only_use_rate_idx][2])
 end
 
+# Used to promote a vector to the appropriate type. Takes the `vec == Any[]` case into account by
+# returning an empty vector of the appropriate type.
+function promote_reaction_vector(vec, type)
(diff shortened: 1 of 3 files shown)


src/reaction.jl
```diff
@@ -134,19 +138,19 @@ Notes:
 - The three-argument form assumes all reactant and product stoichiometric coefficients
   are one.
 """
-struct Reaction{S, T}
+struct Reaction{T}
     """The rate function (excluding mass action terms)."""
     rate::Any
     """Reaction substrates."""
-    substrates::Vector
+    substrates::Vector{BasicSymbolic{Real}}
```
TorkelE, line 145: We should know that species are `BasicSymbolic{Real}`.

Technically, someone could declare a constantspecies parameter, and set it to a non-default type (i.e. `BasicSymbolic{Int64}`). However, we probably should not permit this?

src/reaction.jl
```diff
@@ -85,9 +92,6 @@ function get_netstoich(subs, prods, sstoich, pstoich)
     [el for el in nsdict if !_iszero(el[2])]
 end
 
-# Get the net stoichiometries' type.
-netstoich_stoichtype(::Vector{Pair{S, T}}) where {S, T} = T
```
TorkelE, line 89: No longer needed as we need to check both `S` and `T` (as `S` might now need to be converted from `Num` to `BasicSymbolic{Real}`.

isaacsas:
Actually, now I think this is a bad idea. Someone has made the confusing choice to define `get_variables` completely inconsistently with `get_variables!` in Symbolics, where the latter gets dependent variables (so rhs only). 

I think if we want such a function for `Reaction`s we should come up with a different name to avoid confusion (since we already have `get_variables!` defined on it).

TorkelE:
Do you mean this definition:
```julia
# determine which unknowns a reaction depends on
function MT.get_variables!(deps::Set, rx::Reaction, variables)
    (rx.rate isa Symbolic) && get_variables!(deps, rx.rate, variables)
    for s in rx.substrates
        # parametric stoichiometry means may have a parameter as a substrate
        any(isequal(s), variables) && push!(deps, s)
    end
    deps
end
```
and this PR https://github.com/SciML/Catalyst.jl/pull/855?

To do `get_variables(x)` where you return a vector of all the symbolic variables in `x` (whatever it is) is (to me) the established way to use the function, so `get_variables(rx::Reaction)` makes sense. Practically, I think we should have some function that does this, as it seems like something genuinely useful.

src/reaction.jl
```diff
@@ -72,6 +72,13 @@ function metadata_only_use_rate_check(metadata)
     return Bool(metadata[only_use_rate_idx][2])
 end
 
+# Used to promote a vector to the appropriate type. Takes the `vec == Any[]` case into account by
+# returning an empty vector of the appropriate type.
+function promote_reaction_vector(vec, type)
+    isempty(vec) && (return type[])
+    type[value(v) for v in vec]
+end
+
```
TorkelE, line 81: helper function to ensure something have the best type

src/reaction.jl
isaacsas: Let me know when you've reverted this and updated the PR as we discussed.

TorkelE: The decision is to change to
```julia
substrates::Vector{S}
products::Vector{R}
```
for now

(sorry, was a bit unsure what the final conclusion was if there was one, more than happy to make this switch)

isaacsas: No, it was to just leave them untyped as they currently are.

PR #868, opened 2024-05-23
change to not depend on equality after seed in stoch tests

TorkelE:


test/reactionsystem_core/events.jl
```diff
@@ -313,6 +309,66 @@ end
 
 ### Additional Correctness Tests ###
 
+# Tests that events are properly triggered for SDEs.
```
TorkelE, line 312: Makes stochastic models where all types of events should trigger. Trigger updates a parameter, which is used to check that they were triggered.

test/reactionsystem_core/symbolic_stoichiometry.jl
```diff
@@ -192,64 +192,68 @@ end
...
 
     # Test SDE simulations with integer coefficients.
-    sprob_int = SDEProblem(rs_int, u0, tspan, ps_int)
-    sprob_int_ref = SDEProblem(rs_ref_int, u0, tspan, ps_int)
-    @test solve(sprob_int, ImplicitEM(); seed) ≈ solve(sprob_int_ref, ImplicitEM(); seed)
+    sprob_int = SDEProblem(rs_int, u0_int, tspan_stoch, ps_int)
+    sprob_int_ref = SDEProblem(rs_ref_int, u0_dec, tspan_stoch, ps_int)
+    ssol_int = solve(sprob_int, ImplicitEM(); seed)
+    ssol_int_ref = solve(sprob_int_ref, ImplicitEM(); seed)
+    @test mean(ssol_int[:X1]) ≈ mean(ssol_int_ref[:X1]) atol = 2*1e0
```
TorkelE, line 240: Checks that stochastic model means are approximately the same (tolerance set so this holds over as many simulations I could bother trialing)

test/simulation_and_solving/simulate_jumps.jl
```diff
@@ -16,12 +16,17 @@ include("../test_networks.jl")
 
 # Compares jump simulations generated through hCatalyst, and manually created systems.
 let
-    # Manually declares jumps to compare Catalyst-generated jump simulations to.
+    # Declares vectors to store information about each network in. Initial conditions are approximately
```
TorkelE, line 19: I have added manually checked parameter sets and initial conditions, and for each checked that the mean are reproducibly similar across very many simualtions.

test/reactionsystem_core/events.jl
```diff
@@ -243,10 +243,6 @@ let
     sol_dsl = solve(ODEProblem(rn_dsl, u0, tspan, ps), Tsit5())
     sol_prog = solve(ODEProblem(rn_prog, u0, tspan, ps), Tsit5())
     @test sol_dsl == sol_prog
-
-    sol_dsl = solve(SDEProblem(rn_dsl, u0, tspan, ps), ImplicitEM(); seed = 1234)
-    sol_prog = solve(SDEProblem(rn_prog, u0, tspan, ps), ImplicitEM(); seed = 1234)
-    @test sol_dsl == sol_prog
```
TorkelE, line 249: These are not strictly necessary. I mostly added them because I figured they could catch some underlying issue in MTK (which I never will exclude). Hopefully it won't happen, and if it does, we try SDE simulations all over the tests, so one of those might catch it.

test/reactionsystem_core/symbolic_stoichiometry.jl
```diff
@@ -192,64 +192,68 @@ end
...
+    # Test Jump simulations with integer coefficients.
+    dprob_int = DiscreteProblem(rs_int, u0_int, tspan_stoch, ps_int)
+    dprob_int_ref = DiscreteProblem(rs_ref_int, u0_int, tspan_stoch, ps_int)
     jprob_int = JumpProblem(rs_int, dprob_int, Direct(); rng)
     jprob_int_ref = JumpProblem(rs_ref_int, dprob_int_ref, Direct(); rng)
-    sol_int = solve(jprob_int, SSAStepper(); seed)
-    sol_int_ref = solve(jprob_int_ref, SSAStepper(); seed)
-    @test mean(sol_int[:Y]) ≈ mean(sol_int_ref[:Y]) atol = 1e-2 rtol = 1e-2
+    jsol_int = solve(jprob_int, SSAStepper(); seed)
+    jsol_int_ref = solve(jprob_int_ref, SSAStepper(); seed)
```
isaacsas, line 255: Just be careful about how many jumps are occurring as this is saving for every jump as configured...

test/reactionsystem_core/symbolic_stoichiometry.jl
isaacsas: What is the expected value of the mean here? Is it comparable to 2? (I'm trying to understand what accuracy you are requesting and if the test is worthwhile with such a large `atol`...) 

Same for below.

isaacsas (changes requested):

TorkelE:
The typical error in those situations are about 0.1, I have seen it go up to 0.3 (over maybe 50 simulations?). I had it at 1e1, but then added a 2 at the end just to be thoroughly sure.

test/reactionsystem_core/symbolic_stoichiometry.jl
TorkelE: updated

PR #874, opened 2024-05-23
Proposed FAq doc updates

TorkelE:
reuploaded due to a rebase issue

### DIFF
--- a/docs/src/faqs.md
+++ b/docs/src/faqs.md
@@ -5,13 +5,10 @@ One can directly use symbolic variables to index into SciML solution objects.
 Moreover, observables can also be evaluated in this way. For example,
 consider the system
 ```@example faq1
-using Catalyst, DifferentialEquations, Plots
+using Catalyst, OrdinaryDiffEq, Plots
 rn = @reaction_network ABtoC begin
   (k₊,k₋), A + B <--> C
 end
-
-# initial condition and parameter values
-setdefaults!(rn, [:A => 1.0, :B => 2.0, :C => 0.0, :k₊ => 1.0, :k₋ => 1.0])
 nothing    # hide
 ```
 Let's convert it to a system of ODEs, using the conservation laws of the system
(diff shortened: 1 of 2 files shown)


docs/src/faqs.md
```diff
@@ -28,33 +26,35 @@ observed(osys)
 Let's solve the system and see how to index the solution using our symbolic
 variables
 ```@example faq1
+u0 = [rn.A => 1.0, rn.B => 2.0, rn.C => 0.0]
+ps = [rn.k₊ => 1.0, rn.k₋ => 1.0]
```
TorkelE, line 30: Without this rewriting, running thsi example takes crazy (infinite? I have a feeling that we have completed docs with it recently, so might finish, never bothered waiting that long in the retests though, but I have waited a good bit) long times, see this issue: https://github.com/SciML/Catalyst.jl/issues/886

isaacsas: Instead of changing our docs can you make an MTK level example and open issue there so this can get fixed?

isaacsas: I assume the same issue occurs with ODESystems and is related to the new symbolic indexing stuff.

TorkelE: So right now I am trying to figure out what is going on enough that I can actually make an MTK level issue. Given that I have over 50 unfixed MTK issues right now, I don't feel very confident that this will be fixed anytime soon, unfortunately. It was not my intention to simply change the docs so that the issue can be ignored (hence I raised the Catalyst issue of a starter).

TorkelE: (The change is only proposed, showing that some type of update is required here and what that could be, it is not meant as a final suggestion how the doc should be)

docs/src/faqs.md
```diff
@@ -28,33 +26,35 @@ observed(osys)
...
+```julia
 @unpack A, B = osys
 sol(t, idxs = [A, B])
 ```
 Plotting multiple variables using the SciML plot recipe can be achieved
 like
-```@example faq1
+```julia
 plot(sol; idxs = [A, B])
 ```
```
isaacsas, line 59: Why are these no longer dynamic?

docs/src/faqs.md
isaacsas: Did this issue ever get resolved?

TorkelE: not yet

docs/src/faqs.md
TorkelE: I was trying to reduce runtime, sorry it got confused.

If you are happy with the change above I can reenable these, but if not we will have to figure something else out.

isaacsas: Yeah, if you reenable these I am fine to merge. I'd prefer we keep everything dynamic if we can.

PR #884, opened 2024-05-27
Doc merge updates

TorkelE:
Rebuilding docs with all project updates takes ages (but fine without), trying changes 1 by 1.


TorkelE:
Ok, I finally managed to figure out how to trim down doc building times (and update to use the latest correct version of the various packages). This one does this (but only for the newly added docs). It also adds missing references between these ones. Would (hopefully) be ready to merge, but https://github.com/SciML/ModelingToolkit.jl/issues/2754 hit, so that one needs to be sorted out before we can do pretty much anything.

PR #887, opened 2024-05-29
[WIP] Fix Symbolics related stuff

TorkelE:
Update, right now this one covers most stuff symbolics/equation related.
- Replaced `Symbolics.istree` with `TermInterface.iscall`.
- Reactivated tests that should no longer be broken due to a recent Symbolcis update.
- Updates compats for new Symbolics.
- Remove SymboliUtils dependency, and all function calls of it (where possible).
- Introduces the new `hasnode` and `replacenode` symbolic functions (instead of `_occursin` and `replace`).


TorkelE:
Issue for the SDE thing: https://github.com/SciML/ModelingToolkit.jl/issues/2754

isaacsas (commented):
Fine to merge once runtests.jl is updated.

isaacsas:
Should this be good now that MTK has been updated (modulo updating runtests here)?

TorkelE:
Turns out that StructuralIdentifiability is not compatible with the latest versions (although just checking the project.toml file it should be, so might be some of its dependencies which is causing it): https://github.com/SciML/StructuralIdentifiability.jl/issues/326

Later today I plan to temporarily disable the SI extension and stuff so that it all passes. Then when SI is updated (which I imagine and hope will be soon), it should be a very quick thing to add it back in.

Project.toml
```diff
@@ -22,8 +22,8 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 RuntimeGeneratedFunctions = "7e49a35a-f44a-4d26-94aa-eba1b4ca6b47"
 Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
-SymbolicUtils = "d1185830-fcd6-423d-90d6-eec64667417b"
 Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
+TermInterface = "8ea1fca8-c5ef-4a55-8b96-4e9afe9c9a3c"
```
isaacsas, line 26: Why? Isn’t `iscall` available via Symbolics?

src/reactionsystem_conversions.jl
```diff
@@ -556,8 +556,8 @@ function nonlinear_convert_differentials_check(rs::ReactionSystem)
         # If the lhs upper level function is not a differential w.r.t. time.
         # If the contenct of the differential is not a variable (and nothing more).
         # If either of this is a case, throws the warning.
-        if Symbolics._occursin(Symbolics.is_derivative, eq.rhs) ||
-           !Symbolics.istree(eq.lhs) ||
+        if hasnode(Symbolics.is_derivative, eq.rhs) ||
+           !Symbolics.is_derivative(eq.lhs) ||
```
isaacsas, line 560: Is this really what you want? It is different than the meaning of the old code.

isaacsas (commented):
Shouldn’t this also bump MTK and Symbolics to the latest versions that have SymbolicUtils V2 support?

Project.toml
TorkelE: It is, I basically just took what was in the depreciation warning, but updating to this one now

src/reactionsystem_conversions.jl
TorkelE: Yes, but I figured that this one would be stricter and also better. I.e. the only case where a differential equation is allowed (without additional check) is if the lhs is a pure differential with respect to the iv (and the rhs contain no diffs).

Previously I just checked if the lhs was a function, and if it was, that it correspond to the differential. Now we check if it is a differential operator directly (happy to swap back though)

TorkelE:
yes it should

docs/Project.toml
```diff
@@ -68,5 +68,5 @@ SpecialFunctions = "2.4"
 StaticArrays = "1.9"
 SteadyStateDiffEq = "2.2"
 StochasticDiffEq = "6.65"
-StructuralIdentifiability = "0.5.7"
+StructuralIdentifiability = "0.5.8"
 Symbolics = "5.28"
```
isaacsas, line 72: Update MTK and Symbolics here too?

TorkelE: done

isaacsas:
Cool, feel free to merge when you are done then. Once this is merged I will continue with some doc updates, but I want to wait until I can build with SymbolicUtils V2.

TorkelE:
Will merge as soon as everything has passed.

PR #889, opened 2024-05-30
Documentation page on how to access model properties

TorkelE:
Not actually sure what to call this one, but this is the tutorial that goes through how to access stuff within a `ReactionSystem`. 

- For a starter, it goes through basic stuff like `species(rs)`. 
- For now, I also put our main section on how accessing symbolic variables via the model (e.g. `rs.X`) works.
- There is also a bit on hierarchical models, i.e. the difference between `get_species` and `species`, and `rs.submodelname.X`.


docs/src/model_creation/reactionsystem_content_accessing.md
```diff
@@ -0,0 +1,249 @@
+# [Accessing model properties](@id model_accessing)
+Catalyst is based around the creation, analysis, and simulation of chemical reaction network models. Catalyst stores these models in [`ReactionSystem](@ref) structures. This page describes some basic functions for accessing the content of these structures. This includes retrieving lists of species, parameters, or reactions that a model consists of. An extensive lists of relevant functions for working with `ReactionSystem` models can be found in Catalyst's [API](@ref ref).
+
+!!! warn
+    Generally, the field of Julia structures can be accessed through `struct.field`. E.g. a simulation's time vector can be retrieved using `simulation.t`. While Catalyst `ReactionSystem`s are structures, one should *never* access their fields using this approach, but rather using the accessor functions described below and in the [API](@ref ref) (direct accessing of fields can yield unexpected behaviours). E.g. to retrieve the species of a `ReactionsSystem` `rs`, use `Catalyst.get_species(rs)`, *not* `rs.species`.
+
+## [Direct accessing of symbolic model parameter and species](@id model_accessing_symbolic_variables)
```
TorkelE, line 7: I think it is good to have a dedicated section going through this, and I think this is a place where it makes sense to put it (the other alternatives would be the advanced introduction or the compositional modelling part, right now I feel this was the most fitting place).

docs/src/model_creation/reactionsystem_content_accessing.md
```diff
@@ -0,0 +1,249 @@
...
+There exist several other functions for accessing model properties. 
+
+The `observed`, `continuous_events`, `discrete_events` functions can be used to access a model's [observables](@ref ref), [continuous events](ref ref), and [discrete events](@ref ref), respectively.
+
+The `ModelingToolkit.get_iv` function can be used to retrieve a [model's independent variable](@ref ref):
+```@example model_accessing_basics
+ModelingToolkit.get_iv(sir)
+```
+
+## [Accessing properties of hierarchical models](@id model_accessing_hierarchical)
```
TorkelE, line 146: There is quite a lot on this, but I felt it was good to go through it properly (as I personally also have been a bit unfamiliar with it). Definitely check this part in case I have made a misstake.

docs/src/model_creation/reactionsystem_content_accessing.md
```diff
@@ -0,0 +1,249 @@
+# [Accessing model properties](@id model_accessing)
+Catalyst is based around the creation, analysis, and simulation of chemical reaction network models. Catalyst stores these models in [`ReactionSystem](@ref) structures. This page describes some basic functions for accessing the content of these structures. This includes retrieving lists of species, parameters, or reactions that a model consists of. An extensive lists of relevant functions for working with `ReactionSystem` models can be found in Catalyst's [API](@ref api).
```
isaacsas, line 2:
```suggestion
Catalyst is based around the creation, analysis, and simulation of chemical reaction network models. Catalyst stores these models in [`ReactionSystem](@ref) structures. This page describes some basic functions for accessing the content of these structures. This includes retrieving lists of species, parameters, or reactions that a model consists of. An extensive list of relevant functions for working with `ReactionSystem` models can be found in Catalyst's [API](@ref api).
```

docs/src/model_creation/reactionsystem_content_accessing.md
```diff
@@ -0,0 +1,249 @@
...
+
+!!! warn
```
isaacsas, line 4:
```suggestion
!!! warning
```

docs/src/model_creation/reactionsystem_content_accessing.md
```diff
@@ -0,0 +1,249 @@
...
+    Generally, the field of Julia structures can be accessed through `struct.field`. E.g. a simulation's time vector can be retrieved using `simulation.t`. While Catalyst `ReactionSystem`s are structures, one should *never* access their fields using this approach, but rather using the accessor functions described below and in the [API](@ref api_accessor_functions) (direct accessing of fields can yield unexpected behaviours). E.g. to retrieve the species of a `ReactionsSystem` `rs`, use `Catalyst.get_species(rs)`, *not* `rs.species`.
```
isaacsas, line 5:
```suggestion
    Generally, a field of a Julia structure can be accessed through `struct.fieldname`. E.g. a simulation's time vector can be retrieved using `simulation.t`. While Catalyst `ReactionSystem`s are structures, one should *never* access their fields using this approach, but rather using the accessor functions described below and in the [API](@ref api_accessor_functions) (direct accessing of fields can yield unexpected behaviours). E.g. to retrieve the species of a `ReactionsSystem` `rs`, use `Catalyst.get_species(rs)`, *not* `rs.species`.
```

isaacsas: Maybe explain why this shouldn't be done by pointing out that instead `rs.t` is used to accessible symbolic variables encapsulated by `rs`?

docs/src/model_creation/reactionsystem_content_accessing.md
```diff
@@ -0,0 +1,249 @@
...
+
+## [Direct accessing of symbolic model parameter and species](@id model_accessing_symbolic_variables)
+Previously we have described how the parameters and species (and [variables](@ref constraint_equations_coupling_constraints)) that Catalyst models contain are represented using so-called [*symbolic variables*](@ref introduction_to_catalyst) (and how these enable the forming of [*symbolic expressions*](@ref introduction_to_catalyst)). We have described how, during [programmatic modelling](@ref programmatic_CRN_construction), the user has [direct access to these](@ref programmatic_CRN_construction) and how this can be [taken advantage of](@ref programmatic_CRN_construction). We have also described how, during [DSL-based modelling](@ref dsl_description), the need for symbolic representation can be circumvented by [using `@unpack`](@ref dsl_advanced_options_symbolics_and_DSL_unpack) or by [creating an observable](@ref dsl_advanced_options_observables). However, sometimes, it is easier to *directly access a symbolic variable through the model itself*.
```
isaacsas, line 8:
```suggestion
... the need for symbolic representation can be circumvented by [using `@unpack`](@ref dsl_advanced_options_symbolics_and_DSL_unpack) or by [creating an observable](@ref dsl_advanced_options_observables)...*.
```
I found this part of the paragraph confusing. Maybe separately just recall the ways to access a symbolic variable/parameter that have already been shown? (I'd leave out the observable stuff as I think it will just confuse the discussion.)

docs/src/model_creation/reactionsystem_content_accessing.md
```diff
@@ -0,0 +1,249 @@
...
+rs_prog = complete(rs_prog)
+nothing # hide
+```
+Here, we can confirm that the symbolic variables we access through the model are identical to those we used to create it:
+```@example model_accessing_symbolic_variables
+isequal(rs.k1, k1)
+```
+
+!!! warn
+    When accessing model symbolic variables through the model (using e.g. `rs.X1`), it is important to first ensure that the [*model is complete*](@ref programmatic_CRN_construction).
```
isaacsas, line 59:
```suggestion
    When accessing model symbolic variables through the model (using e.g. `rs.X1`), it is important to first ensure that the [*model has been marked complete*](@ref programmatic_CRN_construction).
```

docs/src/model_creation/reactionsystem_content_accessing.md
```diff
@@ -0,0 +1,249 @@
...
+
+## [Accessing basic model properties](@id model_accessing_basics)
+
+### [Accessing model parameter and species](@id model_accessing_basics_parameters_n_species)
+Previously we showed how to access individual parameters or species of a `ReactionSystem` model. Next, the `parameters` and `species` functions allow us to retrieve *all* parameters and species as vectors:
```
isaacsas, line 64:
```suggestion
Previously we showed how to access individual parameters or species of a `ReactionSystem` model. Next, the [`parameters`](@ref) and [`species`](@ref) functions allow us to retrieve *all* parameters and species as vectors:
```
All these API functions you are introducing should be links to their documentation. (I haven't made comments beyond this point, but it would be good to update them below too.)

docs/src/model_creation/reactionsystem_content_accessing.md
```diff
@@ -0,0 +1,249 @@
...
+using OrdinaryDiffEq, Plots
+u0 = [rs.nucleus.Xᵢ => 0.0, rs.cytoplasm.Xᵢ => 0.0, rs.cytoplasm.Xₐ => 0.0]
+ps = [rs.nucleus.p => 1.0, rs.nucleus.d => 0.2, rs.cytoplasm.kₐ => 5.0, rs.cytoplasm.d => 0.2, rs.kₜ => 0.1]
+oprob = ODEProblem(rs, u0, (0.0, 10.0), ps)
+sol = solve(oprob)
+plot(sol)
+```
+
+!!! note
+    When we access a symbolic variable through a subsystem (e.g. `rs.nucleus.Xᵢ`) that subsystem's name is prepended to the symbolic variable's name (we call this *namespacing*). This is also the case if we access it through the original model, i.e. `nucleus_sys.Xᵢ`. Namespacing is only performed when we access variables of [*incomplete systems*](@ref programmatic_CRN_construction). I.e. `isequal(nucleus_sys.d, nucleus_sys.d)` returns false (as the systems are incomplete and namespacing is performed). However, `isequal(complete(nucleus_sys).d, complete(nucleus_sys).d)` returns true  (as the systems are complete and namespacing is not performed). This is the reason that the system top-level system's name is never prepended when we do e.g. `rs.kₜ` (because here, `rs` is complete).
```
isaacsas, line 215: `isequal(nucleus_sys.d, nucleus_sys.d) == true` for me. Is there a typo in the comparison you are trying to make?

isaacsas (changes requested):
Happy for you to merge after you've addressed my comments.

docs/src/model_creation/reactionsystem_content_accessing.md
TorkelE: yes, I will add a bit about that

docs/src/model_creation/reactionsystem_content_accessing.md
TorkelE: will do

docs/src/model_creation/reactionsystem_content_accessing.md
TorkelE: This was kind of written in the mind that we had a dedicated section on symbolic variables, but I agree that it might be a bit confusing now. I will try to reword/improve.

Ideally I'd still have a link to the observables (although maybe rather and example where it is used). For DSL-based models, in most cases, I think this is probably the best way to do it (i.e. if there is a specific quantity like this you are using, it probably should be an observable, in other situations `@unpack` might be better)

docs/src/model_creation/reactionsystem_content_accessing.md
TorkelE: The reason is that, as shown [below](@ref model_accessing_symbolic_variables), Catalyst reserves this type of accessing for accessing symbolic variables stored in the system (i.e. `rs.X` refers to the `X` symbolic variable, not a field in `rs` named "X").

isaacsas: Sounds good, but maybe say "Catalyst, and more generally ModelingToolkit for any system type, reserves"

isaacsas: It isn't really a policy that we made.

docs/src/model_creation/reactionsystem_content_accessing.md
TorkelE: you are right, I was meaning to do
```julia
isequal(nucleus_sys.d, cytoplasm_sys.d)
```
and
```julia
isequal(complete(nucleus_sys).d, complete(cytoplasm_sys).d)
```

changed now

docs/src/model_creation/reactionsystem_content_accessing.md
TorkelE: New section

## [Direct accessing of symbolic model parameter and species](@id model_accessing_symbolic_variables)
Previously we have described how the parameters and species that Catalyst models contain are represented using so-called [*symbolic variables*](@ref introduction_to_catalyst) (and how these enable the forming of [*symbolic expressions*](@ref introduction_to_catalyst)). We have described how, during [programmatic modelling](@ref programmatic_CRN_construction), the user has [direct access to these](@ref programmatic_CRN_construction) and how this can be [taken advantage of](@ref programmatic_CRN_construction). I.e. here we declare the symbolic variables associated with a two-state model, and also creates the model programmatically.
```@example model_accessing_symbolic_variables_intro
using Catalyst
t = default_t()
@parameters k1 k2
@species X1(t) X2(t)

rxs = [
    Reaction(k1, [X1], [X2]),
    Reaction(k2, [X2], [X1])
]
@named rs = ReactionSystem(rxs, t)
nothing # hide
```
We have also described how, during [DSL-based modelling](@ref dsl_description), the need for symbolic representation can be circumvented by [using `@unpack`](@ref dsl_advanced_options_symbolics_and_DSL_unpack) or by [creating an observable](@ref dsl_advanced_options_observables). However, sometimes, it is easier to *directly access a symbolic variable through the model itself*.

TorkelE: Actually, didn't like how this turned out, especially since the same model is declare later. Now I use 
```
Previously we have described how the parameters and species that Catalyst models contain are represented using so-called [*symbolic variables*](@ref introduction_to_catalyst) (and how these enable the forming of [*symbolic expressions*](@ref introduction_to_catalyst)). We have described how, during [programmatic modelling](@ref programmatic_CRN_construction), the user has [direct access to these](@ref programmatic_CRN_construction) and how this can be [taken advantage of](@ref programmatic_CRN_construction). We have also described how, during [DSL-based modelling](@ref dsl_description), the need for symbolic representation can be circumvented by [using `@unpack`](@ref dsl_advanced_options_symbolics_and_DSL_unpack) or by [creating an observable](@ref dsl_advanced_options_observables). However, sometimes, it is easier to *directly access a symbolic variable through the model itself*, something which we will describe here.
```
which is not that different from what we had originally. I think the best solution is to create a model creation example which illustrates this, and then simply link to that.

PR #890, opened 2024-05-30
Advanced introduction to Catalyst and its internal

TorkelE:
Again not sure what to call this one.

This is basically a third introduction page. While the two previous ones are more or less parallel (i.e. have the same purpose, but for people who know/do not know Julia), this one is a continuation. It doesn't really go through any core stuff per see, but have two functions:
- Describe some of the internals, i.e. how Catalyst creates various types of MTK systems, and how stuff are internally represented.
- Provide short introductions to modelling workflows in the same place. I.e. there is a section that goes through DSL-based, programmatic, and compositional modelling. Since these are mostly kept separate, it is nice to have them next to each other same place, so that they can be easily compared.

### DIFF
--- a/docs/src/introduction_to_catalyst/advanced_intro_to_catalyst.md
+++ b/docs/src/introduction_to_catalyst/advanced_intro_to_catalyst.md
@@ -1,5 +1,5 @@
 # [How Catalyst works and advanced introduction](@id advanced_intro_to_catalyst) 
-We have [previously](@ref introduction_to_catalyst) introduced Catalyst, and how to use it to create and simulate simple chemical reaction network (CRN) models. Here, we provide a more in-depth introduction. We will explain how Catalyst is constructed (i.e. what packages it builds on and how), provide a comprehensive overview of the ways Catalyst models can be created, and describe how these are different.
+We have [previously](@ref introduction_to_catalyst) introduced Catalyst, and how to use it to create and simulate simple chemical reaction network (CRN) models. Here, we provide a more in-depth introduction. We will explain how Catalyst is constructed (i.e. what packages it builds on and how), provide a comprehensive overview of the different ways Catalyst models can be created, and describe how these are different.
 
 This advanced introduction is not required for understanding the remaining documentation, nor for using any of Catalyst's features. It is not expected that new users will read it. Instead, it aims to provide additional context and understanding for more experienced users. Various section here are also used as references at later part of the documentation (for users who wants additional insight). For those who expect to use Catalyst extensively, we believe that, at some point, reading this tutorial will be useful.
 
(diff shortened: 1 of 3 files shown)


docs/src/introduction_to_catalyst/advanced_intro_to_catalyst.md
```diff
@@ -0,0 +1,356 @@
...
+    `@variables` (and similar [macros introduced later on](@ref ref), like `@parameters` and `@species`) create new variables in the current scope. E.g. if you have previously defined `x = 5` and then run `@variables x`, this will overwrite the previous value of `x`.
+
+Symbolics is used internally within Catalyst to represent all models. One advantage of this is that complicated algebraic expressions can be automatically simplified. E.g. if we create a model with a (trivially) unnecessarily complicated rate and check what it is, we note that it has been automatically simplified
+```@example advanced_intro_to_catalyst_symbolics
+using Catalyst
+rn = @reaction_network begin
+    (k1*k2)/k2, X --> 0
+end
+reactions(rn)[1].rate
+```
```
TorkelE, line 39: Here I had more on symbolic simplifications performed, but looking at examples, it seems that these are never really done (even with `structural_simplify`).

docs/src/introduction_to_catalyst/advanced_intro_to_catalyst.md
```diff
@@ -0,0 +1,356 @@
...
+```@example advanced_intro_to_catalyst_designations
+sol_syms[2rs_dsl.X + rs_dsl.p]
+```
+
+This approach for representing species and parameters is primarily used in combination with [hierarchical or compositional modelling](@ref compositional_modeling) (as it enables us to designate species that have identical names, but are stored in different subsystems). System-based symbolic accessing is discussed in more detail [here](@ref ref).
+
+!!! warn
+    When accessing species and parameters that are stored in a system using the above approach it is important to ensure that the model is *complete*. Model completeness is described [here](@ref ref).
+
+## [Independent variables](@id advanced_intro_to_catalyst_independent_vars)
```
TorkelE, line 325: We should have a short section on independent variables somewhere (I hadn't really heard of this before MTK started using these), and somewhere here seemed natural. The main purpose of this one is that whenever we mention "independent variable" somewhere, we can link this section, and those who are unfamiliar with the term can read up on what it means.

docs/src/introduction_to_catalyst/advanced_intro_to_catalyst.md
```diff
@@ -0,0 +1,356 @@
...
+sol_symb[2X + p]
+```
+Indeed, we can confirm that this returns the expected output:
+```@example advanced_intro_to_catalyst_designations
+sol_symb[2X + p] == 2*sol[X] .+ 1.0
+```
+
+Symbolic variables-based representation is strictly superior to symbol-based representation. However, it is generally only used in combination with programmatic modelling (as during DSL-based modelling we never actually declare the symbolic variables, and thus cannot use them directly). To use symbolic variables-based representation in combination with the DSL, the [`@unpack` macro can be used](@ref ref).
+
+### [System-based symbolic designation](@id advanced_intro_to_catalyst_designations_sys_symb)
```
TorkelE, line 293: The new section on how to access content of models got a more extensive description of this.

TorkelE:
I think our current plan is to, rather than adding additional introductory sections, expand the main one with any needed extra content. So I will close this one for now, and if there is anything in it we need it can always be fetched.

PR #891, opened 2024-05-30
New events + callbacks page

TorkelE:
New doc page for events and callbacks. Goes through the various types, and how to declare and use them. Also describes some common pitfuls one can encounter while working with these.

### DIFF
--- a/docs/src/model_creation/events_and_callbacks.md
+++ b/docs/src/model_creation/events_and_callbacks.md
@@ -1,0 +1,497 @@
+# [Simulation events](@id events)
+
+Sometimes one wishes to model events, describing things that can happen to a system during its simulation. Examples can include:
+ - A chemical system where an amount of some species is added at a time point after the simulation's initiation.
+ - A simulation of a circadian rhythm, where light is turned on/off every 12 hours.
+ - A cell divides when a size variable reaches a certain threshold (subsequently halving the size).
+  
+Generally, there are two approaches for creating events:
+1. Including them as a part of the model when it is created.
+2. Creating a separate *callback*, which is then supplied to the `solve` function.
+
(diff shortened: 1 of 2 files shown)


docs/src/model_creation/events_and_callbacks.md
```diff
@@ -0,0 +1,497 @@
...
+However, if we instruct `solve` to specifically stop at time `5.0`, it *is* triggered:
+```@example events_considerations
+sol = solve(oprob; tstops = [5.0])
+plot(sol)
+```
+Here, `tstops` takes a vector with all the time points we wish to stop at (here we only designate `5.0`).
+
+Since continuous events (and callbacks) actively find the exact time point when they are triggered, supplying `tstops` is not required for these.
+
+## [Programmatic creation of models containing events](@ref events_programmatic)
```
TorkelE, line 458: Right now, I start with going through how events work with the DSL. Then, at the end I have a separate section describing the details one needs to change when using a programmatic workflow (here and most other cases it is straightforward, except for the base syntax the two approaches are really similar).

docs/src/model_creation/events_and_callbacks.md
```diff
@@ -0,0 +1,497 @@
...
+```
+
+To deal with these situations, it is recommended to not pass your `ODEProblem` directly into `solve`, but rather a [`deepcopy`](https://docs.julialang.org/en/v1/base/base/#Base.deepcopy) of it:
+```@example events_considerations
+oprob = ODEProblem(bd_model, u0, tspan, ps)
+sol = solve(deepcopy(oprob), Tsit5())
+nothing # hide
+```
+
+### [Callbacks during jump simulations](@id events_additional_considerations_jump_parameters)
```
TorkelE, line 396: This is handled properly for events, right? Not that the user can do anything if it isn't, but still.

PR #893, opened 2024-05-31
Adding cycles/flux mode calculation 

vyudu:
Computing cycles/flux modes of reaction networks, working on caching equations with a similar format to the functionality for conservation laws.


isaacsas (changes requested):
Looks good. Can you add some test examples with known answers that cover various cases?

src/network_analysis.jl
```diff
@@ -886,3 +886,45 @@ function treeweight(t::SimpleDiGraph, g::SimpleDiGraph, distmx::Matrix)
...
+        all(r -> r <= 0, Nrow) && (Nrow .*= -1)
+    end
+
+    # check we haven't overflowed
+    iszero(nsm * N) || error("Calculation of the cycle matrix was inaccurate, "
+          * "likely due to numerical overflow. Please use a larger integer "
+          * "type like Int128 or BigInt for the net stoichiometry matrix.")
+
+    T(N)
+end
```
isaacsas, line 920: I think it would make sense to factor out the code that is common to conservation laws and cycles, and just handle passing it the transposed matrix for conservation laws. i.e. make a common function both calculations wrap given they are essentially identical up to the transposes.

test/network_analysis/network_properties.jl
isaacsas, line 1: It would probably be good to get a few more tests for systems with/without cycles if possible (even simple cases).

isaacsas (changes requested):

src/network_analysis.jl
```diff
@@ -623,22 +623,9 @@ end
...
-
-    # check we haven't overflowed
-    iszero(N' * nsm) || error("Calculation of the conservation law matrix was inaccurate, "
-          * "likely due to numerical overflow. Please use a larger integer "
-          * "type like Int128 or BigInt for the net stoichiometry matrix.")
-
-    T(N')
+function conservationlaws(nsm::Matrix; col_order = nothing)
+    conslaws = positive_nullspace(nsm'; col_order = col_order)
+    Matrix(conslaws)
```
isaacsas, line 628: Does `conslaws` need to be transposed again here? (Just looking at the old code).

vyudu: I think positive_nullspace does the transposition automatically in the `T(N)` line because the type we're passing in is an adjoint.

isaacsas:
If you can update to master I'll merge this then.

isaacsas (approved):

isaacsas:
CI is failing?

vyudu:
> CI is failing?

Hmm yeah something weird seems to have happened to the test file when I merged master, looking into it

isaacsas:
Still seems there are conflicts with master?

isaacsas:
OK, LGTM. Feel free to merge at your leisure!

PR #901, opened 2024-06-02
change doc collapse level

TorkelE:
Given that the number of doc pages are increasing, we might want to change so that the top sections start collapsed, e.g. like
image

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -32,7 +32,7 @@ include("pages.jl")
 
 makedocs(sitename = "Catalyst.jl",
          authors = "Samuel Isaacson",
-         format = Documenter.HTML(; analytics = "UA-90474609-3",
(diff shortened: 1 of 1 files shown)


TorkelE:
I mostly agree.

As we get more content though, we the list might get very long. However, long-term I think it might make sense to kind of introduce another layer? 

E.g. having upper-level section like model creation, model analysis, model simulation. Then steady state stuff could become a sub-section of model analysis, and there could be another more network analysis stuff. Spatial simulation could be a sub section of simulation, etc. etc. I don't think it is something to act upon now, I say we go with what we have now and keep an open mind that at some point we might want to change it again (also trying to keep changes absolutely minimal)

isaacsas:
Yes, if/when things get too big and crowded we can reassess the layout.

PR #902, opened 2024-06-02
Documentation of file loading/saving

TorkelE:
Briefly discusses:
- Catalyst's native serialization functionality.
- Serialization.jl
- ReactionNetworkImporters.
- SbmlaImporter/SBMLToolkit


docs/src/model_creation/model_file_loading_and_export.md
```diff
@@ -0,0 +1,163 @@
...
+
+
+!!! note
+    It should be noted that .net files support a wide range of potential model features, not all of which are currently supported by ReactionNetworkImporters. Hence, there might be some .net files which `loadrxnetwork` will not be able to load.
+
+A more detailed description of ReactionNetworkImporter's features can be found in its [documentation](https://docs.sciml.ai/ReactionNetworkImporters/stable/).
+
+## Loading SBML files using SBMLImporter.jl and SBMLToolkit.jl
+The Systems Biology Markup Language (SBML) is the most widespread format for representing CRN models. Currently, there exist two different Julia packages, [SBMLImporter.jl](https://github.com/sebapersson/SBMLImporter.jl) and [SBMLToolkit.jl](https://github.com/SciML/SBMLToolkit.jl), that are able to load SBML files to Catalyst `ReactionSystem` structures. SBML is able to represent a *very* wide range of model features. Some of these are not supported by these packages (or Catalyst). Hence, there exist SBML files (typically containing obscure model features such as events with time delays) that currently cannot be loaded into Catalyst models).
+
```
sebapersson, line 99: SBML is able to represent a *very* wide range of model features, and both packages support most features. However, there exist SBML files

As SBMLImporter supports a lot of weird features (more than most other SysBio tools), I think it should be highlighted here

docs/src/model_creation/model_file_loading_and_export.md
```diff
@@ -0,0 +1,163 @@
...
+SBMLImporter's `load_SBML` function can be used to load SBML files. Here, we load a [Brusselator](@ref basic_CRN_library_brusselator) model stored in the "brusselator.xml" file:
+```julia
+using SBMLImporter
+prn, cbs = load_SBML("brusselator.xml")
```
sebapersson, line 103: Would change to;
```julia
prn, cbs = load_SBML("brusselator.xml", massaction=true)
```
For next release I plan to enforce `massaction=true` for mass-action models. This is because trying to infer this proves hard, and sometimes is incorrectly inferred for ODE models (which still are the most common type imported).

And then maybe add some text similar to this:

Here `massaction=true` means that the model follow mass-action principles, and thus can be efficiently simulated with jump-solvers. This option should only be provided for mass-action models.

docs/src/model_creation/model_file_loading_and_export.md
```diff
@@ -0,0 +1,163 @@
...
+tspan = (0.0, 50.0)
+oprob = ODEProblem(prn.rn, prn.u0, tspan, prn.p)
+sol = solve(oprob; callback = cbs)
+plot(sol)
+```
+![Brusselator Simulation](../assets/brusselator_sim_SBMLImporter.svg)
+
+Note that, while ReactionNetworkImporters adds initial condition and species values as default to the imported model, SBMLImporter does not do this. These must hence be provided to the `ODEProblem` directly.
+
+A more detailed description of SBMLImporter's features can be found in its [documentation](https://docs.sciml.ai/ReactionNetworkImporters/stable/).
```
sebapersson, line 117: I would use the link to SBMLImporter for the docs :)
https://sebapersson.github.io/SBMLImporter.jl/stable/

docs/src/model_creation/model_file_loading_and_export.md
```diff
@@ -0,0 +1,163 @@
...
+
+### SBMLImporter and SBMLToolkit
+Above, we described how to use SBMLImporter to import SBML files. Alternatively, SBMLToolkit can be used instead. It has a slightly different syntax, which is described in its [documentation](https://github.com/SciML/SBMLToolkit.jl). A short comparison of the two packages can be found [here](https://github.com/sebapersson/SBMLImporter.jl?tab=readme-ov-file#differences-compared-to-sbmltoolkit). Generally, while they both perform well, we note that for *jump simulations* SBMLImporter is preferable (its way for internally representing reaction event enables more performant jump simulations).
```
sebapersson, line 120: This is a bit tricky section. But, I think it should be mentioned that SBMLImporter supports more features, and allow species to have unit amount or concentration (SBMLToolkit only allow unit amount). But consider this optional to add.

sebapersson (changes requested):
Thanks for adding SBMLImporter to the docs! I just have some minor comments.

docs/src/model_creation/model_file_loading_and_export.md
sebapersson: For SBMLImporter this is not breaking, as we handle concentration `only_use_rate=true` for non-mass action models. While for for a mass-action model aimed for Gillespie simulators what I seen in SBML unit amount is typically used.

But would be nice with chemical concentration in MTK.

docs/src/model_creation/model_file_loading_and_export.md
TorkelE: update

docs/src/model_creation/model_file_loading_and_export.md
TorkelE: whoops! Thanks for noting

PR #903, opened 2024-06-02
Specialised conservation law section on the network analysis doc page

TorkelE:
Adds a dedicated section on the network analysis page which goes through conservation laws and their various API functions. Also adds a short section about this in the ODE performance doc (mentioning that, in some situations, conservation law performance can have an impact on performance).


docs/pages.jl
```diff
@@ -15,7 +15,7 @@ pages = Any[
         #"model_creation/parametric_stoichiometry.md",# Distributed parameters, rates, and initial conditions.
         # Loading and writing models to files.
         "model_creation/model_visualisation.md",
-        #"model_creation/network_analysis.md",
+        "model_creation/network_analysis.md",
```
TorkelE, line 18: This one didn't actually generate any errors, so shouldn't actually have been commented out in the first place.

docs/src/model_creation/examples/basic_CRN_library.md
```diff
@@ -54,12 +54,12 @@ The two-state model describes a component (here called $X$) which can exist in t
...
 using Catalyst, OrdinaryDiffEq, StochasticDiffEq, Plots
 two_state_model = @reaction_network begin
-    (k1,k2), X₁ <--> X₂
+    (k₁,k₂), X₁ <--> X₂
 end
 
 u0 = [:X₁ => 50.0, :X₂ => 50.0]
 tspan = (0.0, 1.0)
-ps = [:k1 => 2.0, :k2 => 3.0]
+ps = [:k₁ => 2.0, :k₂ => 3.0]
```
TorkelE, line 62: I don't really care about `k₁` and `k1` (the first is pretty, but the second one you can actually write everywhere). I think on this page I think it is nice to use the pretty form in some case (I think it is nice to use different approaches in the examples). In the more of actual docs the more sense it makes to just write `k1`.

TorkelE: Here I chaneged it purely for consistency in the model (i.e. we use `X₁`, then `k₁` makes sense as well)

docs/src/model_creation/network_analysis.md
```diff
@@ -8,7 +8,7 @@ section
 Note, currently API functions for network analysis and conservation law analysis
 do not work with constant species (currently only generated by SBMLToolkit).
 
-## Network representation of the Repressilator `ReactionSystem`
+## [Network representation of the Repressilator `ReactionSystem`](@id network_analysis_repressilator_representation)
```
TorkelE, line 11: I added in the reference tags, in case we might need them it is nice to have them in place (per previous reasoning).

docs/src/model_creation/network_analysis.md
```diff
@@ -500,6 +500,94 @@ Catalyst.reset_networkproperties!(rn)
...
+```
+Finally, the `conservationlaws` functions yield a $m \times n$ matrix, where $n$ is a system's number of species, $m$ its number of conservation laws, and element $(i,j)$ correspond to the copy number of the $j$th species that occur in the $i$th conserved quantity:
+```@example network_analysis_conservation_laws
+conservationlaws(rs)
+```
+I.e. in this case we have a single conserved quantity, which contains a single copy each of the system's two species.
+
+!!! note
+    The `remove_conserved = true` option is available when creating `SDEProblem`s, `NonlinearProblem`s, and `SteadyStateProblem`s (and their corresponding systems). However, it cannot be used when creating `JumpProblem`s.
+
```
TorkelE, line 590: I was thinking of adding a final example where I used a much less trivial model, just showing off what we can actually do.

docs/src/model_simulation/ode_simulation_performance.md
```diff
@@ -172,6 +172,26 @@ Finally, we note that when using preconditioners with a matrix-free method (like
...
+end
+
+# Generate and simulate and ODE (while eliminating any conserved quantities).
+u0 = [:X₁ => 2.0, :X₂ => 3.0]
+ps = [:k₁ => 1.0, :k₂ => 2.0]
+oprob = ODEProblem(rs, u0, (0.0, 10.0), ps; remove_conserved = true)
+sol = solve(oprob)
+nothing # hide
+```
+Conservation law elimination should never impact performance negatively. However, its effect on performance can be limited, and is very model-dependent. It is primarily important when performing simulation with [implicit solvers](@ref ode_simulation_performance_stiffness) (as these requires computing system Jacobians, and conservation law eliminations typically transform these from singular Jacobians to non-singular Jacobians).
```
TorkelE, line 192: I did some light tests, saw a (very  minor but at least noticeable) speed up when I had a model with several conservation laws and implicit solver.

TorkelE: At some point it would be nice to do more exhaustive tests.

sebapersson: Conservation law elimination can prove **very** beneficial for models with steady-state simulations where we want to compute gradients. Briefly, if the ODE Jacobian is invertiable, we can obtain the gradient via a simple linear-solve, they perform some benchmarks on this here https://arxiv.org/pdf/2405.16524.

TorkelE: Interesting! Currently, Catalyst does not default to `remove_conserved = true` when it creates ODEs (there are some discussions to do so, but right now there are some niche annoyances which make me reluctant).

However, it might make sense for PEtab to default to using this feature? Since PEtab handles stuff internally, e.g. that eliminated variables are not plotted by default in ODE simulations should not be a problem.

sebapersson: Ahh, I agree. When I have updated PEtab to MTKv9 I have some models where it would be interesting to try this out.

TorkelE: excellent, happy to help (and for the conservation law feature to get more used. I have no part in implementing it, but it is really cool)

docs/src/model_creation/network_analysis.md
```diff
@@ -500,6 +500,97 @@ Catalyst.reset_networkproperties!(rn)
...
+If we simulate it, we note that while the concentrations of $X₁$ and $X₂$ change throughout the simulation, the total concentration of $X$ ($= X₁ + X₂$) is constant:
+```@example network_analysis_conservation_laws
+using OrdinaryDiffEq, Plots
+u0 = [:X₁ => 80.0, :X₂ => 20.0]
+ps = [:k₁ => 10.0, :k₂ => 2.0]
+oprob = ODEProblem(rs, u0, (0.0, 1.0), ps)
+sol = solve(oprob)
+plot(sol; idxs = [rs.X₁, rs.X₂, rs.X₁ + rs.X₂], label = ["X₁" "X₂" "X₁ + X₂ (a conserved quantity)"])
+```
+This makes sense, as while $X$ is converted between two different forms ( $X₁$ and $X₂$), it is neither produced nor degraded. That is, it is a *conserved quantity*. Next, if we consider the ODE that our model generates:
```
isaacsas, line 523:
```suggestion
This makes sense, as while $X$ is converted between two different forms ($X₁$ and $X₂$), it is neither produced nor degraded. That is, it is a *conserved quantity*. Next, if we consider the ODE that our model generates:
```

docs/src/model_creation/network_analysis.md
```diff
@@ -500,6 +500,97 @@ Catalyst.reset_networkproperties!(rn)
...
+we note that it essentially generates the same equation twice (i.e. $\frac{dX₁(t)}{dt} = -\frac{dX₂(t)}{dt}$). By designating our conserved quantity $X₁ + X₂ = Γ$, we can rewrite our differential equation model as a [differential-algebraic equation](https://en.wikipedia.org/wiki/Differential-algebraic_system_of_equations) (with a single differential equation and a single algebraic equation):
+```math
+\frac{dX₁(t)}{dt} = - k₁X₁(t) + k₂(-X₁(t) + Γ) \\
+X₂(t) = -X₁(t) + Γ
+```
+Using Catalyst, it is possible to detect any such conserved quantities and eliminate them from the system. Here, when we convert our `ReactionSystem` to an `ODESystem`, we provide the `remove_conserved = true` argument to instruct Catalyst to perform this elimination:
+```@example network_analysis_conservation_laws
+osys = convert(ODESystem, rs; remove_conserved = true)
+```
+We note that the output system only contains a single (differential) equation. The second (algebraic) equation is stored as an [*observable*](@ref dsl_advanced_options_observables), and can be retrieved using the `observed` function:
```
isaacsas, line 537:
```suggestion
We note that the output system only contains a single (differential) equation and can hence be solved with an ODE solver. The second (algebraic) equation is stored as an [*observable*](@ref dsl_advanced_options_observables), and can be retrieved using the `observed` function:
```

docs/src/model_creation/network_analysis.md
```diff
@@ -500,6 +500,97 @@ Catalyst.reset_networkproperties!(rn)
...
+```@example network_analysis_conservation_laws
+observed(osys)
+```
+Using the `unknowns` function we can confirm that the ODE only have a single unknown variable:
```
isaacsas, line 541:
```suggestion
Using the `unknowns` function we can confirm that the ODE only has a single unknown variable:
```

docs/src/model_creation/network_analysis.md
```diff
@@ -500,6 +500,97 @@ Catalyst.reset_networkproperties!(rn)
...
+Here, while `Γ[1]` becomes a parameter of the new system, it has a [default value](@ref dsl_advanced_options_default_vals) equal to the corresponding conservation law. Hence, its value is computed from the initial condition `[:X₁ => 80.0, :X₂ => 20.0]`, and does not need to be provided in the parameter vector. Next, we can simulate and plot our model using normal syntax:
+```@example network_analysis_conservation_laws
+sol = solve(oprob)
+plot(sol)
+```
+!!! note
+    Any species eliminated using `remove_conserved = true` will not be plotted by default. However, it can be added to the plot using [the `idxs` plotting option](@ref simulation_plotting_options). E.g. here we would use `plot(sol; idxs = [:X₁, :X₂])` to plot both species.
+
+!!! danger
+    Currently, there is a bug where the values associated with conservation laws are not updated properly in response to [`remake`](@ref simulation_structure_interfacing_problems_remake) (or [other problem-updating functions, such as `getu`](@ref simulation_structure_interfacing_functions)). Hence, problems created using the `remove_conserved = true` *should not* be modified.
```
isaacsas, line 568:
```suggestion
    Currently, there is a bug in MTK where the values associated with conservation laws are not updated properly in response to [`remake`](@ref simulation_structure_interfacing_problems_remake) (or [other problem-updating functions, such as `getu`](@ref simulation_structure_interfacing_functions)). Hence, problems created using `remove_conserved = true` *should not* be modified, i.e. by changing initial conditions or the values of the `Γ`'s directly. Instead, to change these values new problems must be generated with the new initial condition map.
```

docs/src/model_creation/network_analysis.md
```diff
@@ -500,6 +500,97 @@ Catalyst.reset_networkproperties!(rn)
...
+sol = solve(oprob)
+plot(sol)
+```
+!!! note
+    Any species eliminated using `remove_conserved = true` will not be plotted by default. However, it can be added to the plot using [the `idxs` plotting option](@ref simulation_plotting_options). E.g. here we would use `plot(sol; idxs = [:X₁, :X₂])` to plot both species.
+
+!!! danger
+    Currently, there is a bug where the values associated with conservation laws are not updated properly in response to [`remake`](@ref simulation_structure_interfacing_problems_remake) (or [other problem-updating functions, such as `getu`](@ref simulation_structure_interfacing_functions)). Hence, problems created using the `remove_conserved = true` *should not* be modified. 
+
+While `X₂` is an observable (and not unknown) of the ODE, we can [access it](@ref simulation_structure_interfacing_problems) just like if `remove_conserved = true` has not been used:
```
isaacsas, line 570:
```suggestion
While `X₂` is an observable (and not unknown) of the ODE, we can [access it](@ref simulation_structure_interfacing_problems) just like if `remove_conserved = true` had not been used:
```

docs/src/model_creation/network_analysis.md
```diff
@@ -500,6 +500,97 @@ Catalyst.reset_networkproperties!(rn)
...
+```@example network_analysis_conservation_laws
+sol[:X₂]
+```
+!!! note
+    Generally, the `remove_conserved = true` should not change any model workflows. I.e. anything that works without this option should also work when an `ODEProblem` is created using `remove_conserved = true`.
+
+For any given `ReactionSystem` model, we can use the `conservationlaw_constants` to compute all a system's conserved quantities:
```
isaacsas, line 577:
```suggestion
For any given `ReactionSystem` model, we can use `conservationlaw_constants` to compute all of a system's conserved quantities:
```

docs/src/model_creation/network_analysis.md
```diff
@@ -500,6 +500,97 @@ Catalyst.reset_networkproperties!(rn)
...
+```@example network_analysis_conservation_laws
+conservationlaw_constants(rs)
+```
+Next, the `conservedequations` function can be used to compute the algebraic equations produced when a system's conserved quantities are eliminated:
+```@example network_analysis_conservation_laws
+conservedequations(rs)
+```
+Finally, the `conservationlaws` functions yield a $m \times n$ matrix, where $n$ is a system's number of species, $m$ its number of conservation laws, and element $(i,j)$ correspond to the copy number of the $j$th species that occur in the $i$th conserved quantity:
```
isaacsas, line 585:
```suggestion
Finally, the `conservationlaws` function yields a $m \times n$ matrix, where $n$ is a system's number of species, $m$ its number of conservation laws, and element $(i,j)$ corresponds to the copy number of the $j$th species that occurs in the $i$th conserved quantity:
```

docs/src/model_simulation/ode_simulation_performance.md
```diff
@@ -171,6 +171,26 @@ Finally, we note that when using preconditioners with a matrix-free method (like
 Generally, the use of preconditioners is only recommended for advanced users who are familiar with the concepts. However, for large systems, if performance is essential, they can be worth looking into.
 
 
+## [Elimination of system conservation laws](@id ode_simulation_performance_conservation_laws)
+Previously, we have described how Catalyst, when it generates ODEs, is able to [detect and eliminate conserved quantities](@ref network_analysis_conservation_laws). In certain cases, doing this can improve performance. E.g. in the following example we will eliminate the single conserved quantity in a [two-state model](@ref basic_CRN_library_two_states). This results in a differential algebraic equation with a single differential equation and a single algebraic equation (as opposed to two differential equations). Conservation laws can be eliminated by providing the `remove_conserved = true` option to `ODEProblem`:
```
isaacsas, line 175:
```suggestion
Previously, we have described how Catalyst, when it generates ODEs, is able to [detect and eliminate conserved quantities](@ref network_analysis_conservation_laws). In certain cases, doing this can improve performance. E.g. in the following example we will eliminate the single conserved quantity in a [two-state model](@ref basic_CRN_library_two_states). This results in a differential algebraic equation with a single differential equation and a single algebraic equation (as opposed to two differential equations). However, as the algebraic equation is fully determined by the ODE solution, Catalyst moves it to be an observable and our new system therefore only contains one ODE that must be solved numerically. Conservation laws can be eliminated by providing the `remove_conserved = true` option to `ODEProblem`:
```

isaacsas (changes requested):

docs/src/model_simulation/ode_simulation_performance.md
```diff
@@ -171,6 +171,26 @@ Finally, we note that when using preconditioners with a matrix-free method (like
...
+end
+
+# Generate and simulate and ODE (while eliminating any conserved quantities).
+u0 = [:X₁ => 2.0, :X₂ => 3.0]
+ps = [:k₁ => 1.0, :k₂ => 2.0]
+oprob = ODEProblem(rs, u0, (0.0, 10.0), ps; remove_conserved = true)
+sol = solve(oprob)
+nothing # hide
+```
+Conservation law elimination should never impact performance negatively. However, its effect on performance can be limited, and is very model-dependent. It is primarily important when performing simulation with [implicit solvers](@ref ode_simulation_performance_stiffness) (as these requires computing system Jacobians, and conservation law eliminations typically transform these from singular Jacobians to non-singular Jacobians).
```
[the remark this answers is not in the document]
TorkelE: makes sense, thanks for pointing this out

PR #906, opened 2024-06-03
Add comments to network analysis file

TorkelE:
I have gone through the network analysis file and added comments where appropriate. I am not super familiar with it, so I basically went through ti from start to finish, and where there were stuff that was not obvious, I tried to figure out what was going on and added a comment (the file is generally mostly readable without comments though, so not super much added).

The only changes that are not comments are a 4 instances of variable renaming for clarity (I think all should be fairly non-controversial, but am happy to change any back):
- One case where there were some inconsistencies how the variables in a function was named.
- One case where where we used `sts` as name (which makes less sense now when `states` is no longer used as a term in MTK. Since these corresponded to species I renamed it `sps`).
- One case where a function I wrote used variable names inconsistent with the rest of the package.
- One line where I thought the names where non-descriptive. You will have to judge this one when you see it.

In all cases, I have marked these.

### DIFF
--- a/src/network_analysis.jl
+++ b/src/network_analysis.jl
@@ -4,6 +4,8 @@
 function filter_constspecs(specs, stoich::AbstractVector{V}, smap) where {V <: Integer}
     isempty(specs) && (return Vector{Int}(), Vector{V}())
 
+    # If any species are constant, go through these manually and add their indices and 
+    # stoichiometries to `ids` and `filtered_stoich`.
     if any(isconstant, specs)
         ids = Vector{Int}()
         filtered_stoich = Vector{V}()
@@ -44,11 +46,15 @@ function reactioncomplexmap(rn::ReactionSystem)
     !isempty(nps.complextorxsmap) && return nps.complextorxsmap
     complextorxsmap = nps.complextorxsmap
 
+    # Retrieves system reactions and a map from species to their index in the species vector.
     rxs = reactions(rn)
     smap = speciesmap(rn)
     numreactions(rn) > 0 ||
         error("There must be at least one reaction to find reaction complexes.")
+
     for (i, rx) in enumerate(rxs)
+        # Create the `ReactionComplex` corresponding to the reaction's substrates. Adds it 
+        # to the reaction complex dictionary (recording it as the substrates of the i'th reaction). 
         subids, substoich = filter_constspecs(rx.substrates, rx.substoich, smap)
         subrc = sort!(ReactionComplex(subids, substoich))
         if haskey(complextorxsmap, subrc)
@@ -57,6 +63,8 @@ function reactioncomplexmap(rn::ReactionSystem)
             complextorxsmap[subrc] = [i => -1]
         end
 
+        # Create the `ReactionComplex` corresponding to the reaction's products. Adds it 
+        # to the reaction complex dictionary (recording it as the products of the i'th reaction). 
         prodids, prodstoich = filter_constspecs(rx.products, rx.prodstoich, smap)
         prodrc = sort!(ReactionComplex(prodids, prodstoich))
         if haskey(complextorxsmap, prodrc)
@@ -94,7 +102,10 @@ function reactioncomplexes(rn::ReactionSystem; sparse = false)
     isempty(get_systems(rn)) ||
         error("reactioncomplexes does not currently support subsystems.")
     nps = get_networkproperties(rn)
+
+    # If the complexes have not been cached, or the cached complexes uses a different sparsity.
     if isempty(nps.complexes) || (sparse != issparse(nps.complexes))
+        # Computes the reaction complex dictionary. Use it to create a sparse/dense matrix.
         complextorxsmap = reactioncomplexmap(rn)
         nps.complexes, nps.incidencemat = if sparse
             reactioncomplexes(SparseMatrixCSC{Int, Int}, rn, complextorxsmap)
@@ -105,8 +116,11 @@ function reactioncomplexes(rn::ReactionSystem; sparse = false)
     nps.complexes, nps.incidencemat
 end
 
+# Creates a *sparse* reaction complex matrix.
 function reactioncomplexes(::Type{SparseMatrixCSC{Int, Int}}, rn::ReactionSystem,
                            complextorxsmap)
+    # Computes the I, J, and V vectors used for the sparse matrix (read about sparse matrix 
+    # representation for more information).
     complexes = collect(keys(complextorxsmap))
     Is = Int[]
     Js = Int[]
@@ -122,6 +136,7 @@ function reactioncomplexes(::Type{SparseMatrixCSC{Int, Int}}, rn::ReactionSystem
     complexes, B
 end
 
+# Creates a *dense* reaction complex matrix.
 function reactioncomplexes(::Type{Matrix{Int}}, rn::ReactionSystem, complextorxsmap)
     complexes = collect(keys(complextorxsmap))
     B = zeros(Int, length(complexes), numreactions(rn))
@@ -158,6 +173,9 @@ function complexstoichmat(rn::ReactionSystem; sparse = false)
     isempty(get_systems(rn)) ||
         error("complexstoichmat does not currently support subsystems.")
     nps = get_networkproperties(rn)
+
+    # If the complexes stoichiometry matrix has not been cached, or the cached one uses a
+    # different sparsity, computes (and caches) it.
     if isempty(nps.complexstoichmat) || (sparse != issparse(nps.complexstoichmat))
         nps.complexstoichmat = if sparse
             complexstoichmat(SparseMatrixCSC{Int, Int}, rn, keys(reactioncomplexmap(rn)))
@@ -168,7 +186,10 @@ function complexstoichmat(rn::ReactionSystem; sparse = false)
     nps.complexstoichmat
 end
 
+# Creates a *sparse* reaction complex stoichiometry matrix.
 function complexstoichmat(::Type{SparseMatrixCSC{Int, Int}}, rn::ReactionSystem, rcs)
+    # Computes the I, J, and V vectors used for the sparse matrix (read about sparse matrix 
+    # representation for more information).
     Is = Int[]
     Js = Int[]
     Vs = Int[]
@@ -182,6 +203,7 @@ function complexstoichmat(::Type{SparseMatrixCSC{Int, Int}}, rn::ReactionSystem,
     Z = sparse(Is, Js, Vs, numspecies(rn), length(rcs))
 end
 
+# Creates a *dense* reaction complex stoichiometry matrix.
 function complexstoichmat(::Type{Matrix{Int}}, rn::ReactionSystem, rcs)
     Z = zeros(Int, numspecies(rn), length(rcs))
     for (i, rc) in enumerate(rcs)
@@ -213,6 +235,9 @@ function complexoutgoingmat(rn::ReactionSystem; sparse = false)
     isempty(get_systems(rn)) ||
         error("complexoutgoingmat does not currently support subsystems.")
     nps = get_networkproperties(rn)
+
+    # If the outgoing complexes matrix has not been cached, or the cached one uses a
+    # different sparsity, computes (and caches) it.
     if isempty(nps.complexoutgoingmat) || (sparse != issparse(nps.complexoutgoingmat))
         B = reactioncomplexes(rn, sparse = sparse)[2]
         nps.complexoutgoingmat = if sparse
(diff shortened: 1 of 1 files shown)


src/network_analysis.jl
```diff
@@ -436,18 +487,23 @@ subnetworks(sir)
...
     p = parameters(rs)
     t = get_iv(rs)
     spatial_ivs = get_sivs(rs)
     complextorxsmap = [map(first, rcmap) for rcmap in values(reactioncomplexmap(rs))]
     subnetworks = Vector{ReactionSystem}()
+
+    # Loops through each sub-graph of connected reaction complexes. For each, create a 
+    # new `ReactionSystem` model and pushes it to the subnetworks vector.
     for i in 1:length(lcs)
-        reacs, specs, newps = subnetworkmapping(lcs[i], rxs, complextorxsmap, p)
```
TorkelE, line 447: Here, I found it weird that `rxs` was used for the old reactions, and `reacs` was used for the new reactions (while `newps` and `p` was used for the new/old parameters). In this, and the previous `subnetworkmapping` function, I propose we change to `newrxs`, `newspecs`, and `newps` for the new reactions, species, and parameters.

src/network_analysis.jl
```diff
@@ -387,32 +427,43 @@ rcs,incidencemat = reactioncomplexes(sir)
...
+
+    # Find the parameters that are part of the sub-reaction network.
     newps = Vector{eltype(p)}()
-    for rx in rxs
+    for rx in newrxs
         Symbolics.get_variables!(newps, rx.rate, p)
     end
-    rxs, specs, newps   # reactions and species involved in reactions of subnetwork
+
+    newrxs, newspecs, newps
```
TorkelE, line 466: New names, used in this and the next function, see it for clarity.

src/network_analysis.jl
```diff
@@ -387,32 +427,43 @@ rcs,incidencemat = reactioncomplexes(sir)
...
+    # Finds the reactions that are part of teh sub-reaction network.
     rxinds = sort!(collect(Set(rxidx for rcidx in linkageclass
                                for rxidx in complextorxsmap[rcidx])))
-    rxs = allrxs[rxinds]
-    specset = Set(s for rx in rxs for s in rx.substrates if !isconstant(s))
-    for rx in rxs
+    newrxs = allrxs[rxinds]
+
+    # Find the species that are part of the sub-reaction network.
+    specset = Set(s for rx in newrxs for s in rx.substrates if !isconstant(s))
```
TorkelE, line 452: Mark as new line. It is not, it has simply been pushed down one step and uses `newrxs` instead of `rxs`.

src/network_analysis.jl
```diff
@@ -645,25 +709,43 @@ end
...
         scaleby = (N[i, depidx] != 1) ? N[i, depidx] : one(eltype(N))
         (scaleby != 0) || error("Error, found a zero in the conservation law matrix where "
-              *
-              "one was not expected.")
+            * "one was not expected.")
+
+        # Creates, for this conservation law, the sum of all independent species (weighted by
+        # the ratio between the coefficient of the species and the species which is elimianted.
         coefs = @view N[i, indepidxs]
-        terms = sum(p -> p[1] / scaleby * p[2], zip(coefs, indepspecs))
```
TorkelE, line 666: This was probably the function i had to look the most at to understand what was going on. In the end I thought that
```
terms = sum((coef, sp) -> coef / scaleby * sp, zip(coefs, indepspecs))
```
(`p[1]` to `coef` and `p[2]` to `sp`)
was clearer enough that I figured I could propose the change.

src/network_analysis.jl
```diff
@@ -708,13 +792,13 @@ Compute conserved quantities for a system with the given conservation laws.
...
+# If u0s are not given while conservation laws are present, throw an error.
 # Used in HomotopyContinuation and BifurcationKit extensions.
-# Currently only checks if any u0s are given
-# (not whether these are enough for computing conserved quantitites, this will yield a less informative error).
+# Currently only checks if any u0s are given (not whether these are enough for computing
+# conserved quantities, this will yield a less informative error).
 function conservationlaw_errorcheck(rs, pre_varmap)
     vars_with_vals = Set(p[1] for p in pre_varmap)
-    any(s -> s in vars_with_vals, species(rs)) && return
+    any(sp -> sp in vars_with_vals, species(rs)) && return
```
TorkelE, line 801: We don't really use `s` as much throughout, changed this to `sp` which is more commonly used (and also more clearly a species).

src/network_analysis.jl
```diff
@@ -645,25 +709,43 @@ end
 
 # Used in the subsequent function.
 function cache_conservationlaw_eqs!(rn::ReactionSystem, N::AbstractMatrix, col_order)
+    # Retrieves nullity (the number of conservation laws). `r` is the rank of the netstoichmat.
     nullity = size(N, 1)
-    r = numspecies(rn) - nullity     # rank of the netstoichmat
-    sts = species(rn)
```
TorkelE, line 650: `sts` kind of suggests `states`, which is no longer a thing. I would change to `us`, but since these are species specifically, I changed this (here and in two calls over the next few lines) to `sps`.

src/network_analysis.jl
```diff
@@ -122,6 +136,7 @@ function reactioncomplexes(::Type{SparseMatrixCSC{Int, Int}}, rn::ReactionSystem
     complexes, B
```
TorkelE, line 136: Here, and in a lot more cases, I wonder if there are more descriptive names for the matrices? I am really not familiar with standards for network analysis though, and I only think it happened like once or twice that I had to do a actual retake to go around to look what kind of matrix I was dealing with. 
(I have no strong opinions, but a general thought I had)

src/network_analysis.jl
TorkelE, line 666: This was probably the function i had to look the most at to understand what was going on. In the end I thought that
  ```
  terms = sum(coef / scaleby * sp for (coef, sp) in zip(coefs, indepspecs))
  ```
  (`p[1]` to `coef` and `p[2]` to `sp`)
  was clearer enough that I figured I could propose the change.

src/network_analysis.jl
```diff
@@ -628,6 +692,7 @@ conservation laws, each represented as a row in the output.
 function conservationlaws(nsm::T; col_order = nothing) where {T <: AbstractMatrix}
 
     # compute the left nullspace over the integers
+    # The `nullspace` function updates the `col_order`.
```
TorkelE, line 695: Is this true @isaacsas? I tried tracking this variable throughout, and looked into the `nullspace` function, int this is my impression of what happens.
(I was unsure since they call it `nullspace` and not `nullspace!`)

src/reactionsystem.jl
```diff
@@ -76,25 +76,106 @@ Base.Sort.defalg(::ReactionComplex) = Base.DEFAULT_UNSTABLE
...
+    i'th species as a result of the j'th reaction.
+    """
     netstoichmat::Union{Matrix{Int}, SparseMatrixCSC{Int, Int}} = Matrix{Int}(undef, 0, 0)
+    """
+    The reaction network's conservation law matrix. It is an MxN matrix where M is its number of 
+    conservation laws and N its number of species. Element i,j is the coefficient of species 
+    j in the i'th conservation law.
+    """
     conservationmat::Matrix{I} = Matrix{I}(undef, 0, 0)
     col_order::Vector{Int} = Int[]
```
TorkelE, line 96: `col_order` was the only one which I felt unsure of, could you confirm what it is @isaacsas

src/reactionsystem.jl
```diff
@@ -74,28 +74,109 @@ Base.Sort.defalg(::ReactionComplex) = Base.DEFAULT_UNSTABLE
...
+    i'th species as a result of the j'th reaction.
+    """
     netstoichmat::Union{Matrix{Int}, SparseMatrixCSC{Int, Int}} = Matrix{Int}(undef, 0, 0)
+    """
+    The reaction network's conservation law matrix. It is an MxN matrix where M is its number of 
+    conservation laws and N its number of species. Element i,j is the coefficient of species 
+    j in the i'th conservation law.
+    """
     conservationmat::Matrix{I} = Matrix{I}(undef, 0, 0)
     cyclemat::Matrix{I} = Matrix{I}(undef, 0, 0)
```
TorkelE, line 94: `cyclemat` I am unsure about

src/reactionsystem.jl
```diff
@@ -74,28 +74,109 @@ Base.Sort.defalg(::ReactionComplex) = Base.DEFAULT_UNSTABLE
...
+    i to complex j.
+    """
     incidencegraph::Graphs.SimpleDiGraph{Int} = Graphs.DiGraph()
+    """
+    A vector of the connected components of the incidence graph. Each element of the
+    `linkageclasses` corresponds to a connected component, and is a vector listing all the
+    reaction complexes in that connected component.    
+    """
     linkageclasses::Vector{Vector{Int}} = Vector{Vector{Int}}(undef, 0)
     stronglinkageclasses::Vector{Vector{Int}} = Vector{Vector{Int}}(undef, 0)
```
TorkelE, line 172: @vyudu At some point it would be good to have short description of these two, and how they are different from normal linkage classes

src/network_analysis.jl
```diff
@@ -420,32 +462,41 @@ end
...
+# that constitute the corresponding sub-reaction network.
 function subnetworkmapping(linkageclass, allrxs, complextorxsmap, p)
-    rxinds = sort!(collect(Set(rxidx for rcidx in linkageclass
-    for rxidx in complextorxsmap[rcidx])))
-    rxs = allrxs[rxinds]
-    specset = Set(s for rx in rxs for s in rx.substrates if !isconstant(s))
-    for rx in rxs
+    # Finds the reactions that are part of teh sub-reaction network.
+    rxinds = sort!(collect(Set(
+        rxidx for rcidx in linkageclass for rxidx in complextorxsmap[rcidx])))
```
TorkelE, line 483: Old:
```julia
rxinds = sort!(collect(Set(rxidx for rcidx in linkageclass
for rxidx in complextorxsmap[rcidx])))
```
and new
```julia
rxinds = sort!(collect(Set(
    rxidx for rcidx in linkageclass for rxidx in complextorxsmap[rcidx])))
```
are identical, but I moved the new line a bit to make the formater less bad (i.e. starting `for` without indent)

src/network_analysis.jl
```diff
@@ -420,32 +462,41 @@ end
...
+    newrxs = allrxs[rxinds]
+    specset = Set(s for rx in newrxs for s in rx.substrates if !isconstant(s))
+    for rx in newrxs
```
TorkelE, line 486: rxs to newrxs. No other change to this function

src/network_analysis.jl
```diff
@@ -657,27 +716,45 @@ function conservationlaws(nsm::T; col_order = nothing) where {T <: AbstractMatri
...
+    # for each conserved quantity
     for (i, depidx) in enumerate(depidxs)
+        # finds the coefficient (in the conservation law) of the species that is eliminated
+        # by this conservation law
         scaleby = (N[i, depidx] != 1) ? N[i, depidx] : one(eltype(N))
-        (scaleby != 0) || error("Error, found a zero in the conservation law matrix where "
-              *
-              "one was not expected.")
+        (scaleby != 0) ||
+            error("Error, found a zero in the conservation law matrix where one was not expected.")
```
TorkelE, line 750: Chaneg to a single error message on a single new line (rather than having it spread out over 3). Again to avoid wonky formating from the formater

TorkelE:
@vyudu If you want some practise in reviewing PRs, this could be a good try. Basically, I have gone through the old network analysis functions, and added additional comments in a few places (as I understand what the code is doing). You probably know this as well as me, so if you had a look you might spot if any of my comments are inaccurate.

Don't feel that you have to/should do though. But if you thought it would be fun to try and review a PR (if only a little of it), this could be one to try.

isaacsas (changes requested):
This is adding a ton of comments lengthening the code that violate a major rule of commenting: “Comments should not duplicate the code.”

Please make sure your comments are about explaining what a complicated block of code is doing, or why a given choice was made in code, not simply restating in words what relatively simple code already shows.

isaacsas:
Perhaps look at some articles about commenting for guidelines on their use: https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/

TorkelE:
Thanks for the link, I read through it. It is good and I mostly agree.

I don't think this is a problem with the PR though? I read through the file from start to finish, and if there was a bit which was not directly clear to me what it did from the code, I added comments to it. There areen't cases of
```
    # returns the complex stoichiometry matrix.
    nps.complexstoichmat
end
```
or similar.

If there are places where you think a comment is superfluous, please tell me. But I really have tried to keep a balance between adding comments adding superfluous stuff, and the file is not significantly longer thorough this PR.

isaacsas:
Ok. I think there are a lot of superfluous comments here, so will leave a detailed review when I have time.

TorkelE:
If there is any that you see directly you can point it out and I can try to adapt with that info, or explain. But yes, otherwise just go through when you have time, there is no hurry.

src/network_analysis.jl
vyudu: Maybe we could go through and use incidencemat, complexstoichmat, netstoichmat, etc. for these variable names, or maybe standardize the letters we use for each of the matrices and have that somewhere in the documentation (I believe the most common I have seen are Y or Z for the complex stoichiometry matrix, D for the incidence matrix, and S or \Gamma for the reaction stoichiometry matrix)

isaacsas: Except you don't want a variable name to be the same as function names we provide so they should be something different.

isaacsas: Perhaps we should have made the functions `get_netstoichmat` and such but it is too late now for that as it would be breaking. (And our naming is not inconsistent with how base Julia names many functions.)

PR #907, opened 2024-06-04
Expand conservation law tests

TorkelE:
Adds additional tests of conservation laws, especially relating to when attempting to simulate them. 

Only non-test change is that 
```julia
convert(JumpSystem, rs; remove_conserved = true)
```
now throws an `ArgumentError` (as opposed to a generic error). The reason was that this was I could write a test to check that it does throw a specific error (and not just one from wrong input).

### DIFF
--- a/src/reactionsystem_conversions.jl
+++ b/src/reactionsystem_conversions.jl
@@ -653,7 +653,7 @@ function Base.convert(::Type{<:JumpSystem}, rs::ReactionSystem; name = nameof(rs
     spatial_convert_err(rs::ReactionSystem, JumpSystem)
 
     (remove_conserved !== nothing) &&
-        error("Catalyst does not support removing conserved species when converting to JumpSystems.")
+        throw(ArgumentError("Catalyst does not support removing conserved species when converting to JumpSystems."))
 
     flatrs = Catalyst.flatten(rs)
     error_if_constraints(JumpSystem, flatrs)
--- a/test/dsl/dsl_basic_model_construction.jl
+++ b/test/dsl/dsl_basic_model_construction.jl
@@ -71,7 +71,7 @@ let
           Set([:p, :k1, :k2, :k3, :k4, :k5, :k6, :d])
     basic_test(reaction_networks_hill[1], 4, [:X1, :X2],
                [:v1, :v2, :K1, :K2, :n1, :n2, :d1, :d2])
-    basic_test(reaction_networks_constraint[1], 6, [:X1, :X2, :X3],
+    basic_test(reaction_networks_conserved[1], 6, [:X1, :X2, :X3],
                [:k1, :k2, :k3, :k4, :k5, :k6])
     basic_test(reaction_networks_real[1], 4, [:X, :Y], [:A, :B])
     basic_test(reaction_networks_weird[1], 2, [:X], [:p, :d])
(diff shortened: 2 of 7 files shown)


src/reactionsystem_conversions.jl
```diff
@@ -653,7 +653,7 @@ function Base.convert(::Type{<:JumpSystem}, rs::ReactionSystem; name = nameof(rs
     spatial_convert_err(rs::ReactionSystem, JumpSystem)
 
     (remove_conserved !== nothing) &&
-        error("Catalyst does not support removing conserved species when converting to JumpSystems.")
+        throw(ArgumentError("Catalyst does not support removing conserved species when converting to JumpSystems."))
```
TorkelE, line 656: Ensuring that this is a `ArgumentError` allow us to write a better test later on (since we can test for that as opposed to `Exception`, which could be generated by other stuff).

test/dsl/dsl_basic_model_construction.jl
```diff
@@ -71,7 +71,7 @@ let
           Set([:p, :k1, :k2, :k3, :k4, :k5, :k6, :d])
     basic_test(reaction_networks_hill[1], 4, [:X1, :X2],
                [:v1, :v2, :K1, :K2, :n1, :n2, :d1, :d2])
-    basic_test(reaction_networks_constraint[1], 6, [:X1, :X2, :X3],
+    basic_test(reaction_networks_conserved[1], 6, [:X1, :X2, :X3],
```
TorkelE, line 74: I named the `reaction_networks_constraint` networks that when I didn't really know about the terminology that well. Changde the names to something better.

test/test_networks.jl
```diff
@@ -160,69 +160,69 @@ end
...
 ### Reaction networks were some linear combination concentrations remain fixed (steady state values depends on initial conditions). ###
 
-reaction_networks_constraint[1] = @reaction_network rnc1 begin
+reaction_networks_conserved[1] = @reaction_network rnc1 begin
     (k1, k2), X1 ↔ X2
     (k3, k4), X2 ↔ X3
     (k5, k6), X3 ↔ X1
 end
-reaction_network_constraints[1] = [1 1 1]
+reaction_network_conslaws[1] = [1 1 1]
```
TorkelE, line 168: Also a new better name

test/network_analysis/conservation_laws.jl
```diff
@@ -132,19 +160,82 @@ let
...
+    ps_2 = [osys.k1 => 2.0, osys.k2 => 3.0, osys.Γ[1] => 4.0]
+    oprob1 = ODEProblem(osys, u0, 10.0, ps_1)
+    oprob2 = ODEProblem(osys, u0, 10.0, ps_2)
+
+    # Checks that the solutions generates the correct conserved quantities.
+    sol1 = solve(oprob1; saveat = 1.0)
+    sol2 = solve(oprob2; saveat = 1.0)
+    @test all(sol1[osys.X1 + osys.X2] .== 2.0)
+    @test all(sol2[osys.X1 + osys.X2] .== 4.0)
+end
```
TorkelE, line 228: I also wanted to have a test checking
```julia
rn = @reaction_network begin
    @parameters k[1:2]
    @species X(t)[1:2]
    (k[1],k[2]), X[1] <--> X[2]
end
```
when conservation laws are eliminated (would be cool if it worked, but think there's an error). However, testing this took really long time. Might be related to the problem we saw earlier, but not sure.

isaacsas (approved):

PR #915, opened 2024-06-06
Misc updates to reaction system serialisation function

TorkelE:


isaacsas:
Can we delete this branch?

TorkelE:
Yes

I think previously I have been really bad at deleting branches (branches sticking around is not just a thing I haven't really thought of). Once v14 is released I need to go through all old branches and delete all those that are sticking around but shouldn't.

PR #919, opened 2024-06-07
Update smol tutorial

isaacsas:
Need https://github.com/SciML/Catalyst.jl/pull/917 merged first and then to be updated.


docs/src/assets/Project.toml
```diff
@@ -5,7 +5,6 @@ CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
 Catalyst = "479239e8-5488-4da2-87a7-35f2df7eef83"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 DiffEqParamEstim = "1130ab10-4a5a-5621-a13d-e4788d82bd4c"
-DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
```
isaacsas, line 8: @tork I'm not sure why this was added as a dependency suddenly again after the last PR, but no compat was added to go with it so I'm assuming it isn't wanted / needed.

isaacsas:
@TorkelE happy to merge this; it renders locally. If you want to review let me know, otherwise will merge tonight (assuming tests pass).

PR #923, opened 2024-06-07
Fix runtest file

TorkelE:
When I tried to parallelise the tests I introduced some groupings, which were then commented out when it turned out more complicated than expected. Since we do not expect these to be used anytime soon, they probably should be dropped. 

I know David Anthoff is working on an updated version of https://github.com/julia-vscode/TestItemRunner.jl (which currently exists, but is experimental) which will include good parallelisation support. Once that happens, we should probably take a look at this again.


isaacsas (approved):
Out of curiosity what was the issue with grouping (i.e. why couldn’t we mimic OrdinaryDiffEq)?

TorkelE:
To be honest, I have forgotten exactly. I tried just copying stuff, and I did manage so that there were separate test runs for each, however, they would all fail. I asked around in the JuliaLab for help, but the only one who had any input was Avik, who told me to just switch to ReTestItems (I think he is also lobbying to make that a general SciML thing).

PR #925, opened 2024-06-08
[v14 - Ready] Remove empty cross references

TorkelE:
Removes cross-references like `[text](@ref ref)` to places which will not be there at the release of v14.


docs/src/model_creation/examples/basic_CRN_library.md
```diff
@@ -79,7 +79,7 @@ oplt = plot(osol; idxs = X₁ + X₂, title = "Reaction rate equation (ODE)")
 splt = plot(ssol; idxs = X₁ + X₂, title = "Chemical Langevin equation (SDE)")
 plot(oplt, splt; lw = 3, ylimit = (99,101), size = (800,450), layout = (2,1))
 ```
-Catalyst has special methods for working with conserved quantities, which are described [here](@ref ref).
+Catalyst has special methods for working with conserved quantities, which are described here.
```
isaacsas, line 82: This doesn’t make sense without a link as written.

TorkelE: it doesn't thanks for double checking

TorkelE:
It's a good one, adding it

PR #928, opened 2024-06-08
[v14 - Ready] Fix deprecated method usage and unnecessary prints while tests run

TorkelE:
With the test log now mostly cleaned of Symbolics stuff, I have been able to go through the other stuff in there and fix it. This PR fixes a couple of incorrect method usage (a few cases where I do `sys.field` and not `get_field(sys)`). It also fixes some usage in the test files which leads to unnecessary warnings. Finally, I remove some cases where log messages are printed unnecessarily (producing unnecessary test in the test log).

Edit: There are some cases that still remain, I am looking into these, but might take a while to figure out some stuff which is produced by other packages.


Project.toml
```diff
@@ -67,6 +67,7 @@ DomainSets = "5b8099bc-c8ec-5219-889f-1d9e522a28bf"
 Graphviz_jll = "3c863552-8265-54e4-a6dc-903eb78fde85"
 HomotopyContinuation = "f213a82b-91d6-5c5d-acf7-10f1c761b327"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
```
TorkelE, line 70: Logging is required to turn of logging in StructuralIdentifiability. I would suggest to them to not require it to be explicitly loaded, but I feel a bit bad to give further comments on how they set it up (especially since I was the one who suggested they add an option to remove logging in the first place). Since Logging is a base package I am not to worried about adding it here, but removing it is no problem either.

test/reactionsystem_core/coupled_equation_crn_systems.jl
```diff
@@ -47,7 +47,7 @@ let
     # Checks that the correct steady state is found through ODEProblem.
     oprob = ODEProblem(coupled_rs, u0, tspan, ps)
     osol = solve(oprob, Vern7(); abstol = 1e-8, reltol = 1e-8)
-    @test osol[end] ≈ [2.0, 1.0]
+    @test osol.u[end] ≈ [2.0, 1.0]
```
TorkelE, line 50: `osol[end]` now gives deprecation warnings.

test/reactionsystem_core/reactionsystem.jl
```diff
@@ -356,7 +356,7 @@ let
     oprob1 = ODEProblem(osys, u0map, tspan, pmap)
     sts = [B, D, E, C]
     syms = [:B, :D, :E, :C]
-    ofun = ODEFunction(f!; syms)
+    ofun = ODEFunction(f!; sys = ModelingToolkit.SymbolCache(syms))
```
TorkelE, line 359: Not too familiar with this one, simply copied what the deprecation warning said.

src/reactionsystem.jl
```diff
@@ -1479,4 +1479,4 @@ function validate(rs::ReactionSystem, info::String = "")
 end
 
 # Checks if a unit consist of exponents with base 1 (and is this unitless).
-unitless_exp(u) = istree(u) && (operation(u) == ^) && (arguments(u)[1] == 1)
+unitless_exp(u) = iscall(u) && (operation(u) == ^) && (arguments(u)[1] == 1)
```
TorkelE, line 1482: Not sure why this one and others was missed in the previous on. Literally did a ctrl+f for `istree`, but I guess something was up

isaacsas (approved):

PR #929, opened 2024-06-08
Fix bug where `expand_registered_functions` mutated original reaction system

TorkelE:
Previously `expand_registered_functions` would unintentionally apply the operation on the input system. E.g.
```julia
rs_expanded_funcs = Catalyst.expand_registered_functions(rs)
```
would result in both `rs` and `rs_expanded_funcs` having their functions expanded. This fixes this, and adds a test to ensure this remains the case.

### DIFF
--- a/src/registered_functions.jl
+++ b/src/registered_functions.jl
@@ -116,7 +116,7 @@ expand_registered_functions(expr)
 
 Takes an expression, and expands registered function expressions. E.g. `mm(X,v,K)` is replaced with v*X/(X+K). Currently supported functions: `mm`, `mmr`, `hill`, `hillr`, and `hill`.
 """
-function expand_registered_functions(expr)
+function expand_registered_functions!(expr)
     iscall(expr) || return expr
     args = arguments(expr)
     if operation(expr) == Catalyst.mm
@@ -132,13 +132,13 @@ function expand_registered_functions(expr)
                ((args[1])^args[5] + (args[2])^args[5] + (args[4])^args[5])
     end
     for i in 1:length(args)
-        args[i] = expand_registered_functions(args[i])
+        args[i] = expand_registered_functions!(args[i])
     end
     return expr
 end
 # If applied to a Reaction, return a reaction with its rate modified.
 function expand_registered_functions(rx::Reaction)
-    Reaction(expand_registered_functions(rx.rate), rx.substrates, rx.products,
+    Reaction(expand_registered_functions!(deepcopy(rx.rate)), rx.substrates, rx.products,
         rx.substoich, rx.prodstoich, rx.netstoich, rx.only_use_rate, rx.metadata)
 end
 # If applied to a Equation, returns it with it applied to lhs and rhs
--- a/test/reactionsystem_core/custom_crn_functions.jl
+++ b/test/reactionsystem_core/custom_crn_functions.jl
@@ -155,3 +155,17 @@ let
     @test isequal(Catalyst.expand_registered_functions(eq4), 0 ~ V * (K^N) / (X^N + K^N))
     @test isequal(Catalyst.expand_registered_functions(eq5), 0 ~ V * (X^N) / (X^N + Y^N + K^N))
 end
+
+# Ensures that original system is not modified.
+let
+    # Create model with a registered function.
+    @species X(t)
+    @parameters v K
+    rxs = [Reaction(mm(X,v,K), [], [X])]
+    @named rs = ReactionSystem(rxs, t)
+
+    # Check that `expand_registered_functions` does not mutate original model.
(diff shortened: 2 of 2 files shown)


src/registered_functions.jl
```diff
@@ -132,18 +132,18 @@ function expand_registered_functions(expr)
...
     for i in 1:length(args)
-        args[i] = expand_registered_functions(args[i])
+        args[i] = expand_registered_functions!(args[i])
     end
     return expr
 end
 # If applied to a Reaction, return a reaction with its rate modified.
 function expand_registered_functions(rx::Reaction)
-    Reaction(expand_registered_functions(rx.rate), rx.substrates, rx.products,
+    Reaction(expand_registered_functions!(deepcopy(rx.rate)), rx.substrates, rx.products,
```
TorkelE, line 141: Is this appropriate use of `deepcopy`? I could not come up with a better way to do it. If we really want to avoid it it would be possible to first check the expression whether there are any registered functions, and only apply the function in this case

TorkelE: Also, I think technically this by could be rewritten using `@set` (since expanding functions really shouldn't change anything more of the reaction), however, I was reluctant to introduce any more changes than necessary.

isaacsas: It would be more memory efficient to first check if the given expression would be expanded, and only then deepcopy and expand, but otherwise this is fine as a less memory efficient version.

isaacsas: (Note, I'm not saying this needs to be done in this PR, just that it would be good to switch to that eventually.)

TorkelE:
Sorry, totally forgot about this one. I think I will add the checks, it should be fairly straightforward with the new symbolic functions.

TorkelE:
Ok, I have changed around the internals (no longer using `deepcopy`, although Symbolics might use something similar internally, I am not sure) and also supporting events.

src/registered_functions.jl
```diff
@@ -117,7 +117,23 @@ expand_registered_functions(expr)
...
+        expr = replacenode(expr, expand_catalyst_function)
+    end
+    return expr
+end
+
+# Checks whether an expression corresponds to a catalyst function call (e.g. `mm(X,v,K)`).
+function is_catalyst_function(expr)
+    iscall(expr) || (return false)
+    return operation(expr) in [
+        Catalyst.mm, Catalyst.mmr, Catalyst.hill, Catalyst.hillr, Catalyst.hillar]
```
isaacsas, line 130: Use a tuple instead of vector to avoid allocations.

src/registered_functions.jl
```diff
@@ -131,23 +147,53 @@ function expand_registered_functions(expr)
...
+    return [Catalyst.expand_registered_functions(element) for element in vec]
+end
+
 # If applied to a ReactionSystem, applied function to all Reactions and other Equations, and return updated system.
+# Currently, `ModelingToolkit.has_X_events` returns `true` even if event vector is empty (hence
+# this function cannot be used).
 function expand_registered_functions(rs::ReactionSystem)
-    @set! rs.eqs = [Catalyst.expand_registered_functions(eq) for eq in get_eqs(rs)]
-    @set! rs.rxs = [Catalyst.expand_registered_functions(rx) for rx in get_rxs(rs)]
+    if isdefined(Main, :Infiltrator)
```
isaacsas, line 186: What’s this?

src/registered_functions.jl
```diff
@@ -131,23 +147,53 @@ function expand_registered_functions(expr)
...
+    end
+
+    @set! rs.eqs = Catalyst.expand_registered_functions(get_eqs(rs))
+    @set! rs.rxs = Catalyst.expand_registered_functions(get_rxs(rs))
+    if !isempty(ModelingToolkit.get_continuous_events(rs))
+        @set! rs.continuous_events = Catalyst.expand_registered_functions(ModelingToolkit.get_continuous_events(rs))
+    end
+    if !isempty(ModelingToolkit.get_discrete_events(rs))
+        @set! rs.discrete_events = Catalyst.expand_registered_functions(ModelingToolkit.get_discrete_events(rs))
+    end
```
isaacsas, line 197:
```suggestion
    end
    reset_networkproperties!(rs)
```

isaacsas (changes requested):

src/registered_functions.jl
isaacsas: This could also just be a global `const` variable.

src/registered_functions.jl
TorkelE: debugging stuff, removing it

src/registered_functions.jl
TorkelE: good point, yes, I am turning it to a const

PR #934, opened 2024-06-09
[v14 - Ready] Migration guide and fixes to history file

TorkelE:
Right now I am creating the migration guide as a hidden page in the documentation, which then can link from e.g. the GitHub home page. 

I will also go through the history file and add anything that was missed, fix formatting/writing errors, and improve parts that are a bit short.


docs/make.jl
```diff
@@ -40,7 +40,6 @@ makedocs(sitename = "Catalyst.jl",
     doctest = false,
     clean = true,
     pages = pages,
-    pagesonly = true,
```
TorkelE, line 43: I know want to build pages that are not actually on the pages list. That also means that I need to move the PEtab doc until we are ready to put that in (when PEtab is updated with respect to the recent changes).

HISTORY.md
```diff
@@ -17,47 +91,9 @@ assess_identifiability(goodwind_oscillator; measured_quantities=[:M])
...
-rn = @reaction_network begin
-    @parameters η
-    k, 2X --> X2, [noise_scaling=η]
-end
-getnoisescaling(rn)
-```
-- `SDEProblem` no longer takes the `noise_scaling` argument (see above for new approach to handle noise scaling).
-- Changed fields of internal `Reaction` structure. `ReactionSystems`s saved using `serialize` on previous Catalyst versions cannot be loaded using this (or later) versions.
-- Simulation of spatial ODEs now supported. For full details, please see https://github.com/SciML/Catalyst.jl/pull/644 and upcoming documentation. Note that these methods are currently considered alpha, with the interface and approach changing even in non-breaking Catalyst releases.
-- LatticeReactionSystem structure represents a spatial reaction network:
```
TorkelE, line 32: I will remove the spatial entry for now, and add it back in when I push the update with the new interface (post v14)

HISTORY.md
```diff
@@ -17,47 +91,9 @@ assess_identifiability(goodwind_oscillator; measured_quantities=[:M])
...
-  X's value will be `1.0` in the first vertex, but `0.0` in the remaining one (the system have 25 vertexes in total). SInce th parameters `p` and `d` are part of the non-spatial reaction network, their values are tied to vertexes. However, if the `D` parameter (which governs diffusion between vertexes) is given several values, these will instead correspond to the specific edges (and transportation along those edges.)
+- A more detailed of how this extension works can be found [here](https://docs.sciml.ai/Catalyst/stable/inverse_problems/structural_identifiability/).
 
-- Update how compounds are created. E.g. use
-```julia
-@variables t C(t) O(t)
-@compound CO2 ~ C + 2O
-```
-to create a compound species `CO2` that consists of `C` and 2 `O`.
-- Added documentation for chemistry related functionality (compound creation and reaction balancing).
```
TorkelE, line 60: Some sections got moved a bit to accommodate for sectioning the new v14 history entry into related sections.

README.md
```diff
@@ -34,7 +34,8 @@ etc).
 **NOTE:** version 14 is a breaking release, prompted by the release of ModelingToolkit.jl version 9. This caused several breaking changes in how Catalyst models are represented and interfaced with.
 
 Breaking changes and new functionality are summarized in the
-[HISTORY.md](HISTORY.md) file.
+[HISTORY.md](HISTORY.md) file. Furthermore, a migration guide on how to adapt your workflows to the 
+the new v14 update can be found [here](https://docs.sciml.ai/Catalyst/stable/v14_migration_guide/).
```
TorkelE, line 38: I hope this works out as intended (making the migration guide a doc page only accessible through this links).

docs/src/v14_migration_guide.md
```diff
@@ -0,0 +1,193 @@
+# Version 14 Migration Guide
```
TorkelE, line 1: I think these are the points we should mention, but tell me if I forgot something.

docs/src/v14_migration_guide.md
```diff
@@ -0,0 +1,193 @@
...
+oprob_remade = remake(oprob; u0 = [:X1 => 5.0]) # NEVER do this.
+sol(oprob)
+# hide
+```
+This might generate a silent error, where the remade problem is different from the intended one. 
+
+This bug was likely present on earlier versions as well, but was only recently discovered. While we hope it will be fixed soon, the bug is in ModelingToolkit, and will not be fixed until its maintainers find the time to do so.
+
+#### Depending on parameter order is even more dangerous than before
+In early versions of Catalyst, parameters and species were provided as vectors (e.g. `[1.0, 2.0]`) rather than maps (e.g. `[p => 1.0, d => 2.0]`). While we have already *strongly* recommended users to use the map form (or they might produce unintended results), the vector form is still (technically). Due to recent internal ModelingToolkit updates, the risk of unexpected behaviour when providing parameter values as vectors is even larger than before.
```
isaacsas, line 188:
```suggestion
In early versions of Catalyst, parameters and species were provided as vectors (e.g. `[1.0, 2.0]`) rather than maps (e.g. `[p => 1.0, d => 2.0]`). While we previously *strongly* recommended users to use the map form (or they might produce unintended results), the vector form was still supported (technically). Due to recent internal ModelingToolkit updates, the purely numeric form is no longer supported and should never be used -- it will potentially lead to incorrect values for parameters and/or initial conditions. Note that if `rn` is a complete `ReactionSystem` you can now specify such mappings via `[rn.p => 1.0, rn.d => 2.0]`.
```

isaacsas (changes requested):
I'd suggest changing the "stable" links to V14 links once the V14 docs are posted. That we the links won't go out of date if we change around the docs later on.

HISTORY.md
```diff
@@ -3,8 +3,82 @@
...
+- The `default_t()` and `default_time_deriv()` functions are now the preferred approaches for creating the default time independent variable and its differential.
+- It is now possible to add metadata to individual reactions, e.g. using:
+```julia
+rn = @reaction_network begin
+    @parameters η
+    k, 2X --> X2, [description="Dimerisation"]
+end
+getdescription(rn)
+```
+a more detailed description can be found [here](https://docs.sciml.ai/Catalyst/dev/model_creation/dsl_advanced/#dsl_advanced_options_reaction_metadata).
```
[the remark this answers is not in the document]
TorkelE: Yes. Implementation of new reaction metadata, and the use of the `hasmetadata` `getmetadata` functions are currently internal and un-documented. If we document these it would be part of the dev-docs on how to create new metadata.

Currently we only support the `noise_scaling`, `description`, and `misc` metatadata, each supported by their own set of getters (MTK style).

docs/src/v14_migration_guide.md
```diff
@@ -0,0 +1,193 @@
...
+- Only complete systems can be used as inputs to simulations or most tools for model analysis.
+- Only incomplete systems can be [composed with other systems to form hierarchical models](@ref compositional_modeling).
+
+A model's completeness depends on how it was created:
+- Models created programmatically (using the `ReactionSystem` constructor) are *incomplete*.
+- Models created using the `@reaction_network` DSL are *complete*.
+- To create *incomplete models using the DSL*, use the `@network_component` macro (in all other aspects identical to `@reaction_network`).
+- Models generated through the `compose` and `extend` functions are *incomplete*.
+
+Furthermore, any systems generated through e.g. `convert(ODESystem, rs)` are also complete.
```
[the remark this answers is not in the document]
TorkelE: no, you are right. This is just an error on my part, should have be "are also incomplete". 

I will change it to "... `convert(ODESystem, rs)` are not marked as complete"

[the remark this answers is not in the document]
TorkelE: Now it is
```
Furthermore, any systems generated through e.g. `convert(ODESystem, rs)`  are *not marked as complete*. 
```

TorkelE:
Updated. You are correct about completeness and `convert`, that was a miss-writing that is now corrected. You are also correct regarding the metadata, the full implementation is kept internal, and we only expose the `noise_scaling`, `description`, and `misc` metadata, each which has their own dedicated MTK-style getters.

PR #937, opened 2024-06-10
Add warning to ReactionSystem field list

TorkelE:


src/reactionsystem.jl
```diff
@@ -204,6 +204,14 @@ end
 
 ### ReactionSystem Structure ###
 
+""" 
+WARNING!!!
+
+The following variable is used to check that code that should be updated when the `ReactionSystem` 
+fields are updated have in fact been updated. Do not just blindly update this without first checking
```
isaacsas, line 211:
```suggestion
fields are updated has in fact been updated. Do not just blindly update this without first checking 
```

isaacsas (approved):

PR #938, opened 2024-06-10
fix jump internals test

isaacsas:


test/reactionsystem_core/reactionsystem.jl
```diff
@@ -159,7 +159,6 @@ end
 
 # Test with JumpSystem.
 let
-    p = rand(rng, length(k))
     @species A(t) B(t) C(t) D(t) E(t) F(t)
     rxs = [Reaction(k[1], nothing, [A]),            # 0 -> A
```
isaacsas, line 163: @TorkelE why is all this code on building the symbolic reaction system within the let block -- isn't it the same system built at the top of this file?

TorkelE: I think it is modified so that a few species at the end are different (i.e. the `E` and the `F` species are only in the second network). Not sure why there is a separate one though, it was originally added here: https://github.com/SciML/Catalyst.jl/pull/535

isaacsas:
Doc tests are failing, but that seems unrelated. @TorkelE ok for me to merge?

PR #940, opened 2024-06-10
Fix parametric stoich tutorial

isaacsas:


isaacsas:
I reset `warnonly = true` in the doc build for now so builds can complete. We need to update this once the remaining link errors are fixed.

PR #943, opened 2024-06-10
[WIP] Fix remaining doc issues

TorkelE:
Fixes remaining things that causes errors/warnings in the docs


docs/src/faqs.md
```diff
@@ -59,7 +59,7 @@ plot(sol; idxs = [A, B])
 ```
 
 ## How to disable rescaling of reaction rates in rate laws?
-As explained in the [Reaction rate laws used in simulations](@ref) section, for
```
TorkelE, line 62: At least in the old docs this yields working links, however, Documenter produces a warning/error from it. writing out the reference, while annoying, removes this while preserving behaviours.

docs/src/inverse_problems/behaviour_optimisation.md
```diff
@@ -1,5 +1,5 @@
 # [Optimization for non-data fitting purposes](@id behaviour_optimisation)
-In previous tutorials we have described how to use [PEtab.jl](@ref petab_parameter_fitting) and [Optimization.jl](@ref optimization_parameter_fitting) for parameter fitting. This involves solving an optimisation problem (to find the parameter set yielding the best model-to-data fit). There are, however, other situations that require solving optimisation problems. Typically, these involve the creation of a custom cost function, which optimum can then be found using Optimization.jl. In this tutorial we will describe this process, demonstrating how parameter space can be searched to find values that achieve a desired system behaviour. A more throughout description on how to solve these problems is provided by [Optimization.jl's documentation](https://docs.sciml.ai/Optimization/stable/) and the literature[^1].
```
TorkelE, line 2: Because the PEtab docs are disabled until Catalyst v14 is out, and then the follow up PEtab release is ready.

docs/src/introduction_to_catalyst/introduction_to_catalyst.md
```diff
@@ -1,7 +1,7 @@
 # [Introduction to Catalyst](@id introduction_to_catalyst)
 In this tutorial we provide an introduction to using Catalyst to specify
 chemical reaction networks, and then to solve ODE, jump, and SDE models
-generated from them. At the end we show what mathematical rate laws and
+generated from them[1]. At the end we show what mathematical rate laws and
```
isaacsas, line 4: If this is supposed to be a reference then I'd just convert the list back to a footnote. I only switched it because it was giving errors when it was unreferenced.

docs/src/model_creation/dsl_advanced.md
```diff
@@ -463,3 +463,77 @@ A reaction's metadata can be accessed using specific functions, e.g. `Catalyst.h
 rx = @reaction p, 0 --> X, [description="A production reaction"]
 Catalyst.getdescription(rx)
 ```
+
+## [Working with symbolic variables and the DSL](@id dsl_advanced_options_symbolics_and_DSL)
```
TorkelE, line 467: Previously I moved these sections to the programmatic modelling section (as it has much better descriptions of symbolic variables and why this can be useful).  Then I planned to just put references to that section here.

I simply copied that section over here, it fits well here as well.

isaacsas (commented):
Can we change the warning settings back to the SciML defaults now then?

docs/src/introduction_to_catalyst/introduction_to_catalyst.md
TorkelE: Ahh, was a bit unsure if that's what you were doing. I will check how the printed docs looks like, but converting back to a list makes sense.

TorkelE:
> Can we change the warning settings back to the SciML defaults now then?

Good point, agreed.

isaacsas:
I switched unreferenced footnote sections to lists, but if they are referenced it makes sense to keep them as footnotes.

isaacsas:
Feel free to merge after you are done with the updates.

isaacsas (approved):

Issue #936, opened 2024-06-10
show in a tutorial how to do periodic events with jumps


isaacsas:
Given they are not likely to support `PeriodicCallback`s any time soon due to https://github.com/SciML/JumpProcesses.jl/issues/417,

isaacsas:
We should just show how to write a callback that updates the `tstop`s correctly for the next event.

PR #946, opened 2024-06-10
Tutorial on periodic events

TorkelE:
I briefly go through periodic events in ODEs as well (primarily as a baseline to compare to, but I think there is some general useful input on these as well. Jump simulation peridoci events is the main thing though).

Intended to fix https://github.com/SciML/Catalyst.jl/issues/936


docs/src/model_simulation/examples/periodic_events_simulation.md
```diff
@@ -0,0 +1,95 @@
...
+
+## [Modelling a circadian periodic event in an ODE simulation](@id periodic_event_simulation_example_ode)
+We will consider a simple circadian model, consisting of a single protein ($X$), which is phosphorylated ($X \to Xᴾ$) in the presence of light ($l$). Here, the light parameter can either be $0$ (night) or $1$ (day). We can model this using a simple periodic event which switches the value of $l$ every 12 hours (here, `%` is the [remainder operator](https://docs.julialang.org/en/v1/manual/mathematical-operations/#Arithmetic-Operators)).
+```@example periodic_event_example
+using Catalyst
+circadian_model = @reaction_network begin
+    @discrete_events begin
+        12 => [l ~ (l + 1)%2]
```
isaacsas, line 10: MTK's events really need some attention. I really don't like this having to hardcode a parameter value (the event time) within the callback. We need a way to specify periodic and preset using symbolic variables to encode times (along with a way to avoid users having to also specify `tstop` values manually for discrete events).

docs/src/model_simulation/examples/periodic_events_simulation.md
```diff
@@ -0,0 +1,95 @@
...
+end
+
+using JumpProcesses
+u0 = [:X => 150, :Xᴾ => 50]
+ps = [:kₚ => 0.1, :kᵢ => 0.1, :l => 1.0]
+dprob = DiscreteProblem(circadian_model, u0, (0.0, 100.0), ps)
+jprob = JumpProblem(circadian_model, dprob, Direct())
+nothing # hide
+```
+Next, we implement our event through a `DiscreteCallback`. It triggers every 12 time units, and when it does, it flips the value of `l`. Since we will interface with $l$'s value multiple times, we [create specific `getl` and `setl` functions](@ref simulation_structure_interfacing_functions) to do this (in practice, due to this model's small size, the performance impact is negliable).
```
isaacsas, line 39:
```suggestion
Next, we implement our event through a `DiscreteCallback`. It triggers every 12 time units, and when it does, it flips the value of `l`. Since we will interface with $l$'s value multiple times, we [create specific `getl` and `setl` functions](@ref simulation_structure_interfacing_functions) to do this (in practice, due to this model's small size, the performance increase is minimal).
```
Just to clarify that the impact should be a performance increase (even if minimal).

docs/src/model_simulation/examples/periodic_events_simulation.md
```diff
@@ -0,0 +1,95 @@
...
+setl = ModelingToolkit.setp(jprob, :l)
+function condition(u, t, integrator)
+ (t % 12) == 0.0
+end
+function affect!(integrator)
+    setl(integrator, (getl(integrator) + 1) % 2)
+    reset_aggregated_jumps!(integrator)
+end
+callback = DiscreteCallback(condition, affect!)
+nothing # hide
```
isaacsas, line 51: Why can't you create a symbolic discrete event with the same condition in the original model? That should work right? (It is just periodic events that are broken.)

docs/src/model_simulation/examples/periodic_events_simulation.md
```diff
@@ -0,0 +1,95 @@
...
+```
+!!! danger
+    Whenever the integrator of a jump simulator is modified, the `reset_aggregated_jumps!` function must be called (with the integrator as its input) afterwards. Omitting to do so will result in erroneous simulations.
```
isaacsas, line 54:
```suggestion
    Whenever the integrator of a jump simulator is modified, the `reset_aggregated_jumps!` function must be called (with the integrator as its input) afterwards. Omitting to do so will result in simulations (incorrectly) using the old values of parameters/states internally instead of the updated values.
```

docs/src/model_simulation/examples/periodic_events_simulation.md
```diff
@@ -0,0 +1,95 @@
...
+```
+The reason is that discrete callbacks' conditions are only checked at the end of each simulation time step, and the probability that these will coincide with the callback's trigger times ($t = 12, 24, 36, ...$) is infinitely small. Hence, we must directly instruct our simulation to stop at these time points. We can do this using the `tstops` argument:
+```@example periodic_event_example
+tstops = 12.0:12.0:dprob.tspan[2]
+sol = solve(jprob, SSAStepper(); callback, tstops)
+plot(sol)
+```
+
+## [Plotting the light level](@id periodic_event_simulation_plotting_light)
+Sometimes when simulating models with periodic parameters, one would like to plot the parameter's value across the simulation. To do this, we must turn our parameter $l$ to a *variable* (so that its value is recorded throughout the simulation):
```
isaacsas, line 69: Shouldn't this example instead use a saving callback (https://docs.sciml.ai/DiffEqCallbacks/stable/output_saving/)? That seems preferred to making the parameter into a variable when it really isn't a variable?

isaacsas (changes requested):

docs/src/model_simulation/examples/periodic_events_simulation.md
TorkelE: I guess there is a problem what to do if the parameter(s) that set the time is updated mid-simulation, but once it is decided how to deal with that it should be straightforward.

docs/src/model_simulation/examples/periodic_events_simulation.md
TorkelE: Yeah, I probably just missed that. Will reimplement using that approach.

docs/src/model_simulation/examples/periodic_events_simulation.md
isaacsas: I don't think there could be a plotting receipe since you can save anything via it if I understand correctly.

TorkelE: right, that makes sense

docs/src/model_simulation/examples/periodic_events_simulation.md
isaacsas: (It might be good to show both versions since this is a nice example of writing a custom callback at a lower-level anyways.)

docs/src/model_simulation/examples/periodic_events_simulation.md
TorkelE: Right now I only make a reference to this, hopefully at some point I can look into it properly and make a full section.

docs/src/model_simulation/examples/periodic_events_simulation.md
TorkelE: yeah, this works much better

PR #953, opened 2024-06-13
Tests for conservation laws and `remake`

TorkelE:
Some stuff is broken, although maybe not exactly the stuff that we thought was.


ChrisRackauckas:
Can you share the MTK system created from this so I can talk about it without grabbing the Catalyst in progress branch? Most of those don't look "broken", it looks like you specified an incompatible system.

TorkelE:
Will do. I think right now I am not 100% certain that the errors are fully in MTK (i.e. some of the failed conservation law issues look odd). Once we know what might actually be an MTK problem I will make corresponding issues over there (using MTK systems) instead.

TorkelE:
Maybe we reopen this in the future, but for now we will probably use https://github.com/SciML/Catalyst.jl/pull/955 instead.

PR #955, opened 2024-06-14
[v14] Conservation law warnings

TorkelE:
Adds a warning to hwenever conservation laws are removed (about `remake` and similar not working), which can be toggled off if required to.


src/reactionsystem_conversions.jl
```diff
@@ -450,6 +450,17 @@ diff_2_zero(expr) = (Symbolics.is_derivative(expr) ? 0 : expr)
 
 COMPLETENESS_ERROR = "A ReactionSystem must be complete before it can be converted to other system types. A ReactionSystem can be marked as complete using the `complete` function."
 
+# Used to, when required, display a warning about conservation law removeal and remake.
+function check_cons_warning(remove_conserved, remove_conserved_warn)
+    (remove_conserved && remove_conserved_warn) || return
+    @warn "You are creating a system while eliminating conserved quantities. While this is possible, 
+        if you use the created system to create a problem (e.g. an `ODEProblem`), you *should not*
+        modify that problem's species values (e.g. using `remake`). Modification of parameter values
```
isaacsas, line 458:
```suggestion
        modify that problem's initial conditions for species (e.g. using `remake`). Changing initial conditions must be done by creating a new Problem from your reaction system or the ModelingToolkit system you converted it into with the new initial condition map. Modification of parameter values 
```

src/reactionsystem_conversions.jl
```diff
@@ -450,6 +450,17 @@ diff_2_zero(expr) = (Symbolics.is_derivative(expr) ? 0 : expr)
 
 COMPLETENESS_ERROR = "A ReactionSystem must be complete before it can be converted to other system types. A ReactionSystem can be marked as complete using the `complete` function."
 
+# Used to, when required, display a warning about conservation law removeal and remake.
+function check_cons_warning(remove_conserved, remove_conserved_warn)
+    (remove_conserved && remove_conserved_warn) || return
+    @warn "You are creating a system while eliminating conserved quantities. While this is possible, 
+        if you use the created system to create a problem (e.g. an `ODEProblem`), you *should not*
+        modify that problem's species values (e.g. using `remake`). Modification of parameter values 
+        is still possible. You might get this warning when creating a problem directly.
```
isaacsas, line 459:
```suggestion
        is still possible, and directly setting numerical values for any conservation law constants will work. You might get this warning when creating a problem directly.
```
We should check this statement is correct.

src/reactionsystem_conversions.jl
```diff
@@ -450,6 +450,17 @@ diff_2_zero(expr) = (Symbolics.is_derivative(expr) ? 0 : expr)
 
 COMPLETENESS_ERROR = "A ReactionSystem must be complete before it can be converted to other system types. A ReactionSystem can be marked as complete using the `complete` function."
 
+# Used to, when required, display a warning about conservation law removeal and remake.
+function check_cons_warning(remove_conserved, remove_conserved_warn)
+    (remove_conserved && remove_conserved_warn) || return
+    @warn "You are creating a system while eliminating conserved quantities. While this is possible,
```
isaacsas, line 456:
```suggestion
    @warn "You are creating a system or problem while eliminating conserved quantities. Please note, due to limitations / design choices in ModelingToolkit
```

src/reactionsystem_conversions.jl
```diff
@@ -450,6 +450,17 @@ diff_2_zero(expr) = (Symbolics.is_derivative(expr) ? 0 : expr)
 
 COMPLETENESS_ERROR = "A ReactionSystem must be complete before it can be converted to other system types. A ReactionSystem can be marked as complete using the `complete` function."
 
+# Used to, when required, display a warning about conservation law removeal and remake.
+function check_cons_warning(remove_conserved, remove_conserved_warn)
+    (remove_conserved && remove_conserved_warn) || return
+    @warn "You are creating a system while eliminating conserved quantities. While this is possible, 
+        if you use the created system to create a problem (e.g. an `ODEProblem`), you *should not*
```
isaacsas, line 457:
```suggestion
        if you use the created system to create a problem (e.g. an `ODEProblem`), or a directly creating a problem, you *should not*
```

isaacsas (changes requested):

isaacsas (approved):

isaacsas:
Oh wait, your other test PR says that setting gamma directly with numerical values doesn’t work either? If so we should modify the error message to not suggest that works…

isaacsas (approved):

PR #963, opened 2024-06-18
Strong and terminal linkage classes

vyudu:
Added ability to detect strong and terminal linkage classes, will be useful for concentration robustness check later on


test/network_analysis/network_properties.jl
```diff
@@ -326,3 +326,81 @@ let
     @test Catalyst.iscomplexbalanced(rn, rates) == true 
 end
 
+### STRONG LINKAGE CLASS TESTS
+
+let
```
TorkelE, line 331: Generally I try to, before each `let ... end` block in the tests, have some kind of comment what is being tested. That can be useful if someone tries to understand the tests. Especially long-term if they need to be modified for some reason.

Here I presume you just test the `stronglinkageclasses` and `terminallinkageclasses` functions? However, it might be good to state that, and also if there are some important differences between the networks. If each test just tests these functions and that is it, it does not hurt to also mention that.

src/network_analysis.jl
```diff
@@ -359,6 +359,51 @@ end
...
+    nps = get_networkproperties(rn)
+    if isempty(nps.terminallinkageclasses)
+        slcs = stronglinkageclasses(rn)
+        tslcs = filter(lc -> isterminal(lc, rn), slcs)
+        nps.terminallinkageclasses = tslcs
+    end
+    nps.terminallinkageclasses
+end
+
+function isterminal(lc::Vector, rn::ReactionSystem)
```
TorkelE, line 394: For any non-trivial function, I usually try to add a comment of what it does. I.e. not being familiar with terminal linkage classes, I am not sure exactly what is being checked for terminal.

src/network_analysis.jl
```diff
@@ -359,6 +359,51 @@ end
...
+    imat = incidencemat(rn)
+
+    for r in 1:size(imat, 2)
```
TorkelE, line 397: These functions are borderline what I think is readable without additional comments in the code. However, in maybe this case I would err on the side of caution and add some comment on what is actually going on.

TorkelE: E.g. check https://github.com/SciML/Catalyst.jl/pull/906/files for some examples of these kinds of functions with comments added.
(a lot of these stuff was written long ago when SciML packages typically had less comments, but we are trying to improve this in Catalyst)

src/reactionsystem.jl
```diff
@@ -93,6 +94,8 @@ Base.@kwdef mutable struct NetworkProperties{I <: Integer, V <: BasicSymbolic{Re
     complexoutgoingmat::Union{Matrix{Int}, SparseMatrixCSC{Int, Int}} = Matrix{Int}(undef, 0, 0)
     incidencegraph::Graphs.SimpleDiGraph{Int} = Graphs.DiGraph()
     linkageclasses::Vector{Vector{Int}} = Vector{Vector{Int}}(undef, 0)
+    stronglinkageclasses::Vector{Vector{Int}} = Vector{Vector{Int}}(undef, 0)
+    terminallinkageclasses::Vector{Vector{Int}} = Vector{Vector{Int}}(undef, 0)
```
isaacsas, line 98: Have you updated `reset!` for these new fields too?

isaacsas (changes requested):

src/reactionsystem.jl
isaacsas: (including `cycles`?)

vyudu: Done - will fix on the cycles PR as well

isaacsas:
@vyudu, if you can merge master into this I'll merge once tests pass.

isaacsas:
Looks like there is a test issue.

PR #964, opened 2024-06-18
Deficiency one and concnetration robustness

vyudu:
Concentration robustness for deficiency one networks. Relies on [strong and terminal linkage classes](https://github.com/SciML/Catalyst.jl/pull/963).


src/reactionsystem.jl
```diff
@@ -137,7 +140,7 @@ function reset!(nps::NetworkProperties{I, V}) where {I, V}
     empty!(nps.linkageclasses)
     empty!(nps.stronglinkageclasses)
     empty!(nps.terminallinkageclasses)
-    nps.deficiency = 0
+    nps.deficiency = -1
```
isaacsas, line 143: `empty!(robustspecies` and reset `checkedrobust` too?

src/network_analysis.jl
```diff
@@ -938,3 +942,79 @@ end
...
+    # A species is concnetration robust in a deficiency one network if there are two non-terminal complexes (i.e. complexes 
+    # belonging to a linkage class that is not terminal) that differ only in species s (i.e. their difference is some 
+    # multiple of s. (A + B, A) differ only in B. (A + 2B, B) differ in both A and B, since A + 2B - B = A + B). 
+
+    if !nps.checkedrobust
+        tslcs = terminallinkageclasses(rn)
+        Z = complexstoichmat(rn)
+
+        # Find the complexes that do not belong to a terminal linkage class 
+        nonterminal_complexes = deleteat!([1:length(complexes);], vcat(tslcs...))
```
isaacsas, line 1005: This feels like it could maybe be done more efficiently by accumulating the ones you want instead of deleting the ones you don't want?

src/network_analysis.jl
```diff
@@ -938,3 +942,79 @@ end
...
+        robust_species = Int64[]
+
+        for (c_s, c_p) in collect(Combinatorics.combinations(nonterminal_complexes, 2))
+            # Check the difference of all the combinations of complexes. The support is the set of indices that are non-zero 
+            supp = findall(!=(0), Z[:, c_s] - Z[:, c_p])
```
isaacsas, line 1010: The column difference could be stored in a pre-calculated vector from outside the loop like
```julia
@. tmp = @views Z[:, c_2] - Z[:, c_p]
```

isaacsas: More generally, seems like this could be written as a non-allocating loop that bails if a second match is found instead of continuing the search, and if a second match isn't found saves it in `robust_species`.

isaacsas (changes requested):

src/network_analysis.jl
```diff
@@ -938,3 +942,79 @@ end
...
+"""
+    satisfiesdeficiencyzero(rn::ReactionSystem)
+
+    Check if a reaction network obeys the conditions of the deficiency zero theorem, which ensures that there is only one equilibrium for every positive stoichiometric compatibility class, this equilibrium is asymptotically stable, and this equilibrium is complex balanced.
+"""
+
+function satisfiesdeficiencyzero(rn::ReactionSystem)
+    δ = deficiency(rn)
+    δ == 0 && isweaklyreversible(rn, subnetworks(rn))
+end
```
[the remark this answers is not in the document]
vyudu: Ah yes good catch, will fix

src/network_analysis.jl
vyudu: It does seem like `deleteat!` is close to optimal here, see https://stackoverflow.com/questions/64957524/how-can-i-obtain-the-complement-of-list-of-indexes-in-julia

src/network_analysis.jl
vyudu: Re the first comment: would I accomplish the same thing by adding `@views` in the `supp` line, so it becomes
```julia
supp = findall(!=(0), Z[:, c_s] - Z[:, c_p]
```
Seems like allocating it entirely outside might take a lot of space (would have to be a 3D array, where each difference vector is indexed by (c_s, c_p))

For the second comment, what do you mean by second match?

src/network_analysis.jl
isaacsas: OK!

src/network_analysis.jl
```diff
@@ -938,3 +942,83 @@ end
...
+
+    if !nps.checkedrobust
+        tslcs = terminallinkageclasses(rn)
+        Z = complexstoichmat(rn)
+
+        # Find the complexes that do not belong to a terminal linkage class 
+        nonterminal_complexes = deleteat!([1:length(complexes);], vcat(tslcs...))
+        robust_species = Int64[]
+
+        for (c_s, c_p) in Combinatorics.combinations(nonterminal_complexes, 2)
```
isaacsas, line 1012: Are `c_s` and `c_p` always scalars? Perhaps I've misunderstood the code.

isaacsas: OK, then couldn't you do something like (note this is not tested):
```julia
        for (c_s, c_p) in Combinatorics.combinations(nonterminal_complexes, 2)
            # Check the difference of all the combinations of complexes. The support is the set of indices that are non-zero 
            suppcnt = 0
            supp = 0
            for i in 1:size(Z,1)
                 (Z[i,c_s] != Z[i, c_p]) && (suppcnt += 1)
                 (suppcnt == 1) && (supp = i)
                 (suppcnt > 1) && continue
            end

            # If the support has length one, then they differ in only one species, and that species is concentration robust. 
            (suppcnt == 1) && (supp ∉ robust_species) && push!(robust_species, supp)
```

isaacsas: Basically if I've understood your code you don't do anything unless there is only one, so you can just stop the loop as soon as a second is found.

isaacsas: (break not continue...)

isaacsas (approved):

isaacsas:
@vyudu LGTM. Feel free to merge at your leisure.

PR #965, opened 2024-07-03
Detailed balance

vyudu:
Code and tests for testing whether a reaction networks with a given set of rate parameters is detailed balanced.


isaacsas:
This needs to be updated to master.

isaacsas:
@vyudu when you have a chance can you update to master? (No rush though as I’m traveling this week as we discussed.)

isaacsas:
bump

vyudu:
Sorry about the delay - will be active again in a week or so

vyudu:
This one should be ready for review @isaacsas

src/network_analysis.jl
```diff
@@ -724,6 +737,87 @@ function conservationlaw_errorcheck(rs, pre_varmap)
...
+    spanning_forest = Graphs.kruskal_mst(undir_img)
+    outofforest_edges = setdiff(collect(edges(undir_img)), spanning_forest)
+
+    # Independent Cycle Conditions: for any cycle we create by adding in an out-of-forest reaction, the product of forward reaction rates over the cycle must equal the product of reverse reaction rates over the cycle.  
+    for edge in outofforest_edges
+        g = SimpleGraph([spanning_forest..., edge])
+        ic = Graphs.cycle_basis(g)[1]
+        fwd = prod([K[ic[r], ic[r + 1]] for r in 1:(length(ic) - 1)]) * K[ic[end], ic[1]]
+        rev = prod([K[ic[r + 1], ic[r]] for r in 1:(length(ic) - 1)]) * K[ic[1], ic[end]]
+        fwd ≈ rev ? continue : return false
```
isaacsas, line 776:
```suggestion
        fwd ≈ rev ? continue : return false
```
Do you really need to calculate the products, or could you check that the sets of rates are the same?

Along those lines, if we really have to use `isapprox` then this should take a absolute and relative tolerance parameters that are passed through to it, and probably the default values should be pretty small (i.e. smaller than it assumes by default).

src/network_analysis.jl
```diff
@@ -724,6 +737,87 @@ function conservationlaw_errorcheck(rs, pre_varmap)
...
+    end
+
+    # Spanning Forest Conditions: for non-deficiency 0 networks, we get an additional δ equations. Choose an orientation for each reaction pair in the spanning forest (we will take the one given by default from kruskal_mst).  
+
+    if deficiency(rs) > 0
+        rxn_idxs = [edgeindex(D, Graphs.src(e), Graphs.dst(e)) for e in spanning_forest]
+        S_F = netstoichmat(rs)[:, rxn_idxs]
+        sols = nullspace(S_F)
```
isaacsas, line 784: Should this use ModelingToolkit's integer nullspace?

src/network_analysis.jl
```diff
@@ -724,6 +737,87 @@ function conservationlaw_errorcheck(rs, pre_varmap)
...
+
+        for i in 1:size(sols, 2)
+            α = sols[:, i]
+            fwd = prod([K[Graphs.src(e), Graphs.dst(e)]^α[i]
+                        for (e, i) in zip(spanning_forest, 1:length(α))])
+            rev = prod([K[Graphs.dst(e), Graphs.src(e)]^α[i]
+                        for (e, i) in zip(spanning_forest, 1:length(α))])
+            fwd ≈ rev ? continue : return false
```
isaacsas, line 792: Similar question to above about this calculation.

src/network_analysis.jl
vyudu: I think you do have to calculate the product, since the forward and reverse rates could be totally different. Sounds good though, will add the tolerance parameters to this function

isaacsas: OK, that makes sense. (I was thinking that they must be the same sets of rates you are multiplying across the two chains, but I guess that doesn't need to be the case in general.)

PR #973, opened 2024-07-07
Expand tests

TorkelE:
Adds tests for various features that were not previously tested properly. 

A few non-test related changes have comments below.


src/network_analysis.jl
```diff
@@ -721,7 +721,8 @@ end
     iscomplexbalanced(rs::ReactionSystem, parametermap)
 
 Constructively compute whether a network will have complex-balanced equilibrium
-solutions, following the method in van der Schaft et al., [2015](https://link.springer.com/article/10.1007/s10910-015-0498-2#Sec3). Accepts a dictionary, vector, or tuple of variable-to-value mappings, e.g. [k1 => 1.0, k2 => 2.0,...]. 
+solutions, following the method in van der Schaft et al., [2015](https://link.springer.com/article/10.1007/s10910-015-0498-2#Sec3). 
+Accepts a dictionary, vector, or tuple of variable-to-value mappings, e.g. [k1 => 1.0, k2 => 2.0,...].
```
TorkelE, line 725: @vyudu This and the next doc string were very long and a single line. I broke them up so that each line is no more than 100 characters (what I try to limit myself to). This should make them more readable when browsing the code. Is this OK with you?

src/network_analysis.jl
```diff
@@ -779,12 +780,12 @@ function iscomplexbalanced(rs::ReactionSystem, parametermap::Dict)
     end
 end
 
-function iscomplexbalanced(rs::ReactionSystem, parametermap::Vector{Pair{Symbol, Float64}})
+function iscomplexbalanced(rs::ReactionSystem, parametermap::Vector{<:Pair})
```
TorkelE, line 783: These functions (and similar for `ratematrix`) could not accept inputs of the form of Symbolics. I made the minimal change so that this (and the tests added below) now should work. Is this OK with you @vyudu? We might want to make them a bit more stringent so that not symbol/symbolic inputs do not work, in this case I tried to make changes minimal though (and on parity with how the `Dict` input was handled.

src/network_analysis.jl
vyudu: Yes! Will keep this in mind.

src/network_analysis.jl
vyudu: This makes sense also - thanks for adding these.

test/reactionsystem_core/reactionsystem.jl
```diff
@@ -325,7 +325,58 @@ let
     end
 end
 
-### Other Tests ###
+### Miscellaneous Getters ###
+
+# Tests spatial independent variables `get_sivs` and `has_sivs`.
+let
+    # Creates a `ReactionSystem`.
+    @variables x
```
isaacsas, line 333:
```suggestion
    @parameters x
```

isaacsas (changes requested):

PR #975, opened 2024-07-09
Cache incidence matrix

vyudu:
For functions that use the incidence matrix (isreversible, isweaklyreversible, deficiency, linkageclasses, etc.) have them construct the incidence matrix internally instead of having the user call reactioncomplexes first.


src/network_analysis.jl
```diff
@@ -511,30 +485,28 @@ end
...
 """
 function isweaklyreversible(rn::ReactionSystem, subnets)
-    im = get_networkproperties(rn).incidencemat
-    isempty(im) &&
-        error("Error, please call reactioncomplexes(rn::ReactionSystem) to ensure the incidence matrix has been cached.")
-    sparseig = issparse(im)
+    nps = get_networkproperties(rn)
+    isempty(nps.incidencemat) && reactioncomplexes(rn)
+    imat = nps.incidencemat
+    sparseig = issparse(imat)
```
TorkelE, line 502: Here I'd probably just do `sparseig = issparse(nps.incidencemat)` to save one line, but mostly a matter of taste.

TorkelE (approved):

isaacsas (changes requested):
Make sure to update the docstrings of any functions that now handle caching (which may mention users previously needed to call reactioncomplexes first...). For example, `incidencematgraph`.

PR #977, opened 2024-07-10
doc tweaks for v14

isaacsas:


TorkelE (approved):

docs/src/index.md
```diff
@@ -165,19 +165,19 @@ cell_model = @reaction_network begin
     kᵢ/V, Gᴾ --> G
```
TorkelE, line 165: I remember what you said about these and that subscript would be much clearer. Always used superscript for phosphorylation, but might not be grounded in the literature. I was meaning to at some point go and look how it is typically written and possibly change then (but didn't want to change back and fourth, nor wait on it, hence I merged with P still as superscript. Probably no big consideration either way, but I think we hope to get to the number of users where we should care about small stuff like this as well)

PR #979, opened 2024-07-11
Revert SDE change

TorkelE:
Currently, MTK requires a matrix to be `Matrix{Num}`, which is not good. However, and update which changes so that `Matrix{Any}` will work again seems like it is out soon. When it is we can merge this and things should be all good again.


isaacsas (approved):

isaacsas:
@TorkelE I think this is now fixed given that tests are passing. Feel free to merge when you are happy with it.

PR #984, opened 2024-07-12
Improve conservation law handling in `find_identifiable_functions`

TorkelE:


docs/src/inverse_problems/structural_identifiability.md
```diff
@@ -116,28 +116,17 @@ print_for_DAISY(si_ode)
...
-```
-E.g. if you run:
-```@example si3
-find_identifiable_functions(rs; measured_quantities = [:X1, :X2])
-```
-we see that `Γ[1]` (`= X1(0) + X2(0)`) is detected as an identifiable expression. If we want to disable this feature for any function, we can use the `remove_conserved = false` option:
-```@example si3
-find_identifiable_functions(rs; measured_quantities = [:X1, :X2], remove_conserved = false)
-```
+contain [conservation laws](@ref network_analysis_deficiency) (in this case $Γ = X1 + X2$, where $Γ = X1(0) + X2(0)$ is a constant). Because the presence of such conservation laws makes structural identifiability analysis prohibitively computationally expensive (for all but the simplest of cases), these are automatically eliminated by Catalyst. This is handled internally, and should not be noticeable to the user. The exception is the `make_si_ode` function. For each conservation law, its output will have one ODE removed, and instead have a conservation parameter (on the form `Γ[i]`) added to its equations. This feature can be disabled through the `remove_conserved = false` option.
```
isaacsas, line 127:
```suggestion
contain [conservation laws](@ref network_analysis_deficiency) (in this case $Γ = X1 + X2$, where $Γ = X1(0) + X2(0)$ is a constant). Because the presence of such conservation laws makes structural identifiability analysis prohibitively computationally expensive (for all but the simplest of cases), these are automatically eliminated by Catalyst. This is handled internally, and should not be noticeable to the user. The exception is the `make_si_ode` function. For each conservation law, its output will have one ODE removed, and instead have a conservation parameter (of the form `Γ[i]`) added to its equations. This feature can be disabled through the `remove_conserved = false` option.
```

isaacsas (approved):

PR #985, opened 2024-07-13
Refactor "srcd/sl.jl" file

TorkelE:
Also does the "src/expression_utils.jl" file, as that primarily support the dsl file (the other relevant file is the chemistry one).

Primarily four types of changes:
- Added comment to code which did not have (primarily older code).
- Rewrite some old code (a lot of stuff in the dsl file is stuff I wrote over 6 years ago, which simply can be rewritten better now).
- Do some organisations for where the options are handled in the main function. Been meaning to do this pretty much since I introduced them, and now when we have all *for the foreseeable future) it seemed like a good time to do it. Now it should also be easier to see where to add new options in the future.
- Rewrite so that the `@reaction_network` and `@network_component` macros reuse minimal amount of code (where as previously it the second was basically ctrl+copied versions of the first.

I also add some tests of all the various ways to create models via the DSL, mostly stuff that was not tested before.


src/Catalyst.jl
```diff
@@ -40,6 +40,7 @@ import ModelingToolkit: check_variables,
 
 import Base: (==), hash, size, getindex, setindex, isless, Sort.defalg, length, show
 import MacroTools, Graphs
+using MacroTools: striplines
```
TorkelE, line 43: this function is used frequenty, and even more frequently when debugging code. Figured it made sense to import directly.

src/chemistry_functionality.jl
```diff
@@ -86,7 +86,7 @@ function make_compound(expr)
     # Cannot extract directly using e.g. "getfield.(composition, :reactant)" because then 
     # we get something like :([:C, :O]), rather than :([C, O]).
     composition = Catalyst.recursive_find_reactants!(expr.args[3], 1,
-        Vector{ReactantStruct}(undef, 0))
+        Vector{ReactantInternal}(undef, 0))
```
TorkelE, line 89: `ReactantInternal` and `ReactionInternal` felt like way better name. Especially now with the `Reaction` struct, having a struct called `ReactionStruct` seemed like it could confuse new users. Since it is used internally only (during the creation of models in the DSL), I figured the name could reflect that.

src/expression_utils.jl
```diff
@@ -2,16 +2,19 @@
...
 # Function that handles variable interpolation.
 function esc_dollars!(ex)
-    if ex isa Expr
-        if ex.head == :$
-            return esc(:($(ex.args[1])))
-        else
-            for i in 1:length(ex.args)
-                ex.args[i] = esc_dollars!(ex.args[i])
-            end
+    # If we do not have an expression: recursion has finished and we return the input.
```
TorkelE, line 5: Add comments

src/expression_utils.jl
```diff
@@ -22,28 +25,26 @@ end
...
-        error("The following symbol(s) are used as species or parameters: " *
-              ((map(s -> "'" * string(s) * "', ",
-                  intersect(forbidden_symbols_error, v))...)) *
-              "this is not permitted.")
-    nothing
+function forbidden_symbol_check(sym)
+    if !isempty(intersect(forbidden_symbols_error, sym))
+        used_forbidden_syms = intersect(forbidden_symbols_error, sym)
+        error("The following symbol(s) are used as species or parameters: $used_forbidden_syms, this is not permitted.")
+    end
```
TorkelE, line 32: Just rewritten nicer than what I could do things before (e.g. before I knew about interpolating into strings). No need to have unnecessarily convoluted code.

src/dsl.jl
```diff
@@ -1,64 +1,4 @@
...
-    end
-
-    ### Simulating Reaction Networks ###
-    probODE = ODEProblem(rn, args...; kwargs...)        # Using multiple dispatch the reaction network can be used as input to create ODE, SDE and Jump problems.
-    probSDE = SDEProblem(rn, args...; kwargs...)
-    probJump = JumpProblem(prob,aggregator::Direct,rn)
-"""
-
-### Constant Declarations ###
+### Constants Declarations ###
```
TorkelE, line 1: Now starts of by declaring relevant constants and the species macro.

src/dsl.jl
```diff
@@ -110,158 +50,207 @@ end
 """
     @reaction_network
 
-Generates a [`ReactionSystem`](@ref dsl_description) that encodes a chemical reaction
-network.
+Macro for generating chemical reaction network models. Outputs a [`ReactionSystem`](@ref) structure,
```
TorkelE, line 53: The old description was pretty much the first documentation of the package, and over 6 years old. Given the important to the package, updating this one seemed long overdue.

src/dsl.jl
```diff
@@ -110,158 +50,207 @@ end
...
+- `ReactionSystem`s creates through `@reaction_network`, by default, have a random name. Specific
+names can be designated as a first argument (before `begin`, e.g. `rn = @reaction_network name begin ...`).
+- For more information, please again consider Catalyst's documentation.
 
-ReactionSystems generated through `@reaction_network` are complete.
 """
-macro reaction_network(name::Symbol, ex::Expr)
-    :(complete($(make_reaction_system(
-        MacroTools.striplines(ex); name = :($(QuoteNode(name)))))))
+macro reaction_network(name::Symbol, network_expr::Expr)
```
TorkelE, line 112: Basically, there are 4 different types of input (nothing, name, reactions, name + reactions). Each case is handled by a set of functions, which are now called from both `@eaction_network` and `@network_component`.

src/dsl.jl
```diff
@@ -110,158 +50,207 @@ end
...
-        new(sub, prod, rate, metadata)
+        new(subs, prods, rate, metadata)
     end
 end
 
 # Recursive function that loops through the reaction line and finds the reactants and their
-# stoichiometry. Recursion makes it able to handle weird cases like 2(X+Y+3(Z+XY)).
+# stoichiometry. Recursion makes it able to handle weird cases like 2(X + Y + 3(Z + XY)). The 
+# reactants are stored in the `reactants` vector. As the expression tree is parsed, the 
+# stoichiometry is updated and new reactants added.
```
TorkelE, line 211: Adds comments to the function, explaining what it does.

src/dsl.jl
```diff
@@ -286,112 +277,88 @@ end
 ### DSL Internal Master Function ###
 
 # Function for creating a ReactionSystem structure (used by the @reaction_network macro).
-function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
```
TorkelE, line 289: name symbol generation is now handled where the user (don't) provide the name as input.

src/dsl.jl
```diff
@@ -286,112 +277,88 @@ end
...
 
-    # Reads more options.
-    vars_extracted, add_default_diff, equations = read_equations_options(
-        options, variables_declared)
+    # Extracts all reactions. Extracts all parameters, species, and variables of the system and
+    # creates lists with them.
+    reactions = get_reactions(reaction_lines)
     variables = vcat(variables_declared, vars_extracted)
-
-    # handle independent variables
```
TorkelE, line 323: Part of the organisations so that options are handled together (in two sets, some before and some after the reactions are considered, depending on whichever is possible for which option)

src/dsl.jl
```diff
@@ -286,112 +277,88 @@ end
...
+    spsexpr, spsvar = scalarize_macro(spsexpr_init, "specs")
+    vsexpr, vsvar = scalarize_macro(vsexpr_init, "vars")
+    compound_expr, compsvar = scalarize_macro(compound_expr_init, "comps")
+    rxsexprs = make_rxsexprs(reactions, equations)
+
+    # Assemblies the full expression that declares all required symbolic variables, and
+    # then the output `ReactionSystem`.
     quote
-        $ps
+        $psexpr
```
TorkelE, line 350: just renamed so that everything which is an expression for the code that generates something ends with expr (and not just some of it)

src/dsl.jl
```diff
@@ -286,112 +277,88 @@ end
...
+        $vsexpr
+        $observed_expr
+        $compound_expr
         $diffexpr
 
         Catalyst.remake_ReactionSystem_internal(
             Catalyst.make_ReactionSystem_internal(
-                $rxexprs, $tiv, setdiff(union($spssym, $varssym, $compssym), $obs_syms),
-                $pssym; name = $name, spatial_ivs = $sivs, observed = $observed_eqs,
+                $rxsexprs, $tiv, setdiff(union($spsvar, $vsvar, $compsvar), $obs_syms),
```
TorkelE, line 360: Stuff which is the name of the variable that contains stuff in code ends with var, stuff which is just a list of symbols with syms.

src/dsl.jl
```diff
@@ -517,98 +488,146 @@ function extract_species_and_parameters!(reactions, excluded_syms)
...
 
 ### DSL Output Expression Builders ###
 
-# Given the species that were extracted from the reactions, and the options dictionary, creates the @species ... expression for the macro output.
-function get_sexpr(species_extracted, options, key = :species;
-        iv_symbols = (DEFAULT_IV_SYM,))
+# Given the extracted species (or variables) and the option dictionary, creates the 
+# `@species ...` (or `@variables ..`) expression which would declare these.
+# If `key = :variables`, does this for variables (and not species).
+function get_usexpr(us_extracted, options, key = :species; ivs = (DEFAULT_IV_SYM,))
```
TorkelE, line 515: make the names more general, as it at some point went from handling just species to species and variables (i.e. unknowns)

src/dsl.jl
```diff
@@ -517,98 +488,146 @@ function extract_species_and_parameters!(reactions, excluded_syms)
...
+    # symbolic variables (e.g. `var"##ps#384"`).
     namesym = gensym(name)
-    if nonempty
+
+    # If the input expression is non-emtpy, wraps it with addiional information.
+    if expr_init != :(())
         symvec = gensym()
-        ex = quote
-            $symvec = $ex
+        expr = quote
```
TorkelE, line 555: in some places we used ex and some expr, here I just made it uniform

test/dsl/dsl_basic_model_construction.jl
```diff
@@ -55,6 +55,65 @@ end
...
+    end
+
+    # Check that non-empty networks are correct.
+    isequivalent(rs_1, rs)
+    rs_2 == rs
+    rs_3 == rs
+    isequivalent(rs_empty_1, rs_empty)
+    rs_empty_2 == rs_empty
+    rs_empty_3 == rs_empty
+end
```
TorkelE, line 115: WHen I remade how the macor inputs are handled at the top level (nothing, name, reactions, name + reactions) I wanted to add some tests that all of this works for all cases and both macros.

test/dsl/dsl_advanced_model_construction.jl
```diff
@@ -212,8 +212,8 @@ end
 
 # Checks that repeated metadata throws errors.
 let 
-    @test_throws LoadError @eval @reaction k, 0 --> X, [md1=1.0, md1=2.0] 
-    @test_throws LoadError @eval @reaction_network begin
```
TorkelE, line 216: This was never supposed to be a `LoadError` in the first place. There was a variable in the `error` statement that was undeclared, so that yielded an `LoadError`, which was not the true error.

test/dsl/dsl_advanced_model_construction.jl
```diff
@@ -131,7 +131,7 @@ let
     end
     # Line number nodes aren't ignored so have to be manually removed
     Base.remove_linenums!(ex)
-    @test eval(Catalyst.make_reaction_system(ex)) isa ReactionSystem
+    @test eval(Catalyst.make_reaction_system(ex, name = QuoteNode(:name))) isa ReactionSystem
```
TorkelE, line 134: adapted to the new `make_reaction_system` function (which won't generate names for the user)

test/dsl/dsl_advanced_model_construction.jl
```diff
@@ -131,7 +131,7 @@ let
     end
     # Line number nodes aren't ignored so have to be manually removed
     Base.remove_linenums!(ex)
-    @test eval(Catalyst.make_reaction_system(ex)) isa ReactionSystem
+    @test eval(Catalyst.make_reaction_system(ex, QuoteNode(:name))) isa ReactionSystem
```
TorkelE, line 134: adapted to the new `make_reaction_system` function (which won't generate names for the user)

src/dsl.jl
```diff
@@ -453,33 +422,30 @@ function read_reaction_line(line::Expr)
...
-    # Loops through each reaction encoded by the reaction composites. Adds the reaction to the reactions vector.
-    for i in 1:maximum(lengs)
-        # If the `only_use_rate` metadata was not provided, this has to be infered from the arrow used.
+    # Loops through each reaction encoded by the reaction's different components. 
+    # Creates a `ReactionInternal` representation and adds it to `reactions`.
+    for i in 1:maxl
+        # If the `only_use_rate` metadata was not provided, this must be inferred from the arrow.
         metadata_i = get_tup_arg(metadata, i)
-        if all(arg -> arg.args[1] != :only_use_rate, metadata_i.args)
+        if all(arg.args[1] != :only_use_rate for arg in metadata_i.args)
```
isaacsas, line 442: Is there some reason you are switching from the functional style in many places to using a generator? I can understand doing so when using `map` to avoid allocations, but this shouldn't have any impact in that regard.

src/dsl.jl
```diff
@@ -453,33 +422,30 @@ function read_reaction_line(line::Expr)
...
-        # Checks that metadata fields are unique.
-        if !allunique(arg.args[1] for arg in metadata_i.args)
-            error("Some reaction metadata fields where repeated: $(metadata_entries)")
-        end
-
-        push!(reactions,
-            ReactionStruct(get_tup_arg(sub_line, i),
-                get_tup_arg(prod_line, i), get_tup_arg(rate, i), metadata_i))
+        # Extracts substrates, products, and rates for the i'th reaction.
+        subs_i, prods_i, rate_i = get_tup_arg.([subs, prods, rate], i)
```
isaacsas, line 447:
```suggestion
        subs_i, prods_i, rate_i = get_tup_arg.((subs, prods, rate), i)
```

src/dsl.jl
```diff
@@ -517,98 +486,146 @@ function extract_species_and_parameters!(reactions, excluded_syms)
...
+    foreach(eq -> push!(rxsexprs.args, eq), equations)
+    return rxsexprs
+end
+
+# From a `ReactionInternal` struct, creates the expression which evaluates to the creation
+# of the correponding reaction.
+function get_rxexpr(rx::ReactionInternal)
+    # Initiates the `Reaction` expression.
+    rate = recursive_expand_functions!(rx.rate)
+    rx_constructor = :(Reaction($rate, [], [], [], []; metadata = $(rx.metadata)))
```
isaacsas, line 578: Please revert back to the old initialization. The point of using `nothing` was to avoid allocations. Creating empty arrays of type `Any` like here will result in unneeded allocations I believe (and inside the `Reaction` it will immediately set several of them to `nothing` anyways).

src/dsl.jl
```diff
@@ -733,27 +750,40 @@ function create_differential_expr(options, add_default_diff, used_syms, tiv)
...
+            if !isempty(ivs)
                 error("An observable ($obs_name) was given independent variable(s). These should not be given, as they are inferred automatically.")
-            isnothing(defaults) ||
+            end
+            if !isnothing(defaults)
                 error("An observable ($obs_name) was given a default value. This is forbidden.")
-            (obs_name in forbidden_symbols_error) &&
+            end
+            if in(obs_name, forbidden_symbols_error)
                 error("A forbidden symbol ($(obs_eq.args[2])) was used as an observable name.")
```
isaacsas, line 783: See comments above about dropping `&&` and `||`.

src/dsl.jl
```diff
@@ -733,27 +750,40 @@ function create_differential_expr(options, add_default_diff, used_syms, tiv)
...
-
-            # Error checks.
+            end
             if (obs_name in species_n_vars_declared) && is_escaped_expr(obs_eq.args[2])
+                println("HERE")
```
isaacsas, line 786:
```suggestion
```

src/dsl.jl
```diff
@@ -768,19 +798,18 @@ function read_observed_options(options, species_n_vars_declared, ivs_sorted)
...
                     Symbolics.get_variables($(obs_eq.args[3]))))
-                ivs_get_expr = :(unique(reduce(
-                    vcat, [arguments(MT.unwrap(dep))
-                           for dep in $dep_var_expr])))
-                ivs_get_expr_sorted = :(sort($(ivs_get_expr);
-                    by = iv -> findfirst(MT.getname(iv) == ivs for ivs in $ivs_sorted)))
-                push!(observed_vars.args,
+                ivs_get_expr = :(unique(reduce(vcat,
+                    [arguments(MT.unwrap(dep)) for dep in $dep_var_expr])))
+                sort_func(iv) = findfirst(MT.getname(iv) == ivs for ivs in ivs_sorted)
```
isaacsas, line 810:
```suggestion
                sort_func(iv) = findfirst(==(MT.getname(iv)),  ivs_sorted)
```

src/dsl.jl
```diff
@@ -794,52 +823,68 @@ function read_observed_options(options, species_n_vars_declared, ivs_sorted)
...
 
 ### `@reaction` Macro & its Internals ###
 
 @doc raw"""
 @reaction
 
-Generates a single [`Reaction`](@ref) object.
+Generates a single [`Reaction`](@ref) object using a similar syntax as the `@reaction_network`
+macro (but permiting only a single reaction). A more detailed introduction to the syntax can
+be found in the description of `@reaction_network`.
```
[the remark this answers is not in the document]
isaacsas: More generally, I'd suggest following the approach I mentioned for `@reaction_network` aiming this as a reference on options for users who are looking things up, and referring users to the docs for tutorials / introductions. (Along with keeping some example usage.)

isaacsas (commented):
For future reference, please try not to reorder functions while also modifying them. It would be a lot easier to look at the diff if you had first rewritten them in this PR, and then in a followup PR reordered. 

(Obviously for a change of like 50 lines this can be ignored, but when there are hundreds of lines of diff showing it makes the review a lot longer when things get moved around too.)

TorkelE:
Updated per various changes that have happened since the initial PR. Also incorporated your suggestions from the previous one. Still need to go through it properly, will note when this is actually ready.

docs/src/model_creation/dsl_advanced.md
```diff
@@ -501,6 +500,34 @@ Catalyst.getdescription(rx)
...
+
+Generally, the `@reaction` macro provides a more concise notation to the [`Reaction`](@ref) constructor. One of its primary uses is for [creating models programmatically](@ref programmatic_CRN_construction). When doing so, it can often be useful to use [*interpolation*](@ref dsl_advanced_options_symbolics_and_DSL_interpolation) of symbolic variables declared previously:
+```@example dsl_advanced_reaction_macro
+t = default_t()
+@parameters k b
+@species A(t)
+ex = k*A^2 + t
+rx = @reaction b*$ex*$A, $A --> C
+```
+
```
TorkelE, line 530: Previously there was actually no section on this in the DSL. The updated `@reaction` docstring now references this one.

src/Catalyst.jl
```diff
@@ -72,11 +72,6 @@ const CONSERVED_CONSTANT_SYMBOL = :Γ
 const forbidden_symbols_skip = Set([:ℯ, :pi, :π, :t, :∅])
 const forbidden_symbols_error = union(Set([:im, :nothing, CONSERVED_CONSTANT_SYMBOL]),
     forbidden_symbols_skip)
-const forbidden_variables_error = let
-    fvars = copy(forbidden_symbols_error)
-    delete!(fvars, :t)
-    fvars
-end
```
TorkelE, line 79: Per discussion, this one is not actually used.

src/chemistry_functionality.jl
```diff
@@ -86,7 +86,7 @@ function make_compound(expr)
     # Cannot extract directly using e.g. "getfield.(composition, :reactant)" because then
     # we get something like :([:C, :O]), rather than :([C, O]).
     composition = Catalyst.recursive_find_reactants!(expr.args[3], 1,
-        Vector{ReactantStruct}(undef, 0))
+        Vector{DSLReactant}(undef, 0))
```
TorkelE, line 89: Better name so that it is not confused with other structures (thanks for the name suggestion)

src/expression_utils.jl
```diff
@@ -22,33 +25,21 @@ end
...
-    nothing
-end
-
-# Throws an error when a forbidden variable is used (a forbidden symbol that is not `:t`).
-function forbidden_variable_check(v)
-    !isempty(intersect(forbidden_variables_error, v)) &&
-        error("The following symbol(s) are used as variables: " *
-              ((map(s -> "'" * string(s) * "', ",
-                  intersect(forbidden_variables_error, v))...)) *
-              "this is not permitted.")
```
TorkelE, line 40: Again, the only time this function is issued it don't actually have any use. Old remnant and thus removed.

test/dsl/dsl_advanced_model_construction.jl
```diff
@@ -123,19 +122,6 @@ let
...
-    ex = quote
-        (Ka, Depot --> Central)
-        (CL / Vc, Central --> 0)
-    end
-    # Line number nodes aren't ignored so have to be manually removed
-    Base.remove_linenums!(ex)
-    exsys = Catalyst.make_reaction_system(ex)
-    sys = @eval Catalyst $exsys
-    @test sys isa ReactionSystem
-end
```
TorkelE, line 137: We can discuss this one, but what it actually does is kind of funny. It basically takes the internal function and evaluates it in the current scope somehow. However, when we have clarified the macro output to make it easier to read, e.g. putting
```julia
observed = $obs_eqs
        continuous_events = $continuous_events_expr
        discrete_events = $discrete_events_expr
```
in the macro and then use
```julia
make_ReactionSystem_internal(rx_eq_vec, $tiv, vars, $psvar; name, spatial_ivs,
                observed, continuous_events, discrete_events, combinatoric_ratelaws)
```
we get the trouble that `observed` overwrites `observed` which is an exported function. Now the test is kind of artificial, so Catalyst works perfectly even if this generates an error, and it makes e.g. `@macroexpand` a bit easier to use. The contrary argument for keeping this is that if one wants to copy and evaluate the code then one must modify. Right now I removed the test, but we can discuss.

test/dsl/dsl_options.jl
```diff
@@ -609,11 +616,11 @@ let
         @equations D(V) ~ 1 - V
         d, D --> 0
     end
-    @test_broken false # @test_throws Exception @eval @reaction_network begin
-        #@variables D(t)
-        #@equations D(V) ~ 1 - V
-        #d, X --> 0
-    #end
```
TorkelE, line 616: These ones are now fixed. Will close https://github.com/SciML/Catalyst.jl/issues/1173

test/dsl/dsl_options.jl
```diff
@@ -675,7 +682,7 @@ let
     @test plot(sol; idxs=:X).series_list[1].plotattributes[:y][end] ≈ 10.0
     @test plot(sol; idxs=[X, Y]).series_list[2].plotattributes[:y][end] ≈ 3.0
     @test plot(sol; idxs=[rn.X, rn.Y]).series_list[2].plotattributes[:y][end] ≈ 3.0
-    @test plot(sol; idxs=[:X, :Y]).series_list[2].plotattributes[:y][end] ≈ 3.0 # (https://github.com/SciML/ModelingToolkit.jl/issues/2778)
```
TorkelE, line 678: Think this link has probably remained since when this was broken long ago.

src/dsl.jl
```diff
@@ -110,159 +50,186 @@ end
...
+    return Expr(:block, :(@parameters t), rs_expr)
+end
+
+# When both a name and a network expression is generated, dispatches these to the internal
+# `make_reaction_system` function.
+function make_rs_expr(name, network_expr; complete = true)
+    rs_expr = make_reaction_system(striplines(network_expr), name)
+    complete && (rs_expr = :(complete($rs_expr)))
+    return rs_expr
 end
```
TorkelE, line 159: I tried to rewrite the `@reaction_network` and `@network_component` macros so that we don't provide repeat code in the two macros.

TorkelE: `make_rs_expr` now does all the work (with the `complete` argument being `true` or `false`).

src/dsl.jl
```diff
@@ -110,159 +50,186 @@ end
...
-            metadata_line::ExprValues, rx_line::Expr)
-        sub = recursive_find_reactants!(sub_line, 1, Vector{ReactantStruct}(undef, 0))
-        prod = recursive_find_reactants!(prod_line, 1, Vector{ReactantStruct}(undef, 0))
+    function DSLReaction(sub_line::ExprValues, prod_line::ExprValues,
+            rate::ExprValues, metadata_line::ExprValues, rx_line::Expr)
+        subs = recursive_find_reactants!(sub_line, 1, Vector{DSLReactant}(undef, 0))
+        prods = recursive_find_reactants!(prod_line, 1, Vector{DSLReactant}(undef, 0))
         metadata = extract_metadata(metadata_line)
-        new(sub, prod, rate, metadata, rx_line)
+        new(subs, prods, rate, metadata, rx_line)
```
TorkelE, line 183: subs and prods because there might be multiple ones,

src/dsl.jl
```diff
@@ -110,159 +50,186 @@ end
...
-            idx = findall(x -> x == ex, getfield.(reactants, :reactant))[1]
-            reactants[idx] = ReactantStruct(ex,
-                processmult(+, mult,
-                    reactants[idx].stoichiometry))
+
+        # If the expression corresponds to a reactant on our list, increase its multiplicity.
+        idx = findfirst(r.reactant == ex for r in reactants)
+        if !isnothing(idx)
+            new_mult = processmult(+, mult, reactants[idx].stoichiometry)
+            reactants[idx] = DSLReactant(ex, new_mult)
```
TorkelE, line 202: Instead of one long line that get split into two lines by the line break I just wrote it out into two lines
```julia
new_mult = processmult(+, mult, reactants[idx].stoichiometry)
reactants[idx] = DSLReactant(ex, new_mult)
```

src/dsl.jl
```diff
@@ -298,149 +266,113 @@ end
 ### DSL Internal Master Function ###
 
 # Function for creating a ReactionSystem structure (used by the @reaction_network macro).
-function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
+function make_reaction_system(ex::Expr, name)
 
-    # Handle interpolation of variables
+    # Handle interpolation of variables in the input.
     ex = esc_dollars!(ex)
```
TorkelE, line 273: This main function have now been reordered, trying to put the various option handling into separate sections depending on whether they deal wth declared symbols (must happen first), inferred symbols (happens after that) or other options (mostly must happen at the end). Previously stuff was a little bit all over the place depending on when it got added. Now it should be much clearer what is going on.

src/dsl.jl
```diff
@@ -298,149 +266,113 @@ end
...
+    vsexpr_init = get_usexpr(vs_inferred, options, :variables; ivs)
+    psexpr, psvar = scalarize_macro(psexpr_init, "ps")
+    spsexpr, spsvar = scalarize_macro(spsexpr_init, "specs")
+    vsexpr, vsvar = scalarize_macro(vsexpr_init, "vars")
+    cmpsexpr, cmpsvar = scalarize_macro(cmpexpr_init, "comps")
+    rxsexprs = get_rxexprs(reactions, equations, union(diffs_declared, diffs_inferred))
+
+    # Assemblies the full expression that declares all required symbolic variables, and
+    # then the output `ReactionSystem`.
+    MacroTools.flatten(striplines(quote
```
TorkelE, line 329: Since we already did some additional here to make the output easier to see with `@macroexpand`, I added `MacroTools.flatten(striplines(` as well. I typically always add these manually when doing debugging, and feel they make it way easier to see what is going on.

src/dsl.jl
```diff
@@ -298,149 +266,113 @@ end
...
+        # Stores each kwarg in a variable. Not necessary, but useful when debugging generated code.
+        name = $name
+        spatial_ivs = $sivs
+        rx_eq_vec = $rxsexprs
+        vars = setdiff(union($spsvar, $vsvar, $cmpsvar), $obs_syms)
+        observed = $obs_eqs
+        continuous_events = $continuous_events_expr
+        discrete_events = $discrete_events_expr
+        combinatoric_ratelaws = $combinatoric_ratelaws
+        default_reaction_metadata = $default_reaction_metadata
```
TorkelE, line 348: I added the remaining options here so that they now all can be looked at. Also renamed them to make the final statement less of a mess (although we might want to reert e.g. `observed`. Trivial ones like `combinatoric_ratelaws = $combinatoric_ratelaws` is technically not needed here.

src/dsl.jl
```diff
@@ -298,149 +266,113 @@ end
...
 
-        # Checks the type of arrow used, and creates the corresponding reaction(s). Returns them in an array.
+        # Checks which type of line is used, and calls `push_reactions!` on the processed line.
         if in(arrow, double_arrows)
-            if typeof(rate) != Expr || rate.head != :tuple
+            (typeof(rate) != Expr || rate.head != :tuple) &&
                 error("Error: Must provide a tuple of reaction rates when declaring a bi-directional reaction.")
-            end
+            (typeof(metadata) != Expr || metadata.head != :tuple) &&
+                error("Error: Must provide a tuple of reaction metadata when declaring a bi-directional reaction.")
```
TorkelE, line 374: This should have been here already, but seems it wasn't.

src/dsl.jl
```diff
@@ -516,155 +450,200 @@ end
...
+    else
+        ivs = (DEFAULT_IV_SYM,)
+        ivsexpr = :($(DEFAULT_IV_SYM) = default_t())
+    end
+
+    # Extracts the independent variables symbols (time and spatial), and returns the output.
+    tiv = ivs[1]
+    sivs = (length(ivs) > 1) ? Expr(:vect, ivs[2:end]...) : nothing
+    return tiv, sivs, ivs, ivsexpr
+end
```
TorkelE, line 615: Moved this bit out of the main function into a separate function.

src/dsl.jl
```diff
@@ -516,155 +450,208 @@ end
...
+    return rx_constructor
+end
+
+# Recursively escape functions within equations of an equation written using user-defined functions.
+# Does not expand special function calls like "hill(...)" and differential operators.
+function escape_equation!(eqexpr::Expr, diffsyms)
+    eqexpr.args[2] = recursive_escape_functions!(eqexpr.args[2], diffsyms)
+    eqexpr.args[3] = recursive_escape_functions!(eqexpr.args[3], diffsyms)
+    eqexpr
 end
```
TorkelE, line 602: Moved `escape_equation!` up here since it is just a helper function to the function above and not involved in other stuff. Calls `recursive_escape_functions!` on the lhs as well.

test/dsl/dsl_options.jl
```diff
@@ -1167,16 +1181,16 @@ end
 # Test whether user-defined functions are properly expanded in equations.
 let
     f(A, t) = 2*A*t
+    g(A) = 2*A + 2
 
-    # Test user-defined function
+    # Test user-defined function (on both lhs and rhs).
     rn = @reaction_network begin
-        @equations D(A) ~ f(A, t)
+        @equations D(A) + g(A) ~ f(A, t)
```
TorkelE, line 1188: updated because we now hanldes functions on both sides.

src/dsl.jl
```diff
@@ -717,31 +703,38 @@ function read_equations_options(options, syms_declared, parameters_extracted; re
...
+        (!isempty(vs_inferred) && requiredec) && throw(UndeclaredSymbolicError(
+            "Unrecognized symbolic variables $(join(vs_inferred, ", ")) detected in equation expression: \"$(string(eq))\". Since the flag @require_declaration is declared, all symbolic variables must be explicitly declared with the @species, @variables, and @parameters options."))
+    end
+
+    # If `D` differential is used, add it to differential expression and inferred differentials list.
+    diffs_inferred = Union{Symbol, Expr}[]
+    if add_default_diff && !any(diff_dec.args[1] == :D for diff_dec in diffsexpr.args)
+        diffs_inferred = [:D]
+        push!(diffsexpr.args, :(D = Differential($(tiv))))
     end
```
TorkelE, line 732: Now, if we use a default differential, `diffsexpr` is updated directly in `read_equations_options!`. Made more sense as I wanted to handle the `@differnetials` option at the beginning (since this explicitly declares symbols). Previously this option was handled after `@equations`.

src/dsl.jl
```diff
@@ -298,149 +266,113 @@ end
...
-        cevents = $continuous_events_expr
-        devents = $discrete_events_expr
+    psexpr_init = get_psexpr(ps_inferred, options)
+    spsexpr_init = get_usexpr(sps_inferred, options; ivs)
+    vsexpr_init = get_usexpr(vs_inferred, options, :variables; ivs)
+    psexpr, psvar = scalarize_macro(psexpr_init, "ps")
+    spsexpr, spsvar = scalarize_macro(spsexpr_init, "specs")
+    vsexpr, vsvar = scalarize_macro(vsexpr_init, "vars")
+    cmpsexpr, cmpsvar = scalarize_macro(cmpexpr_init, "comps")
+    rxsexprs = get_rxexprs(reactions, equations, union(diffs_declared, diffs_inferred))
```
TorkelE, line 325: trying to harmonise notation a bit across everything. exprs is the expression that generates the actually code, var is the generated variable we store stuff in. init on things that should not be evaluated. sps are species, vs variables, us unknowns, ps parameters, cmps compounds. Syms are lists of symbols (e.g. the symbols corresponding to observables). Not insisting we use this across the board, just explaining how I have changed things.

src/dsl.jl
```diff
@@ -516,155 +450,212 @@ end
...
 end
 
-# Function called by extract_species_and_parameters!, recursively loops through an
+# Function called by extract_sps_and_ps, recursively loops through an
 # expression and find symbols (adding them to the push_symbols vector).
-function add_syms_from_expr!(push_symbols::AbstractSet, rateex::ExprValues, excluded_syms)
-    if rateex isa Symbol
-        if !(rateex in forbidden_symbols_skip) && !(rateex in excluded_syms)
-            push!(push_symbols, rateex)
+function add_syms_from_expr!(push_symbols::AbstractSet, expr::ExprValues, excluded_syms)
```
TorkelE, line 496: rateex -> expr as this is not necessarily only the rate anymore.

src/dsl.jl
```diff
@@ -516,155 +450,212 @@ end
...
+        cmps_declared = [find_varinfo_in_declaration(arg.args[2])[1]
+                            for arg in cmpexpr_init.args[3].args]
     else  # If option is not used, return empty vectors and expressions.
-        compound_expr = :()
-        compound_species = Union{Symbol, Expr}[]
+        cmpexpr_init = :()
+        cmps_declared = Union{Symbol, Expr}[]
     end
-    return compound_expr, compound_species
+    return cmpexpr_init, cmps_declared
```
TorkelE, line 658: shortened compound to cmp in line with other shortenings, and to avoid having to break certain lines into two lines in some places.

src/expression_utils.jl
```diff
@@ -19,36 +22,9 @@ function is_escaped_expr(expr)
...
-
-# Throws an error when a forbidden symbol is used.
-function forbidden_symbol_check(v)
-    !isempty(intersect(forbidden_symbols_error, v)) &&
-        error("The following symbol(s) are used as species or parameters: " *
-              ((map(s -> "'" * string(s) * "', ",
-                  intersect(forbidden_symbols_error, v))...)) *
-              "this is not permitted.")
-    nothing
-end
```
TorkelE, line 32: This is not needed, because the check is already carried out in `Reaction` and `ReactionSystem`.

src/dsl.jl
```diff
@@ -914,68 +930,58 @@ end
...
     ex = esc_dollars!(ex)
 
-    # Parses reactions, species, and parameters.
+    # Parses reactions. Extracts species and paraemters within it.
     reaction = get_reaction(ex)
-    species, parameters = extract_species_and_parameters!([reaction], [])
+    species, parameters = extract_sps_and_ps([reaction], [])
 
-    # Checks for input errors.
-    forbidden_symbol_check(union(species, parameters))
```
TorkelE, line 925: This check is already carried out in `Reaction()`, so no longer needed in the DSL.

src/expression_utils.jl
```diff
@@ -22,33 +25,26 @@ end
...
-              ((map(s -> "'" * string(s) * "', ",
-                  intersect(forbidden_variables_error, v))...)) *
-              "this is not permitted.")
-end
-
-function unique_symbol_check(syms)
-    allunique(syms) ||
-        error("Reaction network independent variables, parameters, species, and variables must all have distinct names, but a duplicate has been detected. ")
-    nothing
-end
```
TorkelE, line 47: Now only used in one place, there implemented directly.

src/expression_utils.jl
```diff
@@ -22,33 +25,26 @@ end
...
-# Some options takes input on form that is either `@option ...` or `@option begin ... end`. 
+# Many option inputs can be on a form `@option input` or `@option begin ... end`. In both these
+# cases we want to retrieve the third argument in the option expression. Further more, we wish
+# to throw an error if there is more inputs (suggesting e.g. multiple inputs on a single line).
+# Note that there are only some options for which we wish to make this check.
+function get_block_option(expr)
+    (length(expr.args) > 3) &&
+        error("An option input ($expr) is misformatted. Potentially, it has multiple inputs on a single lines, and these should be split across multiple lines using a `begin ... end` block.")
+    return expr.args[3]
+end
```
TorkelE, line 45: Helper function for handling error checks across multiple options.

src/spatial_reaction_systems/spatial_reactions.jl
```diff
@@ -57,9 +57,10 @@ function make_transport_reaction(rateex, species)
     # Checks for input errors.
     forbidden_symbol_check(union([species], parameters))
 
+
     # Creates expressions corresponding to actual code from the internal DSL representation.
-    sexprs = get_sexpr([species], Dict{Symbol, Expr}())
-    pexprs = get_pexpr(parameters, Dict{Symbol, Expr}())
+    sexprs = get_usexpr([species], Dict{Symbol, Expr}())
+    pexprs = get_psexpr(parameters, Dict{Symbol, Expr}())
```
TorkelE, line 63: Uses new names in the DSL file

test/dsl/dsl_options.jl
```diff
@@ -415,36 +415,35 @@ let
...
 
-    @test_throws ArgumentError begin
-        rn = @reaction_network begin
-            @variables K
-            k, K*A --> B
-        end
+# Tests error when disallowed name is used for variable.
+let
+    @test_throws Exception @eval @reaction_network begin
+        @variables π(t)
```
TorkelE, line 423: new test

src/dsl.jl
```diff
@@ -1,64 +1,4 @@
...
-
-    r = @reaction_network MyReactionType begin
-        my_hill_repression(x, v_x, k_x, n_x), 0 --> x
-    end
-
-    ### Simulating Reaction Networks ###
-    probODE = ODEProblem(rn, args...; kwargs...)        # Using multiple dispatch the reaction network can be used as input to create ODE, SDE and Jump problems.
-    probSDE = SDEProblem(rn, args...; kwargs...)
-    probJump = JumpProblem(prob,aggregator::Direct,rn)
-"""
```
TorkelE, line 59: The first documentation of `@reactio_network`, but now not even a docstring and just a weird remnant. remove.

src/dsl.jl
```diff
@@ -1,64 +1,4 @@
...
-
-### Constant Declarations ###
```
TorkelE, line 61: Not removed, now just line 1 and git got confused.

src/dsl.jl
TorkelE: There are basically four cases of `@reaction_network` (or `@network_component`) inputs, we have one macro command handling each one separately.

src/dsl.jl
```diff
@@ -298,149 +266,114 @@ end
...
+    # Extracts the lines with reactions, the lines with options, and the options. Check for input errors.
     reaction_lines = Expr[x for x in ex.args if x.head == :tuple]
     option_lines = Expr[x for x in ex.args if x.head == :macrocall]
+    options = Dict(Symbol(String(arg.args[1])[2:end]) => arg for arg in option_lines)
+    allunique(arg.args[1] for arg in option_lines) ||
+        error("Some options where given multiple times.")
+    (sum(length, [reaction_lines, option_lines]) != length(ex.args)) &&
+        error("@reaction_network input contain $(length(ex.args) - sum(length.([reaction_lines,option_lines]))) malformed lines.")
+    any(!in(option_keys), keys(options)) &&
+        error("The following unsupported options were used: $(filter(opt_in->!in(opt_in,option_keys), keys(options)))")
```
TorkelE, line 283: Step one, read the macro input and check that it is correctly formatted.

src/dsl.jl
```diff
@@ -298,149 +266,114 @@ end
...
+    vs_declared = extract_syms(options, :variables)
+    tiv, sivs, ivs, ivsexpr = read_ivs_option(options)
+    cmpexpr_init, cmps_declared = read_compound_options(options)
+    diffsexpr, diffs_declared = read_differentials_option(options)
+    syms_declared = collect(Iterators.flatten((cmps_declared, sps_declared, ps_declared,
+        vs_declared, ivs, diffs_declared)))
+    if !allunique(syms_declared)
+        nonunique_syms = [s for s in syms_declared if count(x -> x == s, syms_declared) > 1]
+        error("The following symbols $(unique(nonunique_syms)) have explicitly been declared as multiple types of components (e.g. occur in at least two of the `@species`, `@parameters`, `@variables`, `@ivs`, `@compounds`, `@differentials`). This is not allowed.")
+    end
```
TorkelE, line 298: Step 2, read all options that explicitly declares symbols as stuff.

src/dsl.jl
```diff
@@ -298,149 +266,114 @@ end
...
-        error("The following unsupported options were used: $(filter(opt_in->!in(opt_in,option_keys), keys(options)))")
-    forbidden_symbol_check(union(species, parameters))
-    forbidden_variable_check(variables)
-    unique_symbol_check(union(species, parameters, variables, ivs))
+    sps_inferred, ps_pre_inferred = extract_sps_and_ps(reactions, syms_declared; requiredec)
+    vs_inferred, diffs_inferred, equations = read_equations_options!(diffsexpr, options,
+        union(syms_declared, sps_inferred), tiv; requiredec)
+    ps_inferred = setdiff(ps_pre_inferred, vs_inferred, diffs_inferred)
+    syms_inferred = union(sps_inferred, ps_inferred, vs_inferred, diffs_inferred)
+    all_syms = union(syms_declared, syms_inferred)
```
TorkelE, line 308: Step 3, read the reactions, as well as the options from which we infer symbols.

src/dsl.jl
```diff
@@ -298,149 +266,114 @@ end
...
+
+    # Read options not related to the declaration or inference of symbols.
+    obsexpr, obs_eqs, obs_syms = read_observed_options(options, ivs,
+        union(sps_declared, vs_declared), all_syms; requiredec)
+    continuous_events_expr = read_events_option(options, :continuous_events)
+    discrete_events_expr = read_events_option(options, :discrete_events)
+    default_reaction_metadata = read_default_noise_scaling_option(options)
+    combinatoric_ratelaws = read_combinatoric_ratelaws_option(options)
```
TorkelE, line 316: Step 4, read the other options.

src/dsl.jl
```diff
@@ -298,149 +266,114 @@ end
...
-        cevents = $continuous_events_expr
-        devents = $discrete_events_expr
+    psexpr_init = get_psexpr(ps_inferred, options)
+    spsexpr_init = get_usexpr(sps_inferred, options; ivs)
+    vsexpr_init = get_usexpr(vs_inferred, options, :variables; ivs)
+    psexpr, psvar = scalarize_macro(psexpr_init, "ps")
+    spsexpr, spsvar = scalarize_macro(spsexpr_init, "specs")
+    vsexpr, vsvar = scalarize_macro(vsexpr_init, "vars")
+    cmpsexpr, cmpsvar = scalarize_macro(cmpexpr_init, "comps")
+    rxsexprs = get_rxexprs(reactions, equations, all_syms)
```
TorkelE, line 326: Step 5, creates the various expressions to put in the macro.

src/dsl.jl
```diff
@@ -516,155 +451,216 @@ end
...
     for reaction in reactions
         for reactant in Iterators.flatten((reaction.substrates, reaction.products))
             add_syms_from_expr!(species, reactant.reactant, excluded_syms)
-            (!isempty(species) && requiredec) && throw(UndeclaredSymbolicError(
-                                                                               "Unrecognized variables $(join(species, ", ")) detected in reaction expression: \"$(string(reaction.rxexpr))\". Since the flag @require_declaration is declared, all species must be explicitly declared with the @species macro."))
         end
+        (!isempty(species) && requiredec) &&
+            throw(UndeclaredSymbolicError("Unrecognized reactant $(join(species, ", ")) detected in reaction expression: \"$(string(reaction.rxexpr))\". Since the flag @require_declaration is declared, all species must be explicitly declared with the @species option."))
     end
+    union!(excluded_syms, species)
```
TorkelE, line 477: `union!` replaces `foreach(s -> push!(excluded_syms, s), species)`

src/dsl.jl
```diff
@@ -516,155 +451,216 @@ end
...
 
-# Checks if the `@default_noise_scaling` option is used. If so, read its input and adds it as a
-# default metadata value to the `default_reaction_metadata` vector.
-function check_default_noise_scaling!(default_reaction_metadata, options)
-    if haskey(options, :default_noise_scaling)
-        if (length(options[:default_noise_scaling].args) != 3) # Because of how expressions are, 3 is used.
-            error("@default_noise_scaling should only have a single input, this appears not to be the case: \"$(options[:default_noise_scaling])\"")
-        end
-        push!(default_reaction_metadata.args,
-            :(:noise_scaling => $(options[:default_noise_scaling].args[3])))
```
TorkelE, line 651: `check_default_noise_scaling!` moved later to reflect the order with which these are called.

test/dsl/dsl_options.jl
```diff
@@ -1128,55 +1145,184 @@ let
...
+
+    # Test for erroneous inputs (to few, to many, wrong type).
+    @test_throws Exception @eval @reaction_network begin
+        @combinatoric_ratelaws
+        d, 3X --> 0
+    end
+    @test_throws Exception @eval @reaction_network begin
+        @combinatoric_ratelaws false false
+        d, 3X --> 0
+    end
```
TorkelE, line 1306: Checks that new error checks work (don't think the one above actually errored either, but it should)

docs/src/model_creation/dsl_advanced.md
```diff
@@ -575,6 +574,45 @@ nothing # hide
...
+## [Creating individual reaction using the `@reaction` macro](@id dsl_advanced_options_reaction_macro)
+Catalyst exports a macro `@reaction`, which can be used to generate a singular [`Reaction`](@ref) object of the same type which is stored within the [`ReactionSystem`](@ref) structure (which in turn can be generated by `@reaction_network`). Generally, `@reaction` follows [identical rules to those of `@reaction_network`](@ref dsl_description_reactions) for writing and interpreting reactions (however, bi-directional reactions are not permitted). E.g. here we create a simple dimerisation reaction:
+```@example dsl_advanced_reaction_macro
+using Catalyst # hide
+rx_dimerisation = @reaction kD, 2X --> X2
+```
+Here, `@reaction` is followed by a single line consisting of three parts:
+- A rate (at which the reaction occurs).
+- Any number of substrates (which are consumed by the reaction).
+- Any number of products (which are produced by the reaction).
```
TorkelE, line 586: This bit
```
Here, `@reaction` is followed by a single line consisting of three parts:
- A rate (at which the reaction occurs).
- Any number of substrates (which are consumed by the reaction).
- Any number of products (which are produced by the reaction).
```

is essentially a repeat of previous documentation. Probably unnecessary for most people and only relevant if we wish to keep this section self contained. happy to remove though.

test/dsl/dsl_advanced_model_construction.jl
```diff
@@ -269,6 +255,24 @@ let
...
+    # Malformed lhs
+    @test_throws Exception @eval @reaction_network begin
+        d, X --> 0, [misc,description=>"Metadata lhs should be a single symbol."]
+    end
+
+    # Malformed metadata separator.
+    @test_throws Exception @eval @reaction_network begin
+        d, X --> 0, [misc=>:misc; description="description"]
+    end
+end
```
TorkelE, line 274: One of several new sets of tests for stuff which we previously didn't actually test

test/dsl/dsl_basic_model_construction.jl
```diff
@@ -432,11 +491,74 @@ let
...
+
+# Erroneous `@reaction` usage.
+let
+    # Bi-directional reaction using the `@reaction` macro.
+    @test_throws Exception @eval @reaction (k1,k2), X1 <--> X2
 
-    @test_throws LoadError @eval @reaction im, 0 --> B
-    @test_throws LoadError @eval @reaction nothing, 0 --> B
-    @test_throws LoadError @eval @reaction k, 0 --> im
-    @test_throws LoadError @eval @reaction k, 0 --> nothing
```
TorkelE, line 439: forbidden symbol tests now tested more extensively in a test block two blocks down ("# Check that forbidden symbols correctly generates errors"

src/dsl.jl
```diff
@@ -298,149 +266,114 @@ end
...
-    forbidden_variable_check(variables)
-    unique_symbol_check(union(species, parameters, variables, ivs))
+    sps_inferred, ps_pre_inferred = extract_sps_and_ps(reactions, syms_declared; requiredec)
+    vs_inferred, diffs_inferred, equations = read_equations_option!(diffsexpr, options,
+        union(syms_declared, sps_inferred), tiv; requiredec)
+    ps_inferred = setdiff(ps_pre_inferred, vs_inferred, diffs_inferred)
+    syms_inferred = union(sps_inferred, ps_inferred, vs_inferred, diffs_inferred)
+    all_syms = union(syms_declared, syms_inferred)
+    obsexpr, obs_eqs, obs_syms = read_observed_option(options, ivs,
+        union(sps_declared, vs_declared), all_syms; requiredec)
```
TorkelE, line 310: Step 3: Read all reactions and equations. Finds all symbols that are inferred as something.

src/dsl.jl
```diff
@@ -516,155 +450,197 @@ end
...
+
+# Recursively escape functions within equations of an equation written using user-defined functions.
+# Does not escape special function calls like "hill(...)" and differential operators. Does
+# also not escape stuff corresponding to e.g. species or parameters (required for good error
+# for when e.g. a species is used as a differential, or for time delays in the future).
+function escape_equation!(eqexpr::Expr, all_syms)
+    eqexpr.args[2] = recursive_escape_functions!(eqexpr.args[2], all_syms)
+    eqexpr.args[3] = recursive_escape_functions!(eqexpr.args[3], all_syms)
+    eqexpr
 end
```
TorkelE, line 608: escape equations is called by get_rxexprs

src/dsl.jl
```diff
@@ -914,68 +931,61 @@ end
...
 
-    # Parses reactions, species, and parameters.
+    # Parses reactions. Extracts species and parameters within it.
     reaction = get_reaction(ex)
-    species, parameters = extract_species_and_parameters!([reaction], [])
+    species, parameters = extract_sps_and_ps([reaction], [])
 
-    # Checks for input errors.
+    # Checks for input errors. Needed here but not in `@reaction_network` as `ReactionSystem` perform this check but `Reaction` don't.
     forbidden_symbol_check(union(species, parameters))
```
TorkelE, line 942: For `@reaction_network` this check is not needed as `ReactionSystem` does a similar one. We might consider whether `Reaction` should also perform this type of check (and we remove it from here entirely.

E.g. this works:
```julia
@parameters Γ
@species X(t)
Reaction(Γ, [X], [])
```
while this errors:
```julia
@reaction Γ, X --> 0
```

TorkelE:
This should be ready now. However, Plots fails on precompilation, so have to wait until that is fixed.

Not high priority though, doesn't actually change anything (but should make further updates to the DSL easier).

Once this PR is merged, I will have a follow-up one which reorders the option handling functions to follow the order with which they are called.

src/dsl.jl
```diff
@@ -828,83 +805,123 @@ function read_observed_options(options, species_n_vars_declared, ivs_sorted; req
...
+end
+
+# Finds the time independent variable, and any potential spatial independent variables.
+# Returns these (individually and combined), as well as an expression for declaring them.
+function read_ivs_option(options)
+    # Creates the independent variables expressions (depends on whether the `ivs` option was used).
+    if haskey(options, :ivs)
+        ivs = Tuple(extract_syms(options, :ivs))
+        ivsexpr = copy(options[:ivs])
+        ivsexpr.args[1] = Symbol("@", "parameters")
```
isaacsas, line 861:
```suggestion
        ivsexpr.args[1] = Symbol("@", "parameters")
```
We should probably switch to `@independent_variables` as MTK is going to use that in the future (for now it just wraps `@parameters` but that will change).

isaacsas (dismissed):
Please check that non-scalarization is still working ok for vector parameters, and ensure there isn't a performance regression. See the examples in https://github.com/SciML/Catalyst.jl/issues/1051

src/dsl.jl
```diff
@@ -516,155 +450,197 @@ end
...
+        Union{Symbol, Expr}[]
     end
-    return syms
 end
 
 # Function looping through all reactions, to find undeclared symbols (species or
-# parameters), and assign them to the right category.
-function extract_species_and_parameters!(reactions, excluded_syms; requiredec = false)
+# parameters) and assign them to the right category.
+function extract_sps_and_ps(reactions, excluded_syms; requiredec = false)
```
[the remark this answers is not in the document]
TorkelE: This is actually not obvious. Technically, it shouldn't really, it calls `add_syms_from_expr!` which does this. 

However, you are right that it actually does a `union!(excluded_syms, species)` which updates the `excluded_syms` input (which it doesn't hurt to modify, but which is not really an intended effect of  `extract_sps_and_ps`.

We could change
```julia
union!(excluded_syms, species)
```
inside `` to 
```julia
excluded_syms = union(excluded_syms, species)
```
which should prevent `extract_sps_and_ps` from being mutating. But maybe that would have a (very slight) performance impact? Otherwise it would probably be better to reorder its input, `add the `!` and rename `syms_declared` here:
```julia
sps_inferred, ps_pre_inferred = extract_sps_and_ps(reactions, syms_declared; requiredec)
```
(since it is only a vector of declared symbols before it it put into that function)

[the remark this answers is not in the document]
TorkelE: I have quite a good set of benchmarks lined up for checking DSL (and `ReactionSystem` generally) performance. If the PR shows up fine, I'd be slightly partial to changing to `excluded_syms = union(excluded_syms, species)` (since that is the simplest change), and then do a general look at how we can speed up the core Catalyst parts with some profiling etc.

src/dsl.jl
```diff
@@ -481,22 +414,23 @@ function read_reaction_line(line::Expr)
...
-    for i in 1:maximum(lengs)
-        # If the `only_use_rate` metadata was not provided, this has to be infered from the arrow used.
+# Takes a reaction line and creates reaction(s) from it and pushes those to the reaction vector.
+# Used to create multiple reactions from bundled reactions (like `k, (X,Y) --> 0`).
+function push_reactions!(reactions::Vector{DSLReaction}, subs::ExprValues,
+        prods::ExprValues, rate::ExprValues, metadata::ExprValues, arrow::Symbol, line::Expr)
+    # The rates, substrates, products, and metadata may be in a tuple form (e.g. `k, (X,Y) --> 0`).
+    # This finds these tuples' lengths (or 1 for non-tuple forms). Inconsistent lengths yield error.
+    lengs = (tup_leng(subs), tup_leng(prods), tup_leng(rate), tup_leng(metadata))
+    maxl = maximum(lengs)
```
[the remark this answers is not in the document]
TorkelE: Agree that this is better

src/dsl.jl
```diff
@@ -516,155 +450,197 @@ end
...
-    elseif rateex isa Expr
+    elseif expr isa Expr
         # note, this (correctly) skips $(...) expressions
-        for i in 2:length(rateex.args)
-            add_syms_from_expr!(push_symbols, rateex.args[i], excluded_syms)
+        for i in 2:length(expr.args)
+            add_syms_from_expr!(push_symbols, expr.args[i], excluded_syms)
         end
     end
-    nothing
```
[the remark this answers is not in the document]
TorkelE: Great point, makes lots of sense! Didn't realise, you are right, `nothing` should definitely be here.

src/dsl.jl
```diff
@@ -110,159 +50,187 @@ end
...
+    return make_rs_expr(:(gensym(:ReactionSystem)), expr)
 end
 
-# Ideally, it would have been possible to combine the @reaction_network and @network_component macros.
-# However, this issue: https://github.com/JuliaLang/julia/issues/37691 causes problem with interpolations
-# if we make the @reaction_network macro call the @network_component macro.
+# Ideally, it would have been possible to combine the @reaction_network and @network_component
+# macros. However, this issue: https://github.com/JuliaLang/julia/issues/37691 causes problem
+# with interpolations if we make the @reaction_network macro call the @network_component macro.
+# Instead, these use the same input, but pass `complete = false` to `make_rs_expr`.
```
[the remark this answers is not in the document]
TorkelE: Sounds good

src/dsl.jl
```diff
@@ -110,159 +50,187 @@ end
...
-macro network_component(name::Symbol = gensym(:ReactionSystem))
-    return Expr(:block, :(@parameters t),
-        :(ReactionSystem(Reaction[], t, [], []; name = $(QuoteNode(name)))))
+### DSL Macros Helper Functions ###
+
+# For when no reaction network has been designated. Generates an empty network.
+function make_rs_expr(name; complete = true)
+    rs_expr = :(ReactionSystem(Reaction[], t, [], []; name = $name))
+    complete && (rs_expr = :(complete($rs_expr)))
+    return Expr(:block, :(@parameters t), rs_expr)
```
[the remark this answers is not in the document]
TorkelE: Agreed

src/dsl.jl
```diff
@@ -298,149 +266,114 @@ end
...
 
-    # Read lines with reactions and options.
+    # Extract the lines with reactions, the lines with options, and the options. Check for input errors.
     reaction_lines = Expr[x for x in ex.args if x.head == :tuple]
     option_lines = Expr[x for x in ex.args if x.head == :macrocall]
+    options = Dict(Symbol(String(arg.args[1])[2:end]) => arg for arg in option_lines)
+    allunique(arg.args[1] for arg in option_lines) ||
+        error("Some options where given multiple times.")
+    (sum(length, [reaction_lines, option_lines]) != length(ex.args)) &&
+        error("@reaction_network input contain $(length(ex.args) - sum(length.([reaction_lines,option_lines]))) malformed lines.")
```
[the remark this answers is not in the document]
TorkelE: Since the extra computation was in the error I figured it didn't really matter, and this would be better as it compressed the code one line. But if you prefer this version and `(sum(length, [reaction_lines, option_lines]) != length(ex.args))` cannot be done in a non-allocating manner I am happy to change.

src/dsl.jl
```diff
@@ -298,149 +266,114 @@ end
...
+    rxsexprs = get_rxexprs(reactions, equations, all_syms)
+
+    # Assemblies the full expression that declares all required symbolic variables, and
+    # then the output `ReactionSystem`.
+    MacroTools.flatten(striplines(quote
+        # Inserts the expressions which generate the `ReactionSystem` input.
+        $ivsexpr
+        $psexpr
+        $spsexpr
+        $vsexpr
```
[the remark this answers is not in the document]
TorkelE: Honestly forgotten why I flipped the order, but I agree that species with dpendeant/metadata variables seems more likely than the other way around, will flip it back.

For the serialization code I wrote a functionality which detects symbolic variables that depend on other symbolic variables and declares them in order, at some point we should add that to the DSL as well.
(especially a single pass, where we split symbolic variables with/without dependants would be very easy to implement and clear most cases)

src/dsl.jl
```diff
@@ -298,149 +266,114 @@ end
...
+    # Handle interpolation of variables in the input.
     ex = esc_dollars!(ex)
 
-    # Read lines with reactions and options.
+    # Extract the lines with reactions, the lines with options, and the options. Check for input errors.
     reaction_lines = Expr[x for x in ex.args if x.head == :tuple]
     option_lines = Expr[x for x in ex.args if x.head == :macrocall]
+    options = Dict(Symbol(String(arg.args[1])[2:end]) => arg for arg in option_lines)
+    allunique(arg.args[1] for arg in option_lines) ||
+        error("Some options where given multiple times.")
```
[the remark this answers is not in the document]
TorkelE: It doesn't really matter. I checked that it works fine. The advantage is that we can throw the error very slightly faster, but never felt that minuscule performance increases really mattered for error throwing, and then it was better to have the code slightly easier to read. We can still flip them though if you think it is important.

src/dsl.jl
TorkelE: Agreed, changing to `@independent_variables`

src/dsl.jl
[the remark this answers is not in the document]
isaacsas: If you don't want to add a `!` just document that it modifies `excluded_syms` by possibly enlarging it. That should be fine.

src/dsl.jl
TorkelE: agreed, I've added this in now

src/dsl.jl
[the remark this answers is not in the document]
isaacsas: It is just a lot more readable (to me) this way. I can immediately understand the test while the other approach required me to pause and think.

src/dsl.jl
```diff
@@ -516,155 +450,197 @@ end
...
-            ex = quote
-                $namesym = $ex
-            end
+
+    # If the input expression is non-empty, wrap it with additional information.
+    if expr_init != :(())
+        symvec = gensym()
+        expr = quote
+            $symvec = $expr_init
+            $namesym = reduce(vcat, Symbolics.scalarize($symvec))
```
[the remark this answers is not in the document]
isaacsas: Please double check the other changes from that PR to the DSL remain in effect (since they are performance related dropping them wouldn't get caught by any test I think).

src/dsl.jl
```diff
@@ -298,149 +266,114 @@ end
...
+        $diffsexpr
+
+        # Stores each kwarg in a variable. Not necessary, but useful when debugging generated code.
+        name = $name
+        spatial_ivs = $sivs
+        rx_eq_vec = $rxsexprs
+        us = setdiff(union($spsvar, $vsvar, $cmpsvar), $obs_syms)
+        observed = $obs_eqs
+        continuous_events = $continuous_events_expr
+        discrete_events = $discrete_events_expr
```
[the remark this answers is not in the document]
TorkelE: That makes sense as the reason. 

Just running stuff like
```julia
using Catalyst, ModelingToolkit
@reaction_network ...
```
is fine. Stuff like this as well:
```julia
using Catalyst, ModelingToolkit
import ModelingToolkit: discrete_events
discrete_events # discrete_events (generic function with 1 method)
rn = @reaction_network begin
    (p,d), 0 <--> X
end
rn = @reaction_network begin
    @discrete_events 1.0 => [X ~ X/2]
    (p,d), 0 <--> X
end
discrete_events # discrete_events (generic function with 1 method)
discrete_events(rn) # OK
```

 The one case I found is if one try to do something like this:
```julia
ex = quote
    (Ka, Depot --> Central)
    (CL / Vc, Central --> 0)
end
Base.remove_linenums!(ex)
name = QuoteNode(:rs)
exsys = Catalyst.make_reaction_system(ex, name)
sys = @eval Catalyst $exsys
```
essentially calling the internal `make_reaction_system` function and trying to execute the code it creates directly. But that feels really niche and weird.

So yeah, I haven't actually been able to find a way where declaring stuff like this causes troubles. However I am no expert in scopes, so maybe there is something. Happy to revert back to the old approach to be on the safe side if you prefer.

src/dsl.jl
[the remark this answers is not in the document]
TorkelE: Sounds good

src/dsl.jl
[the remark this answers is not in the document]
TorkelE: sounds good

src/dsl.jl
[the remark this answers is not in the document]
isaacsas: If the old approach keeps that example working, which I think maybe people at Pumas were using, let's keep the alternative names (or just add an `_` before the new names to make clear they are internal).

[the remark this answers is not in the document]
isaacsas: Or we need to figure out how to scope it so we can use the normal names.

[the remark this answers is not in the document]
TorkelE: If Pumas is using something like
```julia
ex = quote
    (Ka, Depot --> Central)
    (CL / Vc, Central --> 0)
end
Base.remove_linenums!(ex)
name = QuoteNode(:rs)
exsys = Catalyst.make_reaction_system(ex, name)
sys = @eval Catalyst $exsys
```
then we will have to keep it. I thought that `@eval` was a big no no in pretty much any code though?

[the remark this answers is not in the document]
TorkelE: I have changed to using e.g. `_discrete_events`. This also allows us to reintorudce that test I removed without putting it in a module or something.

src/dsl.jl
```diff
@@ -863,23 +906,23 @@ end
...
     pexprs = get_pexpr(parameters, Dict{Symbol, Expr}())
-    rxexpr = get_rxexprs(reaction)
+    rxexpr = get_rxexpr(reaction)
     iv = :($(DEFAULT_IV_SYM) = default_t())
 
-    # Returns the rephrased expression.
+    # Returns a repharsed expression which generates the `Reaction`.
     quote
         $pexprs
         $iv
```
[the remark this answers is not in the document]
isaacsas: This never seems to have been addressed?

isaacsas (changes requested):

isaacsas:
Ahh, nevermind that comment. For some reason Github showed me an old version of the code... Looks good to me.

isaacsas (approved):

PR #989, opened 2024-07-17
More t is now a parameter test

TorkelE:
Primarily in the test files. These were the last cases I could find, but there might possibly be more that I missed. Need to read through tomorrow as well to check it looks OK.

### DIFF
--- a/test/dsl/dsl_options.jl
+++ b/test/dsl/dsl_options.jl
@@ -436,7 +436,8 @@ let
     end
 
     @parameters k k2
-    @variables s x D(x) E(s) F(s,x)
+    @parameters s x
+    @variables D(x) E(s) F(s,x)
     @species A(s,x) B(s) C(x) C2(s,x)
     rx = Reaction(k*k2*D, [A, B], [C, C2], [E, 1], [F, 1])
     @named ivstest = ReactionSystem([rx], s; spatial_ivs = [x])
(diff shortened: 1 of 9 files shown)


test/dsl/dsl_options.jl
```diff
@@ -436,7 +436,8 @@ let
     end
 
     @parameters k k2
-    @variables s x D(x) E(s) F(s,x)
+    @parameters s x
```
TorkelE, line 439: Generally I did not merge the new independent-variable-paraemter declarations with old parameter declarations. It seems desirable to create a `@independent_variables` declaration or similar for these in the future, so better keep them separate for when that happens.

test/reactionsystem_core/coupled_equation_crn_systems.jl
```diff
@@ -972,7 +975,8 @@ end
 # Checks that various erroneous coupled system declarations yield errors.
 let
     @parameters p1 p2
-    @variables τ  U1(τ) V1(t)
```
TorkelE, line 975: This causes a a `@test_throws` to go from throwing an error to not throwing an error (hence causing test failure). The reason is not directly obvious, but should be possible to figure out since I've been able to look closer at it.

test/dsl/dsl_options.jl
isaacsas: Sounds like that macro will happen, so we can update for it once available. For now we just want to ensure we don't get test failures or such from the change (and even if they change it to just be a warning now it would be good to avoid warnings in the test logs).

test/reactionsystem_core/coupled_equation_crn_systems.jl
isaacsas: Interesting.

isaacsas (approved):
Feel free to merge when you are happy (and I guess make a 14.0.2 release to get the docs updated).

isaacsas:
Approved conditional on the test failure getting fixed that is...

Issue #990, opened 2024-07-17
Add missing model acces functions to API (and add links to it where functions are mentioned)


TorkelE:
Quite a few functions do not have their docstrings printed in the API, here is some I caught from the latest tutorial on adding model functions.
```equations
get_systems
parameters
get_species
get_ps`
get_unknowns
unknowns
get_observed
observed
get_rxs
```

isaacsas:
I've also been wondering if for consistency we should add `get_reactions` and give `get_rxs` a deprecation warning for our next breaking release. The former seems more consistent with `reactions(rn)`.

TorkelE:
That would make sense. In my tutorial I even mistakenly wrote `get_reactions` in one place, and later caught it and changed to `get_rxs`.

TorkelE:
Turns out that, expect of `get_rxs` and `get_species` (which we have docstrings for in Catalyst), MTK lacks docstrings for these (at least what I could tell). This is likely why they were left out in the first place as well.

PR #997, opened 2024-07-21
Api update

TorkelE:
Some updates to the API. Intended to fix https://github.com/SciML/Catalyst.jl/issues/990, but turns out most of those lack docstrings in MTK, which is probably the reason they were left out in the first place, so couldn't actually add those here.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -174,10 +174,6 @@ accessor functions.
 species
 Catalyst.get_species
 nonspecies
-ModelingToolkit.parameters
-ModelingToolkit.get_ps
-ModelingToolkit.unknowns
-ModelingToolkit.get_unknowns
 reactions
 nonreactions
 Catalyst.get_rxs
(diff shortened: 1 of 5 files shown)


TorkelE:
ready

docs/src/api.md
```diff
@@ -156,18 +172,27 @@ accessor functions.
...
+The following system property accessor functions are primarily relevant to reaction system [coupled 
+to differential and/or algebraic equations](@ref constraint_equations).
+```@docs
+ModelingToolkit.has_alg_equations
+ModelingToolkit.alg_equations
+ModelingToolkit.has_diff_equations
+ModelingToolkit.diff_equations
 ```
 
 ## Basic reaction properties
```
isaacsas, line 198: The ordering is a bit awkward now. I'd put the species properties before the reaction properties and metadata.

TorkelE: done

isaacsas (approved):

Issue #994, opened 2024-07-21
Fix long doc build time for single figure in intro to Catalyst


TorkelE:
This figure in the introduction to the docs:
![image](https://github.com/user-attachments/assets/fac5c0a1-3efb-4067-baee-8dca7e938022)
still causes really long doc build times. 

Changing the code block to
```julia
plot(sol)
plot(sol, density = 10000, fmt = :png) # hide
nothing # hide
```
reduced doc build time from 700-800 seconds to around 150 (after the first run which compiles stuff). Just reducing the `density` did not necessarily seem to have that effect though.

Does not need to be fixed soon, and I am not sure what the best approach is, but would be nice to figure out at some point since it makes writing documentation a lot more annoying.

isaacsas:
This seems like a documenter bug — it seems to be ignoring the request in the doc to plot this as a png and is still plotting it as a svg.

isaacsas:
We shouldn’t reduce the density — it leads to a very misleading plot that does not correctly illustrate the noise in the model.

TorkelE:
I checked the plots with/without density and looked fine, but if you saw a difference then Ok.

TorkelE:
The Documenter people suggest that the solution my be to use DisplayAs.jl, and that might work better than simply designating `:png` in the `plot` command. https://tkf.github.io/DisplayAs.jl/dev/#DisplayAs.PNG

isaacsas:
Closed https://github.com/SciML/Catalyst.jl/pull/1006

PR #1006, opened 2024-07-25
Long jump plotting fix

isaacsas:
I just copied over the minimal set of stuff from DisplayAs that we need so we don't have to rely on it. The code is pretty simple so I think should be fine to keep in Catalyst (and DisplayAs is MIT so compatible).

Note that using PNG's means we don't need all the `plotdensity` stuff any more or to use `saveat`, but we do need to up the default `dpi` as otherwise the plots are a bit blurry when converted.


isaacsas:
Closes https://github.com/SciML/Catalyst.jl/issues/994

TorkelE:
That's great, thanks a lot for fixing <3

PR #1008, opened 2024-07-27
Additional reactionsystem conversion file comments

TorkelE:
Adds some additional comments to the `reactionsystem_conversion.jl` file. I tried to go through the file and add comments on what each set of code was doing, to make it faster to read for those unfamiliar (which before I did this did include me to some extent). I think I managed to understand everything in the end, but you should probably still go through carefully (no hurry at all though, this is very low prior).

### DIFF
--- a/src/reactionsystem_conversions.jl
+++ b/src/reactionsystem_conversions.jl
@@ -1,7 +1,7 @@
 ### ODE & SDE Assembly ###
 
 """
-    oderatelaw(rx; combinatoric_ratelaw=true)
+ oderatelaw(rx; combinatoric_ratelaw = true)
 
 Given a [`Reaction`](@ref), return the symbolic reaction rate law used in
 generated ODEs for the reaction. Note, for a reaction defined by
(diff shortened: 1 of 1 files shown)


src/reactionsystem_conversions.jl
```diff
@@ -1,7 +1,7 @@
 ### ODE & SDE Assembly ###
 
 """
-    oderatelaw(rx; combinatoric_ratelaw=true)
+    oderatelaw(rx; combinatoric_ratelaw = true)
```
TorkelE, line 4: My impression is that we now have changed to having space in between `=` and whatever is on each side. If this is not the case, feel free to correct me though.

TorkelE: (there is a couple of changes like this throughout the file)

src/reactionsystem_conversions.jl
```diff
@@ -89,41 +111,52 @@ function assemble_oderhs(rs, ispcs; combinatoric_ratelaws = true, remove_conserv
...
+# the ODEs generated by the reaction rate equation. The `as_odes` argument determine whether the LHS should
+# be differential (when generating e.g. a `ODESystem`) or `0` (when generating e.g. a NonlinearSystem).
+function assemble_drift(rs, ind_sps; combinatoric_ratelaws = true, as_odes = true,
+                        include_zero_odes = true, remove_conserved = false)
+    rhsvec = assemble_oderhs(rs, ind_sps; combinatoric_ratelaws, remove_conserved)
     if as_odes
         D = Differential(get_iv(rs))
         eqs = [Equation(D(x), rhs)
-               for (x, rhs) in zip(ispcs, rhsvec) if (include_zero_odes || (!_iszero(rhs)))]
+               for (x, rhs) in zip(ind_sps, rhsvec) if (include_zero_odes || (!_iszero(rhs)))]
```
TorkelE, line 123: Thought `ind_sps` was when I went through, but feel free to tell me to revert if you do not agree.

src/reactionsystem_conversions.jl
```diff
@@ -89,41 +111,52 @@ function assemble_oderhs(rs, ispcs; combinatoric_ratelaws = true, remove_conserv
...
-# this doesn't work with constraint equations currently
-function assemble_diffusion(rs, sts, ispcs; combinatoric_ratelaws = true,
-        remove_conserved = false)
+# this doesn't work with coupled equations currently
+# assemblies the diffusion equations terms used in the chemical Langevin equation SDE
+function assemble_diffusion(rs, us, ind_sps; combinatoric_ratelaws = true,
+                            remove_conserved = false)
     # as BC species should ultimately get an equation, we include them in the noise matrix
-    num_bcsts = count(isbc, get_unknowns(rs))
+    num_bcus = count(isbc, get_unknowns(rs))
```
TorkelE, line 135: generally I changes `sts` and similar to `us` (to follow the convention change states to unknowns)

src/reactionsystem_conversions.jl
```diff
@@ -43,22 +48,33 @@ function oderatelaw(rx; combinatoric_ratelaw = true)
 end
 
 # Function returning `true` for species which shouldn't change from the reactions, 
-# including non-species variables.
-drop_dynamics(s) = isconstant(s) || isbc(s) || (!isspecies(s))
-
-function assemble_oderhs(rs, ispcs; combinatoric_ratelaws = true, remove_conserved = false)
+# including non-species variables
+drop_dynamics(spec) = isconstant(spec) || isbc(spec) || (!isspecies(spec))
```
TorkelE, line 52: I tried to reduce the number of expressions used to describe a species. Here `spec` was used in all places where `drop_dynamics` was called.

src/reactionsystem_conversions.jl
```diff
@@ -43,22 +48,33 @@ function oderatelaw(rx; combinatoric_ratelaw = true)
...
+
+# for a `ReactionSystem`, create a vector with all the right-hand side expressions in the
+# corresponding ODE system (used by the `assemble_drift` function)
+function assemble_oderhs(rs, isps; combinatoric_ratelaws = true, remove_conserved = false)
```
TorkelE, line 56: Again, `ispcs` to `isps` (previously we used s, sp, spc, and spec to denote species, now we use sp or spec, which is a little bit less versions at least) if you want I can change these back

TorkelE:
Had this around for a while, and was meaning to fix merge into master and stuff now, but didn't finish. Will close for now and re upload at some other point when actually finished.

PR #1010, opened 2024-08-02
test: remove unnecessarily specific parameter type annotations

AayushSabharwal:
With https://github.com/SciML/ModelingToolkit.jl/pull/2908, tunables are being merged into one `Vector{T}`, where `T` is some floating point number. This is because tunables are intended for optimization and sensitivity analysis, but AD doesn't work for integers. Not splitting into `Tuple{Vector{T1}, Vector{T2}}` also has performance benefits for sensitivity analysis, and helps reduce a significant number of bugs. Array tunables are also concatenated into this vector. Accessing these functions is done through a view to prevent unnecessary performance overhead, even in MTK's generated functions.

To this end, parameters are promoted to a common type. Overly specific type annotations are also in general invalid, since the parameters could also be duals.

This PR is a draft so it can be rerun once to aforementioned MTK PR is merged and tagged.

## Checklist

### DIFF
--- a/test/reactionsystem_core/parameter_type_designation.jl
+++ b/test/reactionsystem_core/parameter_type_designation.jl
@@ -17,8 +17,8 @@ t = default_t()
 # Declares a simple model to run tests on.
 begin
     t = default_t()
-    @parameters p1 p2::Float64 p3::Int64 p4::Float32 p5::Rational{Int64}
-    @parameters d1 d2::Float64 = 1.2 d3::Int64 = 2 [description = "A parameter"] d4::Rational{Int64} d5::Float32
+    @parameters p1 p2 p3::Int64 p4 p5::Rational{Int64}
+    @parameters d1 d2 = 1.2 d3::Int64 = 2 [description = "A parameter"] d4::Rational{Int64} d5
     @species X1(t) X2(t) X3(t) X4(t) X5(t)
 
     rxs = [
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
>  Not splitting into Tuple{Vector{T1}, Vector{T2}} also has performance benefits for sensitivity analysis, and helps reduce a significant number of bugs. Array tunables are also concatenated into this vector. Accessing these functions is done through a view to prevent unnecessary performance overhead, even in MTK's generated functions.

And not to mention, no AD or optimization library supports having the tuning vector be different types, and I don't see that changing any time soon, and I don't see a case where that's a real optimization.

isaacsas:
Thanks for helping with any migration this will entail. Would this then mean we can’t have integer parameters that stay integer or is that not an issue as they aren’t tunable?

Also, isn’t this a breaking change at some level? I’d imagine type specific user registered functions may stop working if parameters are being silently promoted now. Will this be held for a MTK 10 release?

isaacsas:
Note there are likely doc updates needed here too as I believe @TorkelE wrote about typing parameters for the V14 docs too.

ChrisRackauckas:
It's not breaking if integer-valued parameters are not marked as tunable. This is just changing the internal representation of tuanbles. @AayushSabharwal confirm that `@parameters p=1` has that default to non-tunable? Only AbstractFloats should default to tuanble.

AayushSabharwal:
> Thanks for helping with any migration this will entail. Would this then mean we can’t have integer parameters that stay integer or is that not an issue as they aren’t tunable?

They will still be integers, they just won't be in the tunables portion. `@parameters p::Integer = 1` will still be stored as an integer.

> @parameters p=1 has that default to non-tunable? Only AbstractFloats should default to tuanble.

Only `AbstractFloat`s would default to tunable yes, but `@parameters p = 1` is by default a `BasicSymbolic{Real}` which we treat as a float. To get an integer, you need the type annotation.

ChrisRackauckas:
That would be breaking then, because we previously respected it as an integer there. I really think we should set tunable more carefully by default. I don't think we can go forward with this without it, since otherwise you very easily get rid of people's integers and bools.

AayushSabharwal:
It's not breaking. `@parameters p = 1` was always a float, since v9

AayushSabharwal:
Parameters that users have explicitly marked as integers/bools will still be integers/bools, just in a different partition.

ChrisRackauckas:
Oh, I thought we had a conversion. Okay, if that was already respecting the symtype then we're good.

isaacsas:
We already assume that if a parameter isn’t explicitly annotated as an integer it gets converted so that case shouldn’t be an issue.

TorkelE:
Did these additionally specific typings cause error in tests with new MTK? The tests here was not really intended to be optimised code for AD, but rather to test various really niche cases and that these works, to potentially catch niche errors.

PR #1021, opened 2024-08-14
`arguments` -> `sorted_arguments` updates

TorkelE:
fixes some remaining cases where `arguments` should be `sorted_arguments` with the new Symbolics update

### DIFF
--- a/src/Catalyst.jl
+++ b/src/Catalyst.jl
@@ -23,7 +23,7 @@ using RuntimeGeneratedFunctions
 RuntimeGeneratedFunctions.init(@__MODULE__)
 
 import Symbolics: BasicSymbolic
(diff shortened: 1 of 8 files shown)


src/Catalyst.jl
```diff
@@ -23,7 +23,7 @@ using RuntimeGeneratedFunctions
 RuntimeGeneratedFunctions.init(@__MODULE__)
 
 import Symbolics: BasicSymbolic
-using Symbolics: iscall
```
TorkelE, line 26: I figured we used `sorted_arguments` enough that it might make sense to import the function (happy to revert, not sure if we have any guidelines when we want to import/not import a function)

isaacsas (approved):
Thanks!

PR #1024, opened 2024-08-14
cleanup DSL generated expression for readability

isaacsas:
This introduces a bunch of intermediate variables to store inputs to `make_ReactionSystem_internal`, with the goal of making the generated expression easier to read when debugging. It also makes the creation of `t` the first thing in the generated code, which should be a first step to allowing time-dependent parameters (something MTK is working on better supporting).

### DIFF
--- a/test/dsl/dsl_advanced_model_construction.jl
+++ b/test/dsl/dsl_advanced_model_construction.jl
@@ -132,7 +132,8 @@ let
     # Line number nodes aren't ignored so have to be manually removed
     Base.remove_linenums!(ex)
     exsys = Catalyst.make_reaction_system(ex)
-    @test (@eval Catalyst $exsys) isa ReactionSystem
+    sys = @eval Catalyst $exsys
+    @test sys isa ReactionSystem
 end
 
 # Miscellaneous interpolation tests. Unsure what they do here (not related to DSL).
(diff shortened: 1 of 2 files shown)


test/dsl/dsl_advanced_model_construction.jl
```diff
@@ -131,7 +131,9 @@ let
     end
     # Line number nodes aren't ignored so have to be manually removed
     Base.remove_linenums!(ex)
-    @test eval(Catalyst.make_reaction_system(ex)) isa ReactionSystem
+    exsys = Catalyst.make_reaction_system(ex)
+    sys = @eval Catalyst $exsys
```
isaacsas, line 135: This evaluates the expression within the Catalyst module's scope.

TorkelE:
Looks good.

Ideally I'd prefer to not do any non-required changes to the DSL  file until after https://github.com/SciML/Catalyst.jl/pull/985 is merged (I've already had to do one messy merge thing to get that one back functioning with master). The removing `Catalyst.` things might work with a simple merge, but not sure about the other changes. If you thing there won't be any problems then feel free to merge, but otherwise I wouldn't might waiting until https://github.com/SciML/Catalyst.jl/pull/985 (with neither of these being very urgent)

Issue #1022, opened 2024-08-14
DSL issues with equations


isaacsas:
1. Using an external function in a reaction works fine but has issues in equations
```julia
# works
f(A,t) = 2*A*t
rn = @reaction_network begin
    f(A,t), A --> 0
end 

# does not work
rn = @reaction_network begin
    @equations D(A) ~ f(A,t)
end
```
with the latter giving the error
```
ERROR: UndefVarError: `f` not defined
Stacktrace:
 [1] top-level scope
   @ ~/.julia/dev/Catalyst/src/dsl.jl:391
```

2. This doesn't work 
```julia
rn = @reaction_network begin
    @equations begin
        D(A) ~ Iapp
        Iapp ~ f(A,t)
    end
end
```
with the error
```
ERROR: UndefVarError: `Iapp` not defined
Stacktrace:
 [1] top-level scope
   @ ~/.julia/dev/Catalyst/src/dsl.jl:391
```

3. This also doesn't work
```julia
rn = @reaction_network begin
    @variables Iapp(t)
    @equations begin
        D(A) ~ Iapp
        Iapp ~ f(A,t)
    end
end
```
with a similar error to case 1.

isaacsas:
Expanding the macro we see that function is becoming `Catalyst.f` when used in an equation vs. when used in a reaction, i.e.
```julia
f(A,t) = 2*A*t
@macroexpand @reaction_network begin
    f(A,t), A --> 0
end 
```
gives
```julia
julia> @macroexpand @reaction_network begin
           f(A,t), A --> 0
       end
:(Catalyst.complete(begin
          #= /Users/isaacsas/.julia/dev/Catalyst/src/dsl.jl:383 =#
          var"#402###ps#349" = Catalyst.Num[]
          #= /Users/isaacsas/.julia/dev/Catalyst/src/dsl.jl:384 =#
          var"#403#t" = Catalyst.default_t()
          #= /Users/isaacsas/.julia/dev/Catalyst/src/dsl.jl:385 =#
          var"#404###vars#350" = Catalyst.Num[]
          #= /Users/isaacsas/.julia/dev/Catalyst/src/dsl.jl:386 =#
          begin
              #= /Users/isaacsas/.julia/dev/Catalyst/src/dsl.jl:591 =#
              var"#405###352" = begin
[...]

PR #1041, opened 2024-09-03
expanding equations when passed to `@reaction_network`

vyudu:
Closes [#1022. ](https://github.com/SciML/Catalyst.jl/issues/1022)


isaacsas:
Can you add some tests in the `dsl_options.jl` file?

src/dsl.jl
```diff
@@ -376,7 +376,8 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
         push!(rxexprs.args, get_rxexprs(reaction))
     end
     for equation in equations
-        push!(rxexprs.args, equation)
+        equation = expand_equation_RHS!(equation)
+        push!(rxexprs.args, (equation))
```
isaacsas, line 380:
```suggestion
        push!(rxexprs.args, equation)
```

TorkelE:
It seems like this extracts parameters from DSL expressions (checking on my phone, so have not gone through properly)? Previously we decided quite strongly against doing any such extractions, and generally keep it to a minimum with the exception of fir reactions (we could change this of course, but it would be a rather dramatic shift)

vyudu:
Ah, I can revert that. I wanted to make 
```julia
rn = @reaction_network
    @equations D(A) ~ hill(A, v, K, n)
end
```
work without throwing an error that v is undefined, but I see that it could be complicated to allow this inference generally. What was the main reason for not letting this, out of curiosity?

TorkelE:
The reason has to do that it can be ambiguous what is a species, a variable, and a parameter, from just looking at the expression. And if there is some misunderstanding, there could be nasty errors.

Given that equations in ReactionSystems is relatively niche, we figured it would be better to have the user manually declare in these situations.

test/dsl/dsl_options.jl
```diff
@@ -152,6 +153,56 @@ let
...
+    g(A, K, n) = A^n + K^n
+    rn4 = @reaction_network begin
+        @equations D(A) ~ hill(A, v, K, n)*g(A, K, n)
+    end
+    @test length(equations(rn4)) == 1
+    @test equations(rn4)[1] isa Equation
+    @parameters v n
+    @test isequal(Catalyst.expand_registered_functions(equations(rn4)[1]), D(A) ~ v*(A^n))
+end
+
```
isaacsas, line 205: Can we also add similar tests using a registered function for `f(A,t)`? (Just to make sure that case works too.)

isaacsas: Ahh right, forgot about that! But maybe make one you explicitly create in case we change that in the future? (I've been debating whether we should stop registering those functions and just let MTK trace -- registration is currently breaking unit checks in them.)

vyudu: As in explicitly define the `hill` function inside the test block?

isaacsas: Sure, but don't call it `hill`.

vyudu:
Hmm, escaping both sides of the equation seems to let you write more complicated LHS differential expressions a la #1044. This should be fine right? If we want users to be able to define custom `D`'s they should be resolved in their call scope anyway right?

test/dsl/dsl_options.jl
```diff
@@ -152,6 +153,68 @@ let
...
+    # Test Catalyst function
+    rn2r = @reaction_network begin
+        @parameters v K n
+        @equations D(A) ~ hill2(A, v, K, n)
+    end
+    @test length(equations(rn2r)) == 1
+    @test equations(rn2r)[1] isa Equation
+    @parameters v K n
+    @test isequal(equations(rn2r)[1], D(A) ~ hill2(A, v, K, n))
+
```
vyudu, line 190: Is this what you had in mind for the registered/unregistered function test @isaacsas?

src/dsl.jl
```diff
@@ -688,25 +692,41 @@ function read_equations_options(options, variables_declared)
             error("Malformed equation: \"$eq\". Equation's left hand and right hand sides should be separated by a \"~\".")
         end
 
-        # Checks if the equation have the format D(X) ~ ... (where X is a symbol). This means that the
-        # default differential has been used. X is added as a declared variable to the system, and
-        # we make a note that a differential D = Differential(iv) should be made as well.
```
TorkelE, line 693: seems some stuff here got removed

TorkelE:
This does a lot of stuff with how equations in the DSL are handled which I don't think we should touch here (e.g. permits f(d(X)) LHS). We did have some reasons to be restrictive with how we infer variables, but I really don't mind expanding this either. However, I think we should let this just handle the function errors, and then we can discuss separately which additional cases we wish to support and not, and handle those in a separate PR.

test/dsl/dsl_options.jl
isaacsas: This is what I had in mind for the two ways of specifying external functions, yes. 

I'd suggest making two versions of the system with the same name (one constructed via the symbolic interface), then you can test if they are equal. That would establish that the DSL interface is generating the same system as the symbolic interface.

src/dsl.jl
```diff
@@ -688,25 +692,41 @@ function read_equations_options(options, variables_declared)
...
+        push!(diff_vars, lhs.args[2])
+    end
+
+    for arg in lhs.args
+        isa(arg, Expr) && uses_default_differential!(arg, diff_vars)
+    end
+
+    isempty(diff_vars) ? (return false) : (return true)
+end
+
```
isaacsas, line 729: @vyudu are these changes needed for the function bug fix? If not, can we move them to a followup PR to expand the supported LHS's in ODEs? Let's keep this focused on just fixing the bug with the current setup.

src/dsl.jl
isaacsas: It was refactored to a function that looks like it is trying to handle more general LHS. All of that makes sense to do as we should try to support any valid ODE that MTK can support, but that should really be a in a followup PR to generalize what we support (keeping this focused on just fixing the function bug).

isaacsas (approved):
This looks good to me assuming docs build ok. @TorkelE can you confirm it is ok with you before we merge?

TorkelE:
I will have a look later today, there is a thing I want to check on my computer.

src/dsl.jl
```diff
@@ -904,17 +908,28 @@ end
...
+# Recursively escape functions in the right-hand-side of an equation written using user-defined functions. Special function calls like "hill(...)" are not expanded.
+function escape_equation_RHS!(eqexpr::Expr)
+    # lhs = recursive_escape_functions!(eqexpr.args[2])
+    # eqexpr.args[2] = lhs
+
+    rhs = recursive_escape_functions!(eqexpr.args[3])
+    eqexpr.args[3] = rhs
+
+    eqexpr
+end
```
TorkelE, line 931: I'd probably remove the commented-out lines of code, and remove the superfluous empty lines in the function to compress it a bit. But a matter of test, up to you.

TorkelE:
If you ctrl + f search for "### " in the dsl_options.jl test file you will find a section with tests for adding coupled equations to the DSL, I'd move the tests there. Otherwise looks good.
(we have been meaning to have a short dev doc page somewhere with stuff like this, without it is is quite hard to know about stuff likes that!)

PR #1048, opened 2024-09-10
test master

isaacsas:


isaacsas:
Ahh, I see the issue. 

@TorkelE this isn't a consistent initialization. At `t = 0` the equations should hold, and as such `X` and `Y` are fully deteremined by `S(0)`, however, you are specifying values for them. Can you update any such examples in the tests? There seem to be multiple failures now due to such issues.

ChrisRackauckas:
Yes, that would be from https://github.com/SciML/ModelingToolkit.jl/pull/3032. It fixes @TorkelE 's issue of https://github.com/SciML/ModelingToolkit.jl/issues/2619 which there were cases we let inconsistent initial conditions through before. Now that we plugged that, yes that would error now 😅

isaacsas:
Thanks!

Issue #1051, opened 2024-09-12
Extemely slow creation of `ODEProblem` from `ReactionSystem`


johannesnauta:
**Describe the bug 🐞**

Me again. I am simulating Lotka-Volterra systems, and have recently jumped back on `Catalyst.jl` for the flexibility and comparison with other solvers (SDEs, SSAs, etc.) other than solving ODEs. However, I find that creating (not solving) the `ODEProblem` from a `ReactionSystem` that is sufficiently large takes an incredible amount of time. The time it takes reaches the point where trying to create a problem of medium sizes (~50 species) basically does not finish. Additionally, when using an `ODESystem` using `ModelingToolkit.jl` directly, and creating the `ODEProblem` does not seem to have this issue.

**Minimal Reproducible Example 👇**
To illustrate, consider the following functions, one using `Catalyst.jl`, and one using `ModelingToolkit.jl`. They both create a simple Lotka-Volterra system, that is

$$
\frac{dx_i}{dt} = r_i x_i ( 1 - \sum_j A_{ij} x_j )
$$

#### Using `Catalyst.jl`
(note: if there are other/better ways to generate the `ReactionSystem`, let me know!)

```julia
function create_rs(S::Int)
    t = default_t()
    @parameters r[1:S] A[1:S,1:S]
    @species (x(t))[1:S]

    #/ Allocate and fill
    growthrxs = Array{Reaction}(undef, S)
    interactionrxs = Array{Reaction}(undef, S, S)
    for i in 1:S
        #~ Growth
        growthrxs[i] = Reaction(r[i], [x[i]], [x[i]], [1], [2])
        #~ Diagonal terms
        interactionrxs[i,i] = Reaction(r[i]*A[i,i], [x[i]], [x[i]], [2], [1])
        for j in (i+1):S
            #~ Off-diagonal terms
            interactionrxs[i,j] = Reaction(r[i]*A[i,j], [x[i], x[j]], [x[j]], [1, 1], [1])
            interactionrxs[j,i] = Reaction(r[i]*A[j,i], [x[j], x[i]], [x[i]], [1, 1], [1])
        end
    end

    #/ Gather reactions
    rxs = [growthrxs..., interactionrxs...]
    @named glvrs = ReactionSystem(rxs, t, combinatoric_ratelaws=false)
    return Catalyst.complete(glvrs)
end
```

#### Using `ModelingToolkit.jl`
```julia
function create_ODESystem(S::Int)
[...]

PR #1052, opened 2024-09-13
drop scalarizing

isaacsas:
Closes https://github.com/SciML/Catalyst.jl/issues/1051

Not handled yet
- [x] parameters are not being left as arrays/vectors when using the two-argument constructor
- [x] scalarizing in the DSL
- [ ] ~scalarizing in spatial code~ I'm not sufficiently familiar with the spatial code, and tests are passing I think, so I will leave this for @TorkelE to handle when he has time. See https://github.com/SciML/Catalyst.jl/issues/1059

WIth this PR the DSL will still scalarize non-parameters, but I believe this is consistent with MTK now.

### DIFF
--- a/src/reactionsystem.jl
+++ b/src/reactionsystem.jl
@@ -485,11 +485,10 @@ function make_ReactionSystem_internal(rxs_and_eqs::Vector, iv, us_in, ps_in;
         spatial_ivs = nothing, continuous_events = [], discrete_events = [],
         observed = [], kwargs...)
 
-    # Filters away any potential observables from `states` and `spcs`.
+    # Error if any observables have been declared a species or variable
     obs_vars = Set(obs_eq.lhs for obs_eq in observed)
     any(in(obs_vars), us_in) &&
         error("Found an observable in the list of unknowns. This is not allowed.")
-    # us_in = filter(u -> !any(isequal(u, obs_var) for obs_var in obs_vars), us_in)
 
     # Creates a combined iv vector (iv and sivs). This is used later in the function (so that 
     # independent variables can be excluded when encountered quantities are added to `us` and `ps`).
--- a/src/reactionsystem.jl
+++ b/src/reactionsystem.jl
@@ -487,7 +487,8 @@ function make_ReactionSystem_internal(rxs_and_eqs::Vector, iv, us_in, ps_in;
 
     # Filters away any potential observables from `states` and `spcs`.
     obs_vars = Set(obs_eq.lhs for obs_eq in observed)
-    any(in(obs_vars), us_in) && error("Found an observable in the list of unknowns. This is not allowed.")
+    any(in(obs_vars), us_in) &&
+        error("Found an observable in the list of unknowns. This is not allowed.")
     # us_in = filter(u -> !any(isequal(u, obs_var) for obs_var in obs_vars), us_in)
 
     # Creates a combined iv vector (iv and sivs). This is used later in the function (so that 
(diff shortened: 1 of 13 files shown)


test/reactionsystem_core/reactionsystem.jl
```diff
@@ -126,15 +127,11 @@ let
...
+    kvals[1] = 5.0
+    pmap = [k => kvals]
     prob = ODEProblem(rs, u0map, (0, 10.0), pmap)
     @test prob.ps[k[1]] == 5.0
     @test prob.u0[1] == 5.0
-    u0 = [10.0, 11.0, 12.0, 13.0]
-    ps = [float(x) for x in 100:119]
-    prob = ODEProblem(rs, u0, (0, 10.0), ps)
-    @test  [prob.ps[k[i]] for i in 1:20] == ps
-    @test prob.u0 == u0
```
isaacsas, line 137: These are not appropriate inputs since they aren't mappings, hence I removed them.

src/reactionsystem.jl
```diff
@@ -485,16 +485,17 @@ function make_ReactionSystem_internal(rxs_and_eqs::Vector, iv, us_in, ps_in;
         spatial_ivs = nothing, continuous_events = [], discrete_events = [],
         observed = [], kwargs...)
 
-    # Filters away any potential observables from `states` and `spcs`.
-    obs_vars = [obs_eq.lhs for obs_eq in observed]
-    us_in = filter(u -> !any(isequal(u, obs_var) for obs_var in obs_vars), us_in)
```
isaacsas, line 490: @TorkelE I dropped this as I think it makes more sense to disallow including observable variables in the unknowns, and to instead give users an error (rather than filtering them out). This also seems less likely to lead to user code bugs where a user thinks something is an unknown but it is implicitly an observable.

isaacsas:
@TorkelE if tests pass I am happy with this (I'd like to get a new release with this out). Can you let me know if it looks OK to you?

I guess the only other question is if this requires a breaking release? I think it may since we are changing how parameters are handled to drop scalarizing, which will break peoples existing codes unfortunately. If you are OK I'll make this V15 and add a note that removing scalarizing is the only change.

test/dsl/dsl_advanced_model_construction.jl
```diff
@@ -340,10 +340,23 @@ let
         k[1]*a+k[2], X[1] + V[1]*X[2] --> V[2]*W*Y + B*C
     end
 
-    @parameters k[1:3] a B
```
TorkelE, line 343: I think this one is intentionally different (i.e. giving different `k` lengths, but only using some, should give the same result). However, there should be a comment pointing this out though.

test/dsl/dsl_advanced_model_construction.jl
```diff
@@ -340,10 +340,23 @@ let
...
+    @parameters k[1:2] a B
     @variables (V(t))[1:2] W(t)
     @species (X(t))[1:2] Y(t) C(t)
     rx = Reaction(k[1]*a+k[2], [X[1], X[2]], [Y, C], [1, V[1]], [V[2] * W, B])
     @named arrtest = ReactionSystem([rx], t)
-    arrtest == rn
```
TorkelE, line 348: small detail 😅

test/dsl/dsl_advanced_model_construction.jl
isaacsas: Yeah, but they won't be equal any more as the underlying symbolic array variables are no longer considered the same. I can just not test equality if you prefer then?

TorkelE: No, in that case all is good

PR #1053, opened 2024-09-14
Added docs and assets for interactive simulation plotting tutorial

jonathanfischer97:
Added tutorial on how to simulate and plot a `Catalyst.jl` generated `ODEProblem` using `GLMakie.jl`. 

Additional embedded assets are SVG screenshots and an mp4 demoing the full interactive window.

image
image (1)
image (2)
image (3)


docs/src/model_simulation/examples/interactive_brusselator_simulation.md
```diff
@@ -0,0 +1,273 @@
+# [Interactive Simulation and Plotting](@id interactive_brusselator)
+
+Catalyst can utilize the [GLMakie.jl](https://github.com/JuliaPlots/GLMakie.jl) package for creating interactive visualizations of your reaction network dynamics. This tutorial provides a step-by-step guide to creating an interactive visualization of the Brusselator model, building upon the basic [Brusselator](@ref basic_CRN_library_brusselator) example.
+
+!!! note
+    This tutorial assumes you have GLMakie.jl installed. If not, you can install it by running `using Pkg; Pkg.add("GLMakie")` in your Julia REPL.
```
TorkelE, line 6: this note is probably not required, as we do not have similar notes elsewhere (however, if you think it is better, having it still here does not hurt either)

docs/src/model_simulation/examples/interactive_brusselator_simulation.md
```diff
@@ -0,0 +1,273 @@
...
+    u0 = [:X => X0, :Y => Y0]
+    newprob = remake(prob, p=p, u0=u0)
+    solve(newprob, Tsit5(), saveat = 0.1) # saveat = 0.1 saves the solution every 0.1 time units. It's important to set this so that the size of the solution doesn't change during interactivity.
+end
+```
+
+This code sets up our Brusselator model using Catalyst.jl's `@reaction_network` macro. We also define initial parameters, initial conditions, create an `ODEProblem`, and define a function to solve the ODE with given parameters. 
+
+!!! tip
+    Be sure to set `saveat` to a value that is appropriate for your system; otherwise, the size of the solution can change during interactivity, which will cause dimension mismatch errors once we add our interactive elements.
```
TorkelE, line 44: This feel more like a note than a tip

docs/src/model_simulation/examples/interactive_brusselator_simulation.md
```diff
@@ -0,0 +1,273 @@
...
+
+```julia
+lines!(ax_time, lift(sol -> sol.t, solution), lift(sol -> sol[:X], solution), 
+       label = "X", color = :green, linewidth = 2, linestyle = :dash)
+```
+
+## [Extending the interactive visualization](@id extending_interactive_visualization)
+
+You can further extend this visualization by:
+- Adding other interactive elements, such as buttons or dropdowns to control different aspects of the simulation or visualization.
```
TorkelE, line 269: If there are good places in Makie's docs explaining how to do e.g. buttons, I would add a link to that.

TorkelE:
This looks really nice! I think my main comment is that it would be nice to run as much of the docs as possible dynamically, e.g. by using
```
'''@example dynamic_brusselator_example
# code
'''
```
rather than

```
'''julia
# code
'''
```
(see rest of docs)

That way, if there is an error in the code (of if an error is produced later on due to package updates) that is caught. I imagine this will become problems for the code doing the interactivity stuff, but some code blocks should be possible to make this way.

docs/src/model_simulation/examples/interactive_brusselator_simulation.md
```diff
@@ -0,0 +1,273 @@
...
+tspan = (0.0, 50.0)
+
+oprob = ODEProblem(brusselator, u0, tspan, p)
+
+# Function to solve the ODE
+function solve_brusselator(A, B, X0, Y0, prob = oprob)
+    p = [:A => A, :B => B]
+    u0 = [:X => X0, :Y => Y0]
+    newprob = remake(prob, p=p, u0=u0)
+    solve(newprob, Tsit5(), saveat = 0.1) # saveat = 0.1 saves the solution every 0.1 time units. It's important to set this so that the size of the solution doesn't change during interactivity.
```
isaacsas, line 37:
```suggestion
    solve(newprob, Tsit5(), saveat = 0.1) 
```

docs/src/model_simulation/examples/interactive_brusselator_simulation.md
```diff
@@ -0,0 +1,273 @@
...
+end
+```
+
+This code sets up our Brusselator model using Catalyst.jl's `@reaction_network` macro. We also define initial parameters, initial conditions, create an `ODEProblem`, and define a function to solve the ODE with given parameters.
```
isaacsas, line 41:
```suggestion
This code sets up our Brusselator model using Catalyst.jl's `@reaction_network` macro. We also define initial parameters, initial conditions, create an `ODEProblem`, and define a function to solve the ODE with given parameters.  Setting `saveat = 0.1` in the call to `solve` ensures the solution is saved with the desired temporal frequency we want for our later plots.
```

isaacsas (changes requested):
As @TorkelE said it would be great if you could update the first static plot to use example blocks that build it dynamically. That would at least let us catch if for some reason the beginning of the tutorial stops work.

jonathanfischer97:
Made all the requested suggestions, including dynamically building the plot from the code blocks!

The reason I didn't do this originally is because `GLMakie` displays the frame buffer in a separate window, so I wasn't sure how this would be displayed as an embedded plot in the HTML. But it seems there are arguments to plot inline and prevent the window, so I added that as a hidden line in the example code.

The last code block doesn't display the dynamically generated figure because I have the mp4 of the interaction instead

jonathanfischer97:
Looks like some tests and the documentation build weren't successful. This is the first time I've PRed docs, do I need to fix anything on my end?

isaacsas:
There are some issues with the Catalyst tests currently. @TorkelE is fixing them in https://github.com/SciML/Catalyst.jl/pull/1070

Once that gets merged if you merge master into your local pr branch and then push it to Github that should hopefully allow your tests to go through here.

isaacsas:
(I usually use Gitkraken for such things as you can just drag master onto your local branch and tell it to merge master into your branch.)

isaacsas:
@TorkelE no worries -- I hope you feel better soon!

isaacsas:
@jonathanfischer97 can you merge master into this PR? I think that should allow tests to pass now.

isaacsas:
@jonathanfischer97 looks like the StructuralIdentifiability extension is causing issues here still. I'll try to update master later today to completely remove it so things can work again.

isaacsas:
@jonathanfischer97 just removed all the StructuralIdentifiability stuff if you want to merge master again. That will hopefully get this to pass CI for you.

isaacsas:
@jonathanfischer97 looks like there is a link error in the tutorial.

jonathanfischer97:
Issue with GLMakie in CI without GPU, apparently requires using `xvfb` in the workflow. 

There's an example in the `GLMakie.jl` tests, will look at it next week and push the fix

.github/workflows/Documentation.yml
```diff
@@ -15,14 +15,17 @@ jobs:
...
       - name: Install dependencies
         run: julia --project=docs/ -e 'ENV["JULIA_PKG_SERVER"] = ""; using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
       - name: Build and deploy
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
           DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }} # For authentication with SSH deploy key
           GKSwstype: "100" # https://discourse.julialang.org/t/generation-of-documentation-fails-qt-qpa-xcb-could-not-connect-to-display/60988
-        run: julia --project=docs/ --code-coverage=user docs/make.jl
+        run: |
+          DISPLAY=:0 xvfb-run -s '-screen 0 1024x768x24' julia --project=docs/ --code-coverage=user docs/make.jl
```
isaacsas, line 28: What practical impact does this have? Does it mean all our docs are now being built under the assumption they live in a virtual 1024x768 framebuffer?

jonathanfischer97: No, the use of `xvfb` with `-screen 0 1024x768x24` does not mean all the plots rendered during doc build will be constrained to that resolution, if that's what you mean. This setup only applies during the CI build to provide a virtual display for `GLMakie` plots, which require `OpenGL`, but also doesn't dictate the final resolution of the `GLMakie` plots, which instead are determined by the plot settings themselves.

Plots rendered by `Plots.jl` (using GR) or `CairoMakie.jl` aren't affected, cause these libraries use software-based rendering that doesn't depend on an `OpenGL` display. Their output is generated independently of the `xvfb` virtual framebuffer, so that those final doc artifacts are unaffected by this temporary display configuration.

jonathanfischer97: However, if you'd rather not implement this change in the the docs CI, another option is I can just use `CairoMakie` as the backend for all the inline rendered plots, but make the presented code blocks look like they use `GLMakie`. 

But this would obviously break some of the spirit of making the plots dynamically to ensure `GLMakie` functionality in the tutorial.

isaacsas: No that's fine. I just wanted to understand the code.

If tests pass is this ready to merge on your end?

isaacsas:
@jonathanfischer97 if you feel this is done I'm happy to merge it.

jonathanfischer97:
@isaacsas This is ready on my end! I think this should fix most everything, so merge at your convenience!

PR #1056, opened 2024-09-20
DO NOT MERGE - test master

isaacsas:


isaacsas:
@TorkelE tests seem to be failing from a test that is relying on Symbolics.jl internals.

isaacsas:
Looking at this more carefully it seems to me like this is a bug and that variables arising from observables are not being created correctly by the DSL. They should be completely equivalent to a normal variable, even generating the same code, but when I look at what the DSL generates it is not producing the same thing.

TorkelE:
Yeah, there is a comment on it
```
# Declares observables implicitly/explicitly.
# Cannot test `isequal(rn1, rn2)` because the two sets of observables have some obscure Symbolics
# substructure that is different.
```

Last time when i had a look at it, I could not figure it out though, e.g.
```julia
isequal(rn1.X, rn2.X) # false
isequal(Symbolics.unwrap(rn1.X), Symbolics.unwrap(rn2.X)) # false
isequal(Symbolics.unwrap(rn1.X).metadata, Symbolics.unwrap(rn2.X).metadata) # true
```

However, the remaining fields checked with
```julia
println(fieldnames(typeof(Symbolics.unwrap(rn1.X))))
```
is nothing like normal Julia types
```
(:metadata, Symbol("###simplified###2"), Symbol("###Any###3"), Symbol("###Any###4"), Symbol("###Any###5"), Symbol("###Any###6"), Symbol("###Any###7"), Symbol("###tag###8"))
```
and I'd really need a Symbolics expert to explain to me what is going on and what these weird stuff are.

PR #1069, opened 2024-09-25
See if non-bifurcationkit non-homotopycontinuation tests pass

isaacsas:


isaacsas:
@TorkelE seems like all three extension-related tests are now broken (HC and BK) or taking an enormous amount of time (SI). The latest MTK changes seem to have really messed them up.

PR #1070, opened 2024-09-27
Fix broken tests

TorkelE:
Fixes the tests that have been broken by recent updates. 

The long running for identifiability analysis has to do with us using array variables for representing unknown quantities, and there seems to be a problem with this (likely related to recent updates). Might be easiest to simple rewrite code to not use that, will have a look.

### DIFF
--- a/test/extensions/bifurcation_kit.jl
+++ b/test/extensions/bifurcation_kit.jl
@@ -56,32 +56,35 @@ end
 # Checks that the same bifurcation problem is created as for BifurcationKit.
 # Checks with Symbolics as bifurcation and plot vars.
 # Tries setting `jac=false`.
+# Note: Only one parameter used, as tests technically depended on internal parameter ordering
+# (Potentially the test should also be removed as it tests internal parameter stuff, however, test
+# was written a while ago when I paid less attention to this kind of stuff.)
 let 
     # Creates BifurcationProblem via Catalyst.
     bistable_switch = @reaction_network begin
-        0.1 + hill(X,v,K,n), 0 --> X
-        d, X --> 0
+        0.1 + hill(X,5.0,K,3), 0 --> X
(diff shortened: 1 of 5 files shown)


test/extensions/bifurcation_kit.jl
```diff
@@ -56,32 +56,35 @@ end
...
-        0.1 + hill(X,v,K,n), 0 --> X
-        d, X --> 0
+        0.1 + hill(X,5.0,K,3), 0 --> X
+        1.0, X --> 0
     end
-    @unpack X, v, K, n, d = bistable_switch
+    @unpack X, K = bistable_switch
     u0_guess = [X => 1.0]
-    p_start = [v => 5.0, K => 2.5, n => 3, d => 1.0]
+    p_start = [K => 2.5]
```
isaacsas, line 70: Do multiple parameters that are symbolic no longer work with the extension? Why do you need to change the parameters to hard-coded values?

TorkelE: No, still work (see other tests).

The problem is that I wrote a test which evaluates some internal MTK functions and check that they return the same stuff. However, this kind of depended on ordering, and the test fails due to the parameter ordering not being right. I changed the this test so that there was only a single parameter (so that parameter ordering is not really a worry).

Probably the test should just be removed instead, but for now I just tried to make a minimal change.

TorkelE:
There is still some weird stuff with SI. I have been looking at it for a while and untangled it a bit, but still not fully sure how to finally fix it. SI being broken is probably what holds up the docs as well.

isaacsas:
Sounds good. Feel free to merge this then when you want.

isaacsas:
@TorkelE I'm going to merge this as I don't want to hold off much longer on getting PRs merged and making a release with the array fixes.

PR #1071, opened 2024-10-01
fix: turn off parameter splitting for `BifurcationProblem`

AayushSabharwal:
`BifurcationProblem` does not support symbolic indexing and as a result, MTK (somewhat hackily) just removes `sys.index_cache` in the `BifurcationProblem` constructor to force `split = false` behavior. However, with `complete` now reordering parameters, this breaks assumptions and makes analysis of these problems difficult.

In hindsight, `BifurcationProblem` should have just errored if the system was created with `split=true`, but changing that behavior is breaking. I will add a warning in MTK. For now, this should fix CI

## Checklist

### DIFF
--- a/ext/CatalystBifurcationKitExtension/bifurcation_kit_extension.jl
+++ b/ext/CatalystBifurcationKitExtension/bifurcation_kit_extension.jl
@@ -24,7 +24,7 @@ function BK.BifurcationProblem(rs::ReactionSystem, u0_bif, ps, bif_par, args...;
     Catalyst.conservationlaw_errorcheck(rs, vcat(ps, u0))
     nsys = convert(NonlinearSystem, rs; defaults = Dict(u0),
         remove_conserved = true, remove_conserved_warn = false)
-    nsys = complete(nsys)
+    nsys = complete(nsys; split = false)
 
     # Makes BifurcationProblem (this call goes through the ModelingToolkit-based BifurcationKit extension).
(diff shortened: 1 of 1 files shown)


isaacsas:
@AayushSabharwal just circling back to this now that Catalyst CI finally works again. Is this still needed? I think MTK's BifurcationKit extension was updated some since you submitted this to Catalyst?

isaacsas:
I don't see `split = true` being used there at all (in the tests for example).

AayushSabharwal:
It is still needed. `complete` with `split = true` (the default) reorders parameters. Since `BifurcationProblem` doesn't support symbolic indexing, the order matters for user accessibility.

ChrisRackauckas:
We can just make BifurcationProblem support SII? That seems like the better long term solution here.

AayushSabharwal:
I agree, this was written to get tests passing quickly. A quick look at BifurcationKit internals says it shouldn't be too difficult to support SII so I'll see what I can do.

isaacsas:
Tests seem to pass now in either case. MTK seems to explicitly not set this though: https://github.com/SciML/ModelingToolkit.jl/blob/1f2d9430cfc6b7ebe2044dfb1b9a078eac4af87e/ext/MTKBifurcationKitExt.jl#L152

So that needs updating I guess (until BifurcationKit is made to work with SII)?

PR #1074, opened 2024-10-09
disable SI for now

vyudu:
Removing SI from the Project.toml and docs for now


isaacsas:
docs/pages.jl needs updating too.

isaacsas (approved):
If the docs build ok too with no errors feel free to merge once CI passes.

PR #1089, opened 2024-10-25
Try reenabling PETab tutorial

isaacsas:
It would be nice to get this back into the docs.

### DIFF
--- a/docs/pages.jl
+++ b/docs/pages.jl
@@ -46,7 +46,7 @@ pages = Any[
     ],
     "Inverse problems" => Any[
         "inverse_problems/optimization_ode_param_fitting.md",
-        # "inverse_problems/petab_ode_param_fitting.md",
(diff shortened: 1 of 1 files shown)


isaacsas:
@sebapersson should PETab work with the latest Catalyst and MTK now?

isaacsas:
Ahh, I see PEtab doesn't support the latest optimization yet which we've updated to, so this isn't going to work.

isaacsas:
Awesome, let us know and we will reenable as soon as a release that works with the bumped optimizations libs is out.

ChrisRackauckas:
If you're not using the data iterator or the extra cost function argument results it should just be a quick bump. I have a few tutorial bumps to look into (DiffEqFlux, SciMLSensitivty), but they've now made it to the top of my inbox so if you need help, see those PRs from Vaibhav before I get the merge.

PR #1092, opened 2024-10-28
try to disable bifkit extension

isaacsas:
To see if we can then test against the current MTK.


isaacsas:
If https://github.com/SciML/Catalyst.jl/pull/1094 can be made to work I'll close this since it should resolve the issue of not letting MTK test against Catalyst by breaking things into groups.

Issue #1095, opened 2024-10-28
check BifKit extension in the context of conservation law removal


isaacsas:
We need to check if it is actually working when one is using an initial condition as the bifurcation parameter. I suspect that it has no way to know it needs to update the conserved constant in this case (i.e. it is just changing that parameter in the underlying parameter vector), and hence will give incorrect results.

isaacsas:
This doesn't work, but at least it errors:
```julia
rn = @reaction_network begin
              @parameters M
              @species A(t) = M
              (1.0,1.0), A <--> B
              end

osys = complete(convert(ODESystem, rn; remove_conserved = true))
u_guess = [osys.B => 2.0]
p_start = [osys.M => 3.0]
bprob = BifurcationProblem(osys, u_guess, p_start, osys.M; plot_var = osys.A)
```
giving
```julia
ERROR: ArgumentError: SymbolicUtils.BasicSymbolic{Real}[M] are either missing from the variable map or missing from the system's unknowns/parameters list.
Stacktrace:
 [1] throw_missingvars_in_sys(vars::Vector{SymbolicUtils.BasicSymbolic{Real}})
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/XHRvt/src/utils.jl:765
 [2] promote_to_concrete(vs::Vector{SymbolicUtils.BasicSymbolic{Real}}; tofloat::Bool, use_union::Bool)
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/XHRvt/src/utils.jl:784
 [3] varmap_to_vars(varmap::Vector{…}, varlist::Vector{…}; defaults::Dict{…}, check::Bool, toterm::Function, promotetoconcrete::Nothing, tofloat::Bool, use_union::Bool)
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/XHRvt/src/variables.jl:171
 [4] BifurcationProblem(::NonlinearSystem, ::Vector{…}, ::Vector{…}, ::Num; plot_var::Num, record_from_solution::Function, jac::Bool, kwargs::@Kwargs{})
   @ MTKBifurcationKitExt ~/.julia/packages/ModelingToolkit/XHRvt/ext/MTKBifurcationKitExt.jl:104
 [5] BifurcationProblem(::ODESystem, ::Vector{Pair{Num, Float64}}, ::Vararg{Any}; kwargs::@Kwargs{plot_var::Num})
[...]

PR #1101, opened 2024-10-29
BifurcationKit test

vyudu:
Wrote a test for the dependent parameters for BK from #1095, that can be uncommented once that works


isaacsas:
Let’s merge this once MTK releases with your update since then v.4 should work here. 

You can use `@test_broken` for now as that would let us know when the test starts working.

isaacsas:
@vyudu if you want to merge master into this the SciMLBase cap will hopefully allow it to run through to completion.

isaacsas:
@vyudu the doc folder Project.toml needs updating too.

Project.toml
```diff
@@ -42,7 +42,7 @@ CatalystHomotopyContinuationExtension = "HomotopyContinuation"
 # CatalystStructuralIdentifiabilityExtension = "StructuralIdentifiability"
 
 [compat]
-BifurcationKit = "0.3"
+BifurcationKit = "0.4.3"
```
isaacsas, line 45:
```suggestion
BifurcationKit = "0.4.4"
```

docs/Project.toml
```diff
@@ -45,7 +45,7 @@ Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
 
 [compat]
 BenchmarkTools = "1.5"
-BifurcationKit = "0.3.4"
+BifurcationKit = "0.4"
```
isaacsas, line 48:
```suggestion
BifurcationKit = "0.4.4"
```

isaacsas:
OK, seems there are now new intialization related errors. That never happened under the old extension so I guess this is something broken by recent updates to the MTK-BifurcationKit extension (or initialization related changes). 

If someone wants to take a look at what is going on feel free. I think for now I will just plan to disable this extension like we've disabled SI as I don't have the time or capacity to handle these breakages.

ChrisRackauckas:
> That never happened under the old extension so I guess this is something broken by recent updates to the MTK-BifurcationKit extension (or initialization related changes).

BifurcationKit is broken due to Krylov.jl things, completely unrelated to anything MTK?

isaacsas:
And this CI failure is something symbolic related: https://github.com/SciML/Catalyst.jl/actions/runs/11859907434/job/33053954257?pr=1101#step:6:1289

isaacsas:
@vyudu I dropped your new test. It had a parameter that was not used and no value specified which was causing it to fail CI, and the ODESystem test part seems unrelated to BifurcationKit (so while a good test should probably be placed in a different location).

If you want to followup if/when this gets merged and modify those tests feel free (I've left them commented out for now).

Hopefully adding a parent field to `ReactionSystem`s will fix the symbolic indexing issues.

vyudu:
Ahh, I see, leaving `ratechange` in was sloppy sorry. Will revisit this after merge

PR #1114, opened 2024-11-14
Fix observables in DSL

TorkelE:
I think this shoudl be fixable. Technically we know the independent variables for each variable/species within the DSL, so we can extract and use that there (instead of using at compile time).

This PR only really add the test though (I have a new computer set up so want to check that GitKraken is connected properly etc.).


isaacsas (approved):
This failed last I checked. It would be interesting if it now is fixed.

TorkelE:
Superseeded by https://github.com/SciML/Catalyst.jl/pull/1124

PR #1115, opened 2024-11-14
GraphMakie update 

vyudu:
Refactoring the Graphviz code into GraphMakie. Some plots generated: 

Brusselator species-reaction graph
fig1
Brusselator complex graph
fig2
Repressilator species-reaction graph
fig3

Might be worth playing around with the layouts a bit more to see if we can make the edge-lengths for these graphs more consistent and have the layouts be more horizontal. Will work on updating the documentation in a follow-up.


isaacsas:
The first and third look great! The second is a bit strange. Do you want to tweak the layouts more in the PR, or get it merged as is and update later? (I ask as I'll review it sonner in the latter case or wait till you say you are done in the former case).

isaacsas:
Also, is this based on the current master? I thought I had removed BifurcationKit there so the docs build but this seems to be crashing on it.

vyudu:
Probably just merge this one as is and I'll look more at it in a follow-up. Also just rebased it, it seems like the remote branch was based on an earlier master commit

isaacsas:
It may still fail -- I have another PR that is completely removing Bifkit, so it may need one more rebasing once that is merged (but if tests pass and docs build as is I'll merge after reviewing).

isaacsas (changes requested):
Is there any way to add some tests for this?

ext/CatalystGraphMakieExtension.jl
```diff
@@ -1,11 +1,13 @@
 module CatalystGraphMakieExtension
 
 # Fetch packages.
-using Catalyst, GraphMakie
+using Catalyst, GraphMakie, Graphs 
 import Catalyst: lattice_plot, lattice_animation, extract_vals
-import Graphs: AbstractGraph, SimpleGraph
+import Graphs: AbstractGraph, SimpleGraph, SimpleDiGraph, SimpleEdge, src, dst, ne, nv 
+import Catalyst: speciesreactiongraph, incidencematgraph
```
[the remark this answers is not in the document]
vyudu: Some of these imports were there before I believe. Though I thought it'd make it easier to not have to write out Graphs.SimpleEdge, etc. too much in the code since some of these names aren't exported by default, but I can just have them be qualified in the code, no issue there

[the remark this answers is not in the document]
isaacsas: Just use `using` then. (We should eventually get rid of import everywhere anyways I think, but that is not high priority).

isaacsas:
@vyudu probably worth updating this to master now that the BifurcationKit PR is merged. That should hopefully have fixed some issues in tests and doc builds.

ext/CatalystGraphMakieExtension.jl
```diff
@@ -1,11 +1,13 @@
...
 
 # Creates and exports hc_steady_states function.
 include("CatalystGraphMakieExtension/graph_makie_extension_spatial_modelling.jl")
-
+include("CatalystGraphMakieExtension/rn_graph_plot.jl")
+export SRGraph, ComplexGraph
```
isaacsas, line 12: In Catalyst.jl have

```julia
function plot_speciesreaction_graph end
function plot_complex_graph end
export plot_speciesreaction_graph, plot_complex_graph
```
Note you also need to update the doc strings below for the new names.

isaacsas: Actually, for consistency it should probably be
```julia
plot_species_reaction_graph
```

isaacsas: Which admittedly is very long. Can we think of a shorter descriptive name?

vyudu: plot_sr_graph?

isaacsas: What about just `plot_network(...)` and `plot_complexes(...)`?

ext/CatalystGraphMakieExtension/rn_graph_plot.jl
```diff
@@ -0,0 +1,193 @@
...
+end
+
+"""
+    plot_petrinet(rn::ReactionSystem)
+
+    See the documentation for [`plot_network`](@ref).
+"""
+function Catalyst.plot_petrinet(rn::ReactionSystem) 
+    plot_network(rn)
+end
```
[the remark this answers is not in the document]
vyudu: Just thought some people might know them better as Petri nets, but yeah probably better to mention this in the documentation than have another name.

isaacsas (approved):

isaacsas:
@vyudu feel free to merge when done now.

isaacsas:
@vyudu ping -- are you done with this? Can we merge?

vyudu:
I am yeah, we can merge

PR #1122, opened 2024-11-15
Add @no_infer flag for turning off species/variable/parameter inferring

vyudu:
Probably missed some places where we do this, @TorkelE might appreciate some guidance when you can. Also probably need some more tests to catch edge cases


src/dsl.jl
```diff
@@ -286,6 +286,11 @@ end
 ### DSL Internal Master Function ###
 
 # Function for creating a ReactionSystem structure (used by the @reaction_network macro).
+# What should no_infer do? We currently infer in:
+#   equations
+#   observables
+#   differentials
+# make it so that we no longer do so.
```
isaacsas, line 293:
```suggestion
```

src/dsl.jl
```diff
@@ -317,7 +323,7 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
 
     # Reads equations. 
     vars_extracted, add_default_diff, equations = read_equations_options(
-        options, variables_declared)
+        options, variables_declared; noinfer = noinfer)
```
isaacsas, line 326:
```suggestion
        options, variables_declared; noinfer)
```

src/dsl.jl
```diff
@@ -341,13 +347,13 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
 
     # Reads observables.
     observed_vars, observed_eqs, obs_syms = read_observed_options(
-        options, [species_declared; variables], all_ivs)
+        options, [species_declared; variables], all_ivs; noinfer = noinfer)
```
isaacsas, line 350:
```suggestion
        options, [species_declared; variables], all_ivs; noinfer)
```

src/dsl.jl
```diff
@@ -341,13 +347,13 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
...
 
     # Collect species and parameters, including ones inferred from the reactions. 
     declared_syms = Set(Iterators.flatten((parameters_declared, species_declared,
         variables)))
     species_extracted, parameters_extracted = extract_species_and_parameters!(
-        reactions, declared_syms)
+        reactions, declared_syms; noinfer = noinfer)
```
isaacsas, line 356:
```suggestion
        reactions, declared_syms)
```

src/dsl.jl
```diff
@@ -711,9 +720,12 @@ function read_equations_options(options, variables_declared)
             diff_var = lhs.args[2]
             if in(diff_var, forbidden_symbols_error)
                 error("A forbidden symbol ($(diff_var)) was used as an variable in this differential equation: $eq")
+            elseif (!in(diff_var, variables_declared)) && noinfer
+                error("Unrecognized symbol $(diff_var) was used as a variable in this differential equation. Since the @no_infer flag is set, all variables in equations must be explicitly declared via @variables, @species, or @parameters.")
```
isaacsas, line 724:
```suggestion
                error("Unrecognized symbol $(diff_var) was used as a variable in an equation. Since the @no_infer flag is set, all variables in equations must be explicitly declared via @variables, @species, or @parameters.")
```

isaacsas (changes requested):
Maybe we want to instead have our own error type `UndeclaredSymbolicError` that we `throw` instead of calling `error`? Then we can more precisely `@test_throws UndeclaredSymbolicError` to verify this is working in all the different cases?

It might also be helpful if the equation/reaction can be displayed in some way too to actually show a user precisely where the error is occuring.

src/dsl.jl
vyudu: Does this actually work? We need the variable name since it's a kwarg right?

vyudu:
> It might also be helpful if the equation/reaction can be displayed in some way too to actually show a user precisely where the error is occuring.

Might need write some more pretty-print utilities for `ReactantStruct` and `ReactionStruct` if we want to do this, right now the way they show is very long and hard to parse. I'm happy to add these as part of this PR if it wouldn't get too bloated

src/dsl.jl
isaacsas: If a kwarg is the same as the variable name you don’t need to repeat it (when you use “;”).

isaacsas:
Yeah, adding a `show` for them sounds useful.

TorkelE:
The itnended behaviour is that all species/parameters/variables have to be declared using the corresponding options? Wouldn't a better name bet `require_declaration`?

Also, a small request, can we hold of merging this one a bit? I am trying to sort out the observables PR (think it is working locally, but need to test during the weekend),  working in the DSL file is increasingly messy, and this changes some of the relevant code. Ideally I had wanted https://github.com/SciML/Catalyst.jl/pull/985 to be merged before any further changes to the DSL code (since that makes it way easier to modify), but it is getting further and further away. if possible, after this and the osbervablwes fix, could we hold of further DSL changes until that one gets sorted out?

TorkelE:
(this is definitely useful stuff though, I'll have a look tomorrow morning)

isaacsas:
I'm fine with holding off on further changes after the observables fix and this one get merged (and with merging observables first), if you will be actively working on the older one afterwards.

TorkelE:
Yes, that is the plan. Although I just remembered we decided to change how un-declared symbolic variables in equations is handled (always set to variables, instead of for special cases). Since that is breaking I probably should do that one asap as well.
(and ofcourse, if we discover any more bugs, that might have to take residence as well, but hopefully it can all be sorted out quickly)

TorkelE:
In the main `make_reaction_system` function we already have variables
```julia
parameters_extracted
species_extracted
vars_extracted
add_default_diff
```
listing all symbolic variables we inferred from the input. I think it would be easier (since we don't have to pass an additional argument into each function and add separate checks to add a but to the main function like
```julia
if noinfer && (!isempty(parameters_extracted) || ...)
    error("...")
end
```
Then we can also make it more granular, checking each separately with a designated error messages saying which symbolic variables must be designated.

This would also solve and old issue that I think @gaurav-arya raised some time (but I maybe it was just mentioned over lunch and never made it to a actual issue, couldn't find it when I looked)

isaacsas:
@TorkelE I don't feel super strongly about this, but there are a couple advantages to the approach here:

1. It is eager, so as soon as a problem arises a user is informed. For large networks this could be useful to save waiting.
2. It offers the possiblity to provide useful error messages on where the inferred variable is coming from. 

I think 2. is the main reason to keep this approach. If we only check after collecting everything we can only report something was improperly inferred, but not where it was inferred from. If we keep this approach we should try to display for users enough context to easily find the line that is causing a problem.

vyudu:
`@require_declaration` does sound like a better name, happy to roll with that

TorkelE:
@isaacsas (As you say) I don't think (1) is much concern here. I could see how 2 could be an advantage if there is a situation where you have mistakenly used a symbolic name which you do not need to use (`x` instead of `X`), checking only in a bunch is useful only in the case when you mean to declare `x` but forgot about it.

We probably should modify the error messages to give the equations/reaction in question then.

src/dsl.jl
```diff
@@ -682,7 +703,7 @@ end
 # `vars_extracted`: A vector with extracted variables (lhs in pure differential equations only).
 # `dtexpr`: If a differential equation is defined, the default derivative (D ~ Differential(t)) must be defined.
 # `equations`: a vector with the equations provided.
-function read_equations_options(options, variables_declared)
+function read_equations_options(options, variables_declared; requiredec = false)
```
TorkelE, line 706: Making `require_declaration` is probably not necessary, since these are internal functions (almost) always called using `requiredec` as input?

src/Catalyst.jl
```diff
@@ -122,7 +122,7 @@ export symmap_to_varmap
 # reaction_network macro
 include("expression_utils.jl")
 include("dsl.jl")
-export @reaction_network, @network_component, @reaction, @species
+export @reaction_network, @network_component, @reaction, @species, UndeclaredSymbolicError
```
isaacsas, line 125: Why does this need to be exported?

src/dsl.jl
```diff
@@ -283,6 +286,17 @@ function extract_metadata(metadata_line::Expr)
...
+
+
+struct UndeclaredSymbolicError <: Exception
+    msg::String
+end
+
+function Base.showerror(io::IO, err::UndeclaredSymbolicError) 
+    print(io, "UndeclaredSymbolicError: ")
+    print(io, err.msg)
+end
```
TorkelE, line 298: Is there a reason we are creating a new error types for these? Don't really hurt, but also seems a bit... Overkill

isaacsas: Yes, we can be a bit more explicit in the tests with catching it.

isaacsas: Plus the name is more informative to a user about what the error is.

TorkelE: Sounds good. (long term) we probably should be better at errors, both creating designated ones, but also using `throw` instead of just `error`.

src/dsl.jl
vyudu: Sorry not understanding, what do you mean by making `require_declaration`?

src/Catalyst.jl
vyudu: I thought newly-defined errors needed to be exported for some reason but will remove this.

test/dsl/dsl_options.jl
```diff
@@ -1022,3 +1022,73 @@ let
     @parameters v n
     @test isequal(Catalyst.expand_registered_functions(equations(rn4)[1]), D(A) ~ v*(A^n))
 end
+
+### test that @no_infer properly throws errors when undeclared variables are written
+
+import Catalyst: UndeclaredSymbolicError
+let
+    # Test error when species are inferred
+    @test_throws UndeclaredSymbolicError @macroexpand rn = @reaction_network begin
```
TorkelE, line 1031: Technically, `rn =` is not needed in these cases

isaacsas:
@vyudu please don't forget the followup doc and HISTORY.md updates.

PR #1124, opened 2024-11-17
Fix dsl obs try again

TorkelE:
Improves how observables are created in the DSL (not using the hack we had previously). This should solve the problem of models created with the DSL turning out differently.

This will require the latest version of Symbolics, when it is released, as it has a fix that make this work. On other plus side this approach is way simpler than the alternative I was working on (and also simpler than what we currently have).


TorkelE:
Does require upping to the latest Symbolics, but this one should be sorted out now at least.

isaacsas (approved):
Nice, thanks!

PR #1125, opened 2024-11-17
Attempt to fix possible multiple PR docs build error

jonathanfischer97:
### Restrict concurrency so that only one doc is built at a time

I restricted the concurrency group so that multiple doc builds don't proceed at the same time, hopefully fixing the intermittent build error that possibly results from said concurrency. 

If the issue persists, then I have some more potential fixes involving `sleep` and retry, so that `xvfb` is initialized correctly.


isaacsas:
Fingers crossed this fixes it!

PR #1127, opened 2024-11-20
Uncap those bases

isaacsas:
Let's see where things now stand.


isaacsas:
@TorkelE this looks like the first failure in the tests on the latest SciMLBase / DiffEqBase: https://github.com/SciML/Catalyst.jl/blob/ecd28d606281683e1776e4c5310c6b38b1a6ef5e/test/reactionsystem_core/coupled_equation_crn_systems.jl#L593-L608 seems like something related to initialization and DAEs.

ChrisRackauckas:
What is that doing? It says it's testing something in the comments but there is no test?

ChrisRackauckas:
@AayushSabharwal that looks like a bug, state values shouldn't be called on the odefunction but on the prob

TorkelE:
The test are a little bit below. basically, the same model is declared twice, but me basically scrambling up the declaration as much as possible. Since the models are not actually different, their simulations should be identical, which does not seem to be the case anymore. Will have a look. Should be relatively straightforward to remove stuff until they are identical, and figure out which step do not work anymore.

AayushSabharwal:
> that looks like a bug, state values shouldn't be called on the odefunction but on the prob

Yeah. https://github.com/SciML/OrdinaryDiffEq.jl/pull/2541 fixes it, but I'm still figuring out how _none_ of the tests caught that.

AayushSabharwal:
Yeah so OrdinaryDiffEq doesn't test OverrideInit at all, it should have lit up red for multiple reasons

ChrisRackauckas:
Only MTK hits overrideinit since it requires building the init system to test it. We probably want to section those out so that SciMLBase can have that in its downstream tests.

ChrisRackauckas:
I see, that's just a twisted circle

isaacsas:
@TorkelE this is the stability failure I mention in the TODO fix issue. I think that may be the last on the list.

Can you take a look at it? I think I'll merge this unless you object and we can get that fixed in a followup?

TorkelE:
Yes, let's merge this and I will get to the stability thing over the weekend (away tomorrow, but will fix it on Sunday). The spatial one is under way, and the serialization one is ready once I have figured out the new way hierarchical systems are stored. Once I have that figured out I might add some normal tests to our file to ensure that is fully covered in the future as well.

ChrisRackauckas:
> is the stability failure I mention in the TOFIX issue. I think that may be the last item on the list besides rebuild (but perhaps it needs rebuild?).

SteadyStateProblem / NonlinearProblem just needs to be setup to handle the initialization system, which should be done shortly.

PR #1129, opened 2024-11-20
`@require_declaration` docs

vyudu:
Some starting docs for it. Might be a better spot than where I put it (end of the "[Explicit specification of network species and parameters](https://docs.sciml.ai/Catalyst/stable/model_creation/dsl_advanced/#dsl_advanced_options_declaring_species_and_parameters)" heading)

### DIFF
--- a/docs/src/model_creation/dsl_advanced.md
+++ b/docs/src/model_creation/dsl_advanced.md
@@ -278,9 +278,12 @@ In some cases it may be desirable for Catalyst to not infer species and paramete
 ```@example dsl_advanced_no_infer
 using Catalyst
 # The following case will throw an UndeclaredSymbolicError.
-rn = @reaction_network begin
+try @macroexpand @reaction_network begin
     @require_declaration
     (k1, k2), A <--> B
+end
+catch e
+    println(e.msg)
 end
 ```
 In order to avoid an error in this case all the relevant species and parameters will have to be declared.
@@ -300,7 +303,7 @@ The following cases in which the DSL would normally infer variables will all thr
 - Inferring a parameter in a reaction rate expression, as in the reaction line `k*n, A --> B`
 - Inferring a parameter in the stoichiometry of a species, as in the reaction line `k, n*A --> B`
 - Inferring a differential variable on the LHS of a coupled differential equation, as in `A` in `@equations D(A) ~ A^2`
-- Inferring an [observable](@dsl_advanced_options_observables) that is declared using `@observables`
+- Inferring an [observable](@ref dsl_advanced_options_observables) that is declared using `@observables`
 
 ## [Naming reaction networks](@id dsl_advanced_options_naming)
 Each reaction network model has a name. It can be accessed using the `nameof` function. By default, some generic name is used:
--- a/docs/src/model_creation/dsl_advanced.md
+++ b/docs/src/model_creation/dsl_advanced.md
@@ -604,8 +604,3 @@ latexify(rn; form = :ode)
 ```
 has rate $d[X]/dt = -2 k$.
 
-## Turning off species, parameter, and variable inferring
-In some cases it may be desirable for Catalyst to not infer species and parameters from the DSL, as in the case of reaction networks with very many variables, or as a sanity check that variable names are written correctly. To turn off inferring, simply add the `@require_declaration` macro to one of the lines of the `@reaction_network` declaration. Having this macro means that every single variable, species, or parameter will have to be explicitly declared using the `@variable`, `@species`, or `@parameter` macro. In the case that the DSL parser encounters an undeclared symbolic in this case, it will error with an `UndeclaredSymbolicError` and print the reaction or equation that the undeclared symbolic was found in. 
-
-```@example dsl_advanced_defaults
-```
--- a/docs/src/model_creation/dsl_advanced.md
+++ b/docs/src/model_creation/dsl_advanced.md
@@ -88,7 +88,6 @@ Generally, there are four main reasons for specifying species/parameters using t
 !!! warning
     Catalyst's DSL automatically infer species and parameters from the input. However, it only does so for *quantities that appear in reactions*. Until now this has not been relevant. However, this tutorial will demonstrate cases where species/parameters that are not part of reactions are used. These *must* be designated using either the `@species` or `@parameters` options (or the `@variables` option, which is described [later](@ref constraint_equations)).
 
-
 ### [Setting default values for species and parameters](@id dsl_advanced_options_default_vals)
 When declaring species/parameters using the `@species` and `@parameters` options, one can also assign them default values (by appending them with `=` followed by the desired default value). E.g here we set `X`'s default initial condition value to $1.0$, and `p` and `d`'s default values to $1.0$ and $0.2$, respectively:
 ```@example dsl_advanced_defaults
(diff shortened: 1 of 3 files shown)


docs/src/model_creation/dsl_advanced.md
```diff
@@ -272,6 +272,36 @@ sol = solve(oprob)
 plot(sol)
 ```
 
+### [Turning off species, parameter, and variable inferring](@id dsl_advanced_options_require_dec)
+In some cases it may be desirable for Catalyst to not infer species and parameters from the DSL, as in the case of reaction networks with very many variables, or as a sanity check that variable names are written correctly. To turn off inferring, simply add the `@require_declaration` macro to one of the lines of the `@reaction_network` declaration. Having this macro means that every single variable, species, or parameter will have to be explicitly declared using the `@variable`, `@species`, or `@parameter` macro. In the case that the DSL parser encounters an undeclared symbolic in this case, it will error with an `UndeclaredSymbolicError` and print the reaction or equation that the undeclared symbolic was found in.
```
isaacsas, line 276:
```suggestion
In some cases it may be desirable for Catalyst to not infer species and parameters from the DSL, as in the case of reaction networks with very many variables, or as a sanity check that variable names are written correctly. To turn off inferring, simply add the `@require_declaration` macro to one of the lines of the `@reaction_network` declaration. Having this macro means that every single variable, species, or parameter will have to be explicitly declared using the `@variable`, `@species`, or `@parameter` macro. In the case that the DSL parser encounters an undeclared symbolic, it will error with an `UndeclaredSymbolicError` and print the reaction or equation that the undeclared symbolic was found in. 
```

docs/src/model_creation/dsl_advanced.md
```diff
@@ -272,6 +272,36 @@ sol = solve(oprob)
...
+### [Turning off species, parameter, and variable inferring](@id dsl_advanced_options_require_dec)
+In some cases it may be desirable for Catalyst to not infer species and parameters from the DSL, as in the case of reaction networks with very many variables, or as a sanity check that variable names are written correctly. To turn off inferring, simply add the `@require_declaration` macro to one of the lines of the `@reaction_network` declaration. Having this macro means that every single variable, species, or parameter will have to be explicitly declared using the `@variable`, `@species`, or `@parameter` macro. In the case that the DSL parser encounters an undeclared symbolic in this case, it will error with an `UndeclaredSymbolicError` and print the reaction or equation that the undeclared symbolic was found in. 
+
+```@example dsl_advanced_no_infer
+using Catalyst
+# The following case will throw an UndeclaredSymbolicError.
+rn = @reaction_network begin
+    @require_declaration
+    (k1, k2), A <--> B
+end
```
isaacsas, line 284: To keep Documenter.jl from failing the build you probably need to catch the error here. (Otherwise just don't make this dynamic and manually paste in the output.)

isaacsas (changes requested):

docs/src/model_creation/dsl_advanced.md
```diff
@@ -278,10 +278,13 @@ In some cases it may be desirable for Catalyst to not infer species and paramete
 ```@example dsl_advanced_no_infer
 using Catalyst
 # The following case will throw an UndeclaredSymbolicError.
-rn = @reaction_network begin
+try @macroexpand @reaction_network begin
```
isaacsas, line 281: Why does this need `@macroexpand`? Won't that not run the code but expand it into an expression?

isaacsas: So that actually runs the code and displays the correct error? That seems like a strange behavior. Are you sure you shouldn’t be wrapping in an `eval`?

isaacsas: Ok, I see. OK I guess the error is not thrown during running of the macro generated expression but during the expression construction, hence why this works.

docs/src/model_creation/dsl_advanced.md
```diff
@@ -272,6 +272,39 @@ sol = solve(oprob)
...
+
+```@example dsl_advanced_no_infer
+using Catalyst
+# The following case will throw an UndeclaredSymbolicError.
+try @macroexpand @reaction_network begin
+    @require_declaration
+    (k1, k2), A <--> B
+end
+catch e
+    println(e.msg)
```
TorkelE, line 286: Maybe it would just be easier to make it a
'''julia block
then no need to involve catch and try. Think that is what we have used previously.

docs/src/model_creation/dsl_advanced.md
```diff
@@ -272,6 +272,39 @@ sol = solve(oprob)
...
+    @require_declaration
+    @species A(t) B(t)
+    @parameters k1 k2
+    (k1, k2), A <--> B
+end
+```
+
+The following cases in which the DSL would normally infer variables will all throw errors if `@require_declaration` is set and the variables are not explicitly declared.
+- Inferring a species in a reaction, as in the example above
+- Inferring a parameter in a reaction rate expression, as in the reaction line `k*n, A --> B`
```
TorkelE, line 303: period at the end of these and following pulletpoint

docs/src/model_creation/dsl_advanced.md
```diff
@@ -272,6 +272,39 @@ sol = solve(oprob)
 plot(sol)
 ```
 
+### [Turning off species, parameter, and variable inferring](@id dsl_advanced_options_require_dec)
```
TorkelE, line 275: "Turning off species, parameter, and variable **inference**"

sounds better to me (but I am not actually a native speaker, so defer actual judgment to you guys)

docs/src/model_creation/dsl_advanced.md
```diff
@@ -272,6 +272,39 @@ sol = solve(oprob)
...
+In some cases it may be desirable for Catalyst to not infer species and parameters from the DSL, as in the case of reaction networks with very many variables, or as a sanity check that variable names are written correctly. To turn off inferring, simply add the `@require_declaration` macro to one of the lines of the `@reaction_network` declaration. Having this macro means that every single variable, species, or parameter will have to be explicitly declared using the `@variable`, `@species`, or `@parameter` macro. In the case that the DSL parser encounters an undeclared symbolic, it will error with an `UndeclaredSymbolicError` and print the reaction or equation that the undeclared symbolic was found in.
```
TorkelE, line 276: "To turn off inferring, simply use the `@require_declaration` option when using the `@reaction_network` DSL. This will require every variable, species, or parameter used within the DSL to be explicitly declared using the `@variable`, `@species`, or `@parameter` options."

Usage of "@..." within the DSL are not macro, and really don't have anything with them to do. It is simply the notation we have opted to use within the DSL to designate options (i.e. every input to the macro that is not a reaction line)

docs/src/model_creation/dsl_advanced.md
```diff
@@ -272,6 +272,39 @@ sol = solve(oprob)
...
+rn = @reaction_network begin
+    @require_declaration
+    @species A(t) B(t)
+    @parameters k1 k2
+    (k1, k2), A <--> B
+end
+```
+
+The following cases in which the DSL would normally infer variables will all throw errors if `@require_declaration` is set and the variables are not explicitly declared.
+- Inferring a species in a reaction, as in the example above
```
TorkelE, line 302: I would probably write "occurrence of an undeclared species in a reaction, as in the example above".  As the problem is not about not being able to infer, but rather that stuff is not declared. But maybe I am nitpicking, it is a very slight opinion.

PR #1133, opened 2024-11-22
Update spatial models for new MTK release

TorkelE:
Detecting hierarchical models is different now, so needed to update how this is checked. Since these now are flattened, _in principle_ it should work for spatial modelling. However, we really haven't tested it, so nothing I recommend. Still, I changed it to a warning for the new version.

### DIFF
--- a/src/spatial_reaction_systems/lattice_reaction_systems.jl
+++ b/src/spatial_reaction_systems/lattice_reaction_systems.jl
@@ -8,7 +8,7 @@ const GridLattice{N, T} = Union{Array{Bool, N}, CartesianGridRej{N, T}}
 """
 $(TYPEDEF)
 
-A representation of a spatial system of chemical reactions on a discrete (lattice) space. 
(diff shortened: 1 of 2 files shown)


src/spatial_reaction_systems/lattice_reaction_systems.jl
```diff
@@ -8,7 +8,7 @@ const GridLattice{N, T} = Union{Array{Bool, N}, CartesianGridRej{N, T}}
 """
 $(TYPEDEF)
 
-A representation of a spatial system of chemical reactions on a discrete (lattice) space.
```
TorkelE, line 11: Suddenly all my PRs have started to remove random spaces from any files I touch. Nothing bad, but does clog up the PR changes. Have seen it with other PRs earlier as well.

isaacsas: This is probably some VSCode setting you want to turn off.

isaacsas:
Feel free to merge if you are happy with this / done.

PR #1134, opened 2024-11-23
Add API functions for the other kinds of matrixes that a CRN ODE system can be factored into

### LINKED ISSUE #1132: Further decompostion of a system of ODEs coming from mass-action chemical reaction networks
As suggested [here](https://discourse.julialang.org/t/possibility-of-factorization-for-odes-produced-by-a-mass-action-chemical-reaction-network/122950/2?u=adhalanay) I am opening an issue.

**Is your feature request related to a problem? Please describe.**

In the algebraic approach to  mass action chemical reaction networks there is a very useful factorization of the RHS of the ODE system 
$$
\frac{dx}{dt}=Y\cdot A_k \cdot \Phi(x),
$$ 
where $Y$ is the complex stoichiometric matrix, $A_k=-L_k$ and $\Phi(x)$ is the vector of monomials. In order to define $L_k$ consider the complex graph associated to the CRN as a weighted directed graph with reaction rates as weights and then $L_k$ is its Laplacian matrix. 

While there is a function returning  $Y$ I don't think there are functions for the other two factors.
**Describe the solution you’d like**

It will be great to have exposed functions for the last factors: $A_k$ and $\Phi(x)$

**Describe alternatives you’ve considered**

For $A_k$ it will be maybe enough to have the weighted graph of complexes.

vyudu:
Specifically `laplacianmat`, `fluxmat`, and `massactionvector`. Sorry about the terrible looking diff, I just split a lot of the more chemical reaction network-related tests (detailed/complex balance, deficiency theorems, concentration robustness) into a new file crn_theory.jl and added the tests for this to the bottom of the `network_analysis.jl`. (Realize I should have done the file splitting in a separate PR but was in too deep already). Closes #1132.


src/network_analysis.jl
```diff
@@ -192,6 +192,122 @@ function complexstoichmat(::Type{Matrix{Int}}, rn::ReactionSystem, rcs)
     Z
 end
 
+"""
+    laplacianmat(rn::ReactionSystem, pmap=Dict(); sparse=false)
+
+    Return the negative of the graph Laplacian of the reaction network. The ODE system of a chemical reaction network can be factorized as ``\frac{dx}{dt} = Y A_k Φ(x)``, where Y is the [`complexstoichmat`](@ref) and A_k is the negative of the graph Laplacian, and Φ is the [`massactionvector`](@ref). A_k is an n-by-n matrix, where n is the number of complexes, where A_{ij} = k_{ij} if a reaction exists between the two complexes and 0 otherwise.
```
isaacsas, line 198: The math here needs to be wrapped appropriately so it displays as math.

I would also put the formula for `A_{ij}` as a standalone math block using cases rather than including it within the text. That latter is easier to read.

src/network_analysis.jl
```diff
@@ -192,6 +192,122 @@ function complexstoichmat(::Type{Matrix{Int}}, rn::ReactionSystem, rcs)
...
+"""
+function fluxmat(rn::ReactionSystem, pmap::Dict = Dict(); sparse=false) 
+    rates = reactionrates(rn)
+
+    if !isempty(pmap)
+        length(pmap) != length(parameters(rn)) && error("Incorrect number of parameters were specified.")
+        pmap = symmap_to_varmap(rn, pmap)
+        pmap = Dict(ModelingToolkit.value(k) => v for (k, v) in pmap)
+        rates = [substitute(rate, pmap) for rate in rates]
+    end
```
isaacsas, line 220: Can this code be extracted to one function that is then called in the various cases for species or parameters in these functions?

src/network_analysis.jl
```diff
@@ -192,6 +192,122 @@ function complexstoichmat(::Type{Matrix{Int}}, rn::ReactionSystem, rcs)
...
+
+    rcmap = reactioncomplexmap(rn); nc = length(rcmap); nr = length(rates)
+    mtype = eltype(rates) <: Symbolics.BasicSymbolic ? Num : eltype(rates)
+    K = if sparse
+        fluxmat(SparseMatrixCSC{mtype, Int}, rcmap, rates)
+    else
+        fluxmat(Matrix{mtype}, rcmap, rates)
+    end
```
isaacsas, line 228:
```suggestion
    if sparse
        return fluxmat(SparseMatrixCSC{mtype, Int}, rcmap, rates)
    else
        return fluxmat(Matrix{mtype}, rcmap, rates)
    end
```

src/network_analysis.jl
```diff
@@ -192,6 +192,122 @@ function complexstoichmat(::Type{Matrix{Int}}, rn::ReactionSystem, rcs)
...
+        specs = [substitute(s, scmap) for s in specs]
+    end
+
+    vtype = eltype(specs) <: Symbolics.BasicSymbolic ? Num : eltype(specs)
+    Φ = Vector{vtype}()
+    rcmap = reactioncomplexmap(rn)
+    for comp in keys(reactioncomplexmap(rn))
+        subs = map(ce -> getfield(ce, :speciesid), comp)
+        stoich = map(ce -> getfield(ce, :speciesstoich), comp)
+        push!(Φ, prod(vtype[specs[s]^α for (s, α) in zip(subs, stoich)]))
```
isaacsas, line 295: Isn't this resulting in a different rate law then Catalyst usually generates for ODEs? Where are the factorial factors getting put here? If not in the vector then they need to be in the rate `k` right?

Basically this decomposition should probably support the `combinatoric_ratelaws` keyword and have it `true` by default.

test/network_analysis/network_properties.jl
```diff
@@ -649,286 +320,103 @@ let
...
+    @test isequal(Φ[8], MAPK.KK_KKK_)
+    
+    K = fluxmat(MAPK)
+    @test isequal(K[1, 1], MAPK.k₁)
+    @test all(==(0), K[1, 2:end])
+    @test isequal(K[2, 2], MAPK.k₂) 
+    @test all(==(0), vcat(K[2,1], K[2,3:end]))
+    @test isequal(K[3, 2], MAPK.k₃)
+    @test all(==(0), vcat(K[3,1], K[3,3:end]))
+    @test count(k -> !isequal(k, 0), K) == length(reactions(MAPK))
```
isaacsas, line 364: Why not just hand make the matrix and test the two matrices are equal?

test/network_analysis/network_properties.jl
```diff
@@ -649,286 +320,103 @@ let
...
+        (k₂₉, k₃₀), KPP + KPase <--> KKPPKPase
     end
 
-    @test Catalyst.satisfiesdeficiencyone(rn) == false
-end
+    Φ = Catalyst.massactionvector(MAPK)
+    specs = species(MAPK)
+    @test isequal(Φ[1], MAPK.KKK * MAPK.E1)
+    @test isequal(Φ[6], MAPK.KKK * MAPK.E2)
+    @test isequal(Φ[8], MAPK.KK_KKK_)
```
isaacsas, line 355: Why not hand make the full vector and test the two vectors are equal (or `setequal`)?

test/network_analysis/network_properties.jl
```diff
@@ -649,286 +320,103 @@ let
...
-        k6, B + E --> A + C
-    end
-
-    # No longer weakly reversible
-    rn3 = @reaction_network begin
-        k1, A --> 2B
-        (k3, k4), A + C <--> D
-        k5, D --> B + E 
-        k6, B + E --> A + C
+        k1, 2X + Y + 3Z --> ∅
```
isaacsas, line 406: Add tests that the ODEs from the new decomposition are the same as what Catalyst generates in each case of `combinatoric_ratelaws`. I suspect right now they aren't the same.

isaacsas (changes requested):

src/network_analysis.jl
isaacsas: Similar comment for the other docstrings.

vyudu: What's the right way to wrap math? I was using the double backticks ``N`` since I saw it in other docstrings.

isaacsas: Not everything is wrapped here though, for example "A_{ij} = k_{ij}". Look at:

https://github.com/SciML/Catalyst.jl/blob/e3578276888167206dcc006b3a6ae69680718aec/src/network_analysis.jl#L71-L92

and note we needed to use "@doc raw" there to get it to render in the API section correctly.

vyudu:
@isaacsas i think this is done mod the Documentation build thing

test/network_analysis/network_properties.jl
```diff
@@ -649,286 +320,189 @@ let
...
+    crvec = [rn.X^2/2 * rn.Y * rn.Z^3/6,
+             1.,
+             rn.Y^2/2 * rn.Z^2/2,
+             rn.X^3/6]
+    @test isequal(Φ, crvec)
+    ncrvec = [rn.X^2 * rn.Y * rn.Z^3,
+              1.,
+              rn.Y^2 * rn.Z^2,
+              rn.X^3]
+    Φ_2 = Catalyst.massactionvector(rn; combinatoric_ratelaws = false)
```
isaacsas, line 453: OK, but then shouldn't the new API functions also take this kwarg and be consistent with the Catalyst convention? It seems weird for them to give the `combinatoric_ratelaws = false` case when everything else in Catalyst defaults to the `true` case. (They should also respect the system's internal value for this if set by a user.)

test/network_analysis/network_properties.jl
```diff
@@ -649,286 +320,189 @@ let
...
+    @test isequal(Φ_2, ncrvec)
```
isaacsas, line 454: Should these kind of tests be `issetequal`? We don't generally want to make assumptions about ordering.

test/network_analysis/network_properties.jl
```diff
@@ -649,286 +320,189 @@ let
...
-    end
-
-    rcs, D = reactioncomplexes(rn)
-    @test Catalyst.edgeindex(D, 1, 2) == 1
-    @test Catalyst.edgeindex(D, 4, 3) == 6
+        k1, 2X + Y + 3Z --> ∅
+        (k2, k3), 2Y + 2Z <--> 3X
+    end
+
+    Φ = Catalyst.massactionvector(rn); specs = species(rn)
```
isaacsas, line 443:
```suggestion
    Φ = Catalyst.massactionvector(rn)
    specs = species(rn)
```
I'm going to stop commenting on this, but please fix all the places you have multiple assignments on one line like this.

isaacsas (changes requested):

src/network_analysis.jl
```diff
@@ -192,6 +192,122 @@ function complexstoichmat(::Type{Matrix{Int}}, rn::ReactionSystem, rcs)
...
+
+    Return an r×c matrix ``K`` such that, if complex ``j`` is the substrate complex of reaction ``i``, then ``K_{ij} = k``, the rate constant for this reaction. Mostly a helper function for the network Laplacian, [`laplacianmat`](@ref). Has the useful property that ``\frac{dx}{dt} = S*K*Φ(x)``, where S is the [`netstoichmat`](@ref) or net stoichiometry matrix and ``Φ(x)`` is the [`massactionvector`](@ref).
+    Returns a symbolic matrix by default, but will return a numerical matrix if rate constants are specified as a `Tuple`, `Vector`, or `Dict` of symbol-value pairs via `pmap`.
+"""
+function fluxmat(rn::ReactionSystem, pmap::Dict = Dict(); sparse=false) 
+    rates = reactionrates(rn)
+
+    !isempty(pmap) && (rates = substitutevals(rn, pmap, parameters(rn), rates))
+    rcmap = reactioncomplexmap(rn); nc = length(rcmap); nr = length(rates)
+    mtype = eltype(rates) <: Symbolics.BasicSymbolic ? Num : eltype(rates)
```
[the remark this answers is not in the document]
vyudu: I think it has to, doesn't seem like it's actually possible to have zeros in this matrix if the eltype is `BasicSymbolic`

test/network_analysis/network_properties.jl
vyudu: Thinking about this more I think we have to check the order for the output to the mass action vector, otherwise building the ODEs with something like (Y*K*\Phi) will not yield correct results. In general the set of ODEs has to be in the iteration order of `species(rn)`.

src/network_analysis.jl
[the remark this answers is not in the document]
isaacsas: Can't it by type `Any`?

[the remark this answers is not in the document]
vyudu: Is that better than being `Num`? I assumed more specific was better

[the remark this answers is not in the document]
isaacsas: We don't want to return mixtures of Nums and non-Nums across different methods, so we should not re-wrap internal symbolics.

isaacsas (changes requested):
@vyudu just change the returning of `Num`s, and then if tests pass you are welcome to merge.

isaacsas:
It looks like right now tests are failing?

src/network_analysis.jl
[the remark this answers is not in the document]
vyudu: Could I just wrap it in an `Any` when I return it? It's useful to have it be `Num` at least temporarily because I use `zero(T)` and `one(T)` in various places (and product over an empty `Num[]` returns 1, which is useful).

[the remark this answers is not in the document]
vyudu: Actually this creates some issues for the sparse method (can't really do any arithmetic with a `SparseMatrixCSC{Any, T}` since zero(Any) is undefined). Would `Union{Int64, BasicSymbolic}` be okay? If not it might just be better to define two functions, one symbolic and one numeric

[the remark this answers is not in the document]
isaacsas: If you want to internally use Num I guess that is ok, but yeah, we should unwrap before returning to users for consistency.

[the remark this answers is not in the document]
vyudu: What about returning a `Union` as the element type?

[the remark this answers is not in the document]
vyudu: Actually that doesn't work, nevermind, when you find the Laplacian matrix (D*K) type inference infers its eltype as Any anyway. Since it's the sparse matrix support that's causing the issue I think it might just be better to not allow the sparse kwarg for now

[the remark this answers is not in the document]
isaacsas: Alright, I guess just stick with `Num`. But please add documentation about this and why it is needed to the function doctrings.

test/runtests.jl
```diff
@@ -42,13 +42,14 @@ end
...
         @time @safetestset "Network Properties" begin include("network_analysis/network_properties.jl") end
+        @time @safetestset "CRN Theory" begin include("network_analysis/crn_theory.jl") end
 
         # Tests ODE, SDE, jump simulations, nonlinear solving, and steady state simulations.
         @time @safetestset "ODE System Simulations" begin include("simulation_and_solving/simulate_ODEs.jl") end
         @time @safetestset "Automatic Jacobian Construction" begin include("simulation_and_solving/jacobian_construction.jl") end
         @time @safetestset "SDE System Simulations" begin include("simulation_and_solving/simulate_SDEs.jl") end
         @time @safetestset "Jump System Simulations" begin include("simulation_and_solving/simulate_jumps.jl") end
-        @time @safetestset "Nonlinear and SteadyState System Solving" begin include("simulation_and_solving/solve_nonlinear.jl") end
+        # @time @safetestset "Nonlinear and SteadyState System Solving" begin include("simulation_and_solving/solve_nonlinear.jl") end
```
isaacsas, line 52:
```suggestion
        @time @safetestset "Nonlinear and SteadyState System Solving" begin include("simulation_and_solving/solve_nonlinear.jl") end
```

isaacsas:
@vyudu is this good to go now? If so, please merge before something else breaks and we have to cycle through updates of master again.

vyudu:
done

PR #1145, opened 2024-12-17
Fix serialisation tests

TorkelE:
There is still an MTK issue we need fixed (https://github.com/SciML/ModelingToolkit.jl/issues/3275), but it seems like a clear bug, so have just references it and marked the corresponding test as broken.


isaacsas:
@tork this is fine with me to merge. The doc failure is unrelated.

PR #1148, opened 2024-12-17
Add errors for array species/parameters when submitted to `LatticeReactionSystem`s

TorkelE:
Long-term we should probably support it. However, doing so will be a decent chunk of work, and rather low priority right now.


isaacsas:
I think we need to drop GLMakie in docs and tests. It just is too unreliable.

isaacsas:
@jonathanfischer97 we are getting GLMakie issues again. Any suggestions?

isaacsas (approved):

isaacsas:
Feel free to merge as the doc failures are unrelated.

jonathanfischer97:
Sorry for missing this guys. Indeed, it seems that GFLW load issue just won't go away, shame. 

I'll look into it when I get some time, but let's just drop it for now until I find a fix or GLMakie itself is more mature.

PR #1149, opened 2024-12-18
Update Optimization/parameter fitting tutorial

TorkelE:
Removes DiffEqParamEstim from the tutorial. Since we now create a full cost function here, I simply introduce all relevant concepts directly in this tutorial (instead as before, having some in the behaviour optimisation tutorial, which hence have seen a decent amount of change as well).


docs/pages.jl
```diff
@@ -45,8 +45,8 @@ pages = Any[
         "steady_state_functionality/dynamical_systems.md"
     ],
     "Inverse problems" => Any[
-        "inverse_problems/optimization_ode_param_fitting.md",
         "inverse_problems/petab_ode_param_fitting.md",
```
TorkelE, line 48: This feels like the simpler to use and more reliable option, so figured it might make sense to put first. Then if someone cannot use it they can proceed to the next option.

docs/src/inverse_problems/optimization_ode_param_fitting.md
```diff
@@ -24,7 +28,7 @@ ps_true = [:kB => 1.0, :kD => 0.1, :kP => 0.5]
 using OrdinaryDiffEqDefault
 oprob_true = ODEProblem(rn, u0, (0.0, 10.0), ps_true)
 true_sol = solve(oprob_true)
-data_sol = solve(oprob_true; saveat=1.0)
+data_sol = solve(oprob_true; saveat = 1.0)
```
TorkelE, line 31: I think when I wrote these I didn't use to put spaces around `=` in kwargs, but that should be changed to conform to standard. At some point I will have a look at more tutorials to see if there are other places I don't use the SciML standard.

isaacsas:
@TorkelE is this ok to merge?

Issue #1096, opened 2024-10-28
drop OrdinaryDiffEq CI 


isaacsas:
We should just switch to the actual sub-libraries we need to reduce Github action times. (Same for docs.)

ChrisRackauckas:
Yes

isaacsas:
Completed now for docs, just need to do for CI too.

PR #1150, opened 2024-12-18
Remove last OrdinaryDiffEq mention

TorkelE:
Was going to finish this one off https://github.com/SciML/Catalyst.jl/issues/1096, but @vyudu already did. Did the one case remaining in a docstring though. Also, there is this from the `JumpInputs` docstring:
```julia
using Catalyst, OrdinaryDiffEq, JumpProcesses, Plots
rn = @reaction_network begin
    k*(1 + sin(t)), 0 --> A
end
jinput = JumpInputs(rn, [:A => 0], (0.0, 10.0), [:k => .5])
@assert jinput.prob isa ODEProblem
jprob = JumpProblem(jinput)
sol = solve(jprob, Tsit5())
plot(sol, idxs = :A)

rn = @reaction_network begin
    k, 0 --> A
end
jinput = JumpInputs(rn, [:A => 0], (0.0, 10.0), [:k => .5])
@assert jinput.prob isa DiscreteProblem
jprob = JumpProblem(jinput)
sol = solve(jprob)
plot(sol, idxs = :A)
```
Could this be changed from `OrdinaryDiffEq` to `OrdinaryDiffEqTsit5` @isaacsas?

### DIFF
--- a/src/spatial_reaction_systems/lattice_reaction_systems.jl
+++ b/src/spatial_reaction_systems/lattice_reaction_systems.jl
@@ -27,7 +27,7 @@ are possible.
 Example:
 ```julia
 # Fetch packages.
-using Catalyst, OrdinaryDiffEq
+using Catalyst, OrdinaryDiffEqDefault
 import CairoMakie
 
 # Creates the `LatticeReactionSystem` model.
--- a/src/spatial_reaction_systems/lattice_reaction_systems.jl
+++ b/src/spatial_reaction_systems/lattice_reaction_systems.jl
@@ -27,7 +27,7 @@ are possible.
 Example:
 ```julia
 # Fetch packages.
-using Catalyst, OrdinaryDiffEqTsit5
+using Catalyst, OrdinaryDiffEqDefault
 import CairoMakie
 
 # Creates the `LatticeReactionSystem` model.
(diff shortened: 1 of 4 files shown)


isaacsas:
@TorkelE yes, that could probably be changed. Trying running it and if it works go ahead and update it too.

isaacsas:
Let's see if tests are still failing.

src/spatial_reaction_systems/lattice_reaction_systems.jl
```diff
@@ -43,7 +43,7 @@ u0 = [:X => rand(5,5)]
 tspan = (0.0, 1.0)
 ps = [:p => 1.0, :d => 0.5, :D => 0.1]
 oprob = ODEProblem(lrs, u0, tspan, ps)
-osol = solve(oprob)
+osol = solve(oprob, Tsit5())
```
isaacsas, line 46: Are you sure Tsit5 is a good recommendation for a spatial system? I would think you'd want an implicit solver often?

TorkelE:
Good point actually. Not sure at all to be honest. Would the sensible be to just do some tests with various solvers and see what is fastest?

isaacsas:
> Good point actually. Not sure at all to be honest. Would the sensible be to just do some tests with various solvers and see what is fastest?

Yes, I'd think so.

ChrisRackauckas:
Or at least ROCK

isaacsas:
I think if you update to master this should hopefully now build.

TorkelE:
I did a couple of benchmarks. `Tsit5` seems fastest, but default seems to be even better. Also not necessary to specify a solver here and I think going by the default is better in sections not really about solvers anyway. Switching to that.

```julia
@btime solve(oprob)
@btime solve(oprob, ROCK2())
@btime solve(oprob, ROCK4())
@btime solve(oprob, Rodas5P())
@btime solve(oprob, Rosenbrock23())
@btime solve(oprob, Vern7())
@btime solve(oprob, Tsit5())
```

PR #1153, opened 2024-12-19
MTK tests update

TorkelE:
Went through the MTK tests, fixed some things. Unfortunatelye everything major that was broken still is.

### DIFF
--- a/test/reactionsystem_core/coupled_equation_crn_systems.jl
+++ b/test/reactionsystem_core/coupled_equation_crn_systems.jl
@@ -228,7 +228,7 @@ let
         Reaction(d, [X], []),
         Reaction(d, [X], nothing, [2], nothing),
         D(V) ~ X - v*V,
-        W^2 ~ log(V) + X 
+        W^2 ~ log(V) + X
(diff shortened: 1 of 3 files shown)


isaacsas:
Feel free to merge if you are done and tests pass.

It might be good more generally to go through your various open issues on indexing and remake and collate them all into one, new, big issue in MTK? (Closing the old ones.) Just to give a bump on all this and to summarize where things stand today since that may not be clear to everyone.

ChrisRackauckas:
Yes. My understanding is that everything is done, so I would expect most of your issues close? It looks like we missed one symbolic+array part that you just highlighted, but hopefully that's the only part that slipped.

TorkelE:
Yeah, I think I will need to do another write-up with summaries. I think there might have been some issues that was partially solved, but some follow-up sub cases wasn't.

TorkelE:
The test failure seems to be some form of regression on master. Present in other PRs that only affect docs as well.

isaacsas:
@TorkelE should we merge this?

PR #1159, opened 2025-01-08
Test fix

TorkelE:
Turns out that MTK now gives over/under determined system warnings for certain cases when SDEProblems are generated. What is slightly odd is that this is not the case for the corresponding ODEs (does this make sense @AayushSabharwal ?).

We had some tests which depended on checking warnings for when conserved quantities are eliminated. These checks no longer worked when MTK threw additional warnings. I have removed the MTK warnings. This makes tests pass, but we might want to check why these warnings are actually generated.


TorkelE:
Another failure has appeared, apparently affecting `remake` for SDEs (ODEs work fine). This is the example:
```julia
using Catalyst, StochasticDiffEq
model = @reaction_network begin
    @observables XY ~ X + Y
    (kp,kd), 0 <--> X
    (k1,k2), X <--> Y
end
@unpack XY, X, Y, kp, kd, k1, k2 = model

# Creates problem.
u0_vals = [X => 4, Y => 5]
tspan = (0.0, 10.0)
p_vals = [kp => 1.0, kd => 0.1, k1 => 0.25, k2 => 0.5]
sprob = SDEProblem(model,u0_vals, tspan, p_vals)

# tests remake.
rp = remake(sprob; u0 = [Y => 7])
rp[[X, Y]] == [4, 7] # false (should be true).
rp[X] # -7.0 (!, should be 4.0, X is never changed)
rp[Y] # 7.0 (correct)

oprob = ODEProblem(model,u0_vals, tspan, p_vals)
rp = remake(oprob; u0 = [Y => 7])
rp[X] # 4.0, correct.
```

I will raise with MTK

TorkelE:
A different case, where we cannot even call `remake`
```julia
using Catalyst, StochasticDiffEq
model = @reaction_network begin
    (p1,2.0), 0 <--> X
    (p2,2.0), 0 <--> Y
end
@unpack X, Y, p1, p2 = model

# Creates problem.
u0_vals = [X => 4, Y => 5]
tspan = (0.0, 10.0)
p_vals = [p1 => 1.0, p2 => 0.1]
sprob = SDEProblem(model,u0_vals, tspan, p_vals)

# tests remake.
rp = remake(sprob; u0 = [Y => 7] # ERROR: Initialization incomplete. Not all of the state variables of the DAE system can be determined by the initialization. 
```

TorkelE:
MTK issue: https://github.com/SciML/ModelingToolkit.jl/issues/3295

TorkelE:
Another one! Turns out that parameter defaults are broken for `DiscreteProblem`s (but not other problem types:
```julia
using Catalyst
model = @reaction_network begin
    @parameters Y0
    @species Y(t) = Y0
    (k1,k2), X <--> Y
end
@unpack X, Y, Y0, k1, k2 = model

u0 = [X => 4]
tspan = (0.0, 10.0)
p = [k1 => 0.25, k2 => 10, Y0 => 2]

oprob = ODEProblem(model, u0, tspan, p) # Works.
sprob = SDEProblem(model, u0, tspan, p) # Works.
dprob = DiscreteProblem(model, u0, tspan, p) # Errors.
```

ChrisRackauckas:
What error does that spit out?

TorkelE:
```
ERROR: MethodError: no method matching unhack_observed(::Vector{Equation}, ::RecursiveArrayTools.ArrayPartition{Any, Tuple{…}})
The function `unhack_observed` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  unhack_observed(::Vector{Equation}, ::Vector{Equation})
   @ ModelingToolkit ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/nonlinear/initializesystem.jl:352

Stacktrace:
  [1] generate_initializesystem(sys::JumpSystem{…}; u0map::Dict{…}, pmap::Dict{…}, initialization_eqs::Vector{…}, guesses::Dict{…}, default_dd_guess::Float64, algebraic_only::Bool, check_units::Bool, check_defguess::Bool, name::Symbol, extra_metadata::@NamedTuple{…}, kwargs::@Kwargs{})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/nonlinear/initializesystem.jl:17
  [2] generate_initializesystem
    @ ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/nonlinear/initializesystem.jl:6 [inlined]
  [3] ModelingToolkit.InitializationProblem{…}(sys::JumpSystem{…}, t::Float64, u0map::Dict{…}, parammap::Dict{…}; guesses::Dict{…}, check_length::Bool, warn_initialize_determined::Bool, initialization_eqs::Vector{…}, fully_determined::Nothing, check_units::Bool, use_scc::Bool, kwargs::@Kwargs{…})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/diffeqs/abstractodesystem.jl:1307
  [4] InitializationProblem
    @ ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/diffeqs/abstractodesystem.jl:1285 [inlined]
  [5] #_#1075
    @ ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/diffeqs/abstractodesystem.jl:1263 [inlined]
  [6] InitializationProblem
    @ ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/diffeqs/abstractodesystem.jl:1262 [inlined]
  [7] #InitializationProblem#1073
    @ ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/diffeqs/abstractodesystem.jl:1251 [inlined]
  [8] InitializationProblem
    @ ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/diffeqs/abstractodesystem.jl:1250 [inlined]
  [9] maybe_build_initialization_problem(sys::JumpSystem{…}, op::Dict{…}, u0map::Dict{…}, pmap::Dict{…}, t::Float64, defs::Dict{…}, guesses::Dict{…}, missing_unknowns::Set{…}; implicit_dae::Bool, kwargs::@Kwargs{…})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/problem_utils.jl:546
 [10] maybe_build_initialization_problem
    @ ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/problem_utils.jl:526 [inlined]
 [11] process_SciMLProblem(constructor::Type, sys::JumpSystem{…}, u0map::Vector{…}, pmap::Vector{…}; build_initializeprob::Bool, implicit_dae::Bool, t::Float64, guesses::Dict{…}, warn_initialize_determined::Bool, initialization_eqs::Vector{…}, eval_expression::Bool, eval_module::Module, fully_determined::Nothing, check_initialization_units::Bool, tofloat::Bool, use_union::Bool, u0_constructor::typeof(identity), du0map::Nothing, check_length::Bool, symbolic_u0::Bool, warn_cyclic_dependency::Bool, circular_dependency_max_cycle_length::Int64, circular_dependency_max_cycles::Int64, substitution_limit::Int64, use_scc::Bool, kwargs::@Kwargs{})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/problem_utils.jl:669
 [12] DiscreteProblem(sys::JumpSystem{…}, u0map::Vector{…}, tspan::Tuple{…}, parammap::Vector{…}; use_union::Bool, eval_expression::Bool, eval_module::Module, kwargs::@Kwargs{})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/jumps/jumpsystem.jl:428
 [13] DiscreteProblem(sys::JumpSystem{…}, u0map::Vector{…}, tspan::Tuple{…}, parammap::Vector{…})
    @ ModelingToolkit ~/.julia/packages/ModelingToolkit/iQ7So/src/systems/jumps/jumpsystem.jl:414
 [14] DiscreteProblem(::ReactionSystem{…}, ::Vector{…}, ::Tuple{…}, ::Vector{…}; name::Symbol, combinatoric_ratelaws::Bool, checks::Bool, kwargs::@Kwargs{})
    @ Catalyst ~/.julia/dev/Catalyst/src/reactionsystem_conversions.jl:855
 [15] DiscreteProblem(::ReactionSystem{…}, ::Vector{…}, ::Tuple{…}, ::Vector{…})
    @ Catalyst ~/.julia/dev/Catalyst/src/reactionsystem_conversions.jl:846
 [16] top-level scope
    @ ~/Desktop/Julia Playground/Environment - Catalyst only/catalyst_playground.jl:57
Some type information was truncated. Use `show(err)` to see complete types.
```

TorkelE:
New error, but this one seems codecov related
```
Run codecov/codecov-action@v5
Run CC_ACTION_VERSION=$(cat ${GITHUB_ACTION_PATH}/src/version)
==> Running Action version 5.1.2
Run git config --global --add safe.directory /home/runner/work/Catalyst.jl/Catalyst.jl
Run CC_FORK="false"
Run if [ "false" == 'true' ] && [ "$CC_FORK" != 'true' ];
==> Token set from input
Run if [ -z "$CC_BRANCH" ] && [ -z "$CC_TOKEN" ] && [ "$CC_FORK" == 'true' ]
Run if [ -z "$CC_SHA" ];
Run ${GITHUB_ACTION_PATH}/dist/codecov.sh
     _____          _
    / ____|        | |
   | |     ___   __| | ___  ___ _____   __
   | |    / _ \ / _` |/ _ \/ __/ _ \ \ / /
   | |___| (_) | (_| |  __/ (_| (_) \ V /
    \_____\___/ \__,_|\___|\___\___/ \_/
                            Wrapper-0.0.31
                                  
==> Detected linux
 -> Downloading https://cli.codecov.io/latest/linux/codecov
==> Finishing downloading linux:latest
      Version: v9.1.1
 
gpg: directory '/home/runner/.gnupg' created
gpg: keybox '/home/runner/.gnupg/pubring.kbx' created
gpg: /home/runner/.gnupg/trustdb.gpg: trustdb created
gpg: key 806BB28AED779869: public key "Codecov Uploader (Codecov Uploader Verification Key) <security@codecov.io>" imported
gpg: Total number processed: 1
gpg:               imported: 1
==> Verifying GPG signature integrity
 -> Downloading https://cli.codecov.io/latest/linux/codecov.SHA256SUM
 -> Downloading https://cli.codecov.io/latest/linux/codecov.SHA256SUM.sig
 
gpg: Signature made Sat Dec  7 16:07:53 [20](https://github.com/SciML/Catalyst.jl/actions/runs/12675329977/job/35326113772#step:8:21)24 UTC
gpg:                using RSA key 27034E7FDB850E0BBC2C62FF806BB28AED779869
gpg: Good signature from "Codecov Uploader (Codecov Uploader Verification Key) <security@codecov.io>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: [27](https://github.com/SciML/Catalyst.jl/actions/runs/12675329977/job/35326113772#step:8:29)03 4E7F DB85 0E0B BC2C  62FF 806B B28A ED77 9869
codecov: FAILED
sha256sum: WARNING: 1 computed checksum did NOT match
codecov: FAILED
==> Could not verify SHASUM. Please contact Codecov if problem continues
    Exiting...
Error: Process completed with exit code 1.
```

isaacsas:
What's the new HC doc build failure?

TorkelE:
Yeah, the symbolic stoichiometry tests are broken now. Which is funny as these are before the MTK tests that were failing (and fixes), so we shouldn't even have gotten there if those weren't working. I tried to reproduce locally, but worked fine. Pkg marked everything as up to date, but maybe there has been a very recent release which CI pulls but my local Pkg is unaware of.

ChrisRackauckas:
You're not looking at the generated code right? The only release of today is a fix to change ^ to NaNMath.pow, but that shouldn't break anything except tests that expect ^ specifically

TorkelE:
Not sure really. I have a bunch of other stuff to sort out, and there is already another bunch of tests failing. Will have a look at this again tomorrow, if I cannot reproduce locally then I will have to look into it deeper.

TorkelE:
Cannot see any HC problems, but what is causing issues is stuff like
```
MethodError: no method matching factorial(::Float64)
```
and
```
 MethodError: no method matching binomial(::Float64, ::Float64)
```

I think we had some problem like this long ago, but should have been sorted by now.

isaacsas:
I'll make another PR later today to cap MTK at the previous release to see if that fixes stuff (and hopefully allow us to work on other PRs).

isaacsas:
Sounds like jump initial conditions or integer parameters are now getting promoted to floating point?

TorkelE:
Sounds like a good guess, hopefully we'll know for sure soon.

isaacsas:
We don't do anything special like that in Catalyst.

AayushSabharwal:
For the SDE `remake` bug:
```julia
using Catalyst, StochasticDiffEq
model = @reaction_network begin
    @observables XY ~ X + Y
    (kp,kd), 0 <--> X
    (k1,k2), X <--> Y
end
@unpack XY, X, Y, kp, kd, k1, k2 = model

# Creates problem.
u0_vals = [X => 4, Y => 5]
tspan = (0.0, 10.0)
p_vals = [kp => 1.0, kd => 0.1, k1 => 0.25, k2 => 0.5]
sprob = SDEProblem(model,u0_vals, tspan, p_vals)

# tests remake.
rp = remake(sprob; u0 = [Y => 7])
rp[[X, Y]] == [4, 7] # false (should be true).
rp[X] # -7.0 (!, should be 4.0, X is never changed)
rp[Y] # 7.0 (correct)
```

I don't yet know how, but this is the equations in the initialization system before `structural_simplify`:
```julia
 Y(t) ~ 7
 XY(t) ~ Y(t) + X(t)
```

And after:
```julia
 Y(t) ~ 7.0
 X(t) ~ -0.0
 XY(t) ~ Y(t) + X(t)
```

Somehow it's adding `X ~ 0`

ChrisRackauckas:
Can we turn the alias elimination issue into an MTK issue to track?

TorkelE:
> Can we turn the alias elimination issue into an MTK issue to track?

https://github.com/SciML/ModelingToolkit.jl/issues/3302

TorkelE:
Ok, there's the issue ou mentioned again Sam. Still cannot reproduce locally even when I think I am on the same environment, so might take a bit to figure out.

Project.toml
```diff
@@ -58,7 +58,7 @@ JumpProcesses = "9.13.2"
 LaTeXStrings = "1.3.0"
 Latexify = "0.16.5"
 MacroTools = "0.5.5"
-ModelingToolkit = "< 9.60"
+ModelingToolkit = "< 9.60, =9.60"
```
isaacsas, line 61:
```suggestion
ModelingToolkit = "9.60"
```

Project.toml
```diff
@@ -67,7 +67,7 @@ RuntimeGeneratedFunctions = "0.5.12"
 SciMLBase = "2.57.2"
 Setfield = "1"
 # StructuralIdentifiability = "0.5.8"
-SymbolicUtils = "< 3.8"
+SymbolicUtils = "< 3.8.1, = 3.8.1"
```
isaacsas, line 70:
```suggestion
SymbolicUtils = "3.8.1"
```

docs/Project.toml
```diff
@@ -61,7 +61,7 @@ IncompleteLU = "0.2"
 JumpProcesses = "9.13.2"
 Latexify = "0.16.5"
 LinearSolve = "2.30"
-ModelingToolkit = "< 9.60"
+ModelingToolkit = "< 9.60, =9.60"
```
isaacsas, line 64:
```suggestion
ModelingToolkit = "9.60"
```

isaacsas (changes requested):
Isn't the point of this PR to have a place to test against the latest versions of everything?

Also, what are you referring to by the problem that is reappearing? Tests were passing?

TorkelE:
I tried to mark the latest version (which we are testing against). Figured `ModelingToolkit = "9.60"` didn't make sense, as we aren't sure that versions beyond 9.60 actually work. Since 9.60 is the highest MTK version we have tested against, we shouldn't make u as compatible with later versions until that is actually tested. Wasn't this what we wanted to do with the capping?

isaacsas:
No, the goal is to not have master be capped. Otherwise we won't see when things break. The point is to cap releases so they can't end up getting versions of MTK/SciMLBase that don't work with them.

TorkelE:
Sounds good

TorkelE:
This was the problem you were talking about: https://github.com/SciML/Catalyst.jl/issues/1170#issuecomment-2580209002, but seems to dealt with it (for now)already)

isaacsas:
I'm not seeing that it is fixed? I'm getting errors about it as of my most recent PR today.

TorkelE:
No, it is not fixed. My commented was that the problem is still there, but I could not reproduce locally so might not be able to make an immediate fix. Then I referenced https://github.com/SciML/Catalyst.jl/issues/1170 meaning that you were already taking care of it (as it commenting it out for now, although I referenced the issue, not the PR).

isaacsas:
Could it be you are using different package versions than the doc builds?

TorkelE:
Something is definitely off. I have some other stuff I really need to sort out asap, but will sit down to have a closer look at it.

TorkelE:
Ok, this one is ready now

isaacsas:
I am a bit hesitant about part of the approach in this PR. By just setting `@test_broken` I feel like we will now just forget about these issues and go on with them potentially not getting fixed at this time.  

I think I'd prefer to keep MTK capped on master until such a time as these get fixed in MTK, as that keeps us aware that these bugs have not been fixed in MTK (i.e. if we try uncapping MTK we will presumably get these test failures again, so we will see these are not fixed). 

If there are tests where there are bugs / incorrect MTK usage on our end, we should certainly fix and merge those. But things that are supposed to work but are now broken should probably not be ignored via `test_broken`.

Aside: the strategy of setting `@test_broken false` is also problematic since that will never indicate to us that an issue has been fixed (it really isn't any different than just commenting out the test completely). It would be better to actually keep the broken test tested so we can see that it has been fixed (as `@test_broken` will then start failing).

TorkelE:
Don't minding keeping this open for now. I guess the capping to some extent replaces `@test_broken`. 

I think the `@test_broken false` have a (limited) role though. Yes, it is basically commenting out the test, however, since my main way of figuring out what is broken is doing a search through the package on `@test_broken", I catch it (and a reference to the relevant issue) at that time. Note that I am only using it in situations where `@test_broken statement` is not possible due to loops where `statement` is sometimes `true` and sometimes `false`. Having issues is also good, but by ensuring that whenever something is broken there is a `@test_broken` statement I can keep track of it (and it also shows up in the test sets that there is something going on). The alternative is just scratching those statements, which does not seem like an improvement.

isaacsas:
@TorkelE I guess if you want to merge master into this we can merge this when tests pass. But we need to get all these open MTK issues that are blocking V15 collated in one Catalyst issue so we can more easily track their state and know whether they are fixed.

TorkelE:
I have no strong opinions really. Might as well see if we can get those others fixed first. If later MTK updates got useful changes and there seem to be no progress maybe we should. But I will give it a bit longer.

TorkelE:
Superseeded by changes in https://github.com/SciML/Catalyst.jl/pull/1186.

PR #1160, opened 2025-01-08
Create DocsCleanup.yml

andreasnoack:
To reduce the size of the `gh-pages` branch. It's currently huge (~2.8GB). See https://documenter.juliadocs.org/dev/man/hosting/#Cleaning-up-gh-pages


isaacsas:
Thanks!

PR #1162, opened 2025-01-08
Remove old previews

andreasnoack:
Follow-up to #1160. The action only deletes the preview of the PR being closed so it is necessary to manually delete all the old previews. Initially, this will keep them in the git history but once the clean-up action has run after this PR has been merged then the repo size should much smaller.


isaacsas:
Thanks for helping with this!

PR #1163, opened 2025-01-08
Cap mtk

isaacsas:


isaacsas:
@vyudu @TorkelE this should fix the test and doc failures if you want to merge master into your PRs that have gotten blocked. We can keep @TorkelE’s PR with fixes as the place to uncap once this is all resolved.

Project.toml
```diff
@@ -67,7 +67,7 @@ RuntimeGeneratedFunctions = "0.5.12"
 SciMLBase = "2.57.2"
 Setfield = "1"
 # StructuralIdentifiability = "0.5.8"
-SymbolicUtils = "2.1.2, 3.3.0"
+SymbolicUtils = "< 3.8"
```
ChrisRackauckas, line 70: This one is incorrect? It doesn't do anything about the fact that it was a patch to NaNMath that was already released?

TorkelE:
Once this gets established, it might make sens to list capped packaged (and where they are capped) in https://docs.sciml.ai/Catalyst/dev/devdocs/dev_guide/ somewhere. Just for an easy checklist which ones need checking an updating?

Project.toml
isaacsas: This was merged before any fix was released or made (and while the fix was being put in originally as a SymbolicUtils PR).

I'll remove it in https://github.com/SciML/Catalyst.jl/pull/1168 if tests are now passing.

PR #1172, opened 2025-01-10
network_analysis cleanup: Updating `ratematrix` to be more similar to new matrix API functions

vyudu:
- Rename `ratematrix` to `adjacencymat`, change its implementation to be more similar to the new API matrix functions in #1134 
- Use the new network analysis API functions in the `iscomplexbalanced`

### DIFF
--- a/src/network_analysis.jl
+++ b/src/network_analysis.jl
@@ -206,9 +206,6 @@ function laplacianmat(rn::ReactionSystem, pmap::Dict = Dict(); sparse = false)
     D * K
 end
 
-Base.zero(::Type{Union{R, Symbolics.BasicSymbolic{Real}}}) where R <: Real = zero(R)
-Base.one(::Type{Union{R, Symbolics.BasicSymbolic{Real}}}) where R <: Real = one(R)
-
 @doc raw"""
     fluxmat(rn::ReactionSystem, pmap = Dict(); sparse=false)
 
@@ -225,8 +222,6 @@ function fluxmat(rn::ReactionSystem, pmap::Dict = Dict(); sparse=false)
     end
 
     rcmap = reactioncomplexmap(rn)
-    nc = length(rcmap)
-    nr = length(rates)
     mtype = eltype(rates) <: Symbolics.BasicSymbolic ? Num : eltype(rates)
     if sparse
         return fluxmat(SparseMatrixCSC{mtype, Int}, rcmap, rates)
@@ -936,7 +931,7 @@ function isdetailedbalanced(rs::ReactionSystem, parametermap::Dict; abstol=0, re
     complexes, D = reactioncomplexes(rs)
     img = incidencematgraph(rs)
     undir_img = SimpleGraph(incidencematgraph(rs))
-    K = ratematrix(rs, pmap)
+    K = adjacencymat(rs, pmap)
 
     spanning_forest = Graphs.kruskal_mst(undir_img)
     outofforest_edges = setdiff(collect(edges(undir_img)), spanning_forest)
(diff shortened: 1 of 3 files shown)


src/network_analysis.jl
```diff
@@ -206,9 +206,6 @@ function laplacianmat(rn::ReactionSystem, pmap::Dict = Dict(); sparse = false)
     D * K
 end
 
-Base.zero(::Type{Union{R, Symbolics.BasicSymbolic{Real}}}) where R <: Real = zero(R)
-Base.one(::Type{Union{R, Symbolics.BasicSymbolic{Real}}}) where R <: Real = one(R)
```
isaacsas, line 210: Oh wow, I had missed that was in a earlier PR. Yes, we definitely shouldn't add dispatches like that (that is the definition of type piracy).

src/network_analysis.jl
```diff
@@ -993,52 +988,26 @@ function isdetailedbalanced(rs::ReactionSystem, parametermap)
...
-            rxn = rxns[r]
-            if dir == -1
-                K[r, c] = rates[r]
-            end
-        end
-    end
-
-    L = -D * K
-    S = netstoichmat(rs)
+    L = laplacianmat(rs, parametermap; sparse)
```
isaacsas, line 1004: Is `L` ever used subsequently?

src/network_analysis.jl
```diff
@@ -1069,50 +1038,74 @@ function iscomplexbalanced(rs::ReactionSystem, parametermap)
 end
 
 """
-    ratematrix(rs::ReactionSystem, parametermap)
+    adjacencymat(rs::ReactionSystem, pmap = Dict(); sparse = false)
```
isaacsas, line 1041: Does this require `ismassaction`, and if so is that being checked anywhere?

isaacsas: More generally, it would be good if any of these methods that only make sense with fixed rate constants (in the Catalyst sense) return an informative error if they get an invalid system.

isaacsas (changes requested):

src/network_analysis.jl
vyudu: Yeah sorry, I didn't realize I had actually committed this either

TorkelE: i guess we should consider at some point use https://github.com/JuliaTesting/Aqua.jl? I don't really know much about it though, but supposedly it should sort this kind of things out for us.

isaacsas: I found a lot of strangeness with what Aqua reported when I tried it in JumpProcesses, but I recall @ChrisRackauckas was telling me that is because of CI running Julia with different settings (maybe warnings on?), which can cause a lot more things to get flagged by it. So perhaps they weren't false flags as I was assuming.

src/network_analysis.jl
vyudu: Have added them. I think fluxmat and adjacencymat should work with reactions with fixed rate constants that aren't necessarily mass action (which is the case for generalized mass action systems e.g., the CatalystNetworkAnalysis stuff does some stuff with that) and the massactionvector should work with just mass action reactions

isaacsas: Ok, if they should work more generally then the checks aren't needed. I just want to make sure we don't have people using them for systems they aren't intended / expected to work with.

LMK how you want to proceed (I'm fine keeping the checks or dropping them if you feel they aren't actually needed here).

vyudu: I think the check currently is good, we do still wanna check that stuff like `k*A, A --> B` isn't allowed in the adjacencymat/fluxmat since you could no longer use it in the ODE system decomposition if it is time-dependent. But in theory something like `k, A --> B` that doesn't have mass action kinetics (e.g. the rate law is k*A^2) or something is fine and constructing the adjacencymat/fluxmat for such a reaction is useful in some cases.

isaacsas:
OK, then feel free to merge if you are happy with this.

PR #1174, opened 2025-01-14
test: added an edge case in ReactionSystem

sivasathyaseeelan:
I have added an edge cases to test ReactionSystem which were not tested before.

## Checklist

- [X] Appropriate tests were added
- [X] Any code changes were done in a way that does not break public API
- [X] The new code follows the

### DIFF
--- a/test/reactionsystem_core/reactionsystem.jl
+++ b/test/reactionsystem_core/reactionsystem.jl
@@ -612,6 +612,31 @@ let
     @test_throws Exception rs = ReactionSystem([Reaction(k1, [X1], [])], t, [X1, x], [k1])
 end
 
+# Test for forbidden symbols
+let
+    # Define the forbidden symbols (mocked for the purpose of the test)
+    forbidden_symbols_error = Set([:forbidden1, :forbidden2, :forbidden3])
+    Catalyst.forbidden_symbols_error = forbidden_symbols_error
+
+    # Define parameters and species including forbidden symbols
+    @parameters p1 forbidden1
+    @species S1 forbidden2(t)
+
+    # Define reactions using the species and parameters
+    rxs = [
+        Reaction(p1, [], [S1]),
+        Reaction(forbidden1, [], [S1]),
+    ]
+
+    # Attempt to create a ReactionSystem and capture the error
+    try
+        @named rs = ReactionSystem(rxs, t)
+        @test false  # This line should not be reached
+    catch e
+        @test e isa Exception
+    end
+end
+
 # Tests various erroneous `convert` calls.
 let
     # Conversion of non-autonomous `ReactionSystem` to `NonlinearSystem`.
--- a/test/reactionsystem_core/reactionsystem.jl
+++ b/test/reactionsystem_core/reactionsystem.jl
@@ -597,7 +597,7 @@ let
     @variables V(t)
 
     # System using a forbidden symbol.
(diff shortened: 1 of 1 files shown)


TorkelE:
The test is well written, but I don't see how this tests something relevant to the package? I.e. if I understand it correct, what we want to test is that if the user updates the `Catalyst.forbidden_symbols_error` list with a new forbidden variable, then these cannot be used?

I am mostly surprised that the user can update `Catalyst.forbidden_symbols_error`, which really isn;t something anyone should do.

Is this looking at a specific line in the code which we do not have coverage on? In that case we could have al ook at it to see if there is a better way to cover it.

Even if we do not end up using this, still thanks for your effort with the PR!

test/reactionsystem_core/reactionsystem.jl
```diff
@@ -612,6 +612,31 @@ let
...
 
+# Test for forbidden symbols
+let
+    # Define the forbidden symbols (mocked for the purpose of the test)
+    forbidden_symbols_error = Set([:forbidden1, :forbidden2, :forbidden3])
+    Catalyst.forbidden_symbols_error = forbidden_symbols_error
+
+    # Define parameters and species including forbidden symbols
+    @parameters p1 forbidden1
+    @species S1 forbidden2(t)
```
TorkelE, line 623: Should be `S(t)`, species should always depend on an independent variable.

sivasathyaseeelan:
> The test is well written, but I don't see how this tests something relevant to the package? I.e. if I understand it correct, what we want to test is that if the user updates the `Catalyst.forbidden_symbols_error` list with a new forbidden variable, then these cannot be used?
> 
> I am mostly surprised that the user can update `Catalyst.forbidden_symbols_error`, which really isn;t something anyone should do.
> 
> Is this looking at a specific line in the code which we do not have coverage on? In that case we could have al ook at it to see if there is a better way to cover it.
> 
> Even if we do not end up using this, still thanks for your effort with the PR!

I thought of creating some random forbidden symbols. I will remove it and replace it with any exiting forbidden symbol

Also this covers the below condition 
Screenshot from 2025-01-15 01-54-47

TorkelE:
sounds good

TorkelE:
Also, you might want to use `test_throws` to check whether a line of code generates an error.

TorkelE:
https://docs.julialang.org/en/v1/stdlib/Test/#Test.@test_throws

TorkelE:
Looking closer, this seems like a repeat of
```julia
    # System using a forbidden symbol.
    @test_throws Exception rs = ReactionSystem([Reaction(Γ, [X1], [])], t; named = :rs)
```

from the test block just above yours. 

However, there is an error in that test that means that that specific error is not hit (`named` should be `name`). if you want you can fix that bug error though :)

sivasathyaseeelan:
> Looking closer, this seems like a repeat of
> 
> ```julia
>     # System using a forbidden symbol.
>     @test_throws Exception rs = ReactionSystem([Reaction(Γ, [X1], [])], t; named = :rs)
> ```
> 
> from the test block just above yours.
> 
> However, there is an error in that test that means that that specific error is not hit (`named` should be `name`). if you want you can fix that bug error though :)

I will fix the bug in previous test and remove the new test set

TorkelE:
Thanks, will merge once everything passes.

PR #1175, opened 2025-01-14
Update inference of variables and default differential from `@equations` macro

TorkelE:
Updates what we infer as variables from equations, and also when we detect that the default differential `D` is used. The new formula is a bit more liberal, which will make describing it a bit easier.

### Inferring variables 
The order of inference of species/variables/parameters is now:
    (1) Every symbol explicitly declared using `@species`, `@variables`, and `@parameters` are assigned to the correct category.
    (2) Every symbol used as a reaction reactant is inferred as a species.
    (3) Every symbol not declared in (1) or (2) that occurs in an expression provided after `@equations` is inferred as a variable.
    (4) Every symbol not declared in (1), (2), or (3) that occurs either as a reaction rate or stoichiometric coefficient is inferred to be a parameter.

E.g. in
```julia
@reaction_network begin
    @equations V1 + S ~ V2^2
    (p + S + V1), S --> 0
end
```
`S` is inferred as a species, `V1` and `V2` as variables, and `p` as a parameter. 

The previous special cases for the `@observables` (if not declared as symbol/variable, becomes a variable without additional info), `@compounds` (counts as `@species` declarations), and `@differentials` (used to declare non-D differentials) still hold. 

Finally, we also now have the `@require_declaration` options which can be used to remove any uncertainty in messy cases.

### Inferring use of differential `D` 
When determining whether the default differentials have been used within an `@equations` option. Now, if any expression `D(...)` is encountered (where `...` can be anything), this is inferred as usage of the default differential D. E.g. in the following equations `D` is inferred as a differential with respect to the default independent variable:
```julia
@reaction_network begin
  @equations D(V) + V ~ 1
end
@reaction_network begin
  @equations D(D(V)) ~ 1
end
```

The code handling is getting a bit messier than it has to be. This will be better in the PR which finally fixes the mess that is the DSL file. Primarily we will probably just create lists for stuff declared using `@species`, `@variables`, and `@parameters` and then which is inferred. And then this can just be processed in an orderly manner. But for now, this does all the breaking changes, which we want with v15.

I have also added a couple of tests for good measure.


src/dsl.jl
```diff
@@ -358,17 +358,27 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
 
     # Collect species and parameters, including ones inferred from the reactions.
     declared_syms = Set(Iterators.flatten((parameters_declared, species_declared,
-        variables)))
```
TorkelE, line 361: The bit of code that generates variables is now  moved later (as it need to know the inferred species). Hence we only have (and only need) `variables_declared`.

src/dsl.jl
```diff
@@ -358,17 +353,27 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
...
         reactions, declared_syms; requiredec)
 
     species = vcat(species_declared, species_extracted)
     parameters = vcat(parameters_declared, parameters_extracted)
 
+    # Reads equations.
+    designated_syms = [species; parameters; variables_declared]
+    vars_extracted, add_default_diff, equations = read_equations_options(
+        options, designated_syms)
+    variables = vcat(variables_declared, vars_extracted)
```
TorkelE, line 367: Equations are now read after `species_extracted` has been found.

src/dsl.jl
```diff
@@ -352,23 +347,32 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
         combinatoric_ratelaws = true
     end
 
-    # Reads observables.
-    observed_vars, observed_eqs, obs_syms = read_observed_options(
-        options, [species_declared; variables], all_ivs; requiredec)
-
```
TorkelE, line 358: observables must be handled after `variables` is found, so now moved down.

src/dsl.jl
```diff
@@ -328,11 +328,6 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
     parameters_declared = extract_syms(options, :parameters)
     variables_declared = extract_syms(options, :variables)
 
-    # Reads equations.
-    vars_extracted, add_default_diff, equations = read_equations_options(
-        options, variables_declared; requiredec)
-    variables = vcat(variables_declared, vars_extracted)
-
```
TorkelE, line 335: Equations must now be read after `species_extracted` is found.

src/dsl.jl
```diff
@@ -352,23 +347,32 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
...
     # Collect species and parameters, including ones inferred from the reactions.
     declared_syms = Set(Iterators.flatten((parameters_declared, species_declared,
-        variables)))
+        variables_declared)))
     species_extracted, parameters_extracted = extract_species_and_parameters!(
         reactions, declared_syms; requiredec)
 
+    # Reads equations (and infers potential variables).
+    # Exlucdes any parameters already extracted (if they also was a variable).
+    designated_syms = [declared_syms; species_extracted]
```
TorkelE, line 358: Update `designated_syms` with `species_extracted` so these are excluded from the next step.

src/dsl.jl
```diff
@@ -352,23 +347,32 @@ function make_reaction_system(ex::Expr; name = :(gensym(:ReactionSystem)))
...
     species_extracted, parameters_extracted = extract_species_and_parameters!(
         reactions, declared_syms; requiredec)
 
+    # Reads equations (and infers potential variables).
+    # Excludes any parameters already extracted (if they also was a variable).
+    declared_syms = union(declared_syms, species_extracted)
+    vars_extracted, add_default_diff, equations = read_equations_options(
+        options, declared_syms)
+    variables = vcat(variables_declared, vars_extracted)
+    parameters_extracted = setdiff(parameters_extracted, vars_extracted)
```
TorkelE, line 362: Now handles variables and equations (and removing parameters that should be variables).

src/dsl.jl
```diff
@@ -713,34 +717,46 @@ function read_equations_options(options, variables_declared; requiredec = false)
...
-                in(diff_var, variables_declared) || push!(vars_extracted, diff_var)
-            end
+        # If the default differential (`D`) is used, record that it should be decalred later on.
+
+        if !in(eq, excluded_syms) && find_D_call(eq)
+            requiredec && throw(UndeclaredSymbolicError(
+                "Unrecognized symbol D was used as a differential in an equation: \"$eq\". Since the @require_declaration flag is set, all differentials in equations must be explicitly declared using the @differentials option."))
+            add_default_diff = true
+            excluded_syms = push!(excluded_syms, :D)
+        end
```
TorkelE, line 735: This bits checks if the default differential have been used.

src/dsl.jl
```diff
@@ -713,34 +717,42 @@ function read_equations_options(options, variables_declared; requiredec = false)
...
+            requiredec && throw(UndeclaredSymbolicError(
+                "Unrecognized symbol D was used as a differential in an equation: \"$eq\". Since the @require_declaration flag is set, all differentials in equations must be explicitly declared using the @differentials option."))
+            add_default_diff = true
+            excluded_syms = push!(excluded_syms, :D)
         end
+
+        # Any undecalred symbolic variables encountered should be extracted as variables.
+        add_syms_from_expr!(vars_extracted, eq, excluded_syms)
+        (!isempty(vars_extracted) && requiredec) && throw(UndeclaredSymbolicError(
+            "Unrecognized symbolic variables $(join(vars_extracted, ", ")) detected in equation expression: \"$(string(eq))\". Since the flag @require_declaration is declared, all symbolic variables must be explicitly declared with the @species, @variables, and @parameters options."))
```
TorkelE, line 740: This bits extracts any variables.

src/reactionsystem.jl
```diff
@@ -97,9 +97,9 @@ Base.@kwdef mutable struct NetworkProperties{I <: Integer, V <: BasicSymbolic{Re
     stronglinkageclasses::Vector{Vector{Int}} = Vector{Vector{Int}}(undef, 0)
     terminallinkageclasses::Vector{Vector{Int}} = Vector{Vector{Int}}(undef, 0)
 
-    checkedrobust::Bool = false
```
TorkelE, line 100: I need to fix this thing. I have checked an re-checked and there is no actual difference (more than space removal). I think I put in, and then removed, a debug statement in this file (and the space removal happened). Hence changes appeared here.

test/dsl/dsl_options.jl
```diff
@@ -424,6 +424,74 @@ let
...
+    end
+
+    # Variable + differential.
+    @test_throws Exception @eval @reaction_network begin
+        @variables X(t)
+        @differentials X = Differential(t)
+    end
+
+    # Parameter + observable (species/variable + observable is OK, as this e.g. provide additional observables information).
+    @test_broken false #@test_throws Exception @eval @reaction_network begin
```
TorkelE, line 468: I am not really sure how to make a `@test_throws` as broken. What I found online sounded like a proper mess. Still, if there is something that is broken it seems better that we at least get broken warnings in the test set(?) I have fixing this as relatively high priority.

src/dsl.jl
```diff
@@ -713,34 +717,41 @@ function read_equations_options(options, variables_declared; requiredec = false)
     # Loops through all equations, checks for lhs of the form `D(X) ~ ...`.
     # When this is the case, the variable X and differential D are extracted (for automatic declaration).
     # Also performs simple error checks.
-    vars_extracted = Vector{Symbol}()
+    vars_extracted = OrderedSet{Union{Symbol, Expr}}()
```
TorkelE, line 720: species and parameters in the corresponding cases uses OrderedSet{Union{Symbol, Expr}}(), so figured we should use it here too.

src/dsl.jl
```diff
@@ -713,34 +717,41 @@ function read_equations_options(options, variables_declared; requiredec = false)
...
-                add_default_diff = true
-                in(diff_var, variables_declared) || push!(vars_extracted, diff_var)
-            end
+        # If the default differential (`D`) is used, record that it should be decalred later on.
+
+        if !in(eq, syms_declared) && find_D_call(eq)
+            requiredec && throw(UndeclaredSymbolicError(
+                "Unrecognized symbol D was used as a differential in an equation: \"$eq\". Since the @require_declaration flag is set, all differentials in equations must be explicitly declared using the @differentials option."))
+            add_default_diff = true
+            push!(syms_declared, :D)
```
TorkelE, line 733: This bits checks if the default differential has been used.

test/reactionsystem_core/coupled_equation_crn_systems.jl
```diff
@@ -899,38 +899,6 @@ end
...
-    @test_throws Exception @eval @reaction_network begin
-        @variables V(t)
-        @equations D(V) + 1 ~ 0
-    end
-
-    # Attempting to infer a variable when using a non-default differential.
-    @test_throws Exception @eval @reaction_network begin
-        @differentials Δ = Differential(t)
-        @equations Δ(V) ~ -1,0
-    end
```
TorkelE, line 932: Tests of stuff that is now allowed

HISTORY.md
```diff
@@ -7,6 +7,29 @@
...
+`S` is inferred as a species, `V1` and `V2` as variables, and `p` as a parameter. The previous special cases for the `@observables`, `@compounds`, and `@differentials` options still hold. Finally, the `@require_declaration` options (described in more detail below) can now be used to require everything to be explicitly declared.
+- New formula for determining whether the default differentials have been used within an `@equations` option. Now, if any expression `D(...)` is encountered (where `...` can be anything), this is inferred as usage of the default differential D. E.g. in the following equations `D` is inferred as a differential with respect to the default independent variable:
+```julia
+@reaction_network begin
+  @equations D(V) + V ~ 1
+end
+@reaction_network begin
+  @equations D(D(V)) ~ 1
+end
+```
```
isaacsas, line 31: What happens if someone uses `D` in a reaction as an implicit species?

src/reactionsystem.jl
isaacsas: I think you must have some VSCode setting enabled.

HISTORY.md
isaacsas: Aside from that question this is fine to merge I think.

TorkelE: Using D in any other context (e.g. as a species, variable or parameter) should disable inference of D as a differential. Will clarify this

TorkelE:
Turns out the docs have started failing, something in the oscillation parameter fitting example.

PR #1179, opened 2025-02-05
Fix deprecation warnings in Latexify 0.16.6

devmotion:
Latexify 0.16.6 deprecated the `cdot` option. Instead one has to specify `mult_symbol` now.


isaacsas:
Thanks!

PR #1180, opened 2025-02-06
Add advice on checking built docs to dev documentation 

TorkelE:
Very WIP. wrote something on my laptop and uploaded, will go through and make it nice (and check that it is actually correct) when I get back to my desktop.

Adds section to the dev docs on how to check the docs (if nothing else so I can see how to do it next time I forget). Also added a note of the Code Spell Checker. Had no idea about it for a long time, but as a non-native speaker it helps a lot.

At some point we can think of if there is a nicer way to structure this, but for now I figured it is probably best to put in stuff there.


TorkelE:
Ready

docs/src/devdocs/dev_guide.md
```diff
@@ -33,3 +33,26 @@ been installed by the package resolver had it not been capped, a patch release
 should be preferred. If the new release branch is branched from master, *it
 needs to ensure Project.toml caps are all ≥ to those listed in the previous
 Catalyst release branch*.
+
+## Development advice
+
+### Checking doc builds for errors
+When updating documentation, Catalyst will run any Julia code provided within it to dynamically create figures and outputs. In addition to automatically creating these for us, it also provides an automatic check that all code in documentation is correct. Here, if any of the documentation code throws an error, the build job will fail. The documentation build job can be found at the bottom of a PRs conversation, here is an example of a failed one:
```
isaacsas, line 40:
```suggestion
When updating documentation, Catalyst will run any Julia code provided within example blocks to dynamically create figures and outputs. In addition to automatically creating these for us, it also provides an automatic check that all code in documentation is correct. Here, if any of the documentation code throws an error, the build job will fail. The documentation build job can be found at the bottom of a PRs conversation, here is an example of a failed one:
```

docs/src/devdocs/dev_guide.md
```diff
@@ -33,3 +33,26 @@ been installed by the package resolver had it not been capped, a patch release
...
+![Failed builddocs link](../assets/devdocs/failed_builddocs_link.png)
+To check what errors were produced, clock on the "Details" link of the job. Next, any errors can be found at the bottom of the "Build and deploy" section (which should be opened automatically).
```
isaacsas, line 42:
```suggestion
To check what errors were produced, click on the "Details" link of the job. Next, any errors can be found at the bottom of the "Build and deploy" section (which should be opened automatically).
```

isaacsas (changes requested):

PR #1186, opened 2025-02-17
fixes for MTK changes

TorkelE:
Updates conservation laws to work with new `remake` version.


isaacsas:
This is probably never going to pass tests with the change in semantics for `complete`. We can no longer rely on the parameters of a `complete`d system being the same as what was defined when creating the system, which I think a lot of tests rely on.

isaacsas:
We can also no longer rely on `issetequal(parmeters(sys), get_ps(sys))` for a flattened system.

TorkelE:
Let's see how this turns out. If something like what you suggest is really changing it would be one of the largest changes to Catalyst ever, probably after moving to maps as inputs to problems. Would  probably take ages to figure out all the implications.

isaacsas:
@TorkelE let's just work in this PR to try to get tests working again. 

I've handled the conservation law constants getting scalarized issue, and locally gotten tests in reactionsystem.jl working again.

isaacsas:
TODO:

- [ ] `isequivalent` needs more comprehensive testing (perhaps a followup PR)?

TorkelE:
I had some updates locally to sort out the new parameters, but there are quite a few recent MTK issues that are hard-blocking us so I stopped a bit:
https://github.com/SciML/ModelingToolkit.jl/issues/3411
https://github.com/SciML/ModelingToolkit.jl/issues/3410
https://github.com/SciML/ModelingToolkit.jl/issues/3409
https://github.com/SciML/ModelingToolkit.jl/issues/3408
https://github.com/SciML/ModelingToolkit.jl/issues/3396

isaacsas:
OK, I'll get as much fixed here as I can modulo those issues and then plan to merge. 

FYI, take a look at the changes to `isequivalent`. The nice feature is I added a `debug` kwarg that will printout specifically which fields are not the same. (I seem to always want this functionality when writing tests).

src/reactionsystem_conversions.jl
```diff
@@ -719,8 +731,8 @@ function DiffEqBase.ODEProblem(rs::ReactionSystem, u0, tspan,
     # Handles potential differential algebraic equations (which requires `structural_simplify`).
     if structural_simplify
         (osys = MT.structural_simplify(osys))
-    elseif has_alg_equations(rs)
-        error("The input ReactionSystem has algebraic equations. This requires setting `structural_simplify=true` within `ODEProblem` call.")
+    # elseif has_alg_equations(rs)
+    #     error("The input ReactionSystem has algebraic equations. This requires setting `structural_simplify=true` within `ODEProblem` call.")
```
TorkelE, line 735: ANy reason to remove this? `structural_simplify` is still a mess, but this still seems like a useful error to throw?

test/dsl/dsl_options.jl
```diff
@@ -1468,3 +1469,5 @@ let
         @observables X2 ~ X1
     end
 end
+
+nothing
```
TorkelE, line 1473: ?

src/reactionsystem_conversions.jl
ChrisRackauckas: What's the issue with structural simplify? I don't know of one related to Catalyst

ChrisRackauckas: No and no. First of all, that is no issue of `structural_simplify`, this is just a property of DAEs and index 1 transformations 😅 . So if you're waiting for that to ever change, sorry but it's just the math of DAEs. I'd be happy to hop on a call and explain it since you and @isaacsas keep asking the same question, it's probably just easiest to pull out a whiteboard at this point. But it shouldn't cause any difficulties to the user. 

Here's a simple way to understand it. If your DAE is index 1 (which it will be after transformation), then every equation is either a differential equation or an algebraic equation, and every equation is associated through matching to be solving for one variable (this is equivalent to the index 1 condition and not having a singularity in the Jacobian). So with this, any variable can be called either a differential or algebraic variable. You need as many initial conditions as you have algebraic variables, and any variable that doesn't have an initial condition needs a guess. It doesn't matter which one is chosen by the system as differential or algebraic, but it does matter that you have the right number of conditions. If you have more conditions, it throws you a warning because it's generally not a good idea to have an overdetermined system, but it doesn't mean it's impossible to solve. If it's over or underdetermined it will still try to initialize, with many caveats of course. 

The simplest way to understand this then is that, every variable needs either an initial condition or a guess. You should probably be making every differential variable have an initial condition, and for every algebraic condition you should be introducing a new variable and give that one a guess. 

There is potentially a way to not need guesses if we move beyond Newton methods to more homotopy type solvers, since it's really just a property of Newton that you need a `u0`, but this'll be years of research and not something to rely on.

isaacsas: I just commented this out while debugging. I didn't intend to keep it commented out when merging...

test/dsl/dsl_options.jl
TorkelE: maybe move it to the end of the last `let ... end` block, just to make it clear that it belongs there (at least if that is the block which returns the last output, but I presume that would bet he case?).

TorkelE:
If we want to merge this I am happy with it. Still quite a way to go to update to the latest everything, but not sure if it is possible yet.

isaacsas:
No, we can keep this open. But it has a lot of fixes for the MTK changes, so we should just push to this PR as we work on tests going forward...

TorkelE:
sounds good, thought from your latest comment that we maybe merge this and then proceed, but happy to continue using this branch.

ext/CatalystBifurcationKitExtension/bifurcation_kit_extension.jl
```diff
@@ -22,11 +22,22 @@ function BK.BifurcationProblem(rs::ReactionSystem, u0_bif, ps, bif_par, args...;
...
+# more things here.
+function bkext_make_nsys(rs, u0)
+    cons_eqs = conservationlaw_constants(rs)
+    cons_default = [cons_eq.rhs for cons_eq in cons_eqs]
+    cons_default = Catalyst.get_networkproperties(rs).conservedconst => cons_default
+    defaults = Dict([u0; cons_default])
+    nsys = convert(NonlinearSystem, rs; defaults,
+        remove_conserved = true, remove_conserved_warn = false)
+    return complete(nsys)
+end
```
TorkelE, line 43: ModelingToolkit can no longer handle the conservation parameters in bifurcation kit (I think this might have to do with them now having a `missing` default, which some internal function in MTK handles). For now it seems better to handle this in Catalyst (i.e. by computing their value here, and not leaving it to MTK).

src/reactionsystem.jl
```diff
@@ -763,13 +796,55 @@ function get_indep_sts(rs::ReactionSystem, remove_conserved = false)
...
+            # pdeps. Any other `Initial` parameters should be kept (e.g. initialization
+            # systems)
+            filter!(
+                x -> !iscall(x) || !isa(operation(x), Initial) ||
+                         !has_parameter_dependency_with_lhs(network, only(arguments(x))),
+                         ps)
+        end
+    end
+    return ps
+end
```
TorkelE, line 839: Basically a stop gap for now when we cannot use `get_ps`. It seems that at some point the idea is for e.g. `parameters` (and similar MTK getters) to support a kwarg to only access stuff in the top-level system (will likely be allocating though), at which point we should throw this one out.

src/reactionsystem.jl
```diff
@@ -763,13 +796,55 @@ function get_indep_sts(rs::ReactionSystem, remove_conserved = false)
...
+
 """
     numparams(network)
 
 Return the total number of parameters within the given system and all subsystems.
 """
 function numparams(network)
-    nps = length(get_ps(network))
+    nps = length(parameters_toplevel(network))
```
TorkelE, line 847: I tried to figure out when we should use `parameters_toplevel` and when we still should use `get_ps`.

test/network_analysis/conservation_laws.jl
```diff
@@ -127,20 +127,22 @@ let
...
+        ssprob2 = SteadyStateProblem(rn, u0, p)
+        ssprob3 = SteadyStateProblem(rn, u0, p; remove_conserved = true, remove_conserved_warn)
+        nsol1 = solve(nprob1, NewtonRaphson(); abstol = 1e-8)
+        # Nonlinear problems cannot find steady states properly without removing conserved species.
+        nsol3 = solve(nprob3, NewtonRaphson(); abstol = 1e-8)
+        sssol1 = solve(ssprob1, DynamicSS(Tsit5()); abstol = 1e-8, reltol = 1e-8)
+        sssol2 = solve(ssprob2, DynamicSS(Tsit5()); abstol = 1e-8, reltol = 1e-8)
+        sssol3 = solve(ssprob3, DynamicSS(Tsit5()); abstol = 1e-8, reltol = 1e-8)
+        @test nsol1[sps] ≈ nsol3[sps] ≈ sssol1[sps] ≈ sssol2[sps] ≈ sssol3[sps]
+    end
```
TorkelE, line 145: `NonlinearSystem` + conservation laws is broken, and requires MTK fixes before we can make it work.

test/network_analysis/conservation_laws.jl
```diff
@@ -331,16 +333,19 @@ let
...
-        @test_nowarn XProblem(rn, u0, ps; remove_conserved_warn = false)
-        @test_nowarn XProblem(rn, u0, ps; remove_conserved = true, remove_conserved_warn = false)
+    # Conservation law removal currently not working for NonlinearSystems due to MTK depricating something. https://github.com/SciML/ModelingToolkit.jl/issues/3458, https://github.com/SciML/ModelingToolkit.jl/issues/3411
+    # Currently just checks whether a NonlinearProblem can be created.
+    @test_broken for XProblem in [NonlinearProblem, SteadyStateProblem]
+        XProblem(rn, u0, ps; remove_conserved = true)
+        # @test_nowarn XProblem(rn, u0, ps)
+        # @test_logs (:warn, r"You are creating a system or problem while eliminating conserved quantities. Please *") XProblem(rn, u0, ps; remove_conserved = true)
+        # @test_nowarn XProblem(rn, u0, ps; remove_conserved_warn = false)
+        # @test_nowarn XProblem(rn, u0, ps; remove_conserved = true, remove_conserved_warn = false)
```
TorkelE, line 343: Again `NonlinearSystem` + conservation laws. 

If we are confident that `remake` works then we should remove these tests..

src/reactionsystem.jl
```diff
@@ -763,13 +796,59 @@ function get_indep_sts(rs::ReactionSystem, remove_conserved = false)
...
 end
 
+"""
+    parameters_toplevel(network; initial_parameters = false)
+
+Returns the parameters of the top-level system only. Equivalent to `parameters`, but ignores
+parameters of sub-systems.
+
+WARNING:
+- This function may be removed in non-breaking changes.
```
TorkelE, line 806: Right now I marked this as something that might be removed in non-breaking releases. Which might not be ok since we briefly mentions it in the docs.

src/reactionsystem.jl
```diff
@@ -1388,7 +1467,8 @@ function MT.flatten(rs::ReactionSystem; name = nameof(rs))
         balanced_bc_check = false,
         spatial_ivs = get_sivs(rs),
         continuous_events = MT.continuous_events(rs),
-        discrete_events = MT.discrete_events(rs))
+        discrete_events = MT.discrete_events(rs),
+        metadata = MT.get_metadata(rs))
```
TorkelE, line 1471: Bugfix. Currently, flattening a system removes any system metadata.

docs/src/inverse_problems/examples/ode_fitting_oscillation.md
```diff
@@ -1,5 +1,5 @@
...
+In this example we will use [Optimization.jl](https://github.com/SciML/Optimization.jl) to fit the parameters of an oscillatory system (the [Brusselator](@ref basic_CRN_library_brusselator)) to data. Here, special consideration is taken to avoid reaching a local minimum. Instead of fitting the entire time series directly, we will start with fitting parameter values for the first period, and then use those as an initial guess for fitting the next (and then these to find the next one, and so on). Using this procedure is advantageous for oscillatory systems, and enables us to reach the global optimum. For more information on fitting ODE parameters to data, please see [the main documentation page](@ref optimization_parameter_fitting) on this topic.
```
TorkelE, line 2: This needed some updating due to breakage (in how to update parameters in the loss function). Also took the opportunity to do some additional updates for newer workflows.

docs/src/inverse_problems/examples/ode_fitting_oscillation.md
```diff
@@ -72,45 +79,36 @@ nothing # hide
 
 Next, we will fit a parameter set to the data on the interval `(0, 10)`.
 ```@example pe_osc_example
-p_estimate = optimise_p([5.0, 5.0], 10.0)
```
TorkelE, line 75: just making optimize/optimise consistent within the individual doc

isaacsas (changes requested):

docs/src/api.md
```diff
@@ -124,7 +124,8 @@ direct access to the corresponding internal fields of the `ReactionSystem`)
   entries in `get_species(rn)` correspond to the first `length(get_species(rn))`
   components in `get_unknowns(rn)`.
 * `ModelingToolkit.get_ps(rn)` is a vector that collects all the parameters
-  defined *within* reactions in `rn`.
+  defined *within* reactions in `rn`. This includes initialisation parameters (which
+  are added to the system by ModelingToolkit, and not the user).
```
TorkelE, line 128: This should be updated now, right?

src/reactionsystem.jl
TorkelE: Agreed this is a bit of a mess, but I think this replicated the behaviour MTK has (i.e. flattening used to remove metadata, and now it doesn't anymore).

In practise the main problem is that I have never really got a reply of what the `metadata` field actually is. I use it a bit for a personal project. But if we can confirm whether it is for users benefit, or if MTK at some point want to stick stuff here, that changes things a lot.

test/network_analysis/conservation_laws.jl
TorkelE: yes, although there is hope I might be able to fix it now when a bug got fixed.

test/dsl/dsl_options.jl
```diff
@@ -1059,8 +1059,9 @@ let
         @equations 2X ~ $c - X
     end)
     oprob = ODEProblem(rn, [], (0.0, 100.0); structural_simplify=true)
-    sol = solve(oprob, Tsit5(); abstol=1e-9, reltol=1e-9)
-    @test sol[rn.X][end] ≈ 2.0
+    @test_broken false # Related MTK issue: https://github.com/SciML/ModelingToolkit.jl/issues/3467 
+    # sol = solve(oprob, Tsit5(); abstol=1e-9, reltol=1e-9)
+    # @test sol[rn.X][end] ≈ 2.0
```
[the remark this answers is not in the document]
TorkelE: yes it should, I will reenable

docs/Project.toml
```diff
@@ -63,7 +63,7 @@ IncompleteLU = "0.2"
 JumpProcesses = "9.13.2"
 Latexify = "0.16.5"
 LinearSolve = "2.30, 3"
-ModelingToolkit = "< 9.60"
+ModelingToolkit = "9.65"
```
isaacsas, line 66:
```suggestion
ModelingToolkit = "9.68.1"
```

Project.toml
```diff
@@ -61,17 +61,17 @@ LaTeXStrings = "1.3.0"
 Latexify = "0.16.6"
 MacroTools = "0.5.5"
 Makie = "0.22.1"
-ModelingToolkit = "< 9.60"
+ModelingToolkit = "9.68"
```
isaacsas, line 64:
```suggestion
ModelingToolkit = "9.68.1"
```

src/reactionsystem.jl
isaacsas: Sounds good.

isaacsas:
@AayushSabharwal thanks for all your help with the various issues @TorkelE opened. We've finally got tests passing again! If as of now you find our tests start failing during MTK/SciMLBase downstream testing we would very much appreciate a heads up about what changes might be causing such issues and what we need to update.

AayushSabharwal:
Not a problem! Glad to get everything green again. And yeah, I'll make sure to avoid breaking Catalyst CI or at least inform you if I find outdated code.

Issue #1183, opened 2025-02-11
Change some network graph plotting defaults


TorkelE:
E.g. here: 

![Image](https://github.com/user-attachments/assets/d0b1bbdd-e081-41c4-85db-0853c3bf0c9b)

The ticks and grid lines are not really necessary for this purpose and mostly clutters up the plot. It might make sense to, by default, remove these.

vyudu:
Ah I remember why I didn't do this, the `hidedecorations!` and other functions that remove the ticks and grid lines are in `CairoMakie` and not `GraphMakie` itself. I suppose we could make one of the Makie backends a dependency of the ext?

vyudu:
Or just regular Makie, I'll do that instead.

PR #1189, opened 2025-02-28
Fix graph defaults, plotting graphs

### LINKED ISSUE #1182: Missing plots in network graph plotting docs
Looks like there are cases here: https://docs.sciml.ai/Catalyst/dev/model_creation/model_visualisation/#Customizing-Plots where we would want a plot to be displayed, but where it is not. Probably just add
```julia
plot(f) # hide
```
(or what the correct make syntax is) at the end of some places might be an improvement?

Image

vyudu:
Close #1182, #1183


isaacsas:
Seems like there was some issue with the doc build.

vyudu:
OK, I think it's working now.

isaacsas (approved):

isaacsas:
Thanks!

PR #1191, opened 2025-03-07
Enable functional/time dependent parameters

TorkelE:
Adds docs and tests. Also modifies the main code to enable these to be used within Catalyst. 

Example:
```julia
using Catalyst, DataInterpolations, OrdinaryDiffEqDefault, Plots

# Create input function/data
tend = 120.0
ts = collect(0.0:1.0:tend)
light = sin.(ts/6) .+ 1
light = [max(0.0, l - rand()) for l in light]
interpolated_light = LinearInterpolation(light, ts)
plot(interpolated_light)
```
image

```julia
# Create model
t = default_t()
in_type = typeof(interpolated_light)
@parameters kA kD (light_in::in_type)(..)
@species Pₐ(t) Pᵢ(t)
rxs = [
    Reaction(kA*light_in(t), [Pᵢ], [Pₐ]),
    Reaction(kD, [Pₐ], [Pᵢ])
]
@named rs = ReactionSystem(rxs, t)
rs = complete(rs)

# Simulate the model.
u0 = [Pᵢ => 1.0, Pₐ => 0.0]
ps = [kA => 1.5, kD => 1.0, light_in => interpolated_light]
oprob = ODEProblem(rs, u0, tend, ps)
sol = solve(oprob)
plot(sol)
```
image


test/reactionsystem_core/custom_crn_functions.jl
```diff
@@ -49,7 +49,7 @@ let
         ps = rnd_ps(custom_function_network_1, rng; factor)
         t = rand(rng)
         @test f_eval(custom_function_network_1, u0, ps, t) ≈ f_eval(custom_function_network_2, u0, ps, t)
-        @test jac_eval(custom_function_network_1, u0, ps, t) ≈ jac_eval(custom_function_network_2, u0, ps, t)
+        @test_broken jac_eval(custom_function_network_1, u0, ps, t) ≈ jac_eval(custom_function_network_2, u0, ps, t) # Weird error dur to some SciML update. Reported in https://github.com/SciML/ModelingToolkit.jl/issues/3447.
```
TorkelE, line 52: This change is actually done in https://github.com/SciML/Catalyst.jl/pull/1192 (but this PR is based off that one).

TorkelE:
Builds on the change in https://github.com/SciML/Catalyst.jl/pull/1192, however, but otherwise this is now ready.

docs/src/model_creation/functional_parameters.md
```diff
@@ -0,0 +1,148 @@
...
+```@example functional_parameters_circ_rhythm
+rs == rs_dsl
+```
+
+## [Non-time functional parameters](@id functional_parameters_sir)
+Previously we have demonstrated functional parameters that are a function of time. However, functional parameters can be functions of any variable (however, currently, more than one argument is not supported). Here we will demonstrate this using a [SIR model](@ref basic_CRN_library_sir), but instead of having the infection rate scale linearly with the number of infected individuals, we instead assume we have measured data of the infection rate (as dependent on the number of infected individuals) and wish to use this instead. Normally we use the following infection reaction in the SIR model:
+```julia
+@reaction k1, S + I --> 2I
+```
+In practise, this is identical to
```
isaacsas, line 93:
```suggestion
For ODE models, this would give the same equations as 
```

docs/src/model_creation/functional_parameters.md
```diff
@@ -0,0 +1,148 @@
...
+plot(I_rate; label = "Meassured infection rate")
+plot!(I_grid, I_grid; label = "Normal SIR infection rate")
+```
+Next, we create our model (using the DSL-approach). As `I_rate` will be a function of $I$ we will need to declare this species first as well.
+```@example functional_parameters_sir
+using Catalyst
+@species I(default_t())
+inf_rate = I_rate(I)
+sir = @reaction_network rs begin
+    k1*$inf_rate, S --> I
```
isaacsas, line 118: It would be nicer if we could make something like
```suggestion
    k1*($I_rate)(I), S --> I
```
work.

isaacsas: Having to declare `t` and/or `I(t)` symbolically is a bit clunky.

docs/src/model_creation/functional_parameters.md
```diff
@@ -0,0 +1,148 @@
...
+    k2, I --> R
+end
+nothing # hide
+```
+Finally, we can simualte our model.
+```@example functional_parameters_sir
+using OrdinaryDiffEqDefault
+u0 = [:S => 99.0, :I => 1.0, :R => 0.0]
+ps = [:k1 => 0.002, :k2 => 0.01, inf_rate => I_rate]
```
isaacsas, line 127: Again, why not
```suggestion
ps = [:k1 => 0.002, :k2 => 0.01, :inf_rate => I_rate]
```

docs/src/model_creation/functional_parameters.md
```diff
@@ -0,0 +1,148 @@
...
+oprob = ODEProblem(sir, u0, 250.0, ps)
+sol = solve(oprob)
+plot(sol)
+```
+!!! note
+    When declaring a functional parameter of time, it is easy to set its grid values (i.e. ensure they range from the first to the final time point). For Functional parameters that depend on species concentrations it is trickier, and one must make sure that any potential input-species values that can occur during the simulation are represented in the interpolation. In this example it is fairly straightfor
```
isaacsas, line 133:
```suggestion
    When declaring a functional parameter of time, it is easy to set its grid values (i.e. ensure they range from the first to the final time point). For Functional parameters that depend on species concentrations it is trickier, and one must make sure that any potential input-species values that can occur during the simulation are represented in the interpolation. 
```

test/runtests.jl
```diff
@@ -21,6 +21,7 @@ end
         @time @safetestset "Higher Order Reactions" begin include("reactionsystem_core/higher_order_reactions.jl") end
         @time @safetestset "Symbolic Stoichiometry" begin include("reactionsystem_core/symbolic_stoichiometry.jl") end
         @time @safetestset "Parameter Type Designation" begin include("reactionsystem_core/parameter_type_designation.jl") end
+        @time @safetestset "Functional Parameters" begin include("reactionsystem_core/functional_parameters.jl") end
```
isaacsas, line 24: I would put this a lot lower down in the list. It seems less core than things like events or coupling equations.

test/reactionsystem_core/custom_crn_functions.jl
isaacsas: It should work now right?

isaacsas (changes requested):

docs/src/model_creation/functional_parameters.md
TorkelE: Agreed. had wanted to do this but get a
```julia
ERROR: LoadError: TypeError: in isdefined, expected Symbol, got a value of type Expr
Stacktrace:
  [1] recursive_escape_functions!(expr::Expr, syms_skip::Vector{Any})
    @ Catalyst ~/.julia/dev/Catalyst/src/dsl.jl:980
  [2] (::Catalyst.var"#232#233"{Expr, Vector{Any}})(i::Int64)
    @ Catalyst ~/.julia/dev/Catalyst/src/dsl.jl:978
  [3] foreach
    @ ./abstractarray.jl:3187 [inlined]
  [4] recursive_escape_functions!(expr::Expr, syms_skip::Vector{Any})
    @ Catalyst ~/.julia/dev/Catalyst/src/dsl.jl:978
  [5] recursive_escape_functions!
    @ ~/.julia/dev/Catalyst/src/dsl.jl:977 [inlined]
  [6] get_rxexpr(rx::Catalyst.DSLReaction)
    @ Catalyst ~/.julia/dev/Catalyst/src/dsl.jl:555
  [7] #219
    @ ~/.julia/dev/Catalyst/src/dsl.jl:546 [inlined]
  [8] foreach(f::Catalyst.var"#219#221"{Expr}, itr::Vector{Catalyst.DSLReaction})
    @ Base ./abstractarray.jl:3187
  [9] get_rxexprs(reactions::Vector{Catalyst.DSLReaction}, equations::Vector{Expr}, all_syms::Vector{Union{Expr, Symbol}})
    @ Catalyst ~/.julia/dev/Catalyst/src/dsl.jl:546
 [10] make_reaction_system(ex::Expr, name::QuoteNode)
    @ Catalyst ~/.julia/dev/Catalyst/src/dsl.jl:323
 [11] make_rs_expr(name::QuoteNode, network_expr::Expr; complete::Bool)
    @ Catalyst ~/.julia/dev/Catalyst/src/dsl.jl:153
 [12] make_rs_expr(name::QuoteNode, network_expr::Expr)
    @ Catalyst ~/.julia/dev/Catalyst/src/dsl.jl:152
 [13] var"@reaction_network"(__source__::LineNumberNode, __module__::Module, name::Symbol, network_expr::Expr)
    @ Catalyst ~/.julia/dev/Catalyst/src/dsl.jl:92
in expression starting at /home/loman/Desktop/Julia Playground/Environment - Catalyst Docs (full)/catalyst_docs_playground.jl:71
```

docs/src/model_creation/functional_parameters.md
TorkelE: This is not directly possible in this case, but I can rewrite a bit to enable it.

test/runtests.jl
TorkelE: (of the `ReactionSyste`m structure block)

docs/src/model_creation/functional_parameters.md
```diff
@@ -0,0 +1,148 @@
...
+```
+Next, we create our model, [interpolating](@ref dsl_advanced_options_symbolics_and_DSL_interpolation) the input parameter into it (making it a function of `t`).
+```@example functional_parameters_basic_example
+input = pIn(default_t())
+bd_model = @reaction_network begin
+    $input, 0 --> X
+    d, X --> 0
+end
+```
+Finally, we can simulate our model as normal, but setting the value of the `pIn` parameter to our interpolated data.
```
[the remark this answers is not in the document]
TorkelE: Changed to "Finally, we can simulate our model as normal (but where we set the value of the `pIn` parameter to our interpolated data)."

docs/src/model_creation/functional_parameters.md
```diff
@@ -0,0 +1,148 @@
...
+Let us now go through everything again, but providing some more details. Let us first consider the input parameter. We have previously described how a [time-dependent rate can model a circadian rhythm](@ref dsl_description_nonconstant_rates_time). For real applications, due to e.g. clouds, sunlight is not a perfect sine wave. Here, a common solution is to take real sunlight data from some location and use in the model. Here, we will create synthetic (noisy) data as our light input:
+```@example functional_parameters_circ_rhythm
+using Plots
+tend = 120.0
+ts = collect(0.0:1.0:tend)
+light = sin.(ts/6) .+ 1
+light = [max(0.0, l - rand()) for l in light]
+plot(ts, light; seriestype = :scatter, label = "Experienced light")
+```
+Now this input is only actually defined at the sample points, making it incompatible with a continuous ODE simulation. To enable this, we will use the DataInterpolations package to create an interpolated version of this data, which forms the actual input:
```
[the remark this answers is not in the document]
TorkelE: My idea was to start with as simple as possible (while still being vaguely comprehensible) start-up example. And then have a proper one where I explain what is happening (so as not to have them go through lots of reading before things actually get somewhere). Although I think I might have been overly cautious here. I will try add a note in the first example to make it clear what is going on (and that it will get explained later on).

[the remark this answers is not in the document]
TorkelE: I have modified the inital example so that we actually plot the data interpolation, and added a sentence how DataInterpolation allow us to interpolate the discrete values to a continuous function. I agree that this is a key point.

docs/src/model_creation/functional_parameters.md
TorkelE: fixed

docs/src/model_creation/dsl_basics.md
```diff
@@ -264,6 +264,17 @@ Catalyst comes with the following predefined functions:
...
 
+### [Registration of non-algebraic functions](@id dsl_description_nonconstant_rates_function_registration)
+Previously we showed how user-defined functions [can be used in rates directly](@ref dsl_description_nonconstant_rates_available_functions). For functions containing more complicated syntax (e.g. `for` loops or `if` statements), we must add an additional step: registering it using the `@register_symbolic` macro. Below we define a function which output depends on whether `X` is smaller or larger than a threshold value. Next, we register it using `@register_symbolic`, after which we can use it within the DSL.
+```@example dsl_basics
+threshold_func(X) = (X < 10) ? X : 10.0
+@register_symbolic threshold_func(X)
+rn = @reaction_network begin
+    threshold_func(X), 0 --> X
+    d, X --> 0
+end
```
TorkelE, line 275: currently models in this file are named `rn_1` and so on. This didn't hurt when it was written in the beginning, but annoying to maintain and not sure if there is a point to it. I think I will go in and change them all to `rn`, but not in this PR.

TorkelE: `thresfunc(X) = (X < 10) ? X : 10.0` is maybe not the best example, since this is essentially `min(X,10)` which is supported normally. Still, I liked it because it is something one could plausible use in some situations. Could replace it with `thresfunc(X) = (X < 10) ? 1.0 : 2.0`, which is a bit less relevant (but really do requires `@register_symbolic`).

Potentially we should add a note that this kind of discontinuities might affect numeric performance a bit.

isaacsas: But it should use `ifelse` instead to allow AD to work more generally?

TorkelE: good point. I will think if I can come up with something else so we don't have to get into stuff like this in the most basic tutorial.

TorkelE: I have changed it to
```julia
weirdfunc(x) = round(x) + 2.0
@register_symbolic weirdfunc(X)
rn = @reaction_network begin
    weirdfunc(X), 0 --> X
    d, X --> 0
end
```

docs/src/model_creation/functional_parameters.md
isaacsas: Let’s get an issue to follow up and fix this then. (Post V15.)

isaacsas (commented):
I still think the initial example is likely to just confuse readers as to what a functional parameter is, and would be better off just using a basic function instead of bringing in DataInterpolations (or else add an explanation that you want to mimic an interpolated function through measured data at that earlier point to motivate what is being done). But that said, feel free to merge if you want.

TorkelE:
I added a note at the end of that example to clarify things a bit, but I have now expanded further to
```
!!! note
    For this simple example, $(2 + t)/(1 + t)$ could have been used directly as a reaction rate (or written as a normal function), technically making the functional parameter approach unnecessary. However, here we used this function as a simple example of how discrete data can be made continuous using DataInterpolations, and then have its values inserted using a (functional) parameter.
```
Do you think that is helpful?

isaacsas:
Sure, seems helpful. In any case, finish it as you see fit and then merge. (I should move on to other V15 blocker PRs.)

TorkelE:
Sounds good, I will merge like this (once things pass). Happy to get back to this though. It might even make sense to have a dedicated example tutorial to the various ways of modelling functions.

PR #1192, opened 2025-03-07
Try sort tests out

TorkelE:
Another test failure appeared, not sure how to deal with it. Spent an afternoon isolating it and there is a report at: https://github.com/SciML/ModelingToolkit.jl/issues/3447. Ideally, it gets fixed, but in the meantime (even if this PR one doesn't get merged) I need to have it removed so that we actually can check CI of other stuff (which I will base of this PR).


isaacsas:
Can you push these into that other fixes PR? Let’s work in that branch till tests work.

TorkelE:
No longer an issue.

PR #1193, opened 2025-03-10
Finish dsl src update (reorder functions)

TorkelE:
Follow-up to https://github.com/SciML/Catalyst.jl/pull/985. Builds on https://github.com/SciML/Catalyst.jl/pull/1192 (for tests to pass.

Basically does the reordering of the functions in the DSL file (which was not part of the original PR to make it less messy). With one exception (flipping the order of the `get_psexpr` and `get_usexpr` functions, to have the order with which they appear in the file), all changes are to reorder the option handling functions to appear in the same order in which they were called in the main function.

Also slightly renames two of these option handling functions, so that their names correspond to the option the handle.

### DIFF
--- a/src/dsl.jl
+++ b/src/dsl.jl
@@ -285,7 +285,7 @@ function make_reaction_system(ex::Expr, name)
     ps_declared = extract_syms(options, :parameters)
     vs_declared = extract_syms(options, :variables)
     tiv, sivs, ivs, ivsexpr = read_ivs_option(options)
-    cmpexpr_init, cmps_declared = read_compound_option(options)
+    cmpexpr_init, cmps_declared = read_compounds_option(options)
     diffsexpr, diffs_declared = read_differentials_option(options)
     syms_declared = collect(Iterators.flatten((cmps_declared, sps_declared, ps_declared,
         vs_declared, ivs, diffs_declared)))
@@ -303,7 +303,7 @@ function make_reaction_system(ex::Expr, name)
     ps_inferred = setdiff(ps_pre_inferred, vs_inferred, diffs_inferred)
     syms_inferred = union(sps_inferred, ps_inferred, vs_inferred, diffs_inferred)
     all_syms = union(syms_declared, syms_inferred)
-    obsexpr, obs_eqs, obs_syms = read_observed_option(options, ivs,
+    obsexpr, obs_eqs, obs_syms = read_observables_option(options, ivs,
         union(sps_declared, vs_declared), all_syms; requiredec)
 
     # Read options not related to the declaration or inference of symbols.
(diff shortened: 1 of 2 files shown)


src/dsl.jl
```diff
@@ -614,24 +614,28 @@ end
...
+    # Extracts the independent variables symbols (time and spatial), and returns the output.
+    tiv = ivs[1]
+    sivs = (length(ivs) > 1) ? Expr(:vect, ivs[2:end]...) : nothing
+    return tiv, sivs, ivs, ivsexpr
 end
 
 # When compound species are declared using the "@compound begin ... end" option, get a list
 # of the compound species, and also the expression that creates them.
-function read_compound_option(options)
+function read_compounds_option(options)
```
TorkelE, line 638: the option is called `compounds`, not `compound`.

src/dsl.jl
```diff
@@ -740,36 +736,10 @@ function find_D_call(expr)
...
-    end
-
-    return diffsexpr, diffs_declared
-end
-
 # Reads the observables options. Outputs an expression for creating the observable variables,
 # a vector containing the observable equations, and a list of all observable symbols (this
 # list contains both those declared separately or inferred from the `@observables` option` input`).
-function read_observed_option(options, all_ivs, us_declared, all_syms; requiredec = false)
+function read_observables_option(options, all_ivs, us_declared, all_syms; requiredec = false)
```
TorkelE, line 742: the option is called `observables`, not `observed`. This one is a bit annoying as I named the option differently from argument name in the `ReactionSystem` constructor. At some point we might consider switching the option name.

test/reactionsystem_core/custom_crn_functions.jl
```diff
@@ -49,7 +49,7 @@ let
         ps = rnd_ps(custom_function_network_1, rng; factor)
         t = rand(rng)
         @test f_eval(custom_function_network_1, u0, ps, t) ≈ f_eval(custom_function_network_2, u0, ps, t)
-        @test jac_eval(custom_function_network_1, u0, ps, t) ≈ jac_eval(custom_function_network_2, u0, ps, t)
+        @test_broken jac_eval(custom_function_network_1, u0, ps, t) ≈ jac_eval(custom_function_network_2, u0, ps, t) # Weird error dur to some SciML update. Reported in https://github.com/SciML/ModelingToolkit.jl/issues/3447.
```
isaacsas, line 52: Fixed now?

isaacsas (commented):
What is actually new here vs moving stuff around? Hard to tell...

isaacsas:
If this is nothing more than moving things around and renaming some functions feel free to merge.

TorkelE:
Except for the two mentioned function renamings this just reorders function. Everything else was done in the main PR (the function name inconsistency I only discovered now, but felt to little to make a separate PR).

isaacsas:
Feel free to merge then.

PR #1194, opened 2025-03-17
Add example of how to compute nullclines using bifurcation kit

TorkelE:
WIP


isaacsas (changes requested):
I'm fine with this at a high level modulo my comments above, but this seems like a really complicated approach. Can't you just extract the symbolic equation from the ODE rhs for a given nullcline, and then plot it as an implicit function?

TorkelE:
You might be right about extracting functions. I had some idea why this made sense (for more complicated curves, although here I used a simpler case because I am familiar with it). You might be right that we really can do this generally without involving BifurcationKit. Maybe I will go back to the drawingboard a bit and think if I can streamline things.

isaacsas (approved):
This looks a lot cleaner!

PR #1197, opened 2025-03-19
[WIP] Minor docfixes

TorkelE:
Fixes some minor doc problems I have noted. Still need to figure out how to make this bit
 ```
 @JuliaRegistrator register branch=release-15.0.0
 ```
appear well here (https://docs.sciml.ai/Catalyst/dev/devdocs/dev_guide/#Release-Process). Works in my markdown previewer, but the website will not have it.


docs/src/model_creation/dsl_basics.md
```diff
@@ -103,28 +103,28 @@ end
...
-rn7 = @reaction_network begin
+rn_bidir = @reaction_network begin
     k1, X1 --> X2
     k2, X2 --> X1
 end
 ```
 it is common that reactions occur in both directions (so-called *bi-directional* reactions). Here, it is possible to bundle the reactions into a single line by using the `<-->` arrow. When we do this, the rate term must include two separate rates (one for each direction, these are enclosed by a `()` and separated by a `,`). I.e. the two-state model can be declared using:
 ```@example dsl_basics
-rn7 = @reaction_network begin
+rn_bidir = @reaction_network begin
```
TorkelE, line 113: In some cases these names are nice to show that two models are identical. Now I just tried appending descriptors to their names instead.
(could also do a `rn_this == rn_that` check, but preferred to keep things very straightforward here)

TorkelE:
> Comments on pull requests will not trigger Registrator, as it is disabled. Please try commenting on a commit or issue.

Well this could have become a funny incident...

docs/src/inverse_problems/behaviour_optimisation.md
```diff
@@ -1,4 +1,4 @@
-# [Optimization for non-data fitting purposes](@id behaviour_optimisation)
+# [Optimization for Non-data Fitting Purposes](@id behaviour_optimisation)
```
TorkelE, line 1: I realised that almost all pages have capitalisation throughout the titles, but there were a couple where I hadn't done this. Don't really know, but figured I'd change the ones I hadn't done this to be consistent.

docs/src/model_creation/parametric_stoichiometry.md
```diff
@@ -1,4 +1,4 @@
-# [Symbolic Stochiometries](@id parametric_stoichiometry)
+# [Symbolic Stoichiometries](@id parametric_stoichiometry)
```
TorkelE, line 1: don't recognise the other spelling, but I could be wrong

docs/src/devdocs/dev_guide.md
```diff
@@ -16,13 +16,13 @@ continuing stability of releases. Before making a release one should
...
    issue](https://github.com/SciML/Catalyst.jl/issues/127) with the
-   command:
+   command:  
    
-    ```
-    @JuliaRegistrator register branch=release-15.0.0
-    ```
+   ```
+   @JuliaRegistrator register branch=release-15.0.0
+   ```
```
isaacsas, line 23: Maybe use a raw block?

https://documenter.juliadocs.org/stable/man/syntax/#@raw-format-block

isaacsas (approved):

docs/src/devdocs/dev_guide.md
TorkelE: nope. trying just a `@JuliaRegistrator register branch=release-15.0.0` and checking how that looks

PR #1198, opened 2025-03-19
Add simulation type preservation tests

TorkelE:
Checks that solution values (and time/problem values for good measure) are kept with the input type. Most cases I do not really care about, but if there are sudden changes in behaviours I would like to catch that. Importantly it prevents another case of https://github.com/SciML/ModelingToolkit.jl/issues/3446

Won't pass until https://github.com/SciML/ModelingToolkit.jl/pull/3448 filter though.

### DIFF
--- a/test/simulation_and_solving/simulate_ODEs.jl
+++ b/test/simulation_and_solving/simulate_ODEs.jl
@@ -164,10 +164,13 @@ end
 
 ### Other Tests ###
 
-# Checks that simulations values consistently have Float64 type.
+# Checks that solution values have types consistent with their input types.
+# Check that both float types are preserved in the solution (and problems), while integers are 
+# promoted to floats.
+# Checks that the time types are correct (`Float64` by default or possibly `Float32`), however,
+# type conversion only occurs in the solution, and integer types are preserved in problems.
 let
-    # Create model and check simulation value type (FLoat64 values).
-    # Generally, the tend type should not matter.
+    # Create model. Checks when input type is `Float64` the produced values are also `Float64`.
     rn = @reaction_network begin
         (k1,k2), X1 <--> X2
     end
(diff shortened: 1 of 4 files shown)


test/simulation_and_solving/simulate_ODEs.jl
```diff
@@ -164,6 +164,41 @@ end
...
+    @test eltype(osol.t) == typeof(oprob.tspan[1]) == typeof(oprob.tspan[2]) == Float64
+
+    # Checks that `Int64` values are promoted to `Float64`. 
+    u0 = [:X1 => 1, :X2 => 3]
+    ps = [:k1 => 2, :k2 => 3]
+    oprob = ODEProblem(rn, u0, 1, ps)
+    osol = solve(oprob)
+    @test eltype(osol[:X1]) == eltype(osol[:X2]) == typeof(oprob[:X1]) == typeof(oprob[:X2]) == Float64
+    @test eltype(osol.t) == Float64
+    @test typeof(oprob.tspan[1]) == typeof(oprob.tspan[2]) == Int64
```
isaacsas, line 191: Why do we care about these remaining integers? That seems degenerate since at some point they will have to be converted to floating point.

isaacsas: I feel like this is asking for trouble since converting them earlier seems reasonable and could be implemented at some point.

isaacsas (changes requested):

test/simulation_and_solving/simulate_ODEs.jl
TorkelE: Personally, I really don't care what the type is. However I don't like the idea of theses suddenly jumping around (as it is likely non-intentional and a sign something else might be going on. And if it is intentional it is easy to switch here)

I will ask about it.

TorkelE: So apparently the current policy is for XProblem to not do any type promotion. I'm inclined to adding a comment to the test about the situation, and if how XProblems work changes we change the test (after confirming that it is intentional). The potential drawback seems small.

But if you want to remove I will do so.

isaacsas: I’d still prefer to use problem appropriate types. So use Float32/16/64 for such tests. It just seems like we are testing a degenerate case here that they could reasonably decide to promote at any moment (since ultimately it is getting promoted during solves), and could then break this test.

test/simulation_and_solving/simulate_SDEs.jl
```diff
@@ -383,6 +383,41 @@ end
...
+    @test eltype(ssol.t) == typeof(sprob.tspan[1]) == typeof(sprob.tspan[2]) == Float64
+
+    # Checks that `Int64` values are promoted to `Float64`. 
+    u0 = [:X1 => 1, :X2 => 3]
+    ps = [:k1 => 2, :k2 => 3]
+    sprob = SDEProblem(rn, u0, 1, ps)
+    ssol = solve(sprob, ISSEM())
+    @test eltype(ssol[:X1]) == eltype(ssol[:X2]) == typeof(sprob[:X1]) == typeof(sprob[:X2]) == Float64
+    @test eltype(ssol.t) == Float64
+    @test typeof(sprob.tspan[1]) == typeof(sprob.tspan[2]) == Int64
```
isaacsas, line 410: Same comment about types

TorkelE:
I have removes any checks on whether a time value stored in a problem is an integer. In the cases where integer times are used for creation, we don't perform the check, and just checks that it is promoted to the appropriate time in the solution.

isaacsas (approved):

PR #1200, opened 2025-03-22
Update README.md with installation instructions

isaacsas:
@TorkelE I think we should add some installation instructions in the README that tell people not to install ModelingToolkit versions post 9.59.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,21 @@ Generated models can be used with solvers throughout the broader Julia and
 for sensitivity analysis, parameter estimation, machine learning applications,
 etc).
 
+## Installation 
+Catalyst can be installed as follows. Please note, we suggest only installing ModelingToolkit versions 9.59 and earlier for use with Catalyst at this time as changes in ModelingToolkit as of version 9.60 can break various Catalyst functionality. 
+```julia
+using Pkg
+
+# (optional but recommended) create new environment in which to install Catalyst
+Pkg.activate("catalyst_environment")
+
+# install ModelingToolkit 9.59
+Pkg.add(; name = "ModelingToolkit", version ="9.59")  
+
+# install latest Catalyst release
+Pkg.add("Catalyst")
+```
+
 ## Breaking changes and new features
 
 **NOTE:** Version 14 is a breaking release, prompted by the release of ModelingToolkit.jl version 9. This caused several breaking changes in how Catalyst models are represented and interfaced with.


TorkelE:
I am not a fan, but given people are getting errors due to somehow mixing Catalyst with the latest MTK version we probably have to stick this in there until things clear out.

TorkelE (approved):

ChrisRackauckas:
What is left for the Catalyst updates?

TorkelE:
On the top of my head I think we got:
- https://github.com/SciML/ModelingToolkit.jl/issues/3483 (although can probably just mark as broken as it is not too major.
- This issue https://github.com/SciML/Catalyst.jl/issues/1199 just popped up. There is also a new events related error in the Catalyst tests that was not there a few days ago which I will need to figure out.
- This Catalyst test also started failing suddenly just a few days ago: https://github.com/SciML/Catalyst.jl/blob/master/test/simulation_and_solving/simulate_jumps.jl#L179, which is likely also due to some upstream change.
- We need to figure out what to do about conservation law removal for `NonlinearSystem`s.

I will try to look at this in the next few days. But realistically each of the unreported issues will take me a while to isolate and create reports on, and I was already planning to spend the Sunday catching up on my normal work as I skipped some of it will trying to push through to enable the Catalyst release. Right now it is a bit of an infinite work though, as every time I have figured out some issue another one has popped up.

ChrisRackauckas:
> https://github.com/SciML/ModelingToolkit.jl/issues/3483 (although can probably just mark as broken as it is not too major.

That should be solved now. A new remake function was added for SteadyStateProblem that covers that case now.

> This issue https://github.com/SciML/Catalyst.jl/issues/1199 just popped up. There is also a new events related error in the Catalyst tests that was not there a few days ago which I will need to figure out.
> This Catalyst test also started failing suddenly just a few days ago: https://github.com/SciML/Catalyst.jl/blob/master/test/simulation_and_solving/simulate_jumps.jl#L179, which is likely also due to some upstream change.

Try the latest patch when it comes out in 20 minutes. One change is we removed an old hack and that makes CI more correct https://github.com/SciML/ModelingToolkit.jl/pull/3486 . If one of those don't go away from that update then let me know and I can dive in there.

> We need to figure out what to do about conservation law removal for NonlinearSystems.

Have you setup the codegen to the form of https://github.com/SciML/ModelingToolkit.jl/issues/3458#issuecomment-2725487268  yet? That should be the solution. Now there are some cases where you will need https://github.com/SciML/ModelingToolkit.jl/pull/3452, but we can keep those separate since you already don't handle those cases robustly anyways, so nothing regresses without it, it would just fix new cases (and in fact, that should be the final fix to make conservation laws fully correct, well the previous isn't incorrect but BrownBasic isn't entirely what the user would expect so it would look corrected because it's fixing an ambiguity... long story can elaborate more if you want)

TorkelE:
Saw that https://github.com/SciML/ModelingToolkit.jl/issues/3483 was closed basically as soon I posted. Haven't looked at the solution to https://github.com/SciML/ModelingToolkit.jl/issues/3458#issuecomment-2725487268, but will do and hope for now that the other stuff gets solved as the upstream fixes filters through

isaacsas:
@ChrisRackauckas this test:

https://github.com/SciML/Catalyst.jl/blob/master/test/simulation_and_solving/simulate_jumps.jl#L179

is still failing. Some issue with initialization and variable jump problems (i.e. ODEProblem over JumpProblem when there are no ODEs).

isaacsas:
Here is the error: https://github.com/SciML/Catalyst.jl/actions/runs/14032571458/job/39283295596?pr=1186#step:6:701

PR #1206, opened 2025-03-30
Re-enable conservation laws for `NonlinearSystem`s

TorkelE:
Needs some checking, but this should re-enable the use of conservation law elimination for `NonlinearSystem`s using the approach in https://github.com/SciML/ModelingToolkit.jl/issues/3458#issuecomment-2725487268. Basically requires:
- Re-adding the conservation law to the equations.
- Removing the conservation law from the observables list.
- Creating `initialization_eqs` with the conservation laws (but with species replaced with their initial value parameter.
- Adding the eliminated species back as species of the system.

EDIT, this is not actually the case, my mistake: The updated approach will not enable us to deal with singular Jacobians, which sucks. In the case of Bifurcation diagrams and homotopy continuation this is handled separately though so no problem there (but a problem for normal non-linear system solving).


test/network_analysis/conservation_laws.jl
```diff
@@ -127,22 +127,20 @@ let
     @test osol1[sps] ≈ osol2[sps] ≈ osol3[sps]
 
     # Checks that steady states found using nonlinear solving and steady state simulations are identical.
-    @test_broken begin # Conservation law removal currently not working for NonlinearSystems due to MTK depricating something. https://github.com/SciML/ModelingToolkit.jl/issues/3458, https://github.com/SciML/ModelingToolkit.jl/issues/3411
-        nsys = complete(convert(NonlinearSystem, rn; remove_conserved = true, remove_conserved_warn))
-        nprob1 = NonlinearProblem{true}(nsys, u0, p; guesses = [nsys.Γ => [0.0]])
```
TorkelE, line 132: Providing `guesses` here now yields errors.

ChrisRackauckas:
> The updated approach will not enable us to deal with singular Jacobians, which sucks.

But it gets rid of singular Jacobians?

TorkelE:
No, you are right, I just didn't look through things properly yet 😅

ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
```diff
@@ -36,7 +36,7 @@ Notes:
 ```
   """
 function Catalyst.hc_steady_states(rs::ReactionSystem, ps; filter_negative = true,
-        neg_thres = -1e-20, u0 = [], kwargs...)
+        neg_thres = -1e-15, u0 = [], kwargs...)
```
TorkelE, line 39: On the latest Julai version, the accuracy of HC.jl seems slightly affected. I.e. before values that should be zero could be negative, but _incredibly_ small. Now I have seen values around -1e-17 (which caused test failures). I will tell Paul at HC.jl about it, just in case he is interested,

ChrisRackauckas:
Have you tried using the standard MTK HC interface? I would hope the Catalyst one isn't needed anymore, and it should make things like hill functions now handled through the rationalization passes.

TorkelE:
Doesn't that still eliminate all but one of the solutions though? Last time I checked it only returned a single solution, which mostly defeated the purpose of using it (it might have been updated since, but cannot find anything about it). Long-term using the MTK interface is still the plan though.

isaacsas:
@ChrisRackauckas can you point us to the official documentation on the MTK interface? Searching the last MTK release's docs for homotopy doesn't seem to give any results?

ChrisRackauckas:
It's now just a solver alg in NonlinearSolve.jl

https://docs.sciml.ai/NonlinearSolve/stable/api/homotopycontinuation/

We do need to figure out how to document that better. The AllRoots option controls the return @TorkelE .

ChrisRackauckas:
Though for the depolynomialization you'll want to target it through https://github.com/SciML/ModelingToolkit.jl/blob/master/src/systems/nonlinear/homotopy_continuation.jl#L536.

I'll chalk this up to the fact that MTK kind of needs a complete doc structure overhaul. Maybe we can dedicate one of the upcoming symbolic numerics thrusday meetings to that and just get it done.

@TorkelE did you de-register the Hill functions? Since that would be required for the rationalization.

TorkelE:
> @TorkelE did you de-register the Hill functions? Since that would be required for the rationalization.

still registered, but we can probably change. Not sure what the wider implications for various cases would be though.

isaacsas:
(Also, we should probably move these discussions to issues.)

src/reactionsystem_conversions.jl
```diff
@@ -554,19 +554,41 @@ function Base.convert(::Type{<:NonlinearSystem}, rs::ReactionSystem; name = name
         as_odes = false, include_zero_odes)
     eqs, us, ps, obs, defs = addconstraints!(eqs, fullrs, ists, ispcs; remove_conserved)
 
+    # Makes modifications to handle conservation laws. Currently requires special consideration 
+    # due to MTK updates. Long-term, hopefully we can have a unified solution across all systems.
+    # This, current, NonlinearSystem solution is based on https://github.com/SciML/ModelingToolkit.jl/issues/3458#issuecomment-2725487268.
+    eqs, obs, initialization_eqs = handle_nlsys_cons_laws(rs, eqs, obs, remove_conserved)
+    us = union(species(rs), us)
```
TorkelE, line 561: Essentially, `addconstraints` removes the elimiantesspecies from `us`. But using the new MTK approach these needs to be there, so I add them back in. I think this should work, but you might know some obscure side case where it does not (alternatively we can loop through the conservation equations and fetch them from there).

test/network_analysis/conservation_laws.jl
```diff
@@ -127,22 +127,20 @@ let
...
-        ssprob2 = SteadyStateProblem(rn, u0, p)
-        ssprob3 = SteadyStateProblem(rn, u0, p; remove_conserved = true, remove_conserved_warn)
-        nsol1 = solve(nprob1, NewtonRaphson(); abstol = 1e-8)
-        # Nonlinear problems cannot find steady states properly without removing conserved species.
-        nsol3 = solve(nprob3, NewtonRaphson(); abstol = 1e-8)
-        sssol1 = solve(ssprob1, DynamicSS(Tsit5()); abstol = 1e-8, reltol = 1e-8)
-        sssol2 = solve(ssprob2, DynamicSS(Tsit5()); abstol = 1e-8, reltol = 1e-8)
-        sssol3 = solve(ssprob3, DynamicSS(Tsit5()); abstol = 1e-8, reltol = 1e-8)
-        @test nsol1[sps] ≈ nsol3[sps] ≈ sssol1[sps] ≈ sssol2[sps] ≈ sssol3[sps]
-    end
```
TorkelE, line 145: this shoudl all be unchanged, but Git is getting weird ideas.

test/network_analysis/conservation_laws.jl
```diff
@@ -333,14 +378,12 @@ let
...
         XProblem(rn, u0, ps; remove_conserved = true)
-        # @test_nowarn XProblem(rn, u0, ps)
-        # @test_logs (:warn, r"You are creating a system or problem while eliminating conserved quantities. Please *") XProblem(rn, u0, ps; remove_conserved = true)
-        # @test_nowarn XProblem(rn, u0, ps; remove_conserved_warn = false)
-        # @test_nowarn XProblem(rn, u0, ps; remove_conserved = true, remove_conserved_warn = false)
+        @test_nowarn XProblem(rn, u0, ps)
+        @test_logs (:warn, r"You are creating a system or problem while eliminating conserved quantities. Please *") XProblem(rn, u0, ps; remove_conserved = true)
+        @test_nowarn XProblem(rn, u0, ps; remove_conserved_warn = false)
+        @test_nowarn XProblem(rn, u0, ps; remove_conserved = true, remove_conserved_warn = false)
     end
```
TorkelE, line 387: These tests (and anything with `remove_conserved_warn`) is actually removed in https://github.com/SciML/Catalyst.jl/pull/1203

test/reactionsystem_core/coupled_equation_crn_systems.jl
```diff
@@ -167,7 +167,7 @@ let
     u0 = [X => 0.1, A => 16.1/2]
     nlprob = NonlinearProblem(coupled_rs, u0, ps, structural_simplify = true)
     nlsol = solve(nlprob)
-    @test nlsol ≈ [2.0, 3.0]
```
TorkelE, line 170: because `structural_simplify` sometimes removes one of the variables now.

src/reactionsystem_conversions.jl
isaacsas: I think a better approach would be to have a kwarg to addconstraints that just does the right changes for NonlinearSystems (or a version of it for nonlinear systems), rather than trying to remove and move around stuff.

Also, at this point in the conversion function you should be using `fullrs`, the flattened system, not `rs`.

isaacsas: We can make `addconstraints` return the initialization equations as the final option (which can just be empty if there are none as for other systems). Then we pass them on. 

Another advantage of modifying it to do things the correct way would be that it is subsequently trivial to swap to it for other systems types later on.

isaacsas: Basically, just add a kwarg like `treat_obs_as_eqs`, and have a sub-case for whether it is true or false (the default) in this code: https://github.com/SciML/Catalyst.jl/blob/06aed2e7d13ff79d2895e61651f4ca8a778e5868/src/reactionsystem_conversions.jl#L391-L401

test/reactionsystem_core/coupled_equation_crn_systems.jl
isaacsas: This is better anyways -- we shouldn't rely on ordering of `us` or even what remains a `u`.

test/reactionsystem_core/coupled_equation_crn_systems.jl
```diff
@@ -380,7 +380,7 @@ end
     ps = [A1 => 0.1, B1 => 1.0, C1 => 10.0]
 
     # Create ODE structures.
-    oprob = ODEProblem(coupled_rs, u0, tspan, ps; structural_simplify = true)
```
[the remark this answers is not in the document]
isaacsas: Wait, so it is always going to warn if you pass an initial condition for the eliminated variable, even if there is an associated initialization equation?

isaacsas (changes requested):

src/reactionsystem_conversions.jl
```diff
@@ -554,19 +554,41 @@ function Base.convert(::Type{<:NonlinearSystem}, rs::ReactionSystem; name = name
...
+
     # Throws a warning if there are differential equations in non-standard format.
     # Next, sets all differential terms to `0`.
     all_differentials_permitted || nonlinear_convert_differentials_check(rs)
     eqs = [remove_diffs(eq.lhs) ~ remove_diffs(eq.rhs) for eq in eqs]
 
     NonlinearSystem(eqs, us, ps;
```
isaacsas, line 568: I missed this: I think we need to return a `structural_simplify`'ed system to ensure the conservation law equation gets removed. (And we should document that this is different behavior than other system types...)

src/reactionsystem_conversions.jl
```diff
@@ -509,13 +523,17 @@ Keyword args and default values:
...
-  the number of equations.
+  underlying set of reactions (ignoring coupled ODE or algebraic equations). For each
+  conservation law one steady-state equation is eliminated, and replaced with the
+  conservation law. This ensures a non-singular Jacobian. When using this option, it is
+  recommended to call `ModelingToolkit.structural_simplify` on the converted system to then
+  eliminate the conservation laws from the system equations.
 """
 function Base.convert(::Type{<:NonlinearSystem}, rs::ReactionSystem; name = nameof(rs),
         combinatoric_ratelaws = get_combinatoric_ratelaws(rs),
-        include_zero_odes = true, remove_conserved = false, checks = false,
```
isaacsas, line 517: There is/was no reason to have `include_zero_odes` for nonlinear systems. It just results in a system equation of the form `0 ~ 0` which is not useful (and is probably indicative of an issue, since if the number of unknowns and equations are only balanced with such an equation it means the system is underdetermined).

isaacsas:
@TorkelE I had some merge issues on the tests when updating to master given the divergence between the PR and master now, please double check the tests are ok now.

TorkelE:
Right, yeah, I probably should make a commit making these tests up to date (also running some of the new `remake` tests on `NonlinearSystem`s as well, now when those works). When you are done with the update I will do one more update solely going through that the tests are good.

TorkelE:
Yes! I will have a look at the tests.

test/reactionsystem_core/coupled_equation_crn_systems.jl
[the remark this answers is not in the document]
ChrisRackauckas: Yes because that's an overdetermined system. If you have:

```julia
D(x) ~ x*y
x + y ~ 0
```

and give it `x(0) = 0, y(0) = 1` it's overdetermined. You really only should be supplying one initial condition. So it warns that you're in a not so great case, but then uses nonlinear least squares to see if it's okay. If the system I wrote above, it's an initialization failure, since it's impossible for `x(0) = 0, y(0) = 1` to satisfy `x + y ~ 0`. Normally the first remedy a user should do when they have an overdetermined initialization that is giving failures is simply to remove the extra conditions, because what it generally means is that one of the extra conditions is just wrong. So it's good to throw this warning because you're indicating to the user "if things down the line are failing, the reason is because you have an extra condition that you should really clean up sooner rather than later".

src/reactionsystem_conversions.jl
ChrisRackauckas: Most users should be calling structural simplify anyways before problem generation?

test/reactionsystem_core/coupled_equation_crn_systems.jl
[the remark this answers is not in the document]
isaacsas: Ahh, I missed that this example was a DAE...

test/network_analysis/conservation_laws.jl
```diff
@@ -124,22 +124,20 @@ let
...
-        ssprob3 = SteadyStateProblem(rn, u0, p; remove_conserved = true)
-        nsol1 = solve(nprob1, NewtonRaphson(); abstol = 1e-8)
-        # Nonlinear problems cannot find steady states properly without removing conserved species.
-        nsol3 = solve(nprob3, NewtonRaphson(); abstol = 1e-8)
-        sssol1 = solve(ssprob1, DynamicSS(Tsit5()); abstol = 1e-8, reltol = 1e-8)
-        sssol2 = solve(ssprob2, DynamicSS(Tsit5()); abstol = 1e-8, reltol = 1e-8)
-        sssol3 = solve(ssprob3, DynamicSS(Tsit5()); abstol = 1e-8, reltol = 1e-8)
-        @test nsol1[sps] ≈ nsol3[sps] ≈ sssol1[sps] ≈ sssol2[sps] ≈ sssol3[sps]
-    end
+    nsys = complete(convert(NonlinearSystem, rn; remove_conserved = true))
```
ChrisRackauckas, line 127: Test the structural simplification form, since that's the most standard usage here?

TorkelE: Good point, yeah, that should be changed

test/reactionsystem_core/coupled_equation_crn_systems.jl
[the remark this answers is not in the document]
isaacsas: Outside of specifically testing for degenerate situations, I feel like we shouldn't have tests that intentionally fail initialization due to passing too many initial conditions. i.e. our normal correctness tests should be created for the correct workflow.

src/reactionsystem_conversions.jl
isaacsas: Because of the issue with observed getting removed, we don't currently show it in our docs or enable it by default, so a user would have to know about it to use it.

test/reactionsystem_core/coupled_equation_crn_systems.jl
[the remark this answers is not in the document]
TorkelE: I will check them. However, I think many of these cases are an artefact of how conservation laws are implemented or something (don't remember details, but I do remember looking ta the tests and deciding that the appropriate workflow was used). In all cases, by internal design (i.e. not because values are well selected) the conditions are such that the conservation law works out.

src/reactionsystem_conversions.jl
ChrisRackauckas: But I assumed that was fixed just by putting them into the equation list?

test/network_analysis/conservation_laws.jl
isaacsas: We should test both.

test/reactionsystem_core/coupled_equation_crn_systems.jl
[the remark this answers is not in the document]
isaacsas: Does the normal conservation law workflow result in initialization warnings for users? (For systems without user DAEs, i.e. only the conserved equation(s))? They are going to be quite confused about such warnings I'd think.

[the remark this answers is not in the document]
TorkelE: I will check this through.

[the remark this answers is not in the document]
ChrisRackauckas: Not necessarily. Any complete description of the ODE system by definition is going to be an overdetermined description of the system with conservation laws, but only if the conservation parameter is set. For example, with ROBER as an ODE you have to give [y1,y2,y3] as ODE initial conditions. As an ODE, there is no statement that `y1 + y2 + y3 = 1`. We know from the chemical reactions that the conservation law should be true, but the ODE does not have any notion of that. So you need 3 initial conditions in order to have a complete description of the IVP. However, when you then replace one of the ODE equations with  `y1 + y2 + y3 = 1`, you now have an overdetermined system if you give `[y1,y2,y3]` as the initial conditions, since knowing two of the variables initial conditions determines the 3rd. 

So if you have:

```julia
Initial(y1) ~ const
Initial(y2) ~ const
Initial(y3) ~ const
Initial(y1) + Initial(y2) + Initial(y3) = 1
```

then you clearly have an overdetermined system, 4 equations for 3 unknowns. However, what saves you is that you don't actually know 1. What you have is:

```julia
Initial(y1) ~ const
Initial(y2) ~ const
Initial(y3) ~ const
Initial(y1) + Initial(y2) + Initial(y3) = Gamma
```

where `Gamma = missing` means it's an initialized parameter. This means that the initialization system is to solve for `[Initial(y1), Initial(y2), Initial(y3), Gamma]`, which is a well-determined system.

So your system of computing conservation laws always gives a well-determined system from the same initial condition setup if every algebraic equation is also adding a missing parameter, which I believe every single one here does. So with that, I would believe that you should be able to take any ODE and turn it into the conservation equation and do the eliminations, and the user shouldn't have to know or care because this is well-defined iff the ODE IVP was well-defined. 

I would argue that, since we now have all of this down, it should just do this: the standard Catalyst thing should just make and simplify the DAE, as that would be more numerically accurate and also accelerate a lot of cases.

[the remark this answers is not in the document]
TorkelE: Long-term we can look at creating DAEs. However, right now this is just not possible as `structural_simplify` does not work for Catalyst-generated SDEs. Practically, since we _know_ the conservation laws (which is typically not the case with `structural_simplify` but one of the advantages we get from Catalyst models having specific structures) it seems simpler to build stuff directly here, and then people can use/not use `structural_simplify` depending on their needs (e.g. if they would like to preserve model structures species etc., although I agree that in the long run this might be more narrower cases).

src/reactionsystem_conversions.jl
isaacsas: But do we have any guarantee `structural_simplify` will turn the previous observable equations into observables and not select one or more other equations? If a user model from SBML declares something an observable I don't want us to suddenly turn it into an equation and potentially remove some other equation of the model when `structural_simplify` is called. Such a change also means we are dropping the information from that user that the observable equation is just defining an observable of the state, and is not representing an equation that is supposed to be part of the dynamics. 

I understand that this shouldn't result in any correctness issues, but I'd prefer we not overrule a user's model specification (at least as our default behavior).

test/reactionsystem_core/coupled_equation_crn_systems.jl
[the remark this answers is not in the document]
TorkelE: So I think the warnings are from `remake`. Basically, if you have provided everything first, and then updates a say `\Gamma`, then a warning is thrown. Which probably should not be the case in `remake`. I will try to make a minimal example and make an MTK issue.

src/reactionsystem_conversions.jl
isaacsas: I guess if the variable the observable defines is only used in that one equation we would know that `structural_simplify` would always eliminate it?

TorkelE: I kind of agree with Sam here, it would allow us to actually start using `structural_simplify` a lot more liberally across Catalyst (alternatively we must wait until we have figured out a new uniform way of handling observables across Catalyst-generated systems)

ChrisRackauckas: Read https://github.com/SciML/ModelingToolkit.jl/issues/3536#issuecomment-2776409272. We do not always eliminate because it's not always correct. Assuming it's always correct is exactly what is leading you to bugs. What we guarantee is that every equation in the equation list is always satisfied, and we have checks to ensure that contract can be upheld or error (for example, circular definitions, incorrect types, incorrect clocks, etc.). It's as trivial of a change as saying that we will support you to write x86 assembly code. Before committing to that, can we take a step back and figure out the motivation here? This to me this is an extreme case of over optimizing: you have not pointed out any major performance benefit you can get, so what even is the motivation to do it in the first place?

isaacsas: What is leading us to bugs is MTK changing the long-standing behavior of basic functionality in non-breaking releases.

TorkelE: So, the main motivation to not force `structural_simplify` (if that is what you mean) is that:
- It disables observables.
- It does not leave any guarantees of what is variables, equations, etc. of the new system. Hence we think that users should be able to not use it in cases where this matter (i.e. they might want to do something which is not related to normal simulation.solving workflows. I.e. right now I am not sure how liberal structural simplification will interact with BifurcationKit etc.

ChrisRackauckas: > it would allow us to actually start using structural_simplify a lot more liberally across Catalyst (alternatively we must wait until we have figured out a new uniform way of handling observables across Catalyst-generated systems)

If you always put it in the equations, as has been the only documented way to do it for years, then the observed equations will always be guaranteed to be correctly satisfied in all systems. That is true for every AbstractSystem: ODESystem, SDESystem, OptimizationSystem, NonlinearSystem, DiscreteSystem, PDESystem, ImplicitDiscreteSystem. All of the coming constructions around optimal control are also setup the same way. "What equations should be true" is precisely the equations list, and you should put any equation you want to be true there.

ChrisRackauckas: > What is leading us to bugs is MTK changing the long-standing behavior of basic functionality in non-breaking releases.

No. Can we just get on a call and hash this out? This is getting to utter insanity levels at this point. Yes, I know that in 2021 Yingbo pointed out an undocumented that for MTK v5 could save you a little bit of time. Your insistence to not drop these hacks is still what's leading you to these bugs. I am repeatedly trying to tell you, please please please stop relying on the hacks and please just do the thing that is documented, has correctness checks, and is guaranteed to give you the right behavior. You have now insisted for months that you do not want to use the documented API and instead want to keep mucking with compiler internals, even when I have explained how your mathematical assumptions are not correct. I thoroughly do not understand: if your stated goal is to actually have stability, why are you insisting on not using the stable documented interfaces? Why is this manual optimization so important that you need to have Catalyst keep breaking?

TorkelE: There are problems with your suggested solution e.g. it won't work for `JumpSystem`s or our `SDESystem`s. I do agree we probably should have a call about it (but not now, almost midnight here!).

ChrisRackauckas: > There are problems with your suggested solution e.g. it won't work for JumpSystems or our SDESystems.

"Our SDESystems" our being a key word 😅 https://github.com/SciML/Catalyst.jl/blob/d71d11a75c5fa5593f2b1425ef26558486cdbac1/src/reactionsystem_conversions.jl#L609-L610. 

Okay so then two follow ups:

1. Update the SDE construction to Brownian variables https://docs.sciml.ai/ModelingToolkit/dev/tutorials/stochastic_diffeq/. 
2. Update JumpSystem to allow for equations.

That aligns with the v10  plans to remove all of the extra systems and collapse to just `System` anyways.

TorkelE: So the major blockers right now are observables for stochastic systems. 

I think you are right about the BifurcationKit (or rather, if you say that you are right, then you are, you know this better than me!). However, the "not forcing structural simplify" is more of a future proof thing for other developers. I do believe that in the long term, once everything else is sorted, we can probably throw in `structural_simplify` by default for basically all workflows. However, I am also keen to allow other packages to build upon Catalyst, and here I'd like to keep the option open for non-simplified systems (i.e. in my ideal work `complete` does everything that is required, and `structural_simplify` does the cool stuff on top of that). To be honest I do not know, however, I could imagine there are some obscure network analysis workflow where someone would like to preserve equations. However, for simulation/solving workflows I agree with the idea that one should be able to call `structural_simplify` and nothing is affected.

TorkelE: I discussed this earlier. The problem with the new `SDESystem`s is that they are just `System`s. This causes problems with Catalyst since the idea is that we have a `ReactionSystem` which can be converted to different forms of systems, making `convert(System, rs::ReactionSystem)` weird. We can add another argument to try and sort this out, but I didn't want to do this when `System` was essentially just SDEs and I didn't know where this new interface was going. If MTK is changing to use `System`s more widely I could look at things again. 

However, this and also supporting `structural_simplify` and equations for `JumpSystem`s is going to be loads of work, and I am not sure when I will be able to squeeze that in (I have a more then full time job to do, and right now SciML development is saoking up pretty much most od my evenings and weekends). Right now we have something which actually works up to our specifications. Next when I know what MTK is planning v10 and forward I can try an figure out how to fit Catalyst in this new framework.

ChrisRackauckas: > However, the "not forcing structural simplify" is more of a future proof thing for other developers. I do believe that in the long term, once everything else is sorted, we can probably throw in structural_simplify by default for basically all workflows.

To be clear, the standard workflow is:

```julia
sys = System(...)
ssys = structural_simplify(sys)
prob = Problem(ssys)
```

I wouldn't expect that to change because we need a point where the user says "this is a final system, please do your modifications under the assumption that the system isn't going to change anymore" (since MTK has a lot of features around composing models, so in general you cannot assume that building a system means you're done). So I'm not advocating for it to be done internally and automatically. What I'm saying is, the standard workflow should have just about every user calling structural simplify before building the problem. If that's documented as the standard way to things through all of MTK and Catalyst, I don't think we need auto-simplify. If there is a place to auto-simplify it it's in the problem constructor, though I'm not sure if we're ready to commit to that. Maybe v11...

ChrisRackauckas: > If MTK is changing to use Systems more widely I could look at things again.

The plan is to get to having just one System type. The big idea is:

```julia
sys = System(...)
ssys = structural_simplify(sys)
prob = Problem(ssys)
sol = solve(prob)
```

The user shouldn't have to know what kind of system they are using, what kind of problem type are solving, or what kind of solver they are using. It's all merging into one with complete automation. OptimizationSystem and ODESystem are merging with the optimal control work, ODESystem and SDESystem have merged with the Brownian variables, https://github.com/SciML/ModelingToolkit.jl/issues/3537#issuecomment-2777132166 merges JumpSystem and ODESystem. We will need to of course still support calls like `ODEProblem`, but the key there is to check "do you have brownian variables or cost function? If so, then error". So then:

> Catalyst since the idea is that we have a ReactionSystem which can be converted to different forms of systems, making convert(System, rs::ReactionSystem) weird

I do think we need to reconsider that a bit. `convert(ODESystem, rs::ReactionSystem)` is not uniquely defined. For example, one interpretation of a chemical reaction system as a set of ODEs is to write out the Fokker-Plank equations and semi-discretize that into ODEs. Another interpretations is to take the Chemical Langevin equation and then do the linear noise moment expansion to get ODEs for the means and variances. "I meant the Mass Action Kinetics interpretation of the chemical reaction of the ODE", we're getting to the point where that is only one of the ODE representations of the the chemical reaction, so the interface should be more explicit. `convert(MassActionODE, rs::ReactionSystem)` for example. But that can wait, and it can have a long deprecation time. But it is something I have been thinking about.

ChrisRackauckas: > However, this and also supporting structural_simplify and equations for JumpSystems is going to be loads of work, and I am not sure when I will be able to squeeze that in (I have a more then full time job to do, and right now SciML development is saoking up pretty much most od my evenings and weekends).

Don't worry, you've done your part. We should add these bits to MTK, and then maybe @vyudu can help move the observed over and the update of SDEs to use System. I've also got a few full time jobs and open source software isn't one, but at least we have some hires specifically for these topics 😅

TorkelE: So right now `convert(ODESystem, rs::ReactionSystem)` is defined because we say it correspond to the RRE ODE. But I agree with you that a long-term goal would be stuff like `convert(MassActionODE, rs::ReactionSystem)` (and another reason to implement more CRN interpretations!), and then a workflow like you suggest. It could also allow for more general stuff like where we automatically assign (dynamic) hybrid simulations to different species and reactions (i.e. ODE, Jump etc.), and users could just do `convert(SingleTimeCourse, rs)` or something, that'd be really cool.

Right now I think everything is in place and works for v15 actually, and then we will have to have a discussion about long-term goals and stuff. For now it would be nice to focus on stuff like expanding the hybrid simulations that Sam have implemented while keeping stuff stable. And then once these new stuff in MTK are ready and implemented we can assign a period to making the switch to that.

vyudu: Forgot to respond to this but yes I am happy to help adapt stuff here. (probably will need to get caught up on what's happening in a call though.)

isaacsas:
@TorkelE if everything passes and the doc build looks ok I am happy to merge this (and will do so).

test/network_analysis/conservation_laws.jl
```diff
@@ -408,3 +495,107 @@ let
...
     # @test sol[X[2]][end] ≈ 4.0
 end
+
+# Check conservation law elimination warnings (and the disabling of these) for `NonlinearSystem`s
+# and `NonlinearProblem`s.
+let
+    # Create models.
+    rn = @reaction_network begin
+        (k1,k2), X1 <--> X2
+        (k3,k4), X1 + X2 --> 2X3
```
TorkelE, line 505: I intentionally sat `X1 + X2 --> X3` to try an make the conservation law slightly messier than usual (i.e. to try and avoid always having too trivial systems in the tests).

isaacsas: How is this not erroring? It is using a non-reversible reaction as if it is reversible.

isaacsas: Fell free to add back another test with your more complicated example, but let’s keep this one since it explores a bunch of remake cases.

PR #1209, opened 2025-04-01
Examples of ways to model noise

TorkelE:
A short tutorial of ways to model noise. Builds on https://github.com/SciML/Catalyst.jl/pull/1191.

In principle ready, but I want to read through it one or two more times before I am confident.


docs/pages.jl
```diff
@@ -22,6 +22,7 @@ pages = Any[
         "Examples" => Any[
             "model_creation/examples/basic_CRN_library.md",
             "model_creation/examples/programmatic_generative_linear_pathway.md",
+            "model_creation/examples/noise_modelling_approaches.md",
```
isaacsas, line 25: I would put this second to last in the list. That would keep the ODE-based models next to each other, and to some degree I think they are simpler examples than this one.

isaacsas (changes requested):
One minor comment otherwise feel free to merge (I haven't really read through it carefully, just very quickly skimmed).

docs/pages.jl
TorkelE: sounds good, will update

PR #1210, opened 2025-04-01
Example: Measuring activation time distribution

TorkelE:
Combines a decent couple of concepts from various parts of the docs, and also shows how to use `EnsembleProblem` output functions.


docs/Project.toml
```diff
@@ -40,6 +40,7 @@ SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
 SciMLSensitivity = "1ed8b502-d754-442c-8d5d-10ac956f44a1"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+StatsPlots = "f3b207a7-027a-5e70-b257-86293d7955fd"
```
TorkelE, line 43: A bit annoying to add a dependency for a tutorial. However, the `density` function feels quite important for stochastic -related analysis, so probably something users should know about and which we might want to use in other tutorials as well. Also, it is part of the Plots system, which should make it less of an addition.

isaacsas: We probably should at some point just switch everything to Makie-verse given we are now pulling it in, but that is a longer-term goal I guess.

docs/src/model_simulation/examples/activation_time_distribution_measurement.md
```diff
@@ -0,0 +1,58 @@
...
+```
+Next, to perform stochastic simulations of the system we will create an `SDEProblem`. Here, we will need to assign parameter values to $Kᵢ$ and $Kₐ$, but not to $K$ (as its value is controlled by its default and the event). Also note that we start the simulation at a time $t = -200 < 0$. This ensures that by the input time ($t = 0$), the system has (more or less) reached its (inactive) steady state distribution. It also means that the activation time can be measured exactly as the simulation time at the time of activation (as this will be the time from the input at $t = 0$).
+```@example activation_time_distribution_measurement
+u0 = [:X => 10.0]
+tspan = (-200.0, 2000.0)
+ps = [:v0 => 0.1, :v => 2.5, :Kᵢ => 1000.0, :Kₐ => 40.0, :n => 3.0, :deg => 0.01]
+sprob = SDEProblem(sa_model, u0, tspan, ps)
+nothing # hide 
+```
+We can now create a simple `EnsembleProblem` and perform an ensemble simulation (as described [here](@ref ensemble_simulations)). Please note that the system s an event which modifies its parameters, hence we must add the `safetycopy = true` argument to `EnsembleProblem` (else, subsequent simulations would start with $K = Kₐ$).
```
isaacsas, line 22:
```suggestion
We can now create a simple `EnsembleProblem` and perform an ensemble simulation (as described [here](@ref ensemble_simulations)). Please note that the system has an event which modifies its parameters, hence we must add the `safetycopy = true` argument to `EnsembleProblem` (else, subsequent simulations would start with $K = Kₐ$).
```

docs/src/model_simulation/examples/activation_time_distribution_measurement.md
```diff
@@ -0,0 +1,58 @@
...
+
+Next, we wish to measure the distribution of these activation times. First we will create a [callback](https://docs.sciml.ai/DiffEqDocs/stable/features/callback_functions/) which terminates the simulation once it has reached a threshold. This both ensures that we do not have to expend unnecessary computer time on the simulation after its activation, and also enables us to measure the activation time as the final time point of the simulation. Here we will use a [discrete callback](https://docs.sciml.ai/DiffEqDocs/stable/features/callback_functions/#SciMLBase.DiscreteCallback). By looking at the previous simulations, we determine $X = 100$ as a suitable activation threshold. We use the `terminate!` function to terminate the simulation once this value has been reached.
+```@example activation_time_distribution_measurement
+condition(u, t, integrator) = integrator[:X] > 100.0
+affect!(integrator) = terminate!(integrator)
+callback = DiscreteCallback(condition, affect!)
```
isaacsas, line 35: Can you not use a functional affect and store this in the system?

docs/src/model_simulation/examples/activation_time_distribution_measurement.md
```diff
@@ -0,0 +1,58 @@
...
+nothing # hide
+```
+Next, we will perform our ensemble simulation. By default, for each simulation, these save the full trajectory. Here, however, we are only interested in the activation time. This enables us to utilise an [*output function*](https://docs.sciml.ai/DiffEqDocs/dev/features/ensemble/#Building-a-Problem). This will be called at the end of every single simulation and determine what to save (it can also be used to potentially rerun individual simulations, however, we will not use this feature here). Here we create an output function which saves only the simulation's final time point. We also make it throw a warning if the simulation reaches the end of the simulation time frame ($t = 2000$) (this indicates a simulation that never activated, the occurrences of such simulation would cause us to underestimate the activation times). Finally, just like previously, we must set `safetycopy = true`.
+```@example activation_time_distribution_measurement
+function output_func(sol, i)
+    (sol.t[end] == tspan[2]) && @warn "A simulation did not activate during the given time span."
+    return (sol.t[end], false)
+end
```
isaacsas, line 43: A fancier approach would be to use the integrator interface and keep stepping until a callback calls terminate! (assuming it will always be called eventually, but could with some non-zero probability be after `tspan[2]` -- I haven't checked the problem carefully enough to understand if that is the case here).

docs/src/model_simulation/examples/activation_time_distribution_measurement.md
```diff
@@ -0,0 +1,58 @@
...
+eprob = EnsembleProblem(sprob; output_func, safetycopy = true)
+esol = solve(eprob, ImplicitEM(); trajectories = 250, callback)
+nothing # hide
+```
+Finally, we can plot the distribution of activation times. For this, we will use the [StatsPlots.jl](https://docs.juliaplots.org/latest/generated/statsplots/) package's `density` function (essentially a smoothed histogram). The input to `density` is the activation times (which our output function has saved to `esol.u`).
+```@example activation_time_distribution_measurement
+using StatsPlots
+density(esol.u; label = "Activation time distribution", xlabel = "Activation time")
+```
```
isaacsas, line 52: Why not just use a normal histogram with the option to have it normalize to a pdf? That shows the actual sample noise in the bins instead of smoothing everything out, and avoids the StatsPlots dependency.

isaacsas (changes requested):
Feel free to merge after you've updated per my comments as you see fit.

docs/src/model_simulation/examples/activation_time_distribution_measurement.md
TorkelE: Does system events store non-trivial affects like `termiante`? If it is easier to do I am happy to change, however, it doesn't hurt to show how more complicated events can be implemented through callbacks.

docs/src/model_simulation/examples/activation_time_distribution_measurement.md
TorkelE: I agree that this would be a cooler way. Maybe add a note about it for now? Now in principle that this is doable, but haven't actually seen it in practise.

docs/src/model_simulation/examples/activation_time_distribution_measurement.md
TorkelE: I generally have preferred to use `density` as I feel those plots are better and easier to work with. But if avoiding `StatsPlots` is priority I can plot a histogram and then link to the `desnity` function suggesting that readers should try it out as well?

docs/Project.toml
TorkelE: agreed

PR #1211, opened 2025-04-02
Add additional Jacobian tests (including sparsity)

TorkelE:
After the latest test failures in the spatial tests I checked and realised that we don't actually properly test whether the generated sparse Jacobians are correct. I guess, ironically, a lot of the core feature tests were written way back in time before when we made more assumptions of upstream correctness (and our own, I guess, still think the early tests were quite good though). Hence a lot of stuff is actually caught in extensions/niche feature tests that are actually testing other stuff.

I haven't actually isolated the latest problems, so not sure if they are from Catalyst or somewhere else (or even if they are a problem). However, we get failures here as well, so not related to the spatial stuff that is actually failing. I also only get failures of these tests on the latest version, so it is something recent that has happened.

### DIFF
--- a/test/simulation_and_solving/jacobian_construction.jl
+++ b/test/simulation_and_solving/jacobian_construction.jl
@@ -1,7 +1,7 @@
 ### Prepares Tests ###
 
 # Fetch packages.
-using Catalyst, DiffEqBase, Test
+using Catalyst, DiffEqBase, OrdinaryDiffEqRosenbrock, Test
 
 # Sets stable rng number.
 using StableRNGs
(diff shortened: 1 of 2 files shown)


TorkelE:
This is the likely culprit https://github.com/SciML/OrdinaryDiffEq.jl/pull/2567

test/simulation_and_solving/jacobian_construction.jl
```diff
@@ -67,4 +67,107 @@ let
                     p[k3]*u[B] p[k3]*u[A] -p[k4]-3 * p[k5] * u[C]^2 / 2]
         @test test_jac ≈ real_jac
     end
-end
\ No newline at end of file
+end
+
+# Checks that the Jacobians (dense and sparse) are identical for different system types.
```
isaacsas, line 72: Do you ever test against a custom hand-coded Jacobian we know is right? That would be good to have in at least one case.

isaacsas (changes requested):

test/simulation_and_solving/jacobian_construction.jl
TorkelE: Yes we do (right now that's basically the only Jacobian tests we do actually): https://github.com/SciML/Catalyst.jl/blob/master/test/simulation_and_solving/jacobian_construction.jl

However, we do not do this for `sparse` though.

isaacsas: We should check sparse against the exact Jacobian, but I don't think we need an "exact" Jacobian that returns a sparse matrix (i.e. checking the Catalyst-generated sparse vs. a hand dense is fine).

isaacsas:
Feel free to merge if tests pass and you are happy.

ChrisRackauckas:
I wouldn't worry too much about pre since it's now actually on the pre-release, though that fail is odd . Let blocks are stronger scoped in the next release because of the globals changes, and so we've seen that causes changes in eval / RGF handling. I don't see why that would be related here because I don't see eval at all though...

PR #1221, opened 2025-04-09
test fixes

TorkelE:


TorkelE:
This one seems to fix parts of things. But we still need https://github.com/SciML/Catalyst.jl/pull/1214 (and then there is the pre-release stuff as well)

src/spatial_reaction_systems/spatial_reactions.jl
```diff
@@ -103,28 +103,32 @@ function check_spatial_reaction_validity(rs::ReactionSystem, tr::TransportReacti
...
-function isequivalent(sym1, sym2)
+function isequivalent(sym1, sym2; ignored_metadata = [MT.SymScope])
     isequal(sym1, sym2) || (return false)
-    if any((md1 != ep_metadata) && !(md1 in sym2.metadata) for md1 in sym1.metadata)
+    if any((md1 != ep_metadata) && (md1[1] ∉ ignored_metadata) && (md1 ∉ sym2.metadata)
+            for md1 in sym1.metadata)
         return false
-    elseif any((md2 != ep_metadata) && !(md2 in sym1.metadata) for md2 in sym2.metadata)
+    elseif any((md2 != ep_metadata) && (md2[1] ∉ ignored_metadata) && (md2 ∉ sym1.metadata)
+            for md2 in sym2.metadata)
```
isaacsas, line 131: Can you not just check the metadata you actually care about? Perhaps maintain a list of the important fields and just iterate through them here to check they agree? This current approach seems fragile and likely to cause issues in the future when MTK changes stuff again.

PR #1231, opened 2025-04-13
Drop symmap in docs

isaacsas:
@TorkelE I think these are no longer needed right?`


TorkelE:
yes, this shoudl all be handled by MTK now

TorkelE (approved):

PR #1142, opened 2024-12-02
Refactoring network analysis docs

vyudu:
Probably worth discussing more but my thought here is to make Network Analysis its own section of the Catalyst docs, with subsections for the ODE decompositions, structural properties and chemical reaction network theory, and a third section for advanced stuff that links to CatalystNetworkAnalysis. 

Added sections for the new API functions in #1134 and building symbolic functions from the ODEs to the ODE docs. Added a section on complex/detailed balance to the CRNT docs. 

Need to: add a section on the deficiency one theorem, bring in the CatalystNetworkAnalysis docs, add code examples to the complex/detailed balance section (including stuff about the precision limitations of `isdetailedbalanced`)

PR #1232, opened 2025-04-13
Release 15.0

isaacsas:
@TorkelE this can be the release branch. 

Going forward, in future releases we can add a range of MTK/SciMLBase versions we support (assuming we know they are actually supported). Right now I have no idea how far back in time the current MTK/SciMLBase versions work with Catalyst.


isaacsas:
(Note that MTK and SciMLBase are already fixed in the Project.toml's, but that isn't showing as that is already in the branch on the Catalyst repo.)

isaacsas:
Hmm, I guess we have a chicken and egg issue. Our docs use tools that depend on Catalyst now like PeTab, but they won't build with the latest Catalyst version yet. 

@TorkelE seems like we need to see which libraries depend on us, get them updated too on their master, release Catalyst, immediately have them release, and then make a point release to get the docs working again?

isaacsas:
Maybe for the initial 15.0 release we need to disable PETab docs?

TorkelE:
Yes, there is probably no way around that. In case we want to remove the v14 migration guide, we would be able to turn off the documented option to build doc pages on linked in pages.jl, and simply comment out the PEtab one (but if not, moving it is not much more effort).

For future releases we might consider testing new new version of Catalyst + PEtab directly (or use downstream tests), and if everything is fine check with @sebapersson if he wants to preemptively add the latest Catalyst version as PEtab compatible (if it is possible to do that for unreleased package versions).

sebapersson:
Once there is a migration guide I am happy to test with PEtab to see that everything runs smoothly, and that it can be re-enabled rapidly.

> For future releases we might consider testing new new version of Catalyst + PEtab directly (or use downstream tests), and if everything is fine check with @sebapersson if he wants to preemptively add the latest Catalyst version as PEtab compatible (if it is possible to do that for unreleased package versions).

Would also work in the future if possible.

sebapersson:
> I don't think there should be anything for PEtab to worry about (at least if it has kept up-to date with the latest MTK versions).

This would be great if the case :) 

Is this branch close to release (then I can test it for PEtab and SBMLImporter later this week)

TorkelE:
We might merge https://github.com/SciML/Catalyst.jl/pull/1142 (should be ready now), and then the plan is to release v15.

isaacsas:
@sebapersson please do check the HISTORY.md for the list of breaking changes. My guess is you aren't impacted by most, as they are mainly slight changes in DSL conventions. Maybe the main one to be aware of is that you should not scalarize parameter arrays anymore, i.e. terms like `@parameters k[1:10]` (but maybe you weren't using this anyways). 

The other thing is that we will cap the version of SciMLBase and ModelingToolkit in this release to the latest we've tested with, and going forward only use ranges we have tested against in future releases. So if at any point you need the latest SciMLBase or MTK just let us know and we can make a release once Catalyst works with it. This will hopefully avoid some issue we've had with non-breaking releases of these libraries being incompatible with components of released Catalyst versions.

isaacsas:
We'll probably always need to drop the petab docs when making a breaking release, but if we coordinate well we can quickly put out a .X.0.1 release as soon as you've got it working again.

isaacsas:
@TorkelE if this passes I'll make the release. I've now updated the README and the HISTORY (the latter to make clear what is breaking).

PR #1233, opened 2025-04-14
Disable PEtab doc

TorkelE:
Since PEtab is not compatible with the new Catlayst version, it must be disabled for the v15 release and will be enabled later.

Also removes link to DiffEqParamEstim which we do not really use anymore.


docs/unpublished/petab_ode_param_fitting.md
isaacsas, line 1: Just for future knowledge; you don't need to rename it -- we have the setting turned on that will only build pages that are explicitly in the `pages` variables.

isaacsas:
This is fine with me to merge once the doc build is working. Looks like there is a dangling link issue now?

TorkelE:
right, the problem when we have to disable docs. Fairly sure that we removed links that we forgot to add in. Removing them now.

Issue #1235, opened 2025-04-14
Tests failing for prerelease


isaacsas:
We need to investigate this and see if the issue is Catalyst related or not.

ChrisRackauckas:
I wouldn't worry about it yet. I think it's a bug in the prerelease that is breaking some world age behavior due to some changes in globals. I'll need to talk with @oscardssmith about that.

isaacsas:
Ok, thanks for letting us know.

oscardssmith:
yeah it is a change in globals. 1.12 got a tiny bit stricter in ways that pretty much only affect code that was slightly weird/broken which given the amount of code we have, it isn't exactly surprising that we have some of it.

ChrisRackauckas:
Let's just keep this to the issue https://github.com/SciML/RuntimeGeneratedFunctions.jl/issues/95#issuecomment-2802126751

ChrisRackauckas:
It looks like we know what the RGF one is, but the Catalyst one looks like it's just incorrect macro code.

https://github.com/SciML/Catalyst.jl/blob/005799ce8fc7f9c3b6a121b2569945263da1f590/src/reactionsystem.jl#L1187-L1188

You should never use eval in a macro. The rule is that if you use eval in a macro is probably wrong 😅.

isaacsas:
That was an internal hack to allow unpacking all the Symbolics in a model, and was never exported or advertised/made api. It really isn’t needed anymore now that `Symbol`-based indexing generally works, so we should just remove it.

ChrisRackauckas:
That would be best, because it seems like it's just not going to be easy to support that in Julia v1.12 as it's doing some odd writing to modules at eval time. Working around it seems like a bit more work and is likely not very reliable.

isaacsas:
No longer an issue

PR #1246, opened 2025-04-22
Remove eval in system unpacking

ChrisRackauckas:
This should be at least a partial fix to https://github.com/SciML/Catalyst.jl/issues/1235, but may also need  https://github.com/SciML/RuntimeGeneratedFunctions.jl/pull/96 to fully get pre working


isaacsas:
Just removed in https://github.com/SciML/Catalyst.jl/pull/1247

PR #1251, opened 2025-04-24
test: fix parameter initialization test

AayushSabharwal:


TorkelE:
I will have a look at this this week, however, once I figured out how initialisation is supposed to work again I will probably want to make some modifications to cover some additional cases as well.

PR #1255, opened 2025-04-25
Fix ocasional hybrid test failure

TorkelE:
Increased the timespan of one test. Previously it could actually fail very occasionally. The test is still fast (<1 second). Have run it 10000 times, still no failure.

### DIFF
--- a/test/simulation_and_solving/hybrid_models.jl
+++ b/test/simulation_and_solving/hybrid_models.jl
@@ -33,12 +33,12 @@ let
         @continuous_events begin
             [V ~ 2.0] => [V ~ V/2, A ~ A/2]
         end
-    end        
-    jinput = JumpInputs(rn, [:A => 0, :V => 1.0], (0.0, 10.0), [:k => 1.0, :λ => .4], 
+    end
+    jinput = JumpInputs(rn, [:A => 0, :V => 1.0], (0.0, 10.0), [:k => 1.0, :λ => .4],
         remake_warn = false)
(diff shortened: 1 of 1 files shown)


isaacsas (changes requested):
Please ensure any `JumpProblem`’s that are used in sampling tests take the stable rng as input.

TorkelE:
ready

isaacsas (approved):
Can we re-enable the hybrid remake tests now? I think that is supposedly fixed in the latest MTK.

TorkelE:
I will update it here, and see what happens. Happy to merge if the uncommented tests work?

isaacsas:
Sure, or can save for a follow up to get this merged. Your choice.

TorkelE:
Shoudl be up soon (just need to double check that test to ensure it is actually correct, if it is it will be quick)

TorkelE:
Still gets errors for the `remake` bit, will just merge this then.

isaacsas:
That is unfortunate since the issue was marked as fixed and https://github.com/SciML/ModelingToolkit.jl/issues/3571 was closed. We should get a new MWE showing errors together then.

TorkelE:
Yeah, I will have a closer look at exactly what is going on, whenever MTK didn't fix anything, or I missed something in the test. I did run all the updates and things failed, and at that point it seemed like a decent amount of effort/time which I did not have today.

isaacsas:
No worries, and thanks for looking into it!

PR #1269, opened 2025-05-16
Add finite state projection example

TorkelE:
An example of how to simulate the CME using FiniteStateProjection.jl.

I trialled having a collapsed "quick-start" example at the beginning (in the same way we had the environment set up in the CRN example library doc page).

### DIFF
--- a/docs/src/model_simulation/finite_state_projection_simulation.md
+++ b/docs/src/model_simulation/finite_state_projection_simulation.md
@@ -1,0 +1,86 @@
+# [Solving the chemical master equation using FiniteStateProjection.jl](@id finite-state_projection)
+
+## [Finite state projection simulation of single-species model](@id state_projection_one_species)
+For this example we will use a simple [birth-death model](@ref basic_CRN_library_bd), where a single species ($X$) is created and degraded at constant rates ($p$ and $d$, respectively).
+```@example state_projection_one_species
+using Catalyst
+rs = @reaction_network begin
+    (p,d), 0 <--> X
+end
+```
+Next we perform jump simulations (using the [ensemble simulation interface](@ref ensemble_simulations_monte_carlo)) of the model. Here, we can see how the it develops from an initial condition and reaches a steady state distribution.
+```@example state_projection_one_species
+using JumpProcesses, Plots
+u0 = [:X => 5]
+tspan = (0.0, 10.0)
+ps = [:p => 20.0, :d => 0.5]
+jprob = JumpProblem(JumpInputs(rs, u0, tspan, ps))
+eprob = EnsembleProblem(jprob)
+esol = solve(eprob, SSAStepper(); trajectories = 10)
+plot(esol; ylimit = (0.0, Inf))
+```
+Using chemical master equation simulations, we want to simulate how the *full probability distribution* of these jump simulations develop across the simulation time-frame. 
+
+As a first step, we import the FiniteStateProjection package. Next, we convert our [`ReactionSystem`](@ref) to a `FSPSystem` (from which we later will generate an ODE).
+```@example state_projection_one_species
+using FiniteStateProjection
+fsp_sys = FSPSystem(rs)
+nothing # hide
+```
+Next, we set our initial condition. For normal simulations, $X$'s initial condition would be a single value. Here, however, we will simulate $X$'s probability distribution. Hence, its initial condition will also be a probability distribution. In FiniteStateProjection's interface, the initial condition is a vector, where the $i$'th index is the probability that $X$ have an initial value of $i-1$. The total sum of all probabilities across the vector should be normalised to $1$. Here we assume that $X$'s initial conditions is known to be $5$ (hence the corresponding probability is $1$, and the remaining ones are $0$):
+```@example state_projection_one_species
+u0 = zeros(75)
+u0[6] = 1.0
+bar(u0)
+```
+We also plot the full distribution using the `bar` function. Finally, the initial condition vector also define the finite space onto which we project the chemical master equation. I.e. we will assume that, throughout the entire simulation, the probability of $X$ reaching values outside this initial vector is negligible. 
+!!! warn
+    This last fact is important. Even if the probability seems to ve very small on the boundary provided by the initial condition, there is still a risk that probability will "leak". Here, it can be good to make simulations using different projections, ensuring that the results are consistent (especially for longer simulations).
+
+Now, we can finally create an `ODEProblem` using out `FSPSystem`, our initial conditions, and the parameters declared previously. We can simulate this `ODEProblem` like any other ODE.
+```@example state_projection_one_species
+using OrdinaryDiffEqDefault
+oprob = ODEProblem(fsp_sys, u0, tspan, ps)
+osol = solve(oprob)
+nothing # hide
+```
+Finally, we can plot the $X$'s probability distribution at various time point's of the simulation. Again, we will uset he `bar` function to plot teh distribution, and the interface described [here](@ref simulation_structure_interfacing_solutions) to acess the simulation at various timestamps.
+```@example state_projection_one_species
+bar(osol(1.0);  width = 1.0, linecolor = 1, alpha = 0.7, linealpha = 0.7, label = "t = 1.0")
+bar!(osol(2.0); linecolor = 1, alpha = 0.7, linealpha = 0.7, label = "t = 2.0")
+bar!(osol(5.0); linecolor = 1, alpha = 0.7, linealpha = 0.7, label = "t = 5.0")
+bar!(osol(10.0); linecolor = 1, alpha = 0.7, linealpha = 0.7, label = "t = 10.0")
+```
+
+## [Finite state projection simulation of multi-species model](@id state_projection_multi_species)
+
+```@example state_projection_multi_species
+using Catalyst # hide
+rs = @reaction_network begin
+    (p,d), 0 <--> X
+    (kB,kD), 2X <--> X2
+end
+```
+
+```@example state_projection_multi_species
+ps = [:p => 1.0, :d => 0.2, :kB => 2.0, :kD => 5.0]
+u0 = zeros(25,25)
+u0[1,1] = 1.0
+```
+
+```@example state_projection_multi_species
+using OrdinaryDiffEqRosenbrock, Plots # hide
+fsp_sys = FSPSystem(rs, [:X, :X2])
+oprob = ODEProblem(fsp_sys, u0, 100.0, ps)
+@time osol = solve(oprob, Rodas5P())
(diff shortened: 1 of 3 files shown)


docs/src/model_simulation/finite_state_projection_simulation.md
```diff
@@ -0,0 +1,153 @@
...
+
+[*The chemical master equation*](https://en.wikipedia.org/wiki/Master_equation) (CME) describes the time development of this distribution[^1]. In fact, this equation is at the core of chemical reaction network kinetics, with all other approaches (such as ODE, SDE, and Jump simulations) being derived as various approximations of it. The CME is a system of ODEs, with one variable *for each possible state of the system*. Each of the ODE's variables describes (the rate of change in) the probability of the system being in that state. For a system with a single species $X$, the CME looks like
+```math
+\begin{aligned}
+\frac{dp(x=0)}{dt} &= f_0(p(x=0), p(x=1), ...) \\
+\frac{dp(x=1)}{dt} &= f_0(p(x=0), p(x=1), ...) \\
+\frac{dp(x=2)}{dt} &= f_0(p(x=0), p(x=1), ...) \\
+ &\vdots\\
+\end{aligned}
+```
```
isaacsas, line 48: This is just wrong. In this notation the ODE rhs are all identical... I'd suggest just writing the general chemical master equation as in the math model description (or using that form but replacing vector X/x by scalars).

docs/src/model_simulation/finite_state_projection_simulation.md
```diff
@@ -0,0 +1,153 @@
...
+We note that, since (for almost all chemical reaction networks) there is a non-zero probability that $X$ is at any specific integer value, the CME have an infinite number of variables (corresponding to the infinite number of potential states). Hence, it cannot be solved practically. However, for high enough species values, the probability of the system attaining such values becomes negligibly small. Here, a truncated version of the CME can be solved practically. An approach for this is the *finite state projection*[^2]. Below we describe how to use the [FiniteStateProjection.jl](https://github.com/SciML/FiniteStateProjection.jl) package to solve the truncated CME (with the package's [documentation](https://docs.sciml.ai/FiniteStateProjection/dev/) providing a more extensive description). While the CEM approach can be very powerful, we note that even for systems with few species, the truncated CME typically have too many states to be feasibly solved.
```
isaacsas, line 49:
```suggestion
For chemical reaction networks in which the total population is bounded, the CME corresponds to a finite set of ODEs. In contrast, for networks in which the system can (in theory) become unbounded, such as networks that include zero order reactions like $\varnothing \to A$, the CME will correspond to an infinite set of ODEs. Even in the finite case, the number of ODEs corresponds to the number of possible state vectors, i.e. vectors with components representing the integer populations of each species in the network. Therefore, for even simple reaction networks there can be many more ODEs than can be represented in typical levels of computer memory, and it becomes infeasible to numerically solve the full system of ODEs that correspond to the CME. However, in many cases the probability of the system attaining species values outside some small range can become negligibly small. Here, a truncated, approximating, version of the CME can be solved practically. An approach for this is the *finite state projection method*[^2]. Below we describe how to use the [FiniteStateProjection.jl](https://github.com/SciML/FiniteStateProjection.jl) package to solve the truncated CME (with the package's [documentation](https://docs.sciml.ai/FiniteStateProjection/dev/) providing a more extensive description). While the CME approach can be very powerful, we note that even for systems with a few species, the truncated CME typically has too many states for it to be feasible to solve the full set of ODEs.
```

docs/src/model_simulation/finite_state_projection_simulation.md
```diff
@@ -0,0 +1,153 @@
...
+ps = [:p => 20.0, :d => 0.5]
+jprob = JumpProblem(JumpInputs(rs, u0, tspan, ps))
+eprob = EnsembleProblem(jprob)
+esol = solve(eprob, SSAStepper(); trajectories = 10)
+plot(esol; ylimit = (0.0, Inf))
+Catalyst.PNG(plot(esol; ylimit = (0.0, Inf), fmt = :png, dpi = 200)) # hide
+```
+Using chemical master equation simulations, we want to simulate how the *full probability distribution* of these jump simulations develops across the simulation time frame. 
+
+As a first step, we import the FiniteStateProjection package. Next, we convert our [`ReactionSystem`](@ref) to a `FSPSystem` (from which we later will generate an ODE).
```
isaacsas, line 73:
```suggestion
As a first step, we import the FiniteStateProjection package. Next, we convert our [`ReactionSystem`](@ref) to a `FSPSystem` (from which we later will generate the ODEs that correspond to the truncated CME).
```

docs/src/model_simulation/finite_state_projection_simulation.md
TorkelE: Do you mean
image
?

My intention was basically just to sketch what it looked like in a way which clearly showed that it was infinite (although messing up the `f` indexing doesn't really help...)

TorkelE:
> Should we add a collapsible "installation instructions" too?

Sounds good, will add it.

> Have you double checked you have the axes correct on the heatmaps? (Sometimes some libraries flip the x and y axes I recall.)

Yes (I made a really skewed case and checked that the distribution ended up at the right side). It does flip from what one would expect. However, we probably should add an `!!! warning` statement explicitly pointing this out (as it is not obvious for someone just skimming the tutorial).

> It might be nice to also make the SS heatmap by running SSAs? That would be a nice comparison to show one gets the same result.

Basically just running a long SSA, and plotting the distribution of all values (after some initial time interval where it approaches the steady state? That would make sense.

> Finally, you probably want to pass vectors that label the X and Y ticks in the heatmaps, as otherwise I think they might start plotting at 1 and not zero. (Maybe plot the initial condition to confirm it looks ok and is labeled right?)

Definitely, I will update to do this.

docs/src/model_simulation/finite_state_projection_simulation.md
isaacsas: I meant to mimic https://docs.sciml.ai/Catalyst/stable/introduction_to_catalyst/math_models_intro/#Stochastic-Chemical-Kinetics-Jump-Process-Models where I give the general formula for it (or at least cite that for the general formula for it, and then give it for something like $0 \to A \to 0$ using $P(a,t)$ if you want a simple example that is infinite.

isaacsas:
I was thinking to just run many simulations to a fixed (large enough) time and then calculating the distribution at that time, but you can use the time series too (but that is usually more work to properly construct the distribution from).

docs/src/model_simulation/finite_state_projection_simulation.md
TorkelE: Sounds good. And we should definitely reference that section as well.

TorkelE:
Sounds good. I will check the simulation time of a few 100 Gillespie simulations, and if that is not noteworthy (should be very quick) I will use that.

isaacsas:
Don't forget to only save the final value once you've determined what time things look stationary at.

TorkelE:
will do, sounds like a good oppertunity to make this point generally (and show of a nice feature of the ensemble simulation interface).

docs/src/model_simulation/finite_state_projection_simulation.md
TorkelE: I have changed to it explicitly be for a bd process `(p,d) 0 <--> X`, and linked the main bit about the CME.

TorkelE:
The main example is updated and should be ready, however, an unrelated doc page has started failing. Will have to figure out why.

TorkelE:
The error has been reported in https://github.com/SciML/SciMLSensitivity.jl/issues/1212

isaacsas:
Should we comment out that doc for now to get the build here working and merged? We can open a tracking issue on it to add that tutorial back later once it is fixed.

docs/src/model_simulation/finite_state_projection_simulation.md
```diff
@@ -0,0 +1,191 @@
...
+u0[6,1] = 1.0
+nothing # hide
+```
+In the next step, however, we have to make an additional consideration. Since we have more than one species, we have to define which dimension of the initial condition (and hence also the output solution) corresponds to which species. Here we provide a second argument to `FSPSystem`, which is a vector listing all species in the order they occur in the `u0` array.
+```@example state_projection_multi_species
+using FiniteStateProjection # hide
+fsp_sys = FSPSystem(rs, [:X, :X₂])
+nothing # hide
+```
+Finally, we can simulate the model just like in the 1-dimensional case. In this case, however, we are simulating an ODE with $25⋅25 = 625$ states, which means we need to make some considerations regarding performance. In this case, we will simply specify the `Rodas5P()` ODE solver (more extensive advice on performance can be found [here](@ref ode_simulation_performance)).
```
isaacsas, line 151:
```suggestion
Finally, we can simulate the model just like in the 1-dimensional case. As we are simulating an ODE with $25⋅25 = 625$ states, we need to make some considerations regarding performance. In this case, we will simply specify the `Rodas5P()` ODE solver (more extensive advice on performance can be found [here](@ref ode_simulation_performance)).
```

docs/src/model_simulation/finite_state_projection_simulation.md
```diff
@@ -0,0 +1,191 @@
...
+```@example state_projection_multi_species
+using Plots # hide
+using OrdinaryDiffEqRosenbrock
+oprob = ODEProblem(fsp_sys, u0, 100.0, ps)
+osol = solve(oprob, Rodas5P())
+heatmap(0:24, 0:24, osol[end]; xguide = "X₂", yguide = "X")
+```
+Here we perform a simulation with a long time span ($t = 100.0$) aiming to find the system's steady state distribution. Next, we plot it using the `heatmap` function.
```
isaacsas, line 159: The order was a bit strange to me here -- I kept expecting to see a figure after the text starting with "Here". Putting it before the code that makes the figure makes more sense to me.

docs/src/model_simulation/finite_state_projection_simulation.md
```diff
@@ -0,0 +1,191 @@
...
+using JumpProcesses # hide
+jprob = JumpProblem(JumpInputs(rs, [:X => 0, :X₂ => 0], (0.0, 100.0), ps))
+output_func(sol, i) = (sol[end], false)
+eprob = EnsembleProblem(jprob; output_func)
+esol = solve(eprob, SSAStepper(); trajectories = 10000)
+ss_jump = zeros(25,25)
+for endpoint in esol
+    ss_jump[endpoint[1] + 1, endpoint[2] + 1] += 1
+end
+heatmap(0:24, 0:24, ss_jump ./length(esol); xguide = "X₂", yguide = "X")
```
isaacsas, line 183: Very nice!

Issue #1284, opened 2025-06-20
species graph nodes incorrectly labeled with array species variables


isaacsas:
See https://discourse.julialang.org/t/building-catalyst-jl-reaction-network-from-a-csv-file-error-in-construction/130043

@vyudu any chance you could fix that?

vyudu:
Will look at this tonight

PR #1287, opened 2025-06-20
fix: plot array species better

vyudu:
A start toward #1284. Now the node labels for array species look like 

image


isaacsas:
Yeah, that is a big improvement over `getindex`. Thanks!

isaacsas:
@vyudu can we not generate Latex strings, and so just use the builtin Latexify functionality of Symbolics? It seems like Makie claims to support it: https://docs.makie.org/dev/explanations/latex

vyudu:
Did so, hopefully docs build okay

isaacsas:
Seems this is now breaking some tests?

vyudu:
@isaacsas Sorry about delay, think this is working now, the docs build error shows up elsewhere

isaacsas:
Should be handled in https://github.com/SciML/Catalyst.jl/pull/1396 now.

PR #1297, opened 2025-07-27
Remove Julia v1.9 get_extension compatibility code

ChrisRackauckas:
## Description

This PR removes the compatibility checks for `isdefined(Base, :get_extension)` since all SciML packages now require Julia v1.10+, where package extensions are built-in.

## Changes

- Removed unnecessary version checks in extension loading code
- Simplified extension imports by removing conditional logic
- Cleaned up obsolete compatibility code

## Context

As identified in the SciML-wide analysis, all SciML packages have moved to requiring Julia v1.10+ as the minimum version. This makes the compatibility code for checking whether package extensions are available redundant.

The `isdefined(Base, :get_extension)` checks were used to support Julia v1.9 where extensions were loaded differently. Since we no longer support v1.9, this code can be safely removed.

## Testing

- [ ] Package tests pass locally
- [ ] No changes to functionality, only removal of version checks

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -40,7 +40,7 @@ makedocs(sitename = "Catalyst.jl",
         assets = ["assets/favicon.ico"],
         canonical = "https://docs.sciml.ai/Catalyst/stable/"),
     modules = [Catalyst, ModelingToolkit, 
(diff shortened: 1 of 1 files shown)


isaacsas:
Handled in https://github.com/SciML/Catalyst.jl/pull/1394

PR #1318, opened 2025-08-11
refactor: do not use internal API in several places

AayushSabharwal:
## Checklist


ChrisRackauckas:
https://github.com/SciML/Catalyst.jl/pull/1319 the spatial failures are real

TorkelE:
Currently, we disallow creating spatial lattice models for cases with vector species/parameters. Not because I think it strictly won't work, but because I am not 100% certain that there are not some corner case, and at the time there was not oppertunity to go through and write ll the tests.

Checking the error, I think what have happened is that there is an error check throwing an error when you have a array vriable input, and using the updated code, that error is not triggered. I.e. these should be errors.
```julia
let
    tr = @transport_reaction D Y

    rs1 = @reaction_network begin
        @species X(t)[1:2] Y(t)
        (k1,k2), X[1] <--> X[2]
    end
    @test_throws ArgumentError LatticeReactionSystem(rs1, [tr], CartesianGrid((2,2)))

    rs2 =  @reaction_network begin
        @species Y(t)
        @parameters k[1:2,1:2]
        (k[1,1],k[1,2]), X11 <--> X12
        (k[2,1],k[2,2]), X21 <--> X22
    end
    @test_throws ArgumentError LatticeReactionSystem(rs2, [tr], CartesianGrid((2,2)))
end
```

isaacsas:
Closing as I think this has all been handled, but feel free to let us know if there are more internals we should stop using.

PR #1337, opened 2025-11-12
Support CairoMakie above 0.13

pkofod:
I suppose the restriction is just there because these are breaking changes due to sub 1.0 minor changes.

## Checklist

### DIFF
--- a/test/extensions/lattice_simulation_plotting.jl
+++ b/test/extensions/lattice_simulation_plotting.jl
@@ -73,6 +73,7 @@ let
         for sol in [osol, jsol]
             # Plots the simulation and checks that a stored value is correct.
             fig, ax, hm = lattice_plot(sol, :X, lrs; t = 1.0)
+            @test_broken hm[3].val[1] ≈ sol.u[end][1]
 
             # Attempts to animate the simulation (using various arguments). Deletes the saved file.
             lattice_animation(sol, :X, lrs, "animation_tmp.mp4"; nframes = 10, framerate = 10, colormap = :BuGn_6)


pkofod:
Not sure if any tests are supposed to fail here

Project.toml
jkrumbiegel, line 64: Makie is also restricted here

jkrumbiegel, line 57: graphmakie is also at 0.6 now with Makie 0.24 compat

pkofod: I updated graphmakie and makie, but would it be better to keep 0.5 and 22.2.1 in there in terms of allowing a flexible set of packages if they all work with catalyst?

TorkelE:
I am not too surprised some errors have snuck in into CI over the last while. Catalyst is based on ModelingToolkit, which had a new breaking change a while ago that we have not had time to update to. In tandem with that there have been updates to MTK-related packages, it probably some change there made assumptions that means these errors happens if you are not using the latest ModelingToolkit.

Hopefully we will be able to make an update soon. A major hindrance is that the next ModelingToolkit breaking release is due very soon, so we are simply waiting for that one and skipping a step. Thanks for the PR though, we will integrate as soon as possible once CI works properly again.

devmotion:
Catalyst pins MTK to v9, so CI and releases should be unaffected by breaking changes in MTK.

Similar to how I recently backported support for ForwardDiff@1 and DifferentiationInterface@0.7 to MTK@9, I think it would be good to update/backport compat entries (apart from breaking MTK and SciML releases, of course) to the latest Catalyst version as much as possible.

TorkelE:
So, _in theory_ that should work. But what may have happened is that some other of the many MTK-adjacent packages have made some form of update which implicitly assumed that MTK v10 has happened, but does not explicitly require MTK v10, and hence is pulled in and causes the error. At least that is my guess, as the CI errors I saw don't really relate to Makie, but had something to do with indexing of MTK structures.

devmotion:
If these errors are not caused by CairoMakie, then the same errors should show up when running tests on master or the latest release of Catalyst.

SciML packages are unfortunately pretty bad with stable non-breaking releases in my experience, but that's something to be sddressed separately (eg by fixing incorrect SciML compat entries in Catalyst or other packages).

I think nobody should expect an immediate update to MTK@10, given the amount of breaking and fundamental changes and that MTK@11 is supposed to be released soonish. But I think it's annoying for many users and downstream packages if Catalyst is blocking updates of unrelated dependencies such as plotting packages.

TorkelE:
Didn't think of that this might affect more downstream stuff, that is annoying, definitely. I have started CI, if the error is in MTK indexing, maybe we can try capping SymbolicIndexingInterface. That is the package which does most of the heavy lifting here, could be it.

pkofod:
> Didn't think of that this might affect more downstream stuff, that is annoying, definitely. I have started CI, if the error is in MTK indexing, maybe we can try capping SymbolicIndexingInterface. That is the package which does most of the heavy lifting here, could be it.

thanks!

jkrumbiegel:
Looks like BifurcationKit's compat also has to be updated as it also has a weakdep on Makie. It already has a version updated to Makie 0.24

devmotion:
I see similar (same?) test failures in #1338 which reruns CI on master. I haven't checked all logs yet but e.g. https://github.com/SciML/Catalyst.jl/actions/runs/19298250248/job/55185564290?pr=1338 is mainly a problem with the existing tests in Catalyst being somewhat flaky and probably an upstream fix in a newer version of some SciML package - IMO the tests should just check `SciMLBase.successful_retcode(sol)` instead of explicitly comparing the return code with e.g. `SciMLBase.ReturnCode.Success` or `SciMLBase.ReturnCode.Terminated` (as documented: https://github.com/SciML/SciMLBase.jl/blob/ff00df899cad8438f018705c5da7eec0960e909e/src/retcodes.jl#L16-L23).

TorkelE:
I would have to investiagate closer what is happening, but unless you are talking about a specific case I am not sure exactly what you mean? I.e. in the file which causes the error (1, Core), the only retcode-related test is a
```julia
@test SciMLBase.successful_retcode(sol)
```
and I don't think that is related to the failure here.

devmotion:
No, the problem I was pointing to was in the Hybrid tests and is caused by https://github.com/SciML/Catalyst.jl/blob/64c117f7321b63e61692ebd0bdf98f208d149d53/test/simulation_and_solving/hybrid_models.jl#L109. I opened a PR with a fix: #1339 I haven't checked the other errors yet.

devmotion:
But regardless of the exact nature of the CI failures - if the same errors are present on the master branch and even more importantly the latest release (I assume they are since it seems there are no implementation changes on master apart from an updated compat entry for Optimization: https://github.com/SciML/Catalyst.jl/compare/v15.0.8...master), then there is no harm in merging PRs like this compat update, assuming the relevant plotting tests pass and these versions are actually tested. Users could already run into these issues, so the new release would not make anything worse in that regard; on the contrary, however, users that are currently blocked from e.g. updating CairoMakie would be able to benefit from the latest improvements and bug fixes in CairoMakie.

pkofod:
Is it the current understanding that none of the failures are related to plotting?

TorkelE:
No, nothing plotting related

pkofod:
Okay, so if we relax the related versions that @jkrumbiegel mentioned above it would be possible to make a patch release?

TorkelE:
Yes, however, we would need to actually sort out the errors so that we do not get CI makred as failing and the docs fails to build. There are some fixes here: https://github.com/SciML/Catalyst.jl/pull/1339, but we are not all the way there yet.

isaacsas:
Let's see if tests pass with the newer Makie versions now.

isaacsas:
(If they do I will merge this, check some of the other compat updates that are lingering, and then make a release once we've settled that all.)

isaacsas:
So seems like BifurcationKit will also need a backported update to the latest Makie version if we want to support it here.

devmotion:
I prepared a PR to BifurcationKit: https://github.com/bifurcationkit/BifurcationKit.jl/pull/257 Tests in this PR are currently running with the BifurcationKit version in this PR. The test failures of the extension test group on Julia LTS are expected since Pkg supports `[sources]` only in Julia >= 1.11.

devmotion:
The BifurcationKit dependency issues seem resolved with my changes, but there are some actual errors remaining in the GraphMakie pipeline, triggered by `plot_complexes`: https://github.com/SciML/Catalyst.jl/actions/runs/19757606855/job/56612075226?pr=1337#step:6:1822 It seems some actual fixes on the Catalyst side might be needed.

devmotion:
It seems https://github.com/SciML/Catalyst.jl/pull/1337/commits/ef6e15048e7337eb9a82635f27dae5c37025de30 fixed the previous GraphMakie issue. However, this uncovered that there is another real test error caused by `lattice_animation`: It seems the (seemingly unofficial?) accesss of `.val` in https://github.com/pkofod/Catalyst.jl/blob/ef6e15048e7337eb9a82635f27dae5c37025de30/ext/CatalystCairoMakieExtension/cairo_makie_extension_spatial_modelling.jl#L36 does not work anymore in newer Makie versions. I'd assume that the correct (and Makie-version independent) way to achieve the desired dynamic behaviour would be to make the to-be-updated properties an `Observable` in https://github.com/pkofod/Catalyst.jl/blob/ef6e15048e7337eb9a82635f27dae5c37025de30/ext/CatalystCairoMakieExtension/cairo_makie_extension_spatial_modelling.jl#L27.

test/extensions/lattice_simulation_plotting.jl
```diff
@@ -71,9 +69,8 @@ let
         jsol = solve(jprob, SSAStepper())
 
         for sol in [osol, jsol]
-            # Plots the simulation and checks that a stored value is correct.
+            # Plots the simulation
             fig, ax, hm = lattice_plot(sol, :X, lrs; t = 1.0)
-            @test hm[3].val[1] ≈ sol.u[end][1]
```
TorkelE, line 76: Stuff like this is quite valid to test, though, right? I.e. we want to have some notion that we actually plot the right thing, not just that we are able to make a plot without error.

devmotion: If you want to test the plots, you should compare them to reference plots (as eg done in Makie itself). Relying on internals of the structs used and generated by Makie is not a valid test IMO.

TorkelE: Right, but is there a true and correct way to extract these values? I.e. if we introduced some bug in the plotting recipe, this would make it harder for us to catch it, right?

devmotion: Yes, that's correct. I don't think there's an official way, and so currently in every - even server-non-breaking - release of Makie these tests could break (as shown by the latest update in this PR).

Some errors would still be caught though 😄: https://github.com/SciML/Catalyst.jl/actions/runs/19765753797/job/56638085033?pr=1337

TorkelE: Would it make sense to hcange to `test_broken`, and add a note about this? Then at least I know there is a problem and can try to circle back to it.

devmotion:
It seems with the backports in BifurcationKit (not officially released yet) and my changes to the plotting extensions and tests, all tests pass successfully and the documentation can be built successfully.

test/extensions/lattice_simulation_plotting.jl
devmotion: Sounds good. To reduce CI time, I'll make this change once a new BifurcationKit release is available and I have to update the PR anyway.

Project.toml
```diff
@@ -46,22 +46,22 @@ CatalystStructuralIdentifiabilityExtension = "StructuralIdentifiability"
 
 [compat]
 BifurcationKit = "0.4.4"
-CairoMakie = "0.12, 0.13"
+CairoMakie = "0.15"
```
devmotion: Could we keep these compat bounds a bit more relaxed and allow

```suggestion
CairoMakie = "0.12, 0.13, 0.14, 0.15"
```

?

Project.toml
```diff
@@ -46,22 +46,22 @@ CatalystStructuralIdentifiabilityExtension = "StructuralIdentifiability"
...
 Combinatorics = "1.0.2"
 DataStructures = "0.18, 0.19"
 DiffEqBase = "6.165.0"
 DocStringExtensions = "0.8, 0.9"
 DynamicPolynomials = "0.6"
 DynamicQuantities = "1"
 EnumX = "1"
-GraphMakie = "0.5"
+GraphMakie = "0.6"
```
devmotion: Maybe just
```suggestion
GraphMakie = "0.5, 0.6"
```
?

Project.toml
```diff
@@ -46,22 +46,22 @@ CatalystStructuralIdentifiabilityExtension = "StructuralIdentifiability"
...
 Graphs = "1.4"
 HomotopyContinuation = "2.9"
 JumpProcesses = "9.19.2"
 LaTeXStrings = "1.3.0"
 Latexify = "0.16.6"
 MacroTools = "0.5.5"
-Makie = "0.22.1"
+Makie = "0.24"
```
devmotion: And here?

```suggestion
Makie = "0.22.1, 0.23, 0.24"
```

TorkelE:
If Sam is happy with the new bounds I am happy as well

isaacsas:
It is fine with me.

isaacsas:
As I think I’ve said in the past, we should perhaps again consider moving the SI, BK, and Makie extensions to just be separate libraries, to both keep Catalyst leaner, and to avoid them blocking Catalyst releases and testing as we are seeing here.

docs/Project.toml
```diff
@@ -94,3 +94,6 @@ SteadyStateDiffEq = "2.2"
 StochasticDiffEq = "6.65"
 StructuralIdentifiability = "0.5.11"
 Symbolics = "6.22"
+
+[sources]
+BifurcationKit = { url = "https://github.com/devmotion/BifurcationKit.jl.git", rev = "dmw/backport_makie" }
```
devmotion:
```suggestion
```

test/extensions/Project.toml
```diff
@@ -10,4 +10,7 @@ ModelingToolkit = "961ee093-0014-501f-94e3-6117800e7a78"
 OrdinaryDiffEqDefault = "50262376-6c5a-4cf5-baba-aaf4f84d72d7"
 OrdinaryDiffEqTsit5 = "b1df2697-797e-41e3-8120-5422d3b24e4a"
 SteadyStateDiffEq = "9672c7b4-1e72-59bd-8a11-6ac3964bc41f"
-StructuralIdentifiability = "220ca800-aa68-49bb-acd8-6037fa93a544"
\ No newline at end of file
+StructuralIdentifiability = "220ca800-aa68-49bb-acd8-6037fa93a544"
+
+[sources]
+BifurcationKit = { url = "https://github.com/devmotion/BifurcationKit.jl.git", rev = "dmw/backport_makie" }
```
devmotion:
```suggestion
```

test/extensions/lattice_simulation_plotting.jl
```diff
@@ -31,18 +31,16 @@ let
         jsol = solve(jprob, SSAStepper())
 
         for sol in [osol, jsol]
-            # Plots the simulation and checks that a stored value is correct.
+            # Plots the simulation
```
devmotion:
```suggestion
            # Plots the simulation and checks that a stored value is correct.
```

test/extensions/lattice_simulation_plotting.jl
```diff
@@ -31,18 +31,16 @@ let
...
             fig, ax, plt = lattice_plot(sol, :X, lrs; t = 1.0)
-            @test plt[1].val[1][2] ≈ sol.u[end][1]
```
devmotion, line 36:
```suggestion
            @test_broken plt[1].val[1][2] ≈ sol.u[end][1]
```

test/extensions/lattice_simulation_plotting.jl
```diff
@@ -31,18 +31,16 @@ let
...
 
             # Attempts to animate the simulation (using various arguments). Deletes the saved file.
             lattice_animation(sol, :X, lrs, "animation_tmp.mp4"; nframes = 10, framerate = 10, colormap = :BuGn_6)
             @test isfile("animation_tmp.mp4")
             rm("animation_tmp.mp4")
 
-            # Plots the kymograph and checks that a stored value is correct.
+            # Plots the kymograph
```
devmotion:
```suggestion
            # Plots the kymograph and checks that a stored value is correct.
```

test/extensions/lattice_simulation_plotting.jl
```diff
@@ -31,18 +31,16 @@ let
...
             fig, ax, hm = lattice_kymograph(sol, :X, lrs)
-            hm[3].val[end,1] ≈ sol.u[end][1]
```
devmotion, line 45:
```suggestion
            @test_broken hm[3].val[end,1] ≈ sol.u[end][1]
```

test/extensions/lattice_simulation_plotting.jl
```diff
@@ -71,9 +69,8 @@ let
         jsol = solve(jprob, SSAStepper())
 
         for sol in [osol, jsol]
-            # Plots the simulation and checks that a stored value is correct.
+            # Plots the simulation
```
devmotion:
```suggestion
            # Plots the simulation and checks that a stored value is correct.
```

test/extensions/lattice_simulation_plotting.jl
devmotion, line 76:
```suggestion
            @test_broken hm[3].val[1] ≈ sol.u[end][1]
```

test/extensions/lattice_simulation_plotting.jl
```diff
@@ -122,9 +119,8 @@ let
     jsol = solve(jprob, SSAStepper())
 
     for sol in [osol, jsol]
-        # Plots the simulation and checks that a stored value is correct.
+        # Plots the simulation
```
devmotion:
```suggestion
        # Plots the simulation and checks that a stored value is correct.
```

test/extensions/lattice_simulation_plotting.jl
```diff
@@ -122,9 +119,8 @@ let
...
         fig, ax, plt = lattice_plot(sol, :X, lrs; t = 0.0)
-        @test plt.node_color[] == osol.u[1]
```
devmotion, line 127:
```suggestion
        @test plt.node_color[] == osol.u[1]
```

TorkelE:
Apparently I cannot commit the suggestions :(
image

But please feel fre to go ahead and do them, sorry for the hassle

pkofod:
good to go? :)

ext/CatalystCairoMakieExtension/cairo_makie_extension_spatial_modelling.jl
```diff
@@ -24,18 +24,24 @@ function lattice_animation(
     vals, plot_min, plot_max = Catalyst.extract_vals(sol, sp, lrs, plot_min, plot_max, t)
 
     # Creates the base figure (which is modified in the animation).
```
TorkelE, line 26: This is not true anymore, right? I.e. the animation does not seem to do anything anymore?

TorkelE:
Trying things out, this does seem to work now? Although I have to admit, I don't actually understand the new spatial plotting code anymore.

TorkelE:
> good to go? :)

Seem we wrote at the same time. If Sam also is happy then we should be good so that people can now actually use things downstream. It would be useful with a follow-up to ensure that the code comments are correct/useful, or at least I might bug people with question later on trying to figure stuff out!

isaacsas:
The docs look ok to me, so I'm fine merging this when @TorkelE is happy (i.e. @TorkelE please merge when you are ok with the changes).

TorkelE:
I've merged now, don't want to halt it any more for internal reasons. I might follow up with some quick questions, but for now we should have something where everything works avaiable.

TorkelE:
Thanks a lot for all the help! :)

ext/CatalystGraphMakieExtension/rn_graph_plot.jl
```diff
@@ -232,37 +232,49 @@ with the rate constant for the reaction.
...
+        f = graphplot(cg;
+            layout,
+            ilabels = complexlabels(rn),
+            node_color = :skyblue3,
+            arrow_shift = :end,
+            curve_distance_usage = true,
+            curve_distance = gen_distances(cg),
+            kwargs...
+        )
+    end
```
isaacsas, line 277: Oh wait, isn't this is changing the graph complex plotting? Colors were always used before to distinguish the edge types, but now it seems to only use them when labels are used. These were not intended to be tied together.

TorkelE: good catch

ext/CatalystCairoMakieExtension/cairo_makie_extension_spatial_modelling.jl
devmotion: This is still correct.

devmotion: This is still correct.

devmotion: This is still correct.

PR #1339, opened 2025-11-12
Fix CI

devmotion:
As #1338 shows, tests currently fail on master.

The PR fixes the following test sets:

- [x] docs: Problem with Optimization (v4.8.0?)
- [x] Core (lts, 1, pre): Problem with DynamicSS?
- [x] Extensions (1, pre): Caused by a bug in Arblib, fixed in Arblib >= 1.6.0 but tests use Arblib 1.4.0
- [x] Hybrid (lts, 1, pre): Incorrect checks of solution return code + numerical issues (caused by ForwardDiff@1?)
- [x] Spatial (only pre)

### DIFF
--- a/docs/src/inverse_problems/petab_ode_param_fitting.md
+++ b/docs/src/inverse_problems/petab_ode_param_fitting.md
@@ -435,14 +435,14 @@ And two additional optional argument:
 
 Because `calibrate_multistart` handles initial guess sampling, unlike for `calibrate`, no initial guess has to be provided.
 
-Here, we fit parameters through 10 independent optimisation runs, using QuasiMonteCarlo's `LatinHypercubeSample` method, and save the result to the OptimisationRuns folder:
+Here, we fit parameters through 10 independent optimisation runs, using QuasiMonteCarlo's `SobolSample` method, and save the result to the OptimisationRuns folder:
 ```@example petab1
 using Optim
 using QuasiMonteCarlo
 mkdir("OptimisationRuns") # hide
 res_ms = calibrate_multistart(petab_problem, IPNewton(), 10; dirsave = "OptimisationRuns",
-    sampling_method = QuasiMonteCarlo.LatinHypercubeSample())
-res_ms = calibrate_multistart(petab_problem, IPNewton(), 10; dirsave = "OptimisationRuns", sampling_method = QuasiMonteCarlo.LatinHypercubeSample()) # hide
+    sampling_method = QuasiMonteCarlo.SobolSample())
+res_ms = calibrate_multistart(petab_problem, IPNewton(), 10; dirsave = "OptimisationRuns", sampling_method = QuasiMonteCarlo.SobolSample()) # hide
 nothing # hide
 ```
 The best result across all runs can still be retrieved using `get_ps(res_ms, petab_problem)`, with the results of the individual runs being stored in the `res_ms.runs` field.


isaacsas:
@TorkelE @devmotion let's use this as a PR to work in to get tests working and docs building. Once that is all fixed I am happy to make a release, but I don't want to make one while tests are failing and/or docs don't build.

test/upstream/mtk_structure_indexing.jl
```diff
@@ -52,7 +52,7 @@ begin
     sint = init(sprob, ImplicitEM(); save_everystep = false)
     jint = init(jprob, SSAStepper())
     nint = init(nprob, NewtonRaphson(); save_everystep = false)
-    @test_broken ssint = init(ssprob, DynamicSS(Tsit5()); save_everystep = false) # https://github.com/SciML/SciMLBase.jl/issues/660
```
isaacsas: @TorkelE this doesn't return a boolean or error, so the test is not broken. I assume you meant this as a flag that there is an issue, but it seems to work now. Note though that `ssint` is not used anymore throughout the tests, so nothing is really being tested for it now beyond `init`.

devmotion:
AFAICT we're stuck with Arblib 1.4 which does not work on Julia >= 1.12 (fixed in Arblib >= 1.6) due to interactions with StructuralIdentifiability which is in the same environment for the Extensions tests: StructuralIdentifiability depends on FLINT_jll as well through Nemo, and Nemo pins FLINT_jll quite strictly; however, recent versions of StructuralIdentifiability which support recent versions of Nemo and hence recent versions of FLINT_jll only support MTK@10.

Breaking up the Extensions environment in smaller environments with presumable less conflicts of dependencies could be a solution; alternatively, one could maybe try to backport support for recent versions of Nemo to older MTK@9 compatible releases of StructuralIdentifiability.

isaacsas:
It's surprising they could update StructuralIdentifiability to MTK10 without making a breaking release.

test/upstream/mtk_structure_indexing.jl
TorkelE: This has been fine for 16 months, so I presume that something has changed underneath. I will update to `... false # ...`, which should make things run for now. Then when we do the MTK fixes I will go through all of the changes and implications proiperly again.

TorkelE:
Think SI just really uses MTK as an interface for declaring a model, so it wouldn't surprise me if no workflows have actually changed. If there has been package updates that means that things no longer work, we should introduce (as liberal as possible) compat bounds to ensure that thinsg actually works.

isaacsas:
I think the bigger issue might be that fundamentally we won't work on 1.12 with MTK9 and the versions of the extensions that work with it. It might be that we should just cap testing to 1.11 and put a note on the Catalyst homepage that we currently only support 1.11, with 1.12 support projected in the next breaking release.

isaacsas:
The hybrid test failure is a real issue that I'm trying to debug. There was a typo in the test, and once fixed the test fails as the callback is never triggered. I haven't been able to figure out why that is happening yet though.

isaacsas:
Not sure though why tests were previously passing with that typo.

devmotion:
Maybe due to an upstream fix?

devmotion:
>  It might be that we should just cap testing to 1.11 and put a note on the Catalyst homepage that we currently only support 1.11, with 1.12 support projected in the next breaking release.

It's really not a Catalyst issue IMO. Previous versions of Arblib had a bug. It's a problem of Arblib and a problem you'll always run into if you use Arblib < 1.6 on Julia 1.12. I don't think it should be Catalyst's responsibility to warn users about a bug caused by Arblib.

To unbreak the tests, we could also try to backport this Arblib fix to a backport release of Arblib (probably 1.4.1?).

devmotion:
I asked for the possibility of a backport release of Arblib: https://github.com/kalmarek/Arblib.jl/pull/220#issuecomment-3524168929

isaacsas:
I marked the blocking test as broken but other hybrid tests are still failing locally and I think it is due to continuous event handling. They used to pass when we last released so something seems to have changed to cause them to be broken now in MTK or lower.

isaacsas:
Anyways I can't work more on this till next week now as I have some deadlines the next few days, but I can circle back next week hopefully to investigate the hybrid tests (or perhaps @TorkelE can look into it). It seems that continuous events no longer properly trigger for jumps over ODEProblem, which definitely worked at one point, so will require some work to investigate what is going one and where changes happened.

TorkelE:
I will try to do what I can, but the hybrid stuff looks like it can be a rather deep rabbit hole (but definitely one require solving. The issue with
```
@test_broken ssint = init(ssprob, DynamicSS(Tsit5()); save_everystep = false) # https://github.com/SciML/SciMLBase.jl/issues/660
```
is simply that the broken test have now been fixed (as confirmed by the link issue). 
```
@test_broken error("")
```
passes as any other broken test, but as ` init(ssprob, DynamicSS(Tsit5()); save_everystep = false)` no longer thorws an error, the test correctly notes that it is no longer broken (although I agree that the logic is a bit weird, and if there had been a @test_unintended_error` or similar, that would have been more appropriate).

There are a couple of downstream tests on `ssint` that we couldn't do (because we couldn't create it), I will see if I can reenable them now as well.

devmotion:
Joel backported the fix for Julia 1.12 to Arblib 1.4.1. This indeed fixes the extensions tests.

TorkelE:
That should leave us only with the hybrid stuff and the docs. Checking the doc, there is an issue in SciML/Optimization that it has encountered. I think this has happened before. Will try to isolate and report.

devmotion:
Optimization might be https://github.com/SciML/Optimization.jl/issues/1070. I ran into the same issue in some other project recently, I think there was a big mess in the recent releases. I think something in the registry should be adjusted: https://github.com/SciML/Optimization.jl/issues/1070#issuecomment-3529963406

Edit: I'm checking whether docs can be built with Optimization@4.7.0 in https://github.com/SciML/Catalyst.jl/pull/1339/commits/e4d9a5e44dd897341b9de97de93423167ad64868

Note that https://github.com/SciML/Catalyst.jl/pull/1328 was pretty useless since Optimization@5 requires MTK@10

devmotion:
I spent a bit more time on the Optimization issue, and I think https://github.com/JuliaRegistries/General/pull/142620 should fix it without any changes to Catalyst (so my experiments in the last commits should be reverted once the registry is fixed).

devmotion:
It seems docs will require a new release of DynamicalSystems with https://github.com/JuliaDynamics/DynamicalSystems.jl/pull/259 as well. I wonder if for the docs it could indeed make sense to build them with a fixed Julia versions. To be clear, the fix https://github.com/JuliaDynamics/DynamicalSystems.jl/pull/259 is a bug fix on every Julia versions but only on Julia 1.12 it seems to cause precompilation errors.

Edit: Indeed, restricting SciMLBase to <= 2.120.0 (as in the registry PR) seems to fix the Optimization.jl errors when building the docs. Remaining problems are the DynamicalSystems error on Julia 1.12 and the precompilation errors of PEtab on Julia 1.12 (they were fixed upstream, but those versions only allow for MTK <= 9.80.1, not MTK@9.84; I opened https://github.com/sebapersson/PEtab.jl/pull/330 to see if the compat bound can be changed in PEtab).

devmotion:
PEtab support for MTK@9.84 should be available in an upcoming release, probably today or tomorrow.

TorkelE:
Excellent, thanks with all the help with this! Sorting out the hybrid thing I think is still the only major thing left?

isaacsas:
I unfortunately haven't had much time to work on this. Throwing Claude at it I got the following conclusions with some back and forth:
```
The Bug
JumpProcesses does not forward JumpProblem.kwargs to the solver when solve(jprob, alg) is called.
Evidence
✓ ModelingToolkit.process_events(js) IS called in the JumpProblem constructor (line 577 in jumpsystem.jl)
✓ Continuous event callbacks ARE created successfully
✓ The callbacks are stored in JumpProblem.kwargs[:callback]
✗ solve(jprob, Tsit5()) does NOT pass these kwargs to the underlying ODE solver
✓ solve(jprob, Tsit5(); jprob.kwargs...) DOES work - the event triggers correctly
```
Given these tests used to pass, and I'm not aware of any JumpProcesses changes related to this, I suspect that perhaps something has changed in SciMLBase or DiffEqBase since the last Catalyst release (but haven't had time to try running tests with older versions).

ChrisRackauckas:
I think it's related to that undocumented args change in JumpProcess solve dispatches? I have been meaning to fix up the StochasticDiffEq impact of that. It was previously relying on the undocumented 4 arg alias dispatches to exist.

isaacsas:
That could be it. Claude claims that forwarding works on the versions of SciMLBase, DiffEqBase, and JumpProcesses that were released when we last released Catalyst.

isaacsas:
FWIW, a bit more testing seems to indicate this is indeed a JumpProcesses and down bug. Claude tests directly using JumpProcesses shows callbacks in JumpProblem's kwargs field are no longer getting called (but were previously). So it seems they are now getting dropped somewhere.

docs/Project.toml
```diff
@@ -86,7 +86,7 @@ OrdinaryDiffEqVerner = "1"
 PEtab = "3.8"
 Plots = "1.40"
 QuasiMonteCarlo = "0.3"
-SciMLBase = "2.46"
+SciMLBase = "2.46 - 2.120.0"
```
devmotion:
```suggestion
SciMLBase = "2.46"
```

isaacsas:
https://github.com/SciML/DiffEqBase.jl/pull/1224 Is the first step to fixing hybrid. Once that is merged I’ll update JumpProcesses. I think with that jumps + events should work again in MTK and Catalyst.

isaacsas:
OK, we should have a JumpProcesses fix for the callback issues out later today once tests pass there. With those fixes the hybrid tests now pass for me locally, so hopefully that should get our test suite passing again now. Then there is just the doc issue.

devmotion:
Something seems to be holding back PEtab to an older version in the docs environment, the PEtab-related errors should be fixed in the latest release.

devmotion:
This seems to be the problem:
```julia
(docs) pkg> add PEtab@3.11.0
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package KrylovKit [0b1a1467]:
 KrylovKit [0b1a1467] log:
 ├─possible versions are: 0.1.0 - 0.10.2 or uninstalled
 ├─restricted by compatibility requirements with LinearSolve [7ed4a6bd] to versions: 0.6.0 - 0.10.2 or uninstalled
 │ └─LinearSolve [7ed4a6bd] log:
 │   ├─possible versions are: 0.1.0 - 3.47.0 or uninstalled
 │   ├─restricted to versions 2.30.0 - 3 by project [477c30de], leaving only versions: 2.30.0 - 3.47.0
 │   │ └─project [477c30de] log:
 │   │   ├─possible versions are: 0.0.0 or uninstalled
 │   │   └─project [477c30de] is fixed to version 0.0.0
 │   ├─restricted by compatibility requirements with NonlinearSolve [8913a72c] to versions: 2.36.1 - 3.47.0
 │   │ └─NonlinearSolve [8913a72c] log:
 │   │   ├─possible versions are: 0.1.0 - 4.12.0 or uninstalled
 │   │   ├─restricted to versions 3.12.0 - 4 by project [477c30de], leaving only versions: 3.12.0 - 4.12.0
 │   │   │ └─project [477c30de] log: see above
 │   │   └─restricted by compatibility requirements with ModelingToolkit [961ee093] to versions: 4.3.0 - 4.12.0
 │   │     └─ModelingToolkit [961ee093] log:
 │   │       ├─possible versions are: 0.0.1 - 10.29.0 or uninstalled
 │   │       ├─restricted to versions 9.69.0 - 9 by project [477c30de], leaving only versions: 9.69.0 - 9.84.0
 │   │       │ └─project [477c30de] log: see above
 │   │       ├─restricted by compatibility requirements with Catalyst [479239e8] to versions: 9.72.0 - 9.84.0
 │   │       │ └─Catalyst [479239e8] log:
 │   │       │   ├─possible versions are: 5.0.0 - 15.0.8 or uninstalled
 │   │       │   ├─restricted to versions 15 by project [477c30de], leaving only versions: 15.0.0 - 15.0.8
 │   │       │   │ └─project [477c30de] log: see above
 │   │       │   └─restricted by compatibility requirements with ModelingToolkit [961ee093] to versions: [14.0.0 - 14.4.1, 15.0.6 - 15.0.8] or uninstalled, leaving only versions: 15.0.6 - 15.0.8
 │   │       │     └─ModelingToolkit [961ee093] log: see above
 │   │       └─restricted by compatibility requirements with PEtab [48d54b35] to versions: 9.84.0
 │   │         └─PEtab [48d54b35] log:
 │   │           ├─possible versions are: 1.0.0 - 3.11.0 or uninstalled
 │   │           ├─restricted to versions 3.8.0 - 3 by project [477c30de], leaving only versions: 3.8.0 - 3.11.0
 │   │           │ └─project [477c30de] log: see above
 │   │           └─restricted to versions 3.11.0 by an explicit requirement, leaving only versions: 3.11.0
 │   └─restricted by compatibility requirements with PEtab [48d54b35] to versions: 3.29.0 - 3.47.0
 │     └─PEtab [48d54b35] log: see above
 ├─restricted by compatibility requirements with BifurcationKit [0f109fa4] to versions: [0.7.0 - 0.7.1, 0.8.2 - 0.9.5]
 │ └─BifurcationKit [0f109fa4] log:
 │   ├─possible versions are: 0.1.0 - 0.5.2 or uninstalled
 │   └─restricted to versions 0.4.4 - 0.4 by project [477c30de], leaving only versions: 0.4.4 - 0.4.16
 │     └─project [477c30de] log: see above
 └─restricted by compatibility requirements with LinearSolve [7ed4a6bd] to versions: 0.10.0 - 0.10.2 or uninstalled — no versions left
   └─LinearSolve [7ed4a6bd] log: see above
```

isaacsas:
We may need to update to newer breaking releases of some of the doc deps, but that will require investigating what updates that requires in Catalyst and/or the docs.

devmotion:
Well in this case it's a bit more complicated 😄 Support for KrylovKit@0.10 was added to BifurcationKit in version 0.5.1 (latest release is 0.5.2) but MTK@9 only supports BifurcationKit@0.4 (I was quite surprised to see that ModelingToolkit contains an extension for BifurcationKit, somehow I expected the extension to live in BifurcationKit). That means either we'll need a backport of KrylovKit@0.10 support to BifurcationKit@0.4 (release 0.4.17 in that case) or of BifurcationKit@0.5 support to ModelingToolkit@9 (release 9.85.0 in that case).

isaacsas:
If we are putting in a backport PR to MTK it should probably also update the fixed test from our hybrid model fixes here, as there is a version of it in the tests in MTK with the same bug.

devmotion:
Honestly, I think a backport release of BifurcationKit might be more generally useful. It seems that would solve this particular problem of our docs environment as well. However, given that the remaining issues of our docs environment (PEtab and DynamicVisualizations) all seem to be caused in the first place by problems with Julia 1.12, I think we might generally want to consider building docs on Julia LTS to be not affected by such problems in our dependencies when a new Julia version is released.

isaacsas:
Hybrid seems fixed now, so I think docs are all that is left.

devmotion:
It seems that even on Julia LTS we need a newer version of PEtab >= 3.9.4 that contains https://github.com/sebapersson/PEtab.jl/pull/316. This huge docs environment seems to be a bit of a dependency nightmare 😄

isaacsas:
If docs build without the PEtab tutorial I would be fine removing it for now. We’ve had to do that in the past too at times.

isaacsas:
@TorkelE tests are passing now. Do you think you could look at the docs? The issues seem to be in tutorials that link to PEtab and other dependencies and not the core tutorials.

isaacsas:
I removed the PEtab tutorial, so let's see if the docs now build.

TorkelE:
happy to have a look at it after it has built, still would be ideal to have it working with PEtab though, as parameter fitting is something i imagine that many users are interested in doing.

devmotion:
It seems the docs build almost succeeded in https://github.com/SciML/Catalyst.jl/pull/1339/commits/99bf59e9cf9292d3cf7bf5de265247f19771c75b, one remaining issue seems to be a bug in PEtab introduced in https://github.com/sebapersson/PEtab.jl/pull/318 (the problem is the `@set sampling_method.rng = rng` line since `SobolSample` doesn't have an `rng` field: https://github.com/SciML/QuasiMonteCarlo.jl/blob/83427d1aa4dac4f39f9bc0a3d69d25f04d541271/src/Sobol.jl#L6-L8).

TorkelE:
@devmotion thanks a lot for all the work on this. Let's see if we get a quick reply over there, if not, let's just built this without PEtab later today, and we can at least check that everything else looks fine.

sebapersson:
> @devmotion thanks a lot for all the work on this. Let's see if we get a quick reply over there, if not, let's just built this without PEtab later today, and we can at least check that everything else looks fine.

I can fix this bug related to Sobol-sampling. @TorkelE for the docs, would it be possible to use LatinHypercube sampling in the failing code-block [here](https://docs.sciml.ai/Catalyst/stable/inverse_problems/petab_ode_param_fitting/#petab_multistart_optimisation) (will allow you to rebuild the docs without waiting for PEtab.jl to update)?

TorkelE:
Thanks! I have changed to that sampler (although in the long run I'd like to have a non-default sampler in the example)

isaacsas:
OK, looks like we should be good once we can actually use a BifurcationKit release (and not @devmotion's personal branch).

isaacsas:
@devmotion thanks for the help with this! Are you going to handle making a backport PR for BifurcationKit? If so, please let us know when a new release we can use is available.

devmotion:
Romain is working on a backport release: https://github.com/bifurcationkit/BifurcationKit.jl/issues/255#issuecomment-3579579408

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -435,14 +435,14 @@ And two additional optional argument:
 
 Because `calibrate_multistart` handles initial guess sampling, unlike for `calibrate`, no initial guess has to be provided.
 
-Here, we fit parameters through 10 independent optimisation runs, using QuasiMonteCarlo's `SobolSample` method, and save the result to the OptimisationRuns folder:
+Here, we fit parameters through 10 independent optimisation runs, using QuasiMonteCarlo's `LatinHypercubeSample` method, and save the result to the OptimisationRuns folder:
```
devmotion:
```suggestion
Here, we fit parameters through 10 independent optimisation runs, using QuasiMonteCarlo's `SobolSample` method, and save the result to the OptimisationRuns folder:
```

docs/src/inverse_problems/petab_ode_param_fitting.md
```diff
@@ -435,14 +435,14 @@ And two additional optional argument:
...
 ```@example petab1
 using Optim
 using QuasiMonteCarlo
 mkdir("OptimisationRuns") # hide
 res_ms = calibrate_multistart(petab_problem, IPNewton(), 10; dirsave = "OptimisationRuns",
-    sampling_method = QuasiMonteCarlo.SobolSample())
-res_ms = calibrate_multistart(petab_problem, IPNewton(), 10; dirsave = "OptimisationRuns", sampling_method = QuasiMonteCarlo.SobolSample()) # hide
+    sampling_method = QuasiMonteCarlo.LatinHypercubeSample())
+res_ms = calibrate_multistart(petab_problem, IPNewton(), 10; dirsave = "OptimisationRuns", sampling_method = QuasiMonteCarlo.LatinHypercubeSample()) # hide
```
devmotion:
```suggestion
    sampling_method = QuasiMonteCarlo.SobolSample())
res_ms = calibrate_multistart(petab_problem, IPNewton(), 10; dirsave = "OptimisationRuns", sampling_method = QuasiMonteCarlo.SobolSample()) # hide
```

isaacsas:
Thanks all for the work getting tests and docs working again.

ChrisRackauckas:
Yes amazing. I'm still in the middle of a push to green other tests so this helps a ton.

PR #1345, opened 2025-12-01
Set `edge_color` in `plot_complexes` regardless of whether edge labels are shown or not

### LINKED ISSUE #1344: Fix graph complex plotting color dropping
In https://github.com/SciML/Catalyst.jl/pull/1337/ colors on edges are now only included when using labels.

devmotion:
Fixes #1344.

### DIFF
--- a/ext/CatalystGraphMakieExtension/rn_graph_plot.jl
+++ b/ext/CatalystGraphMakieExtension/rn_graph_plot.jl
@@ -238,44 +238,30 @@ function Catalyst.plot_complexes(rn::ReactionSystem; show_rate_labels::Bool = fa
 
     # Get complex graph and reaction order for edgecolors and edgelabels. rxorder gives the order of reactions(rn) that would match the edge order in edges(cg).
     cg, rxorder = ComplexGraphWrap(rn)
+    edgelabels = show_rate_labels ? [repr(rxs[i].rate) for i in rxorder] : nothing
+    deps = Set()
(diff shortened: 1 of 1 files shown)


isaacsas:
@devmotion thanks! Looks like there is some kind of CI issue currently. I'll make a release with the current master as this isn't a critical issue, and then we can make another release once this is merged. That way we stop holding people up from using Makie.

devmotion:
Hmmm... These CI errors seem very weird and completely unrelated: https://github.com/SciML/Catalyst.jl/actions/runs/19835077827/job/56830272189?pr=1345#step:6:449 Maybe it's a general GHA problem right now? In that case, CI would fail on master as well in the same way.

So I'm not sure if it's worth rushing a new release, I think most likely these tests will pass went they are rerun.

devmotion:
In particular, given that it took 3 weeks until #1337 could be merged, I don't think a few hours more will matter.

devmotion:
The relevant extension test set already passes, so the PR doesn't seem to cause any test failures: https://github.com/SciML/Catalyst.jl/actions/runs/19835077827/job/56830272071?pr=1345 The docs build should show whether it works as intended.

isaacsas:
Yeah, they are quite weird errors. In any case I already started the release, but once this is merged I'll just make another. I don't think it is a big deal to lose the coloring in one (hopefully) short-lived release.

isaacsas:
Thanks!

PR #1349, opened 2026-01-04
Switch from JuliaFormatter to Runic.jl for code formatting

ChrisRackauckas-Claude:
## Summary
- Updated CI workflow to use `fredrikekre/runic-action@v1` instead of JuliaFormatter
- Removed `.JuliaFormatter.toml` configuration file
- Reformatted all source files with Runic.jl

## Changes
This PR migrates the code formatting from JuliaFormatter with SciMLStyle to Runic.jl. Runic.jl is an opinionated formatter (similar to Go's gofmt) that requires no configuration.

Key differences from JuliaFormatter:
- Explicit `return` statements are added where implicit
- Different indentation and alignment for continuation lines
- Consistent multiline formatting for function arguments

## Test plan
- [ ] CI format check passes with Runic.jl
- [ ] All existing tests still pass


isaacsas:
This is  just going to make a huge mess for @TorkelE's work on updating to V11, and isn't something we should deal with here till that is complete.

TorkelE:
Yeah, once Catalyst is up to date and we don't have pending work, let's do this one.

ChrisRackauckas:
Yeah that's fine, we can delay it.

But due to the bot system that's rolling out, note you'll get maybe daily-ish PRs about things like this. Feel free to just keep closing them until v11 is out.

It may be difficult to blacklist a repo that isn't archived but I can see if we can do that when it's in the update phase.

isaacsas:
If the bot won’t merge it we can just leave one open if that is easier.

ChrisRackauckas:
It won't merge but also won't rebase well 😅 and it won't easily check whether another PR on topic exists. That's hard to do without hitting gh rate limits, so each kind of treats the repo as a clean workable state. But I'll see if I can somehow maintain a list of repos to avoid that are in major update mode. But it cannot live in memory so... it might need a file in the repo that it checks. More details to come

PR #1354, opened 2026-01-26
Version 16 jump fixes

isaacsas:
@TorkelE this should now fix all jump and hybrid tests you flagged me on (and a few others).

It requires a release of MTK with https://github.com/SciML/ModelingToolkit.jl/pull/4218 and https://github.com/SciML/ModelingToolkit.jl/pull/4220 before merging.

### DIFF
--- a/test/simulation_and_solving/simulate_jumps.jl
+++ b/test/simulation_and_solving/simulate_jumps.jl
@@ -208,15 +208,20 @@ let
         m2 /= N
         return m1, m2
     end
-    means1 = zeros(3)
-    means2 = zeros(3)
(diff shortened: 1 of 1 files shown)


isaacsas:
After this I can work on refactoring all the conversion stuff to be a thin wrapper over one hybrid conversion function that geenerates systems. The idea is that these functions create a pure ODE/SDE/jump system, and the hybrid one uses physical scales to allow general selection. Then we just wrap that one with the other three and default physical scale choices appropriate for them each.

TorkelE:
Looks good. Agree that refactoring the conversions in such a way sounds like a good idea. I guess that might make it easier to hanlde separately supported jumpa and noise porcesses when we start supporting that as well?

PR #1356, opened 2026-01-27
Ensure docs build on v16

TorkelE:
This PR basically performs the basic set of changes to ensure that the docs actually build on the v16 branch. Once this works, we can do follow-up PRs on separate doc pages adapting them as required for v16.

A few docs are commented out as they fail due to MTK bugs that needs fixing before they can be included.

I have also changed it so that missing references are no treated as errors (as we make referencel inks to now commented out doc pages). Before v16 we should be able to revert that, however, the alternative is to temporarily remove those refeences, which leaves us with no convenient way of reverting those changes and we would have to look at old code.

### DIFF
--- a/docs/src/model_creation/parametric_stoichiometry.md
+++ b/docs/src/model_creation/parametric_stoichiometry.md
@@ -70,10 +70,8 @@ show(stdout, MIME"text/plain"(), equations(osys)) # hide
 ```
 Specifying the parameter and initial condition values,
 ```@example s1
-sim_cond = (
-    revsys.k₊ => 1.0, revsys.k₋ => 1.0, revsys.m => 2, revsys.n => 2,
-    revsys.A => 1.0, revsys.B => 1.0
-)
+sim_cond = (revsys.k₊ => 1.0, revsys.k₋ => 1.0, revsys.m => 2, revsys.n => 2,
+    revsys.A => 1.0, revsys.B => 1.0)
 oprob = ODEProblem(osys, sim_cond, (0.0, 1.0))
 nothing # hide
 ```


docs/src/faqs.md
```diff
@@ -432,47 +432,6 @@ and
...
-println("Correct u0 is: ", (1.0, 2.0, 5.0), "\n", "remade value is: ", nlprob2[(:X₁, :X₂, :X₃)]) 
-```
-and `Γ` becomes
-```@example faq_remake
-println("Correct Γ is: ", 8.0, "\n", "remade value is: ", nlprob2.ps[:Γ][1])
-```
-Unfortunately, as with our first example, trying to enforce that a
-non-eliminated species should have its initial value updated instead of the
-observed species will not work.
-
```
TorkelE, line 475: This kind of doesn't make sense on the new one as for nonlinear system the observed equations do not appear on the new MTKBase mtkcompile. However, I still need to do a proper go through exactly what happens and update this FAQ to ensure that the correct stuff is written down.

docs/make.jl
```diff
@@ -47,7 +47,7 @@ makedocs(sitename = "Catalyst.jl",
     clean = true,
     pages = pages,
     pagesonly = true,
-    warnonly = [:missing_docs])
```
TorkelE, line 50: This is temporary, we can add it back when we make the release.

src/reactionsystem.jl
```diff
@@ -429,7 +429,7 @@ function ReactionSystem(eqs, iv, unknowns, ps;
         throw(ArgumentError("System names must be unique."))
 
     # Process initial_conditions to unwrap Num wrappers.
-    initial_conditions = SymmapT(value(k) => value(v) for (k, v) in pairs(initial_conditions))
+    initial_conditions = SymmapT(value(entry[1]) => value(entry[2]) for entry in initial_conditions)
```
TorkelE, line 432: yielded a bug previously, i.e. `pairs` makes a different type of thing.

docs/src/faqs.md
isaacsas: I think that deserves an MTK issue. I thought we were told that moving an equation of the form `X ~ ...` to be an observable would be trivially handled still as it does not require complicated analysis? If not, we should get them to let us make it an observable ourselves.

src/reactionsystem.jl
isaacsas: I think we should probably be more careful and follow how it is setup in MTK's System constructor:

https://github.com/SciML/ModelingToolkit.jl/blob/2b9b773ec70b800aa4a18f57d30c2834498bd9c2/lib/ModelingToolkitBase/src/systems/system.jl#L426

docs/src/model_creation/constraint_equations.md
```diff
@@ -255,14 +255,15 @@ events, we start by creating reaction equations, parameters, variables, and
 unknowns.
 ```@example ceq3
 t = default_t()
-@parameters k_on switch_time k_off
+@parameters switch_time k_off
+@discretes k_on(t)
 @species A(t) B(t)
 
-rxs = [(@reaction k_on, A --> B), (@reaction k_off, B --> A)]
+rxs = [Reaction(k_on, [A], [B]), Reaction(k_off, [B], [A])]
```
TorkelE: needed because `@reaction` does not make `k_on` a discrete.

docs/src/api/core_api.md
```diff
@@ -117,22 +116,22 @@ retrieve info from just a base [`ReactionSystem`](@ref) `rn`, ignoring
 sub-systems of `rn`, one can use the ModelingToolkit accessors (these provide
 direct access to the corresponding internal fields of the `ReactionSystem`)
 
-* `ModelingToolkit.get_unknowns(rn)` is a vector that collects all the species
+* `ModelingToolkitBase.get_unknowns(rn)` is a vector that collects all the species
```
TorkelE, line 119: At some point we have to figure out what to do with these that technically refer to internal MTK stuff. I do think we have some freedom to request stuff to be made public though.

docs/src/assets/Project.toml
```diff
@@ -0,0 +1,91 @@
...
+OrdinaryDiffEqVerner = "1"
+Plots = "1.40"
+QuasiMonteCarlo = "0.3"
+SciMLBase = "2.46"
+SciMLSensitivity = "7.60"
+SpecialFunctions = "2.4"
+StaticArrays = "1.9"
+SteadyStateDiffEq = "2.2"
+StochasticDiffEq = "6.65"
+StructuralIdentifiability = "0.5.11"
```
TorkelE: This file keeps getting written during local doc builds. We probably should figure out why. Alternatively, or additonally, we should add it to gitignore.

ChrisRackauckas: That's from the reproducibility section, you should delete and .gitignore it

TorkelE: Thanks! will update

docs/src/model_creation/dsl_advanced.md
```diff
@@ -149,14 +149,6 @@ oprob = ODEProblem(rn, u0, tspan, p)
 sol = solve(oprob, Tsit5())
 plot(sol)
 ```
-It is still possible to designate $X$'s value in `u0`, in which case this overrides the default value.
-```@example dsl_advanced_defaults
-u0 = [:X => 0.5]
-p = [:X₀ => 1.0, :p => 1.0, :d => 0.5]
-oprob = ODEProblem(rn, u0, tspan, p)
-sol = solve(oprob, Tsit5())
-plot(sol)
```
isaacsas, line 158: Why did you remove this?

docs/src/api/core_api.md
isaacsas: So what is the way to access the unknowns of just a given system and not its subsystems now?

docs/src/faqs.md
isaacsas: Maybe just comment this out. It may be needed after we get this resolved with MTK?

docs/src/model_creation/constraint_equations.md
isaacsas: Or you could interpolate the discrete version.

docs/src/model_creation/examples/smoluchowski_coagulation_equation.md
```diff
@@ -104,8 +104,7 @@ rs = complete(rs)
 We now convert the [`ReactionSystem`](@ref) into a `ModelingToolkit.JumpSystem`, and solve it using Gillespie's direct method. For details on other possible solvers (SSAs), see the [DifferentialEquations.jl](https://docs.sciml.ai/DiffEqDocs/stable/types/jump_types/) documentation
 ```@example smcoag1
 # solving the system
-jinputs = JumpInputs(rs, u₀map, tspan)
-jprob = JumpProblem(jinputs, Direct(); save_positions = (false, false))
+jprob = JumpProblem(rs, u₀map, tspan; aggregator = Direct(), save_positions = (false, false))
```
isaacsas:
```suggestion
jprob = JumpProblem(rs, u₀map, tspan; save_positions = (false, false))
```

docs/src/model_creation/parametric_stoichiometry.md
```diff
@@ -63,16 +63,18 @@ stoichiometries `(F,2*H,2)`.
...
 ```
 Specifying the parameter and initial condition values,
 ```@example s1
-p  = (revsys.k₊ => 1.0, revsys.k₋ => 1.0, revsys.m => 2, revsys.n => 2)
-u₀ = [revsys.A => 1.0, revsys.B => 1.0]
-oprob = ODEProblem(osys, u₀, (0.0, 1.0), p)
+sim_cond = (
+    revsys.k₊ => 1.0, revsys.k₋ => 1.0, revsys.m => 2, revsys.n => 2,
+    revsys.A => 1.0, revsys.B => 1.0
+)
```
isaacsas:
```suggestion
sim_cond = (revsys.k₊ => 1.0, revsys.k₋ => 1.0, revsys.m => 2, revsys.n => 2,
    revsys.A => 1.0, revsys.B => 1.0)
```

docs/src/model_creation/parametric_stoichiometry.md
```diff
@@ -94,16 +96,19 @@ revsys = @reaction_network revsys begin
...
 ```
 Since we no longer have factorial functions appearing, our example will now run
 with `m` and `n` treated as floating point parameters:
 ```@example s1
-p  = (revsys.k₊ => 1.0, revsys.k₋ => 1.0, revsys.m => 2.0, revsys.n => 2.0)
-oprob = ODEProblem(osys, u₀, (0.0, 1.0), p)
+sim_cond = (
+    revsys.k₊ => 1.0, revsys.k₋ => 1.0, revsys.m => 2.0, revsys.n => 2.0,
+    revsys.A => 1.0, revsys.B => 1.0
+)
```
isaacsas:
```suggestion
sim_cond = (revsys.k₊ => 1.0, revsys.k₋ => 1.0, revsys.m => 2.0, revsys.n => 2.0,
    revsys.A => 1.0, revsys.B => 1.0)
```

docs/src/model_simulation/simulation_structure_interfacing.md
```diff
@@ -124,26 +124,26 @@ sol(1.0; idxs = [:C])
...
 
 There exist four different functions, each returning a function for performing a specific type of interfacing:
-- `ModelingToolkit.getu`: For accessing species values.
-- `ModelingToolkit.getp`: For accessing parameter values.
-- `ModelingToolkit.setu`: For changing species values.
-- `ModelingToolkit.setp`: For changing parameter values.
+- `ModelingToolkitBase.getu`: For accessing species values.
+- `ModelingToolkitBase.getp`: For accessing parameter values.
+- `ModelingToolkitBase.setu`: For changing species values.
+- `ModelingToolkitBase.setp`: For changing parameter values.
```
isaacsas, line 131: These are from SymbolicIndexingInterface and not MTK right? (We should import and qualify with the package they are from and not the reexporting package.)

TorkelE: Yes, I plan to do a general pass to move stuff to SII (and Symbolics etc.) later on. agree we should do this.

docs/src/model_creation/constraint_equations.md
TorkelE: that is a good point

docs/src/api/core_api.md
TorkelE: Think it is still `get_unknowns`, but it is not public anymore, so we probably should request them to make it public (or create a new, public, function with that functionality. I agree this is something that should be possible).

docs/src/model_creation/dsl_advanced.md
TorkelE: This no longer works. if we do
```julia
@species X = X0
```
it becomes a binding, which cannot be overridden, and hence you must define the initial condition using `X0`. If you do
```julia
@species X = 2.0
```
(not symbolic expression) it is a normal default initial condition which can be override (personally would have prefer for default initial conditions to be used with this syntax fully, and bindings defined using some otehr syntax)

docs/src/faqs.md
TorkelE: I would, but I don't think there is a robust way to comment out parts of a doc?

PR #1357, opened 2026-01-29
Unify System creation during conversion

isaacsas:
@TorkelE this is a start on the system conversion updates. I kept pure SDE models with no constraints using noise_eqs, but added a kwarg that allows switching to using Brownians. Hybrid models or models with constraints use Brownians. There is also an explicit `HybridProblem` function if one wants to directly generate the `JumpProblem` associated with a hybrid model -- `JumpProblem` forces everything to be jumps for consistency with ODE and SDE problems (I need to update that for hybrid models without jumps). I also didn't update the nonlinear system code to use the hybrid system function, since that is a different model type.

This should be pretty backwards compatible in that I don't think any non-hybrid model tests need updating.

### DIFF
--- a/src/reactionsystem_conversions.jl
+++ b/src/reactionsystem_conversions.jl
@@ -649,6 +649,7 @@ function make_hybrid_model(rs::ReactionSystem;
     any(==(PhysicalScale.Auto), scales) &&
         error("Unresolved PhysicalScale.Auto scales remain. Provide `default_scale` or per-reaction `physical_scales`.")
 
+    # the || below ensure the empty scale case is handled correctly
     has_ode = any(==(PhysicalScale.ODE), scales) || (_override_all_scales == PhysicalScale.ODE)
     has_sde = any(==(PhysicalScale.SDE), scales) || (_override_all_scales == PhysicalScale.SDE)
     has_jump = any(in(JUMP_SCALES), scales)


isaacsas:
@TorkelE I am happy to merge this if tests pass. I'll handle Brownians + jumps in a followup once MTK adds that functionality.

isaacsas:
(So if you want to review before I merge please let me know.)

src/reactionsystem_conversions.jl
```diff
@@ -47,6 +47,23 @@ end
 # including non-species variables.
 drop_dynamics(s) = isconstant(s) || isbc(s) || (!isspecies(s))
 
+# Check if an SDE system has constraints requiring mtkcompile (algebraic equations or BC species).
+function has_sde_constraints(rs::ReactionSystem)
+    flatrs = Catalyst.flatten(rs)
+    has_alg_equations(rs) || any(isbc, get_unknowns(flatrs))
```
TorkelE: wouldn't `any(isbc, unknowns(rs))` have the same effect, and then you don't need to compute `flatrs` (alterantively, do `has_alg_equations` on `flatrs` instead, forgot if that one checks all or just top level equations)

src/reactionsystem_conversions.jl
```diff
@@ -129,29 +146,89 @@ function assemble_diffusion(rs, sts, ispcs; combinatoric_ratelaws = true,
...
+
+    for (spec, stoich) in rx.netstoich
+        remove_conserved && (spec in nps.depspecs) && continue
+        drop_dynamics(spec) && continue
+        haskey(species_to_idx, spec) || continue
+        i = species_to_idx[spec]
+        coef = _signed_stoich_term(stoich, rlsqrt)
+        f(i, coef)
+    end
+end
```
TorkelE, line 199: The idea here is that `f` is a mutating function? If so, would it make sense to call it `f!`? I am not entirely sure, because where `foreach_noise_coeff` is actually called, no `f` is provided? Or maybe it is the `do` notations which does stuff that I do not fully understand.

src/reactionsystem_conversions.jl
```diff
@@ -525,6 +594,142 @@ COMPLETENESS_ERROR = "A ReactionSystem must be complete before it can be convert
...
+    spatial_convert_err(rs, MT.System)
+
+    flatrs = Catalyst.flatten(rs)
+
+    # Resolve scales: _override_all_scales (internal) takes precedence over everything.
+    if _override_all_scales !== nothing
+        scales = fill(_override_all_scales, length(reactions(flatrs)))
+    else
+        scales = merge_physical_scales(reactions(flatrs), physical_scales, default_scale)
+    end
```
TorkelE, line 648: If you prefer, I am happy for us to use the
```julia
scales =  if ...

else

end
```
in cases like this

src/reactionsystem_conversions.jl
```diff
@@ -525,6 +594,142 @@ COMPLETENESS_ERROR = "A ReactionSystem must be complete before it can be convert
...
+    any(==(PhysicalScale.Auto), scales) &&
+        error("Unresolved PhysicalScale.Auto scales remain. Provide `default_scale` or per-reaction `physical_scales`.")
+
+    has_ode = any(==(PhysicalScale.ODE), scales) || (_override_all_scales == PhysicalScale.ODE)
```
TorkelE, line 653: Is `|| (_override_all_scales == PhysicalScale.ODE)` necessary> I.e. in this case, wouldn't `any(==(PhysicalScale.ODE), scales)` always be `true` anyway?

src/reactionsystem_conversions.jl
```diff
@@ -525,6 +594,142 @@ COMPLETENESS_ERROR = "A ReactionSystem must be complete before it can be convert
...
+    end
+
+    # --- Add constraints (BC species, constraint equations, conserved species) ---
+    if has_continuous
+        eqs, us, ps, obs, defs = addconstraints!(eqs, flatrs, ists, ispcs; remove_conserved)
+    else
+        # Pure jump case.
+        any(isbc, get_unknowns(flatrs)) &&
+            (ists = vcat(ists, filter(isbc, get_unknowns(flatrs))))
+        us = ists
```
TorkelE, line 708: Going forward, we probably should try to use `us` (unknowns) consistently over `sts` (states), since that is what MTK uses (and I think we are consistently using that in new code now anyway). But actually changing old code might be a bit pedantic

src/reactionsystem_conversions.jl
isaacsas: I'm refactoring a bit to just immediately flatten in each function and pass the flattened system around. Then the flatten call has no real cost.

isaacsas: Calling `unknowns` may allocate, so just passing a flattened system from the start is probably more efficient.

src/reactionsystem_conversions.jl
isaacsas: Sure, we could use `f!`. The idea is we have a common case of generating the diffusion term and either saving it in the noise matrix, or adding it to the appropriate SDE as a Brownian. This allows the same code to handle both with just a different `f` to handle the saving basically.

TorkelE:
Looks good. I left a few comments, but I think those are probably more me just not understanding stuff than anything else.

src/reactionsystem_conversions.jl
isaacsas: I don't have a strong preference, and use both, but slightly prefer this form (but again, I use both).

src/reactionsystem_conversions.jl
isaacsas: This is needed for when there are no reactions and so the scales vector is empty.

src/reactionsystem_conversions.jl
TorkelE: make sense, remember this

src/reactionsystem_conversions.jl
isaacsas: We can do a cleanup for notation as a last step before release and have Claude or such fix such inconsistencies.

src/reactionsystem_conversions.jl
TorkelE: the function makes lots of sense. If  `f` is mutating function I think `f!` might make it clearer what is going on, if not, obviously it is a bad idea

src/reactionsystem_conversions.jl
TorkelE: makes sense

src/reactionsystem_conversions.jl
isaacsas: Actually though, it isn't mutating its arguments. It captures the variable to mutate when it is constructed. Maybe there is a more performant way to write it though. This version is probably compiling a new function for each reaction which might not be ideal.

isaacsas: I think we should leave it be for now. It probably is less expensive than the symbolic processing. If we see it has performance issues we can modify this path.

TorkelE: Ok, think it is just me that don't follow the `do` loop well, but happy to keep it for now. When/if we can moe fully to using brownians we could have a look at it again, but for now it works perfectly well so let's roll with it.

isaacsas: This of it as creating a function that takes the arguments after `do`, i.e. `i,coef`, creates a function that takes such arguments and executes what follows on the next line until the `end`, and then passes that function to the foreach function as its first argument. i.e. the following are doing the same thing:

```julia
function double_it(number)
    return number * 2
end

map(double_it, [1, 2, 3]) 
```

```julia
map([1, 2, 3]) do number
    return number * 2
end
```

TorkelE: Ok, still need to kind of wrap my head around this, but that helps a bit, thanks!

PR #1358, opened 2026-01-30
remove performance benchmark tests

TorkelE:
Removes the benchmark tests that we are not running anyway. I have it saved among my Catalyst benchmark scripts anyway, so we can run it at our leisure whenever we want to.


isaacsas:
This is something we should look at setting up a https://github.com/MilesCranmer/AirspeedVelocity.jl CI for in general after V16 releases.

TorkelE:
looks cool, yeah we should

PR #1361, opened 2026-01-30
Designate stoich ps as int in dsl

TorkelE:
With the new MTK update, for stuff like
```julia
rn = @reaction_network begin
    k, n*X --> Xn
end
```
to simualte properly, `n` must be designated as a integers. This update adds a routine to the DSL which detects parameters that occur as stoichimetories, are not explicitly declared, and add the `::Int64` desigantion to their creation.


TorkelE:
Probably, but I will add something just to be on the safe side.

PR #1362, opened 2026-01-30
Reduce `symmap_to_varmap` usage

TorkelE:
I've tried to reduce the usage of `symmap_to_varmap` as much as possible. What I have left is:
- It is used in quite a few tests. If there are any of the tests that you want to skip/remove `symmap_to_varmap` from, feel happy to, but I also did not want to intervene in them too indiscriminately.
- It is used in the HC extension. When we are able to fully move to MTK's HC implementation, we can skip that here, in the meantime, I will let it be. Given that the stability code is fairly related, I am letting that one remain there for now.
- MTK currently does not do symbol conversion for its MTK problems (I have raised an issue). But since we already have the code here, I will let it remain until that issue is fixed. 
- In the spatial code we, given that we want to store Symbolci internally, but accept symbol inputs, we probably have to let it be a thing for the foreseeable future.

### DIFF
--- a/ext/CatalystBifurcationKitExtension/bifurcation_kit_extension.jl
+++ b/ext/CatalystBifurcationKitExtension/bifurcation_kit_extension.jl
@@ -11,13 +11,13 @@ function BK.BifurcationProblem(rs::ReactionSystem, u0_bif, ps, bif_par, args...;
     (bif_par isa Symbol) && (bif_par = ModelingToolkitBase.get_var_to_name(rs)[bif_par])
     (plot_var isa Symbol) && (plot_var = ModelingToolkitBase.get_var_to_name(rs)[plot_var])
     if (u0_bif isa Vector{<:Pair{Symbol, <:Any}}) || (u0_bif isa Dict{Symbol, <:Any})
-        u0_bif = symmap_to_varmap(rs, u0_bif)
+        u0_bif = Catalyst.symmap_to_varmap(rs, u0_bif)
     end
     if (ps isa Vector{<:Pair{Symbol, <:Any}}) || (ps isa Dict{Symbol, <:Any})
-        ps = symmap_to_varmap(rs, ps)
+        ps = Catalyst.symmap_to_varmap(rs, ps)
     end
     if (u0 isa Vector{<:Pair{Symbol, <:Any}}) || (u0 isa Dict{Symbol, <:Any})
-        u0 = symmap_to_varmap(rs, u0)
+        u0 = Catalyst.symmap_to_varmap(rs, u0)
     end
 
     # Creates NonlinearSystem.
--- a/ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
+++ b/ext/CatalystHomotopyContinuationExtension/homotopy_continuation_extension.jl
@@ -52,7 +52,7 @@ function steady_state_polynomial(rs::ReactionSystem, ps, u0)
     # be substituted in later. U's not involved conservation laws get dummy values.
     rs = Catalyst.expand_registered_functions(rs)
     ns = complete(make_rre_algeqs(rs; remove_conserved = true, conseqs_remake_warn = false))
-    pre_varmap = merge(Dict(symmap_to_varmap(rs, u0)), Dict(symmap_to_varmap(rs, ps)))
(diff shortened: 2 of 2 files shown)


isaacsas:
Unfortunately we still need it for jumps. 

Let’s hold off on this as I have a user Brownian and jumps WIP PR that works locally, and I don’t want to have to deal with rebasing since it also makes changes around symmap. I may also try to put in a fix for the jump issue, but I’m still testing if what I have is reasonable.

TorkelE:
Ok, got it. A lot of things actually seems passing though. E.g. for the Hybrid case the only error is a failure to create the problem in this one
```julia
# Checks that various model options (observables, events, defaults and metadata, differential equations,
# non-default_iv) works for hybrid models.
let
    # Creates the model (X species is pure jump, Y is pure ODE, and  Z1,Z2 are mixed).
    # Hybrid species have non-constant rates containing the two other species.
    rn = @reaction_network begin
        @ivs τ
        @differentials Δ = Differential(τ)
        @parameters X0 Y0
        @species X(τ) = X0 [description = "A jump only species"] Y(τ) = Y0 [description = "An ODE only species"]
        @variables Ztot(τ) = 0.0  # Default needed for MTK v10+ callback compilation
        @observables Ztot ~ Z1 + Z2
        @discrete_events [1.0] => [Z1 ~ Pre(Z1) + 1.0]
        @continuous_events [Y ~ 1.0] => [Y ~ 5.0]
        @equations Δ(V) ~ Z1 + X^2 - V
        (p,d), 0 <--> X
        d, Y --> 0, [physical_scale = PhysicalScale.ODE]
        (k*X, k*Y), Z1 <--> Z2, ([physical_scale = PhysicalScale.ODE], [physical_scale = PhysicalScale.Jump])
    end

    # Simulates the model.
    u0 = [:Z1 => 1.0, :Z2 => 2.0, :V => 1.5]
    ps = [:p => 2.0, :d => 0.5, :k => 1.0, :X0 => 0.1, :Y0 => 4.0]
    prob = HybridProblem(rn, u0, (0.0, 10.0), ps)
    sol = solve(prob, Tsit5(); tstops = [1.0 - eps(), 1.0 + eps()])

    # Tests that the model contain the correct stuff.
    @test ModelingToolkitBase.getdescription(rn.X) == "A jump only species"
    @test ModelingToolkitBase.getdescription(rn.Y) == "An ODE only species"
    @test sol[:X][1] == sol.ps[:X0] == 0.1
    @test sol[:Y][1] == sol.ps[:Y0] == 4.0
    @test sol[:V][1] == 1.5
    @test sol[:Ztot] ≈ sol[rn.Z1 + rn.Z2]
    @test minimum(sol[:Y]) ≈ 1.0
    @test maximum(sol[:Y]) ≈ 5.0 atol = 1e-1 rtol = 1e-1
    @test all(isequal([rn.τ], Symbolics.arguments(Symbolics.value(u))) for u in unknowns(rn))
    @test sol(1.0 - eps(); idxs = :Z1) + 1 ≈ sol(1.0 + eps(); idxs = :Z1)
end
```
I don;t think there error is related, but probably something obscure in MTK, but would take some time to investigate. Happy to hold of merging until later though.

isaacsas:
That specific test is the MTK bug I was mentioning. It is due to MTK not properly handling the symbols that are registered as belonging to the generated affect system for the events (the code to handle that was added at some point to ODE and SDE problems but not jump problems). I have a fix for it I will PR to MTK tomorrow. Then we can hopefully fully drop symmap_to_varmap in our conversion functions.

test/compositional_modelling/component_based_model_creation.jl
```diff
@@ -290,8 +290,8 @@ let
     @test isapprox(0, norm(sol[ns.D] .- 2 * sol[A] - 3 * sol[B]), atol = (100 * eps()))
     psyms = [:r₊ => 1.0, :r₋ => 2.0, :ns₊β => 3.0]
     u₀syms = [:A => 1.0, :B => 2.0, :C => 0.0]
-    p = symmap_to_varmap(osys, psyms)
-    u₀ = symmap_to_varmap(osys, u₀syms)
+    p = Catalyst.symmap_to_varmap(osys, psyms)
+    u₀ = Catalyst.symmap_to_varmap(osys, u₀syms)
```
isaacsas: Why is this needed? Can we remove the `symmap_to_varmap` call here?

isaacsas:
(The MTK fix has been released now.)

PR #1365, opened 2026-02-01
CI updates

isaacsas:
CI is getting pretty slow these days for two reasons:

1. Core just has too much, so here I am splitting it into "Models" and "Simulation", which from my review of logs should be roughly the same amount of time.
2. The approach of instantiating the main Catalyst environment (~16 minutes of compilation) and then later instantiating the extension environment (forcing recompilation and another ~20+ minutes) is really slow. Splitting the extension tests to their own CI action should allow us to precompile the full extension environment for those tests, cutting this time in half.


TorkelE:
Not sure I fully understand 2. You mean taht we can do them in parallel? Making things faster sounds all good to me.

isaacsas:
The extensions tests previously had a workflow where we

1. Activated and installed the Catalyst dependencies.
2. After non-extension tests pass we activate a new environment for the extensions tests and run them.

This resulted in double compilation in many cases (i.e. the activation of the extension environment caused invalidation of packages installed from the Catalyst environment, so they were then recompiled I believe). Hence we had 40+ minute of stuff compiling before the extension tests even started running.

WIth this split that should be avoided, halving the compilation time. And now the extension tests should also cache their compiled libraries (so if you make updates to a PR, hopefully there is no recompilation except of the extension itself).

isaacsas:
The main change here is if we add new extension test files they need to be put in the runtests_extensions.jl file. Calling `Pkg.test()` locally will work just like before; it is only in CI that there are any changes.

PR #1367, opened 2026-02-01
Support `@brownians` in DSL

TorkelE:
Adds an `@brownian` option to the DSL:
```julia
    rs = @reaction_network begin
        @parameters η
        @brownians B
        @equations begin
            D(V) ~ X - V + η*B
            D(W) ~ V - W
        end
        (p,d), 0 <--> X
    end
```


test/test_functions.jl
```diff
@@ -37,8 +37,8 @@ end
 
 # Evaluates the the drift function of the ODE corresponding to a reaction network.
 # Also checks that in place and out of place evaluations are identical.
-function f_eval(rs::ReactionSystem, u, p, t; combinatoric_ratelaws = true)
-    prob = ODEProblem(rs, u, 0.0, p; combinatoric_ratelaws)
+function f_eval(rs::ReactionSystem, u, p, t; combinatoric_ratelaws = true, structural_simplify = false)
+    prob = ODEProblem(rs, u, 0.0, p; combinatoric_ratelaws, structural_simplify)
```
TorkelE, line 41: After this I will make a PR to chaneg all mentions of `structrual_simplify` to `mtkcompile`.

isaacsas:
This should probably handle merging these Brownians with autodiscovered ones in the internal make ReactionSystems constructor.

TorkelE:
I started by trying to pass Brownians onto `make_ReactionSystem_internal` which the DSL calls. But that one doesn't take any brownian argument? My impression of the current implementation is that that one will autodiscover all brownians stored in equations, and those are the only one part of the system?

isaacsas:
Just add them as a new fifth arg like the five argument constructor does. Then just handle them like we handle us or ps in that function. If we know them we should use that knowledge rather than rely on auto discovery as we do for the other components.

TorkelE:
I.e.
```julia
function make_ReactionSystem_internal(rxs_and_eqs::Vector, iv, us_in, ps_in, brownians = SymbolicT[];
        spatial_ivs = nothing, continuous_events = [], discrete_events = [],
        observed = [], kwargs...)
```
That should work. Fully agree on the sentiment that adding them directly is a good idea.

isaacsas:
I can make the update if you don't mind me pushing to this PR. But, yes that looks right.

We should probably at some point audit all these `[]` usages since in many cases we can now more concretely type stuff.

isaacsas:
(Not saying to handle [] here. That can be a later cleanup PR when we are further along.)

TorkelE:
Yeah, I will make that update here, also make a test to check that unsued brownians are still stored in the model.

PR #1368, opened 2026-02-01
Try initialization fix for Julia 1.10 with HC and MTKBase

isaacsas:


isaacsas:
@TorkelE this is a hack, but seems to fix the CI bug. But realistically, other orders of loading things could break it again. But this is harmless so I'm going to merge it and hope this doesn't resurface. The lts core tests should be fixed by a similar hack in MTK if they merge the PR I submitted.

PR #1388, opened 2026-02-07
Add @poissonians DSL option for @reaction_network and @network_component

isaacsas:
Requires https://github.com/SciML/ModelingToolkit.jl/pull/4274 and an MTKBase release before we can test/merge.

## Summary

Adds `@poissonians` as a DSL option in `@reaction_network` and `@network_component`, allowing users to declare poissonians inline rather than constructing them externally and interpolating.

```julia
rn = @reaction_network begin
    @parameters λ k d
    @variables X(t)
    @poissonians dN(λ)
    @equations D(X) ~ dN
    (k, d), 0 <--> S
end
```

Multiple poissonians and begin/end block syntax are supported:
```julia
@poissonians begin
    dN₁(λ₁)
    dN₂(λ₂)
end
```

Rate expressions can also be interpolated from external variables:
```julia
rate_expr = k * V
rn = @reaction_network begin
    @poissonians dN($(rate_expr))
    ...
end
```

**Note:** Symbols appearing in poissonian rates must be pre-declared (via `@parameters`, `@species`, etc.) or interpolated — they are not auto-inferred.

**Depends on:** A fix in ModelingToolkitBase that makes `@poissonians` consistently return `Vector{Num}` (matching `@brownians` behavior). Without that fix, the macro returns a scalar `Num` for a single poissonian and a `Tuple` for multiple.

### DIFF
--- a/src/dsl.jl
+++ b/src/dsl.jl
@@ -280,7 +280,6 @@ function make_reaction_system(ex::Expr, name)
     syms_inferred = union(sps_inferred, ps_inferred, vs_inferred, diffs_inferred)
     all_syms = union(syms_declared, syms_inferred)
     validate_poissonian_rate_syms(options, all_syms)
-
     obsexpr, obs_eqs, obs_syms = read_observables_option(options, ivs,
         union(sps_declared, vs_declared), all_syms; requiredec)
 


isaacsas:
@TorkelE please give this a review to make sure you are happy with how this has modified / integrated into the DSL.

isaacsas:
OK, I think this should be good modulo needing that MTK PR + a release.

src/dsl.jl
```diff
@@ -328,8 +334,8 @@ function make_reaction_system(ex::Expr, name)
         _default_reaction_metadata = $default_reaction_metadata
 
         remake_ReactionSystem_internal(
-            make_ReactionSystem_internal(rx_eq_vec, $tiv, us, ps, $brownsvar; name, spatial_ivs,
-                observed = _observed, continuous_events = _continuous_events,
+            make_ReactionSystem_internal(rx_eq_vec, $tiv, us, ps, $brownsvar; poissonians = $poissvar,
```
TorkelE, line 335: is there a reason we have to set it up so that we use `poissonians = $poissvar` for poissonians, but not for `$brownsvar`?

src/dsl.jl
```diff
@@ -278,6 +279,9 @@ function make_reaction_system(ex::Expr, name)
     ps_inferred = setdiff(ps_pre_inferred, vs_inferred, diffs_inferred)
     syms_inferred = union(sps_inferred, ps_inferred, vs_inferred, diffs_inferred)
     all_syms = union(syms_declared, syms_inferred)
+
```
TorkelE:
```suggestion
```

src/dsl.jl
```diff
@@ -278,6 +279,9 @@ function make_reaction_system(ex::Expr, name)
...
+    validate_poissonian_rate_syms(options, all_syms)
+
```
TorkelE:
```suggestion
```

src/dsl.jl
isaacsas: Yes, Brownians are an arg but Poissonians are a kwarg.

src/dsl.jl
```diff
@@ -278,6 +279,9 @@ function make_reaction_system(ex::Expr, name)
     ps_inferred = setdiff(ps_pre_inferred, vs_inferred, diffs_inferred)
     syms_inferred = union(sps_inferred, ps_inferred, vs_inferred, diffs_inferred)
     all_syms = union(syms_declared, syms_inferred)
+
+    validate_poissonian_rate_syms(options, all_syms)
```
TorkelE, line 282: I think we don't actually do this kind of check for other options that do not do inference (i.e. we simply use the default nothing declared that happens). If we do want to do this it might make sense to do it for all cases (but I can do this after v16 some time)

isaacsas: Yeah, we should probably do such checks, and have a way to disable them for performance. I didn't remember what we did so just had it added in given I'm not allowing auto-discovery from the rates within Poissonians.

src/dsl.jl
TorkelE: that makes total sense

PR #1404, opened 2026-02-10
Expand unit checking

### LINKED ISSUE #1400: Expand unit checking
Let's maintain a list of ways units checking could be improved:

- [ ] Check user-provided jumps (Poissonians are handled though within equations).
- [ ] Check events.

calebjubal:
## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
- [x] Any new documentation only uses public API
  
## Additional context

The implementation adds unit validation for user-provided jumps and events in Catalyst.jl's validate function.

For jumps, three new functions were added that check:

- ConstantRateJump: rate must have units of species/time, and affect equations must be unit-consistent
- VariableRateJump: same as ConstantRateJump
- MassActionJump: scaled_rates must have correct units accounting for reactant stoichiometry

For events, two new functions were added that check:

- SymbolicContinuousCallback: condition equations and affect equations are validated using MTK's check_units
- SymbolicDiscreteCallback: same validation for discrete events

The main validate(rs::ReactionSystem) function now loops over all jumps from MT.get_jumps(rs) and all events from MT.get_continuous_events(rs) and MT.get_discrete_events(rs), calling the appropriate validation function for each.

When a unit mismatch is found, a warning is issued describing the expected vs actual units, and the function returns false.

Tests were added covering correct and incorrect units for ConstantRateJump, VariableRateJump, discrete events, and continuous events.

Closes #1400


isaacsas:
Thank you for taking a pass at this. Unfortunately this doesn't look like the right approach. Maybe you are using AI and just let it work on this unsupervised, but this uses parts of the old ModelingToolkit unit checking we no longer use, and generally is not following our revised unit checking approach in a number of ways.

If you want to work on this I suggest spending some time carefully reading the new code and making sure you understand how it works and flows before then trying to add the new cases. Then, start with one of jumps or events, and limit yourself to the case that everything is purely symbolic (i.e. no ImperativeAffects or such).

isaacsas:
Also, a good rule of thumb for getting your PRs reviewed and merged is to keep them short and focused on one thing. The longer a PR is, and the more it tries to do, the longer it will take for us to have time to review it.

calebjubal:
> Thank you for taking a pass at this. Unfortunately this doesn't look like the right approach. Maybe you are using AI and just let it work on this unsupervised, but this uses parts of the old ModelingToolkit unit checking we no longer use, and generally is not following our revised unit checking approach in a number of ways.
> 
> If you want to work on this I suggest spending some time carefully reading the new code and making sure you understand how it works and flows before then trying to add the new cases. Then, start with one of jumps or events, and limit yourself to the case that everything is purely symbolic (i.e. no ImperativeAffects or such).

Hey @isaacsas, I'll take some time to go through the new unit-checking code path more carefully and align this with the revised approach. I’ll focus first on one of the symbolic cases (starting with jumps), clean up the use of the old ModelingToolkit logic, and reopen with a much smaller, focused change that fits the current architecture, thanks again for the clear guidance!

isaacsas:
Sounds good!

Let's break it into small PRs that make it easier to complete stuff, and easier for us to review. Maybe start with just getting `ConstantRateJump`s and `VariableRateJump`s, which should both work with the same code (so don't need separate functions), and ignore `MassActionJump`s.

isaacsas:
Take a look at how we handle checking for `Equation`s. You should be able to write analogous functions for jumps, and even reuse the `Equation` checks for the individual affect equations. The rate function should be checked to have inverse time units, which should be similar, but not identical, to how we check the units of the reaction rate term.

PR #1405, opened 2026-02-10
Update and slim doc dependencies

isaacsas:


docs/Project.toml
```diff
@@ -70,9 +71,9 @@ Latexify = "0.16.5"
 LinearSolve = "2.30, 3"
 ModelingToolkitBase = "1.11.1"
 NetworkLayout = "0.4"
-NonlinearSolve = "3.12, 4"
-Optim = "1.9"
```
TorkelE, line 74: Just for reference, we had this one because we use it for PEtab (which is commented out until the new PEtab version is out, hence it was not used, happy to add it in when we renable that one though)

isaacsas: Does PEtab not support Optimization.jl solvers?

ChrisRackauckas: Optim would be on v2 now though 😅

PR #1406, opened 2026-02-10
Fix some test_broken tests

isaacsas:


src/registered_functions.jl
```diff
@@ -118,18 +118,22 @@ function expand_registered_functions(eq::Equation)
...
 
-# If applied to a discrete event, returns it applied to condition and affects.
+# If applied to a discrete event, returns it applied to condition and affect.
 function expand_registered_functions(de::MT.SymbolicDiscreteCallback)
     conditions = expand_registered_functions(de.conditions)
-    affect = expand_registered_functions(de.affect.affect)
-    return MT.SymbolicDiscreteCallback(conditions, affect = de.affect.discrete_parameters)
+    affect_eqs = isnothing(de.affect) ? nothing : expand_registered_functions(de.affect.affect)
+    disc_ps = isnothing(de.affect) ? MT.SymbolicT[] : de.affect.discrete_parameters
+    return MT.SymbolicDiscreteCallback(conditions, affect_eqs; discrete_parameters = disc_ps)
```
isaacsas, line 136: @TorkelE please check these. The previous version seemed bugged, which is perhaps why the test was failing.

TorkelE: which test? don't follow, sorry, don't follow!

isaacsas: There are some tests in the custom CRN function file marked broken that work with these changes.

Issue #507, opened 2022-05-20
making combinatoric_ratelaws more flexible


isaacsas:
As @paulflang brought up via Slack it would be helpful for SBML if this could be specifed at a per-reaction level, but perhaps just being able to set it when building a `ReactionSystem` would be ok too?

isaacsas:
One problem with making it per reaction is that we'd probably want a way to specify it via the DSL on a per-reaction basis too.

paulflang:
Thanks for creating this issue, @isaacsas . I think it is fine if it is just specified on the `ReactionSystem` level as you suggest.

isaacsas:
OK, great. I'll try to add that in after I finish up the conservation law PR and before making a breaking release.

paulflang:
Thanks!

isaacsas:
Closed by https://github.com/SciML/Catalyst.jl/pull/510

PR #1466, opened 2026-04-26
Bump compat for OrdinaryDiffEq v7 / SciMLBase v3 ecosystem

ChrisRackauckas-Claude:
## Summary

Cascade the major-version compat relaxations required by the OrdinaryDiffEq v7 / SciMLBase v3 ecosystem release:

- **SciML/OrdinaryDiffEq.jl#3562** — cascade major bumps across all 54 monorepo sublibs
- **SciML/OrdinaryDiffEq.jl#3565** — revert OrdinaryDiffEqCore (stays v4) and DiffEqDevTools (stays v3)
- **SciML/OrdinaryDiffEq.jl NEWS.md** — breaking-change reference

### Compat changes

| Package | Before | After |
|---|---|---|
| OrdinaryDiffEqBDF | `"1"` | `"1, 2"` |
| OrdinaryDiffEqCore | `"3.22"` | `"3.22, 4"` |
| OrdinaryDiffEqDefault | `"1"` | `"1, 2"` |
| OrdinaryDiffEqRosenbrock | `"1"` | `"1, 2"` |
| OrdinaryDiffEqSDIRK | `"1"` (docs only) | `"1, 2"` |
| OrdinaryDiffEqTsit5 | `"1"` | `"1, 2"` |
| OrdinaryDiffEqVerner | `"1"` | `"1, 2"` |
| SciMLBase | `"2.84"` | `"2.84, 3"` |
| StochasticDiffEq | `"6.101"` | `"6.101, 7"` |

Applied to: `Project.toml`, `docs/Project.toml`, `test/extensions/Project.toml`.

### Source migrations

Two breaking-change sites were found and fixed:

1. **`verbose = false` removed** (`test/simulation_and_solving/hybrid_models.jl`):
   In v7, `Bool` is no longer accepted for the `verbose` kwarg — must use `ODEVerbosity(...)`. Since these calls were only suppressing output in tests, the kwarg was removed.

2. **EnsembleSolution iteration** (`test/simulation_and_solving/simulate_SDEs.jl`):
   RAT v4 (`RecursiveArrayTools v4`, part of SciMLBase v3) changed `EnsembleSolution` to subtype `AbstractArray`, so `for sol in ensemble_sol` now iterates over scalar elements rather than trajectory solutions. Fixed by iterating `ensemble_sol.u` instead.

## Test plan

- [ ] CI passes on `OrdinaryDiffEqTsit5 v2`, `OrdinaryDiffEqRosenbrock v2`, `OrdinaryDiffEqBDF v2`, `OrdinaryDiffEqDefault v2`, `OrdinaryDiffEqVerner v2`, `StochasticDiffEq v7`, `SciMLBase v3`
- [ ] CI passes on legacy v1/v6/v2 versions too (compat allows both)
- [ ] No regressions in existing tests


test/simulation_and_solving/hybrid_models.jl
```diff
@@ -188,7 +188,7 @@ let
 
     # Solves and checks values of solution (do this before integrator mutation test
     # since integrator mutation affects shared parameter state).
-    sol = solve(prob, Tsit5(); maxiters = 10, verbose = false)
```
TorkelE, line 191: Is `verobes = false` now the default for `solve`?

test/simulation_and_solving/hybrid_models.jl
```diff
@@ -188,7 +188,7 @@ let
...
+    sol = solve(prob, Tsit5(); maxiters = 10)
```
ChrisRackauckas, line 191:
```suggestion
    sol = solve(prob, Tsit5(); maxiters = 10, verbose = ODEVerbosity(SciMLLogging.None()))
```

test/simulation_and_solving/hybrid_models.jl
```diff
@@ -215,7 +215,7 @@ let
     @test prob2.ps[:d] == 0.5   # not changed
 
     # Test that we can solve the remade problem and verify solution values.
-    sol2 = solve(prob2, Tsit5(); maxiters = 10, verbose = false)
+    sol2 = solve(prob2, Tsit5(); maxiters = 10)
```
ChrisRackauckas, line 218:
```suggestion
    sol2 = solve(prob2, Tsit5(); maxiters = 10, verbose = ODEVerbosity(SciMLLogging.None()))
```

isaacsas:
Please remove format related changes from this PR so we can see a clean diff of what is being changed. Format related changes should be restricted to standalone PRs that do not change any code.

test/simulation_and_solving/hybrid_models.jl
```diff
@@ -1013,21 +1031,21 @@ let
         du[1] = p[1]  # k1 (A production drift)
         du[2] = 0.0
         du[3] = 0.0
-        du[4] = 0.0
+        return du[4] = 0.0
```
isaacsas: This and the three subsequent added returns are not correct. If a return is added it should be to return `nothing`.

isaacsas (changes requested):

test/simulation_and_solving/hybrid_models.jl
ChrisRackauckas: No, it doesn't use booleans at all.

ChrisRackauckas:
The rest is just unlocking the downstream repos.

isaacsas:
Why does this require a breaking release and dropping older version support? Is it from `unwrapped_f`?

ChrisRackauckas:
That's just to force the tests forward to ensure nothing is held back. It can be dropped when merging, and would make all go green except the extensions (which seems broken on master?). If you want to merge and continue the extensions separate it can just get pulled back to merge, otherwise it's safe to keep the forced bumps so the tests don't cheat when re-bumping from downstream.

isaacsas:
> That's just to force the tests forward to ensure nothing is held back. It can be dropped when merging, and would make all go green except the extensions (which seems broken on master?). If you want to merge and continue the extensions separate it can just get pulled back to merge, otherwise it's safe to keep the forced bumps so the tests don't cheat when re-bumping from downstream. 

I'm fine either way with regards to merging, but would prefer we keep releases backwards compatible and avoid a breaking release. I'm not sure what the issue is with extensions. They worked when we last released and we haven't really changed anything here. Did the StructuralIdentifiability issue ever get fixed?

ChrisRackauckas-Claude:
Reverted the forced-bump tightening commit per discussion — restored the lower bounds (legacy major ranges + Catalyst stays at v16.1.1) so this is a non-breaking release. Compat is back to:

- `DiffEqBase = "6.165.0, 7"`
- `OrdinaryDiffEqBDF/Default/Rosenbrock/Tsit5/Verner = "1, 2"`
- `OrdinaryDiffEqCore = "3.22, 4"`
- `SciMLBase = "2.84, 3"`
- `StochasticDiffEq = "6.101, 7"`

Same lower bounds restored in `docs/Project.toml` and `test/extensions/Project.toml`. Ready to merge.

ChrisRackauckas-Claude:
## Rebased onto current `master` + local test verification

Rebased onto `master` (tip `6d039ec2`, _Bump version to 16.1.2_). The branch had accumulated revert/re-tighten compat churn (`Drop legacy` → `Restore` → `Re-tighten` → `Revert`) that net out to nothing, so the rebase is flattened into a single commit carrying the final non-breaking compat ranges plus the source/test migrations. The previous `master` vs branch conflict (master inserted `OrdinaryDiffEqNonlinearSolve` into the `docs/Project.toml` compat block) is resolved, and `OrdinaryDiffEqNonlinearSolve` is bumped to `"1, 2"` for consistency with the rest of the v7 ecosystem. PR is now `MERGEABLE`/`CLEAN`.

### Does anything else need to update to unblock the tests? — No.

Resolved the full test environment on both CI stable channels (Julia **1.10 lts** and **1.12**). The whole dependency graph is co-installable at the v7/v3 ecosystem with **nothing held back**:

| Package | Resolved | | Package | Resolved |
|---|---|---|---|---|
| SciMLBase | 3.16.0 | | OrdinaryDiffEqCore | 4.3.0 |
| DiffEqBase | 7.5.3 | | OrdinaryDiffEq{Tsit5,BDF,Default,Rosenbrock,Verner,NLSolve} | 2.x |
| RecursiveArrayTools | 4.3.0 | | StochasticDiffEq | 7.0.0 |
| ModelingToolkitBase | 1.41.0 | | JumpProcesses | 9.29.0 |

MTKBase, JumpProcesses, DiffEqCallbacks, NonlinearSolve, SteadyStateDiffEq, Symbolics, etc. already permit SciMLBase v3 / OrdinaryDiffEq v7, so the compat relaxations here are sufficient on their own — no upstream PR is required to unblock CI.

### Local test results (all green)

Ran each CI group against the resolved v7/v3 environment (`Pkg.test`-equivalent sandbox: package `[deps]` + test `[targets]`):

| Group | Julia 1.10 (lts) | Julia 1.12 ("1") |
|---|---|---|
| Modeling | ✅ | — |
| Simulation | ✅ | ✅ |
| Hybrid | ✅ | ✅ |
| Misc | ✅ | — |
| Spatial | ✅ | ✅ |

Migration-affected groups (Simulation, Hybrid, Spatial) verified on both channels; Modeling/Misc verified on lts. Zero failures; the only `Broken` entries are pre-existing `@test_broken`s. The three migrations all exercise correctly:
- `SciMLBase.unwrapped_f` in `rebuild_spat_internals!` (v7 wraps `f` in a `FunctionWrappersWrapper`) — Spatial passes.
- `SymbolicIndexingInterface.setp` for jump callback param updates — Simulation/`simulate_jumps` passes.
- iterating `EnsembleSolution.u` (RAT v4 makes it an `AbstractArray`) — Simulation/`simulate_SDEs` passes.
- dropping `verbose = false` — Hybrid passes.

### Formatting

The four changed `.jl` files are **format-neutral**: Runic-normalizing both `master` and this branch leaves exactly the four intended migrations as the only diff. (Whole-file Runic v1.7.0 reports churn on pre-existing, untouched lines — a Runic-version question for `master`, not this PR — so those untouched lines were deliberately left alone.)

Please ignore until reviewed by @ChrisRackauckas.

TorkelE:
There is some discussion here: https://github.com/SciML/OrdinaryDiffEq.jl/issues/3667#issuecomment-4577254175 which might affect how we want to do this, will have a look once I have a better idea of that (or next week if that delays)

ChrisRackauckas-Claude:
## Old-major (downgrade) verification + Extensions triage

Following up to confirm the **lower half** of the dual compat ranges (the legacy `SciMLBase v2` / `OrdinaryDiffEq v1` / `OrdinaryDiffEqCore v3` / `StochasticDiffEq v6` / `DiffEqBase v6` ecosystem) still works — not just the new v7/v3 majors.

### Old majors actually resolve, despite the MTKBase floor

`ModelingToolkitBase` 1.17–1.30 *require* SciMLBase v2 (only 1.31+ allow v3), and Catalyst's floor is MTKBase ≥ 1.17, so SciMLBase v2 stays reachable. Forcing the old majors on Julia 1.10 resolves to: **SciMLBase 2.155.1, OrdinaryDiffEqCore 3.33.1, OrdinaryDiffEqTsit5 1.12.0, StochasticDiffEq 6.102.0, DiffEqBase 6.218.0** (newest patches within each old-major line).

### All five groups pass on the old majors (Julia 1.10)

| Group | New majors (v7/v3) | Old majors (v2/v1/v6/v3) |
|---|---|---|
| Modeling | ✅ (CI 1/lts/pre) | ✅ |
| Simulation | ✅ (CI 1/lts/pre) | ✅ |
| Hybrid | ✅ (CI 1/lts/pre) | ✅ |
| Misc | ✅ (CI 1/lts/pre) | ✅ |
| Spatial | ✅ (CI 1/lts/pre) | ✅ |

All four migrations are backward-compatible: `SciMLBase.unwrapped_f` exists in SciMLBase 2.155 (identity fallback on the un-wrapped v2 function), and `setp` / `EnsembleSolution.u` / dropping `verbose=false` all work on the old majors. The 15 main-suite CI jobs (1/lts/pre × 5 groups) are green for the new majors.

### Extensions Tests failure is pre-existing and orthogonal — not from this PR

`Extensions Tests` is red, but it is **already red on `master`** (`6d039ec2`) and has been since ~2026-04-14, independent of this PR:

- **Cause:** upstream **StructuralIdentifiability v0.5.20+** added a restriction (`"States on the left-hand side must appear with derivations"`) that breaks Catalyst's SI extension on conservation-law / DAE models. Failing testset: `Structural Identifiability Extension` (`DomainError ... Any[C(t)]`), 36 pass / 1 error.
- **Identical on master and this PR** — same testset, same error, same SI build (0.5.21), same `ModelingToolkitBase 1.41.0`. The extensions env resolves to SciMLBase **v2** either way (BifurcationKit/HC/SI cap it), so it is unrelated to the SciMLBase-v3 migration.
- The other extension testsets pass on this branch (I ran `Steady State Stability Computations` and `Discrete Space Simulation Plotting` standalone — both green; `dspace_simulation_plotting.jl` doesn't touch the changed spatial code). This PR introduces **no new** extension failures.
- Tracked/fixed by **#1462** ("Fix StructuralIdentifiability extension typing, DAE support, and test coverage") and upstream SI #507.

Net: this PR is non-breaking on both ends of the compat ranges. The remaining red `Extensions` check is a separate, pre-existing SI-v0.5.20 regression and should be resolved by #1462.

test/extensions/Project.toml
isaacsas, line 1: Should these actually have compat entries? I was mimicking other SciML repos in not having them here.

isaacsas (commented):
Other than the extension Project.toml getting combat entries suddenly this looks fine. I agree the extension test failures are preexisting and due to SI still not having a fix for handling the range parameter types in current MTK that broke in that earlier version.

test/extensions/Project.toml
ChrisRackauckas-Claude: Good call — removed (force-pushed 402e15d3), so `test/extensions/Project.toml` now matches master. I'd added them reflexively while mirroring the other Project.tomls, but they're inert here: the extensions env can't reach OrdinaryDiffEq v2 anyway (BifurcationKit/HomotopyContinuation/StructuralIdentifiability cap SciMLBase at v2, so the solvers resolve to v1 — Tsit5 1.12.0, Default 1.14.0, Verner 1.14.0). I confirmed the resolved versions are identical with and without the `[compat]` block.

test/simulation_and_solving/hybrid_models.jl
ChrisRackauckas-Claude: Those `return` additions came from the Runic-formatting commit in the earlier version of this branch; the rebase flattened that churn out. The current branch adds no `return` statements anywhere — the only change in hybrid_models.jl now is dropping `verbose = false` (diff is just those two lines). So this should be resolved; please re-check on 402e15d3.

ChrisRackauckas:
Handled.

test/simulation_and_solving/hybrid_models.jl
TorkelE, line 191: I am still confussed here, `verbose = false` (or whatever is the new syntax) is not the default, right? I.e. we would need to put whatever is the new thing here?

TorkelE:
I am a bit confussed as even after this update we use what I thought was old syntax in the docs, e.g. like
image
However, this does not actually seem to error or generate any problems?

ChrisRackauckas:
Well new docs are blocked by this PR still.

test/simulation_and_solving/hybrid_models.jl
ChrisRackauckas: `verbose = DEVerbosity(SciMLLogging.None())`

ChrisRackauckas: That's if you want literally everything off, though it's probably not a good idea to always do that 😅

TorkelE:
Yeah, but don't we want to suet he new syntax in the docs? Or do we just handle that in a follow-up PR?

test/simulation_and_solving/hybrid_models.jl
TorkelE: To be honest, not sure what the intention is for these tests, I didn't write them

ChrisRackauckas:
The new docs won't build until Catalyst lets it.

test/simulation_and_solving/hybrid_models.jl
ChrisRackauckas: I don't think it needs verbosity turned off.

TorkelE (approved):
Ok, sounds good, happy to merge and handle anything else in follow-ups

PR #1469, opened 2026-05-12
Add some documentation on `@browians`/`@poissnians`

TorkelE:
Adds short sections to the coupled non-CRN model section on these two cases.


docs/src/model_creation/coupled_non_crn_models.md
```diff
@@ -299,3 +299,70 @@ nonspecies(rs)
...
+While brownians model continuous noise, poissonians model discrete noise events. While brownians are associated with stochastic differential processes, poissonians are associated with jump processes. Consider the previous model, but where the volume fluctuates due to discrete noisy events (such as the absorption of vesicles at random time points). The rate/intensity of the poissonian process must be specified explicitly; here we declare it as a parameter `λ`.
+```@example coupled_eqs_noise
+noisy_cell = @reaction_network begin
+    @parameters λ
+    @poissonians dN(λ)
+    @equations D(V) ~ X - V + dN
+    (p,d), 0 <--> X
+end
+```
+The model can be simulated using standard syntax. However, poissonians can only be simulated through `HybridProblem`s. The reason is that the variable subject to the poissonian is both governed by a differential equation and discrete jump process. Below, we perform a hybrid simulation for our model, where the $X$'s reactions and the poissonian are simulated as jumps, while $V$'s governing equation is simulated as an ODE.
```
isaacsas: The V’s equation is probably better called a PDMP (Piecewise Deterministic Markov Process).

isaacsas (commented):
Minor comment — might be nice to in one case write the mathematical equation for the model and point out the noise is a standard Brownian Motion for one SDE case and a time-changed Poisson process for one jump case. I think we have an SDE latex mode now that can generate the equations correctly right?

TorkelE:
Adding the equations and doing that is definitely a good idea, will add it in

TorkelE:
I have added the equations in for the brownian case, but was a bit unsure how to formally write the poissonian case.

PR #1471, opened 2026-05-20
[WIP] Hybrid modelling doc page

TorkelE:
Adds an initial documentation page on hybrid simulations. Can be expanded later, but this gives us something to show that this is a thing that we can do.


isaacsas (approved):
Looks good, but maybe it should also be mentioned that one can pass scales to HybridProblem too to override the model defaults?

TorkelE:
I have added such a section, agree that it is very useful to know

PR #1472, opened 2026-05-24
docs(faqs): explicitly load OrdinaryDiffEqNonlinearSolve

asinghvi17:
## Summary
The first `@example faq1` block in `docs/src/faqs.md` only `using`s `Catalyst, OrdinaryDiffEqTsit5, Plots`, but the third block (`oprob = ODEProblem(osys, ...)` + `solve(oprob, Tsit5())`) requires DAE initialization, which in turn needs a nonlinear solver to be loaded. This currently works in sequential docs builds because some earlier page in build order already pulled in `OrdinaryDiffEqNonlinearSolve`, but the page is not actually self-contained.

When the page is expanded in isolation (e.g. via DocumenterParallel, which runs each page in its own Malt worker), the block fails with:

```
This ODE requires a DAE initialization and thus a nonlinear solve but no nonlinear solve has been loaded.
To solve this problem, do `using OrdinaryDiffEqNonlinearSolve` or pass a custom `nlsolve` choice into the `initializealg`.
```

This PR makes the page self-contained by adding `OrdinaryDiffEqNonlinearSolve` to its `using` and to `docs/Project.toml`.

## Test plan
- [x] `Pkg.instantiate()` resolves clean after the change
- [x] `using Catalyst, OrdinaryDiffEqTsit5, OrdinaryDiffEqNonlinearSolve, Plots` loads on a fresh process
- [ ] CI docs build still passes


asinghvi17:
Huh a funny bit of history...my fork of this repo goes back to when it was called DiffEqBiological.jl!

isaacsas:
Thanks!

PR #1475, opened 2026-06-01
refactor: ignore certain system metadata during serialization

AayushSabharwal:
This refactors system metadata serialization to more robustly ignore `MutableCacheKey`

### DIFF
--- a/src/reactionsystem_serialisation/serialisation_support.jl
+++ b/src/reactionsystem_serialisation/serialisation_support.jl
@@ -245,6 +245,25 @@ const SKIPPED_METADATA = [
     Symbolics.VariableDefaultValue,
     Symbolics.VariableSource]
 
+# List of metadata keys in an `AbstractSystem` which shouldn't be serialized.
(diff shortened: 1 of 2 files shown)


TorkelE:
Looks good and a long term improvement to handle this, thanks! I think this would write a `metadata = Dict([])` in all cases even when there is not metadata though. I think it would be good to not have the case (i.e. right now I have tried to ensure that nothing that the user has not touched appears in the final code, the idea is also to make the code a nice way to check the content of a model). Maybe we canjust add a `has_system_metadata` check ith an `if` statement as well? Happy to ahve a look at it if you are unsure.

AayushSabharwal:
Sure, I can make the change. Should be trivial.

ChrisRackauckas:
We should have serialization functions in ModelingToolkit rather than handling this downstream.

TorkelE (approved):

AayushSabharwal:
Test failures seem to be the same as on master

TorkelE:
yeah, I will fix separately

AayushSabharwal:
Thank you!

PR #1477, opened 2026-06-03
Make docs compatible with Turing 0.45 (handles #1476)

ChrisRackauckas-Claude:
> **Please ignore until reviewed by @ChrisRackauckas.** Opened as a draft.

Handles the code changes required for the Dependabot group bump #1476, which on its own only relaxes compat ranges. This branch is stacked on the Dependabot commit (so it includes those compat bumps) and adds the one source-level change those bumps actually require.

### Turing 0.45 (the only breaking change requiring code edits)
Turing 0.45 makes `FlexiChains` the default chain type and **no longer re-exports MCMCChains**. The Bayesian parameter-fitting tutorial (`docs/src/inverse_problems/turing_ode_param_fitting.md`) relies on MCMCChains-specific behaviour — `chain.value[...]` indexing and the StatsPlots `plot(chain)` recipe. Under 0.45 the default chain is a `FlexiChain`, and `chain.value` throws `type FlexiChain has no field value`.

Fix:
- Pass `chain_type = MCMCChains.Chains` to both `sample(...)` calls.
- `using MCMCChains` explicitly (no longer transitively available via Turing).
- Add `MCMCChains` to `docs/Project.toml` (`[deps]` + compat `"6, 7"`).

Verified locally on Julia 1.10: reproduced the `chain.value` failure on the default chain, and confirmed `chain_type = MCMCChains.Chains` restores the expected `Chains` type and indexing.

### BifurcationKit 0.8 — inert, no code change
The compat bump to `0.8` is currently a no-op. Catalyst's BifurcationKit extension delegates to ModelingToolkitBase's `MTKBifurcationKitExt`, whose weakcompat pins `BifurcationKit = "0.4 - 0.5"` for all released 1.x versions. With the relaxed compat the resolver selects **BifurcationKit 0.5.8** (verified), so the extension and tests run against 0.5 exactly as before. BifurcationKit 0.8 only becomes selectable once ModelingToolkitBase bumps `MTKBifurcationKitExt` and releases — an upstream prerequisite, tracked separately.

### NonlinearSolveFirstOrder 2.1 / LikelihoodProfiler 1.5.1 — no code change
The HH example only does `using NonlinearSolveFirstOrder` + default `solve` (no affected API). LikelihoodProfiler 1.5.1 is additive; the docs use the unchanged `ProfileLikelihoodProblem`/`OptimizationProfiler`/`solve` surface.


ChrisRackauckas (approved):

isaacsas:
@TorkelE can you look at this one?

TorkelE (approved):
For now this seems like a good solution ot make progress, but I think I'd like to modify the tutorial to go back to the default Turing chain type later on.

ChrisRackauckas-Claude:
Prerequisite for the BifurcationKit 0.8 part: SciML/ModelingToolkit.jl#4588 widens `ModelingToolkitBase`'s `BifurcationKit` weakcompat to include 0.8. Until that merges and releases, the `0.8` entry added here is inert (resolver stays on 0.5.8). Verified: with the dev'd MTKBase from that PR + BifurcationKit 0.8.0, Catalyst's full BifurcationKit extension suite passes 35/35 with no Catalyst-side code change.

Issue #1485, opened 2026-06-09
Structural Identifiability analysis broken for models with coupled algebraic equations.


TorkelE:
```julia
rs = @reaction_network begin
    @parameters k c1 c2
    @variables C(t)
    @equations begin
        D(V) ~ k*X - V
        C ~ (c1 + c2) * X/V
    end
    (p/V,d/V), 0 <--> X
end
@unpack p, d, k, c1, c2 = rs

# Tests identifiability assessment when all unknowns are measured.
remove_conserved = false
assess_identifiability(rs; measured_quantities = [:X, :V, :C], loglevel, remove_conserved)
```
gives
```
ERROR: DomainError with States on the left-hand side must appear with derivations (and this is not the case for Any[C(t)]). Did you mean to mtkcompile the model?:

Stacktrace:
  [1] __mtk_to_si(de::System, measured_quantities::Vector{Tuple{…}})
    @ ModelingToolkitSIExt C:\Users\Torkel\.julia\packages\StructuralIdentifiability\31snG\ext\ModelingToolkitSIExt.jl:300
  [2] mtk_to_si
    @ C:\Users\Torkel\.julia\packages\StructuralIdentifiability\31snG\ext\ModelingToolkitSIExt.jl:127 [inlined]
  [3] _assess_identifiability(ode::System; measured_quantities::Vector{…}, funcs_to_check::Vector{…}, known_ic::Vector{…}, prob_threshold::Float64)
    @ ModelingToolkitSIExt C:\Users\Torkel\.julia\packages\StructuralIdentifiability\31snG\ext\ModelingToolkitSIExt.jl:533
  [4] _assess_identifiability
    @ C:\Users\Torkel\.julia\packages\StructuralIdentifiability\31snG\ext\ModelingToolkitSIExt.jl:526 [inlined]
  [5] #62
    @ C:\Users\Torkel\.julia\packages\StructuralIdentifiability\31snG\ext\ModelingToolkitSIExt.jl:516 [inlined]
  [6] with_logstate(f::ModelingToolkitSIExt.var"#62#63"{…}, logstate::Base.CoreLogging.LogState)
    @ Base.CoreLogging .\logging\logging.jl:542
  [7] with_logger
    @ .\logging\logging.jl:653 [inlined]
  [8] #assess_identifiability#60
    @ C:\Users\Torkel\.julia\packages\StructuralIdentifiability\31snG\ext\ModelingToolkitSIExt.jl:515 [inlined]
[...]

PR #1486, opened 2026-06-09
Handle test failures

TorkelE:
The failures seem to be due to https://github.com/SciML/Catalyst.jl/issues/1485. It would be nice to have fixed, but it also feels like a rather obscure case, so for now I am simply disabling it with a more helpful error.


ChrisRackauckas:
Yes thanks, this will be nicer on the automated systems. We can follow up with the issue but this has been persistent so let's mark it broken for now.

PR #1487, opened 2026-06-14
Use SciMLTesting v1.1.0 run_tests harness

ChrisRackauckas-Claude:
Replaces the hand-written `GROUP` if-ladder in `test/runtests.jl` with a single declarative `run_tests` call from [SciMLTesting](https://github.com/SciML/SciMLTesting.jl) v1.1.0.

## What changed
- The six functional groups (`Modeling`, `Simulation`, `Hybrid`, `Misc`, `Spatial`, `Extensions`) are passed as `groups` entries.
- A curated `all = ["Modeling", "Simulation", "Hybrid", "Misc", "Spatial", "Extensions"]` reproduces the original `All` membership **exactly**: bare/`All` runs all six groups in the original order. This repo has no QA group.
- The `Extensions` group keeps its isolated `test/extensions` environment via the `env` kwarg, which performs the same `Pkg.activate` / `Pkg.develop(repo root)` / `Pkg.instantiate` that the old `activate_extensions_env()` helper did.

## Behavior equivalence
This is a behavior-equivalent refactor — no test assertions or logic changed, and the set of testsets run under each `GROUP` value (including bare/`All`) is identical before and after. Verified with a dispatch dry-run:

| GROUP | groups run |
|---|---|
| `All` | Modeling, Simulation, Hybrid, Misc, Spatial, Extensions |
| `Modeling` / `Simulation` / `Hybrid` / `Misc` / `Spatial` / `Extensions` | the named group only |

## Project.toml
- Added `SciMLTesting` (`09d9d899-5365-40a9-917a-5f67fddea283`) to `[extras]`, `[compat] SciMLTesting = "1"`, and the test target.
- Dropped `Pkg` from the test deps — only the old harness used it; `run_tests` bundles `Pkg`.

---
Ignore until reviewed by @ChrisRackauckas.

PR #1488, opened 2026-06-15
Adapt docs to latest SciML releases

TorkelE:
Updates docs tto latest SciML package udpates.


ChrisRackauckas:
This is missing the attractors floor?

TorkelE:
Yeah, there is a few things missing, but I will add that when I update things for when the new OrdinaryDiffEq.jl is available as I need that anyway.

ChrisRackauckas:
It should be available now.

TorkelE:
excellent, will have a go at it thsi afternoon

TorkelE:
Awaiting https://github.com/SciML/OrdinaryDiffEq.jl/pull/3771, but after that this should hopefully be quick work

ChrisRackauckas:
I wouldn't worry about the QA here, https://github.com/SciML/Catalyst.jl/actions/runs/28762798445/job/85281268352?pr=1488 , that will shortly get replaced the moment you finish this https://github.com/SciML/Catalyst.jl/pull/1487 with a more updated QA setup. The segfault is probably the JET thing that this works around. So, I think this is good sans the modeling LTS failure. That's probably from writing the catalyst equivalence in a way that isn't robust to changing symbolic hashes, i.e. the ordering of the variables `x + y` vs `y + x` kind of thing.

TorkelE:
Sounds good, will have a look at that failure and merge later today I hope.

PR #1506, opened 2026-07-10
Initialize conserved constants in ODE and SDE systems

ChrisRackauckas-Claude:
This PR should be ignored until reviewed by @ChrisRackauckas.

## Summary
- Pass conservation-law initialization equations through ODE/SDE/hybrid system construction when conserved species are eliminated.
- Promote the recovered complete-system matrix metadata assertion from `@test_broken` to `@test`.

## Root Cause
With the current post-yank resolver (`SciMLBase v3.33.1`, `ModelingToolkitBase v1.51.0`), conserved constants such as `Γ` were left at default values in ODE/SDE systems with `remove_conserved=true`. The steady-state conversion path already supplied these initialization equations; the ODE/SDE paths did not.

## Validation
- `git diff --check`
- `timeout 3600 env JULIA_DEPOT_PATH=/home/crackauc/sandbox/tmp_20260708_165240_11300/julia_depot_catalyst_clean_modeling GROUP=Modeling JULIA_PKG_PRECOMPILE_AUTO=0 /home/crackauc/.juliaup/bin/julia +1.11 --startup-file=no --color=yes -e 'using Pkg; Pkg.activate(temp=true); if isempty(Pkg.Registry.reachable_registries()); Pkg.Registry.add(Pkg.RegistrySpec(path="/home/crackauc/sandbox/tmp_20260708_165240_11300/General_pre_retrocap_registry")); end; Pkg.develop(Pkg.PackageSpec(path=pwd())); Pkg.test("Catalyst"; coverage=false)'`
  - Result: `Catalyst tests passed`; `Conservation Laws | 250 pass, 17 broken, 267 total`; `ReactionSystem Structure | 419 pass, 2 broken, 421 total`.
- Runic was attempted. Full-file checks are not clean on the current repo state and would require broad unrelated rewrites, so this PR keeps formatting changes scoped to the touched lines.

### DIFF
--- a/src/reactionsystem_conversions.jl
+++ b/src/reactionsystem_conversions.jl
@@ -715,8 +715,12 @@ function hybrid_model(rs::ReactionSystem;
     jumps = vcat(rxn_jumps, user_jumps)
 
     # --- Add constraints (BC species, constraint equations, conserved species) ---
+    initeqs = Equation[]
     if has_continuous
-        eqs, us, ps, obs, ics = addconstraints!(eqs, flatrs, ists, ispcs; remove_conserved)
+        eqs, us, ps, obs, ics, initeqs = addconstraints!(
+            eqs, flatrs, ists, ispcs;
+            remove_conserved, compute_cl_initeqs = remove_conserved
+        )
     else
         # Pure jump case.
         any(isbc, get_unknowns(flatrs)) &&
@@ -729,10 +733,12 @@ function hybrid_model(rs::ReactionSystem;
 
     # --- Construct unified System ---
     # Note: brownians is a positional arg (5th) in the System constructor.
-    MT.System(eqs, get_iv(flatrs), us, ps, brownian_vars;
+    return MT.System(
+        eqs, get_iv(flatrs), us, ps, brownian_vars;
         poissonians = user_poissonians,
         jumps,
         observed = obs,
+        initialization_eqs = initeqs,
         name,
         bindings = MT.get_bindings(flatrs),
         initial_conditions = merge(initial_conditions, ics),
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
CI follow-up after the first run:

- I reproduced the current `master` Modeling failure locally on clean `deb00ddd9b9532370a1b6569c3b6a34d86b2f19c` with Julia 1.10 and the public resolver stack (`SciMLBase v3.33.0`, `ModelingToolkitBase v1.51.0`, `SymbolicUtils v4.39.0`). It fails before the DSL tests with the existing `@test_broken` unexpected pass at `test/reactionsystem_core/reactionsystem.jl:1150`, which this PR promotes to a real `@test`.
- Master CI shows the same earlier failure in Modeling jobs: https://github.com/SciML/Catalyst.jl/actions/runs/29061522873/job/86264219144
- After this PR fixes that first failure, the LTS Modeling job reaches additional DSL option failures. I left those out of this PR to keep the patch focused.
- Extensions and docs are also already failing on master; I filed https://github.com/SciML/Catalyst.jl/issues/1507 for those unrelated failures.
- Some PR jobs are still resolving `SciMLBase v3.33.0` until the General yank PR lands: https://github.com/JuliaRegistries/General/pull/160637

isaacsas:
The different ways we are handling conservation laws is intentional per @AayushSabharwal's instructions. We would need him to review any changes and confirm the modified approach is correct.

AayushSabharwal:
I don't get why this is necessary.

ChrisRackauckas:
Yeah it shouldn't be necessary, it should be handled in initialization?

AayushSabharwal:
Is it not? What's the MWE?

ChrisRackauckas-Claude:
Here is the requested MWE, reduced to pure ModelingToolkitBase:

```julia
using ModelingToolkitBase
using SciMLBase: ODEProblem

t = ModelingToolkitBase.t_nounits
D = ModelingToolkitBase.D_nounits
@variables X1(t) X2(t)
@parameters k1 k2 Γ[1:1] = missing [guess = [Initial(X1 + X2)]]
@named osys = System(
    [D(X1) ~ k1 * (Γ[1] - X1) - k2 * X1],
    t, [X1], [k1, k2, Γ];
    observed = [X2 ~ Γ[1] - X1],
)
osys = complete(osys)
prob = ODEProblem(
    osys,
    [X1 => 1.0, X2 => 2.0, k1 => 0.1, k2 => 0.2],
    (0.0, 1.0),
)
@show prob[X2] prob.ps[Γ[1]]
```

The regression boundary is ModelingToolkitBase 1.48.0 (outputs `2.0`, `3.0`) to 1.48.1 (outputs `0.0`, `1.0`). The specialized IR substitution path does not descend into the `Initial` operator call.

The upstream fix is SciML/ModelingToolkit.jl#4731. Against that fix, the complete Catalyst `GROUP=Modeling` command exits 0 with `Conservation Laws | 250 passed, 17 broken`. This indicates the behavior should be corrected upstream; the Catalyst constructor workaround should not be necessary once the MTKBase fix is merged and released.

PR #1512, opened 2026-07-15
Activate corrected conservation-law tests

ChrisRackauckas-Claude:
## Summary

- Convert seven conservation-law assertions from expected-broken to active tests.
- Record the dependency boundary that corrected the parameter-substitution behavior those assertions exercise.

This branch is rebased directly onto `master` and contains only the conservation-law test change.

## Cause

Catalyst commit `9d3448a6e614f2ddcd7b5c4866a7347266b2b6fd` marked these seven assertions broken on June 29. They now pass on both the current and generated minimum dependency graphs, so `@test_broken` reports seven unexpected-pass errors on clean master.

A public problem-construction probe bisected the original correction to ModelingToolkit commit `1c253adc2aa55934fa110dd664bdf27c4fadc0f1` in #4704, first released in ModelingToolkitBase 1.52.0:

- parent `96275f9cd`: reconstructed observed species `X2 = 0.0`, conservation parameter `Γ = 1.0`;
- correcting commit: `X2 = 2.0`, `Γ = 3.0`.

The change evaluates the operating-point mapping before filling `MTKParameters`, preserving the conservation substitutions these assertions exercise. Catalyst #1506 independently repairs the older MTKBase 1.51 behavior through initialization-equation propagation.

## Local verification

- Julia 1.10.11, clean master, generated minimum graph (MTKBase 1.46.0 / SciMLBase 3.18.0 / OrdinaryDiffEq 7.0.0): 242 passed, 7 unexpected-pass errors, 17 pre-existing broken, exit 1.
- Julia 1.10.11, this branch, the same generated minimum graph: 249 passed, 17 pre-existing broken, 266 total, exit 0.
- Julia 1.10.11 and 1.12.6, current MTKBase 1.53: 249 passed, 17 pre-existing broken, exit 0.
- Julia 1.12.6, MTKBase 1.52.0 / SciMLBase 3.36.0 / Symbolics 7.32.0 / SymbolicUtils 4.40.1: exact `GROUP=Modeling` passed; conservation reported 249 passed and 17 pre-existing broken.
- Runic 1.7 was run on the changed file and repository. It reports pre-existing broad formatting drift; its diff proposes no change to these seven edited assertions, so no unrelated formatting was included.

No assertion was loosened, skipped, deleted, or changed to match buggy output; these seven tests are strengthened from expected-broken to required passing assertions.

Please ignore this draft until it has been reviewed by @ChrisRackauckas.

PR #1511, opened 2026-07-15
Test symbolic stoichiometry through public API

ChrisRackauckas-Claude:
Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

Replace the symbolic-stoichiometry test's construction of private ModelingToolkitBase affect internals with the documented public compilation path:

```julia
compiled_jsys = ModelingToolkitBase.mtkcompile(jsys)
@test isempty(ModelingToolkitBase.equations(compiled_jsys))
```

The test still inspects the generated symbolic jump system, verifies that it contains one jump, and checks that its symbolic affect contains `Pre(V(t))`. This addresses the review request to retain assertions about the generated symbolic jump and its equations.

## Root cause

The failure bisects to ModelingToolkit commit `136462c52d4c983e4ba832893928f270bc399b47` (`refactor: do not rely on metadata when constructing AffectSystem`), whose direct parent `0b01708ffb4c545b888f859bdcbc505350f723ef` passes the test. That refactor made the private `AffectSystem(...; parent_sys)` keyword mandatory, so Catalyst's test errored while reconstructing private internals rather than testing Catalyst behavior.

`ModelingToolkitBase.mtkcompile`, `jumps`, and `equations` are exported by their declaring module; `mtkcompile` has its own docstring and a rendered API-docs entry.

## Local verification

The exact downstream test command used Julia 1.12.6, current ModelingToolkit master `77941f873102904935c798fdb3dcd2bbfbfe8d3a`, and coverage:

```sh
GROUP=Modeling julia +1.12 --project=<downstream-env> --startup-file=no \
  -e 'using Pkg; Pkg.test("Catalyst"; coverage=true)'
```

- Clean Catalyst master: Symbolic Stoichiometry reached 20 passes, then errored with `UndefKeywordError: keyword argument parent_sys not assigned`.
- This PR: Symbolic Stoichiometry passed 30/30.
- The patched full group continued successfully through Reaction Balancing. It then stopped at the independent Conservation Laws baseline issue: 242 passed, 7 pre-existing `@test_broken` assertions unexpectedly passed, and 17 remained broken. That separate issue is tracked by draft PR #1512; this PR does not alter those markers.
- `git diff --check` passed.
- Julia 1.10 and 1.12 parsing passed.
- Runic passed on the changed executable `let` block. The whole-repository Runic check remains nonzero on pre-existing formatting outside this focused diff.

Please ignore this draft until it has been reviewed by @ChrisRackauckas.

PR #1513, opened 2026-07-16
Fix docs cleanup shell quoting

ChrisRackauckas-Claude:
Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- Quote the commit ID returned by `git commit-tree` before passing it to `git branch`.
- Quote the `HEAD^{tree}` revision expression.
- Remove the repo-wide actionlint `SC2046` and `SC1083` findings without changing cleanup behavior.

## Root cause

The faulty cleanup workflow was introduced in `b840ebc9b5d7e8f65391338504420f7e7e0b6569` (`Create DocsCleanup.yml`). Direct history inspection confirms the file is absent from its parent and is introduced with the unquoted command substitution and revision expression.

## Local verification

- `actionlint` from the repository root: passed.
- `git diff --check`: passed.
- Executed the exact rewritten command in a temporary Git repository and verified the created branch has the same tree as `HEAD` and commit subject `delete history`: passed.
- Repo-wide Runic remains red on a clean detached checkout of current `origin/master` (`ae55024af9289d201b353d27859b29909a5e3bde`); this is a widespread pre-existing formatting failure and is intentionally excluded from this focused workflow PR.

### DIFF
--- a/.github/workflows/DocsCleanup.yml
+++ b/.github/workflows/DocsCleanup.yml
@@ -26,7 +26,7 @@ jobs:
               git config user.email "documenter@juliadocs.github.io"
               git rm -rf "${preview_dir}"
               git commit -m "delete preview"
-              git branch gh-pages-new $(echo "delete history" | git commit-tree HEAD^{tree})
+              git branch gh-pages-new "$(echo "delete history" | git commit-tree "HEAD^{tree}")"
               git push --force origin gh-pages-new:gh-pages
           fi
         env:


ChrisRackauckas-Claude:
Final clean-base classification of the red checks:

- The documentation-only shell-quoting change passes repository-wide actionlint and the rewritten cleanup command passes an exact Git-history fixture.
- Current clean `master` reproduces the Modeling failure with the same dependency graph: Symbolic Stoichiometry reaches 20 pass and then errors because the test calls the non-public `ModelingToolkitBase.AffectSystem` constructor without its newly required `parent_sys` keyword. The dependency boundary is ModelingToolkit commit `136462c52`.
- [Catalyst #1511](https://github.com/SciML/Catalyst.jl/pull/1511) replaces only that private reconstruction with the public `mtkcompile`/`equations` path while retaining the one-jump and `Pre(V(t))` assertions. A fresh local Julia 1.12 run passes Symbolic Stoichiometry 30/30.
- The later conservation-law unexpected-pass baseline is independently handled by [#1512](https://github.com/SciML/Catalyst.jl/pull/1512); the older LTS DSL Options regression is handled by [#1503](https://github.com/SciML/Catalyst.jl/pull/1503) and its public-API prerequisite [ModelingToolkit #4714](https://github.com/SciML/ModelingToolkit.jl/pull/4714).
- The recurring QA/Extensions red jobs inspected here ended in runner cancellation or unavailable logs, not assertions caused by this workflow edit.

No test was skipped, weakened, or hidden. The red checks are clean-base/dependency or runner failures, not regressions from this PR.

PR #1523, opened 2026-07-23
Bump Catalyst to v16.2.3

ChrisRackauckas-Claude:
Bumps `Catalyst` **v16.2.2 → v16.2.3** (patch) so the accumulated unreleased `src/` changes on `master` can be registered.

**Rationale:** Update Catalyst strict QA for SciMLTesting 2.4 (#1522) (docs/QA/compat/internal; no public API or behavior break)

Part of a sweep of SciML packages whose source changed since their last registered version without a version bump.

> [!NOTE]
> Please ignore until reviewed by @ChrisRackauckas.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Catalyst"
 uuid = "479239e8-5488-4da2-87a7-35f2df7eef83"
-version = "16.2.2"
+version = "16.2.3"
 
 [deps]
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
@JuliaRegistrator register

Requesting registration of Catalyst v16.2.3 so SciMLBase ReleaseTest can pick up the parent_sys fix from #1509 (registered surface is still v16.2.2).

cc @ChrisRackauckas — please ignore if registration is intentionally held.

Issue #1527, opened 2026-08-01
CI: Tests (1, QA) kills the runner — JET 0.11 report_package peaks at 22 GB RSS on Julia 1.12


ChrisRackauckas-Claude:
## Summary

The `Tests (1, QA)` leg has not completed on any run I checked going back to 2026-07-05. It
does not fail a test — the **runner itself dies** partway through `JET.report_package`, and
GitHub reports it as a lost/cancelled runner. Locally reproduced: `GROUP=QA` on Julia 1.12
peaks at **22.4 GB RSS**, which no `ubuntu-latest` runner (8 GB self-hosted, 16 GB
GitHub-hosted) can survive.

This has been masking real QA regressions — the Symbolics 7.35 reexport failure fixed in
https://github.com/SciML/Catalyst.jl/pull/1526 was found by running QA locally, not by CI.

## It is not a segfault

The Julia process gets `signal 15: Terminated` (SIGTERM), not SIGSEGV. The repeated
`jl_parallel_gc_threadfun` / `pthread_cond_wait` frames in the log are Julia dumping every
GC thread's (idle) stack on SIGTERM, not a crash site. From
https://github.com/SciML/Catalyst.jl/actions/runs/30698976230/job/91366514247:

```
12:34:39  [toplevel-info] Analyzing top-level definitions (progress: 928/1455)
12:37:35  ##[error]The runner has received a shutdown signal...
12:37:35  [toplevel-info] Analyzing top-level definitions (progress: 986/1455)
12:37:35  [2442] signal 15: Terminated
          in expression starting at .../julia-runtest/v1/test_harness.jl:67
```

The workflow timeout is 360 min (`SciML/.github` `tests.yml` default) and the job died at
32 min, so this is not a job timeout either.

## Reproducibility in CI

Every `Tests (1, QA)` job checked, on `master` as well as PR branches:

| job | date | branch | GitHub annotation |
|---|---|---|---|
| [91366514247](https://github.com/SciML/Catalyst.jl/actions/runs/30698976230/job/91366514247) | 08-01 | PR #1526 | The operation was canceled (runner shutdown, GitHub-hosted) |
| [91367946733](https://github.com/SciML/Catalyst.jl/actions/runs/30699521160/job/91367946733) | 08-01 | PR branch | self-hosted runner lost communication … starves it for CPU/Memory |
[...]

PR #1526, opened 2026-08-01
QA: allow the Symbolics 7.35 Laplace reexports

ChrisRackauckas-Claude:
**This PR should be ignored until reviewed by @ChrisRackauckas.**

## Problem

`GROUP=QA` fails on unmodified `master` (Julia 1.12.4, current registry):

```
No unapproved public reexports: Test Failed at ~/.julia/packages/SciMLTesting/YQ45J/src/SciMLTesting.jl:808
  Expression: isempty(reexported)
   Evaluated: isempty([:inverse_laplace, :laplace, :laplace_solve_ode, :partial_frac_decomposition])

Test Summary:                                | Pass  Fail  Broken  Total      Time
Quality Assurance                            |   12     1       7     20  17m51.3s
```

Reproduce with `GROUP=QA julia --project=. -e 'using Pkg; Pkg.test()'` on a clean `master` checkout.

## Cause

Symbolics v7.35.0 (released 2026-07-31) added a Laplace-transform / partial-fractions
API in https://github.com/JuliaSymbolics/Symbolics.jl/pull/1590 ("Laplace Transform",
merged 2026-07-29, 26 minutes after the `v7.34.1` tag commit, so `v7.35.0` is the first
release carrying it). `src/Symbolics.jl` there gained:

```julia
export partial_frac_decomposition
export SymbolicLinearODE, symbolic_solve_ode, solve_linear_ode_system, solve_symbolic_IVP, laplace, inverse_laplace, laplace_solve_ode
```

`v7.34.1` has none of these four names; `v7.35.0` has all four. They reach Catalyst via
`Catalyst: @reexport using ModelingToolkitBase` -> `ModelingToolkitBase: @reexport using
Symbolics`, so `Base.isexported(Catalyst, :laplace)` is now `true`. No Catalyst source
change is involved and nothing about the version resolution changed otherwise
(Symbolics 7.35.0 / SymbolicUtils 4.44.0 / ModelingToolkitBase 1.58.2).

This is not visible in CI yet only because the last `master` Tests run predates the
Symbolics release; the next push will be red.

## What this PR does

Adds exactly those four names to `LEGACY_DEPENDENCY_REEXPORTS` in `test/qa/qa.jl`.
Nothing else.

## Why this (and the decision you may want to override)

The comment above the list says changing it "is a public-API decision, not a consequence
of a dependency adding a new public binding", so this is deliberately a draft rather than
a self-approving fix. The argument for accepting the names:

[...]

PR #1528, opened 2026-08-02
CI: pin the Julia 1 QA leg to a single-threaded 16 GB hosted runner

ChrisRackauckas-Claude:
**This PR should be ignored until reviewed by @ChrisRackauckas.**

Stopgap for https://github.com/SciML/Catalyst.jl/issues/1527. One file changed, workflow only.

## Problem

`Tests (1, QA)` has not completed on any run since at least 2026-07-10, on `master` as well as
PR branches. It is not a test failure and not a segfault — the runner dies mid-run and GitHub
annotates the job with *"The self-hosted runner lost communication with the server ... Anything
in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its
network access can cause this error."* The Julia process only ever receives SIGTERM from the
dying runner.

Cause: JET 0.11.6's `report_package` sweep of Catalyst peaks at **21.8 GB RSS** with 4 analysis
threads (234 GB allocated, 35% of the run in GC, 8.6 GB still live after a full GC). Neither
runner class the leg lands on can hold that — the self-hosted pool is 8 GB, GitHub-hosted is
16 GB. Full measurements and the per-signature breakdown are in #1527.

## What this PR does

Passes two existing inputs of the shared `SciML/.github` `tests.yml` for the `1` + `QA` cell of
the matrix only:

* `num-threads: 1` — JET 0.11 spawns one task per signature, each with its own analyzer and
  local inference caches (aviatesk/JET.jl#785), so peak memory scales with thread count.
* `runner: ["ubuntu-24.04"]` — forces a GitHub-hosted 16 GB runner. The 8 GB self-hosted pool
  squats only the `ubuntu-latest` label, so this is the established way to opt out of it (the
  shared workflow already uses the same trick for legs needing passwordless sudo).

`lts` (resolves JET 0.9, sequential, finishes in 4m11s) and `pre` (skips JET entirely via the
`VERSION < v"1.13-"` gate in `test/qa/qa.jl`) are untouched.

## Verification

Ran full `GROUP=QA` on Julia 1.12.4 inside a cgroup with a hard limit and no swap, to simulate
each runner class:

```
systemd-run --user --scope -p MemoryMax=<N> -p MemorySwapMax=0 \
  env GROUP=QA JULIA_NUM_THREADS=<T> \
  julia --project=. -e 'using Pkg; Pkg.test(; julia_args=["--check-bounds=yes"])'
```

| threads | cap | outcome |
|---|---|---|
| 4 | 16 GB | OOM-killed — `Package Catalyst errored during testing (received signal: KILL)` |
| 1 | 8 GB | OOM-killed — same |
| 1 | 16 GB | **completed** — `Quality Assurance \| 12 passed, 1 failed, 7 broken, 20m17.2s` |

The 4-thread/16 GB row reproduces the CI failure under controlled conditions; the 1-thread/16 GB
row is the configuration this PR selects. `actionlint` passes on the changed workflow.

Note the surviving run's single `Fail` is the Symbolics 7.35 reexport regression fixed by #1526,
which is a separate PR. **On this PR the Julia 1 QA leg is therefore expected to run to
completion and report that one failure** — that is the signal to check: a job that finishes and
reports, rather than a runner that disappears. With #1526 merged the leg should go green.

## Tradeoffs

* Roughly doubles the leg: QA testset goes from 9m35s (4 threads, unconstrained) to 20m17s, and
  that is on a 128-core machine — a 4-vCPU runner will be slower. With the ~20 min this job
  already spends precompiling, expect ~45 min.
* Headroom is thin. The analysis costs ~13 GB at its peak in the 1-thread configuration against a
  16 GB budget, and it grows with Catalyst's and ModelingToolkit's dependency surface. This buys
  time; it does not fix the cost.
* It does nothing for the other suggestion in #1527 — that the whole-package `report_package`
  sweep is the wrong shape of check for this stack in the first place, given
  ModelingToolkitBase's own QA uses targeted `@test_call` / `@test_opt` on concrete call sites
  and Catalyst's sweep findings are marked `jet_broken = true` and thus ignored anyway.

https://claude.ai/code/session_01JhZEGzymbz4sX7KWDEy9xy


ChrisRackauckas-Claude:
## CI verification

The Julia 1 QA leg **ran to completion for the first time since at least 2026-07-10**:
https://github.com/SciML/Catalyst.jl/actions/runs/30746446863/job/91492750137

Both inputs took effect:

```
JULIA_NUM_THREADS: 1
runner_name: GitHub Actions 1001057910      # hosted, not self-hosted-4vcpu-8gb-*
```

Result:

```
Test Summary:                                | Pass  Fail  Broken  Total     Time
Quality Assurance                            |   12     1       7     20  8m42.4s
  No unapproved public reexports             |          1              1     2.2s
##[error]Process completed with exit code 1.
```

That is a reported test failure — the Symbolics 7.35 reexport regression fixed by #1526 —
rather than a runner that disappears. No "lost communication" annotation, no SIGTERM dump, no
cancelled step. With #1526 merged this leg should be green.

Two corrections to the estimates in the description, in the favourable direction:

* The QA testset took **8m42s** on CI, not the ~20 min my locally capped run suggested. My
  local 1-thread measurement was made under a 16 GB cgroup on a heavily loaded 128-core box, so
  it overstated the GC pressure. Total job time was ~31 min, most of it the pre-existing
  precompilation.
* So the wall-clock cost of single-threading is much smaller than I claimed — the leg is now
  comparable to the other groups.

The `lts` and `pre` QA legs still ran on the 8 GB self-hosted pool as intended (untouched by
this change); `lts` fails with the same exit-code-1 reexport failure, not a runner death.

PR #1533, opened 2026-08-14
Raise OrdinaryDiffEq extras floor to 7

ChrisRackauckas-Claude:
**Please ignore this PR until it has been reviewed by @ChrisRackauckas.**

## What changed

Pure `[compat]` lower-bound bump for a test extra:

- `OrdinaryDiffEq = "6, 7"` → `"7"`

No solver code, tests, or package version were changed.

## Why

Catalyst requires `DocStringExtensions = "0.8-0.9"`, which is incompatible with OrdinaryDiffEq 6.0.0 (that release only exists against older DocStringExtensions). The declared `"6, 7"` extra floor therefore cannot resolve.

## Verification

Julia 1.10.11, isolated depot, `Pkg.develop` of this checkout, then `Pkg.add`:

| Pin | Result |
|---|---|
| OrdinaryDiffEq 6.0.0 (old extra floor) | unsatisfiable: DocStringExtensions 0.8–0.9 vs OrdinaryDiffEq 6.0.0 (needs 4.0–5.52 or 6.13.1+) |
| OrdinaryDiffEq 7.0.0 (new extra floor) | `RESOLVE_OK` / `LOAD_OK Catalyst` |

Not verified here: whether some later 6.13+ still resolves, the full test suite, QA, or a live `julia-downgrade-compat` run. Raising to `"7"` matches the other packages whose OrdinaryDiffEq 6.x window is dead against current pins.


ChrisRackauckas-Claude:
Follow-up measurement: OrdinaryDiffEq 6.13.1 (the first 6.x the 6.0.0 unsat log listed as DocStringExtensions-compatible) is also `RESOLVE_FAIL` against current Catalyst pins. 7.0.0 remains the first pin that develops+loads.

PR #1558, opened 2026-09-12
Remove broad ModelingToolkitBase reexport, export only what docs/tests need

ChrisRackauckas-Claude:
## Summary

`@reexport using ModelingToolkitBase` leaked the entire MTKB/Symbolics export surface (~300 names, including transitive reexports like `solve`, `Num`, `structural_simplify`, `ODESystem`, `@syms`, `@mtkbuild`, `unwrap`, ...) through `using Catalyst`.

This PR replaces it with an explicit `export` list of just the dependency-owned names that Catalyst's built documentation and test suite actually exercise through `using Catalyst`:

- DSL macros: `@parameters`, `@variables`, `@independent_variables`, `@discretes`, `@brownians`, `@poissonians`, `@named`, `@nonamespace`, `@mtkcomplete`, `@register_symbolic`, `@unpack`
- Module bindings: `ModelingToolkitBase`, `Symbolics`, `SymbolicUtils`
- Types: `System`, `Equation`, `Differential`, `Initial`, `Pre`, `GlobalScope`, `ParentScope`
- Accessors/builders: `complete`, `compose`, `continuous_events`, `equations`, `extend`, `full_equations`, `getmetadata`, `hasmetadata`, `independent_variables`, `isinitial`, `mtkcompile`, `observed`, `operation`, `parameters`, `setmetadata`, `simplify`, `substitute`, `unknowns`
- Problem constructors Catalyst already exported and extends: `ODEProblem`, `SDEProblem`, `JumpProblem`, `NonlinearProblem`, `SteadyStateProblem` (unchanged pre-existing exports)
- Spatial grids: `CartesianGrid`, `CartesianGridRej` (unchanged pre-existing exports)

`test/qa/qa.jl`'s `LEGACY_DEPENDENCY_REEXPORTS` allowlist is shrunk to the same set, and the `undefined_exports` Aqua exception is removed since the stale `Variable` reexport it tracked is gone. `Reexport` is dropped as a direct dependency. HISTORY.md documents the change under "Unreleased"; **no version bump — not marked breaking.**

## Test plan

- [x] `GROUP=QA julia --project -e 'using Pkg; Pkg.test()'` — passes (14 pass / 6 expected-broken; ExplicitImports reexport check matches the allowlist exactly, Aqua `undefined_exports` now clean)
- [x] `GROUP=Extensions julia --project -e 'using Pkg; Pkg.test()'` — passes
- [x] Every file in the Modeling / Simulation / Misc / Spatial / upstream test sets run individually — all pass (one missing name, `independent_variables`, was caught this way and added)
- [x] `julia --project=docs docs/make.jl` — full docs build completes; all 168 `@example` blocks evaluate
- [x] Runic format check clean; `typos` clean
- [ ] GROUP=All via `Pkg.test()` not run end-to-end (files were run individually instead; the Hybrid lane is separate from `All` and its file passes)

🤖 Generated with Devin CLI (model: SWE-2 Max), harness version 3000.10.21
Session: local session 7cb1e8c5a1cf4f23 (transcript: /home/crackauc/.local/share/devin/cli/summaries/history_7cb1e8c5a1cf4f23.md)


isaacsas:
Wouldn’t this be breaking?

isaacsas:
More generally is there a context where one wants Catalyst but not MTKBase?

ChrisRackauckas:
Yeah we can make this a breaking release. But I think the bigger thing is, Catalyst should probably be choosing what it considers its public API, rather than deriving it from whatever happens downstream (that makes the definition of breaking very loose), so the question is moreso whether this is the right set.

ChrisRackauckas-Claude:
Heads-up on overlap: this and #1559 address the same QA failure — `limited`/`limitnew`/`limitold` leaking through the broad `@reexport using ModelingToolkitBase`.

This PR is the structural fix: it removes the wholesale MTKBase reexport and exports an explicit curated list, so the allowlist entry for `limited` is no longer needed. Note this drops ~300 transitive reexports (`solve`, `Num`, `ODESystem`, `structural_simplify`, ...) from `using Catalyst` — a user-visible breaking surface change, larger in scope than a CI fix typically warrants.

#1559 is the alternative minimal fix: it just approves the three names in `LEGACY_DEPENDENCY_REEXPORTS` and leaves the reexport surface untouched.

These are mutually exclusive — merge at most one. If this PR merges, #1559 should be closed unmerged.

PR #1559, opened 2026-09-12
Approve limited/limitnew/limitold dependency reexports

ChrisRackauckas-Claude:
## Summary

ModelingToolkitBase 1.70 exports the `limited` operator API (`limited`, `limitnew`, `limitold`), which leaks through Catalyst's wholesale `@reexport` and fails the `No unapproved public reexports` QA check on master:

```text
Expression: isempty(reexported)
 Evaluated: isempty([:limited, :limitnew, :limitold])
```

These are documented upstream API names, so this approves them in `LEGACY_DEPENDENCY_REEXPORTS`, matching the precedent of approving the documented ModelingToolkitBase surface.

## Verification

`test/qa/qa.jl` in the QA project on Julia 1.13:

- Without this change: `Quality Assurance | 12 1 6 19` — the reexport check fails on exactly `[:limited, :limitnew, :limitold]`.
- With this change: `Quality Assurance | 13 0 6 19` — all non-broken tests pass.

Generated with Devin CLI 3000.10.21 (model: SWE-2 Max), session: /home/crackauc/.local/share/devin/cli/summaries/history_4c9fddb81d834736.md


ChrisRackauckas-Claude:
Heads-up on overlap: this and #1558 address the same QA failure — `limited`/`limitnew`/`limitold` leaking through the broad `@reexport using ModelingToolkitBase`.

This PR is the minimal fix: approve the three names in `LEGACY_DEPENDENCY_REEXPORTS`, keeping the existing reexport surface unchanged.

#1558 is the alternative, larger change: it removes the wholesale MTKBase reexport entirely and exports an explicit curated list instead — which also makes this allowlist entry unnecessary (`limited` would no longer be reexported at all) but drops ~300 transitive names (`solve`, `Num`, `ODESystem`, `structural_simplify`, ...) from `using Catalyst`, a user-visible breaking surface change.

These are mutually exclusive fixes — merge at most one. If #1558 is preferred, this PR should be closed unmerged; if the minimal fix is preferred for the CI sweep, #1558 can be reconsidered separately on its own merits.

ChrisRackauckas-Claude:
Note: this overlaps with #1558, which fixes the same `No unapproved public reexports` QA failure by removing the broad `ModelingToolkitBase` reexport instead of approving the leaked names. These are alternative remedies — if #1558 merges, this becomes unnecessary; if the minimal fix is preferred to unblock CI quickly, this can merge and #1558 can proceed independently as a cleanup.

ChrisRackauckas:
At least merging this to get green again, though the true follow up is #1558

